PR #2753, opened 2014-12-03
Nalanda rct3 to develop

aronasorman:


kalite/ab_testing/dynamic_assets.py
```diff
@@ -0,0 +1,42 @@
...
+    student_grade_level = fields.IntegerField(default=0)
+    unit = fields.IntegerField(default=1)
+
+
+# modify ds in-place. Use when you're modifying ds rather than defining new dynamic settings
+def modify_dynamic_settings(ds, request=None, user=None):
+
+    user = user or request.session.get('facility_user')
+
+    if user and not user.is_teacher:
```
jamalex: We may want to add a check for the Nalanda package for now here, since it otherwise doesn't need to waste cycles on the rest.

jamalex:
Replaced by #2883.

PR #2779, opened 2014-12-10
Delete khan-exercises folder, since we're using the one within the perseus folder

### LINKED ISSUE #2524: Delete khan-exercises folder, since we're using the one within the perseus folder
The `kalite/distributed/static/khan-exercises` directory should now be almost deletable.

It's still referenced in at least the following places, so these should switch over to using the ke/ files from within the perseus directory:
kalite/distributed/templates/distributed/knowledgemap.html
kalite/playlist/templates/playlist/view_playlist.html
kalite/student_testing/templates/student_testing/test.html
kalite/distributed/templates/distributed/base.html
and possible additional places.

mrpau-richard:
HI @aronasorman 

This fixed #2524.

I changes the `scr path` of the `khan-exercises` to `perseus/ke` static path and delete khan-exercises directory.


jamalex:
This has broken the exercises, so we will be undoing it for now. There are more places that need to be fixed in order to get khan-exercises working from within Perseus.

PR #2819, opened 2014-12-20
Tabular coach reports can now be generated for playlists as well as topics

### LINKED ISSUE #2808: Coach Reports -- Feature Addition
a. Arrange the coach report (exercises being done in class) by playlist

b. Add key (number of questions attempted) for what is meant by “Struggling” and “In 

Progress” for students

c. (?) Struggling students should be those students who have not been able to complete their 

streak after 20 attempts

MCGallaspy:
Is this on the right track to resolve #2808?


MCGallaspy:
@rtibbles, take a look at my latest PR and let me know if it's okay. This test made more sense to me as it's not coupled to the displayed text in the template, but rather to a particular code path which could be reached whether playlists exist or not.

rtibbles:
Looks good, @MCGallaspy - tests pass. Merging!

Issue #2767, opened 2014-12-05
Add timestamp fields to ContentLog model.


cpauya:
Branch: **develop**

We need to add two fields to the `ContentLog` model:
* `start_timestamp = models.DateTimeField()` == this must NOT be null
* `progress_timestamp = models.DateTimeField()`

These will complement these fields:
* `completion_timestamp = models.DateTimeField(blank=True, null=True)`
* `time_spent = models.FloatField(blank=True, null=True)`

The new fields will be helpful when doing date filters for contents and will help in fixing #2271.

PR #2828, opened 2015-01-01
Refactor ContentLog class. Use FloatField instead of IntegerField for progress field.

djallado:
@aronasorman.

Branch:  develop
Issue: https://github.com/learningequality/ka-lite/issues/2767
Related PR: https://github.com/learningequality/ka-lite/pull/2790

I changed `IntegerField` to `FloatField` for percentage use. 

Attached screen shoot:
Sample

### DIFF
--- a/kalite/distributed/static/js/distributed/content/views.js
+++ b/kalite/distributed/static/js/distributed/content/views.js
@@ -130,7 +130,7 @@ window.ContentBaseView = BaseView.extend({
     },
 
     set_progress: function(progress) {
-        if ((progress - (this.log_model.get("completion_counter") || 0)) > this.REQUIRED_PERCENT_FOR_FULL_POINTS) {
+        if ((this.log_model.get("progress") - (this.log_model.get("completion_counter") || 0)) > this.REQUIRED_PERCENT_FOR_FULL_POINTS) {
             this.log_model.set_complete();
         }
         this.log_model.set({
--- a/kalite/distributed/static/js/distributed/content/views.js
+++ b/kalite/distributed/static/js/distributed/content/views.js
@@ -136,6 +136,10 @@ window.ContentBaseView = BaseView.extend({
         this.log_model.set({
             points: Math.min(this.possible_points, Math.floor(this.possible_points * progress))
         });
+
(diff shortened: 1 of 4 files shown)


aronasorman:
@rtibbles any other parts of the code affected by this?

cpauya:
Yikes, I closed the #2767 issue and this PR got closed too.  

@djallado Can you please open another issue re use of FloatField and point this PR on that instead?

cpauya:
Ok, found the `reopen branch` button and will reopen this PR now.  Sorry about that.

kalite/distributed/static/js/distributed/content/views.js
```diff
@@ -136,6 +136,10 @@ window.ContentBaseView = BaseView.extend({
         this.log_model.set({
             points: Math.min(this.possible_points, Math.floor(this.possible_points * progress))
         });
+
```
rtibbles: No need to trigger two change events here, can pass the progress: progress key: value pair in the previous .set method call.

kalite/main/models.py
```diff
@@ -508,11 +508,10 @@ class ContentLog(DeferredCountSyncedModel):
     progress_timestamp = models.DateTimeField(blank=True, null=True)
     content_source = models.CharField(max_length=100, db_index=True)
     content_kind = models.CharField(max_length=100, db_index=True)
-    progress = models.IntegerField(blank=True, null=True) # TODO(djallado): Change this to Floatfield for percentage.
+    progress = models.FloatField(blank=True, null=True)  # (djallado): Changed this to Floatfield for percentage.
```
rtibbles: Remove TODO :)

rtibbles:
Couple of comments.

@aronasorman can't think of anything else that would be affected by this. Mostly just amazed that this didn't totally blow up before without this!

djallado:
@aronasorman. I am replicating this with student account and I view a pdf. I noticed that `complete` field is always return true even if I didn't scroll the pdf.

djallado:
I add `if` condition in `set_complete` method of `ContentLogModel` so that I can assure that the `complete` field will return true if `progress` is 1

Issue #2821, opened 2014-12-21
UI Issue - Improper overriding of search bar


mascot6699:
At the manage acount page(i.e  localhost/management/accounts)
    The green coloured search button's height is not overridden properly
    The current size is 18px while it should have been 34px

![screenshot from 2014-12-22 01 02 03](https://cloud.githubusercontent.com/assets/3908832/5519372/2c21252a-8976-11e4-970d-983df8c44add.png)

PR #2829, opened 2015-01-01
overrided the  search icon size properly

mascot6699:
This request is to correct the issue written below
UI Issue - Improper overriding of search bar #2821

### DIFF
--- a/kalite/control_panel/templates/control_panel/base.html
+++ b/kalite/control_panel/templates/control_panel/base.html
@@ -13,7 +13,7 @@
             margin: 0em;
         }
         #search-button {
-            height: 18px;
+            height: 34px;
         }
         .container {
             max-width: 1030px;
--- a/kalite/control_panel/templates/control_panel/base.html
+++ b/kalite/control_panel/templates/control_panel/base.html
@@ -8,13 +8,6 @@
     <link rel="stylesheet" type="text/css" href="{% static 'css/control_panel/base.css' %}" />
     <!-- Override bootstrap here to fix up search bar -->
     <style type="text/css">
-        input#search {
-            font: -webkit-small-control;
-            margin: 0em;
-        }
-        #search-button {
-            height: 34px;
-        }
         .container {
             max-width: 1030px;
         }


jamalex:
Cool, thanks, @mascot6699! Do you know whether we need to have that #search-button style in the CSS at all? Any sense of what is different about the control panel page vs other pages that requires it?

aronasorman:
Could you post screenshots too @mascot6699?

mascot6699:
screenshot from 2015-01-03 09 07 34

@aronasorman the screenshot

mascot6699:
@jamalex You are correct the overriding is not required at all remove it completely renders the same output as it does after the correction I made.

Also "input#search" has been overridden like this 
        input#search {
            font: -webkit-small-control;
            margin: 0em;
        }
for font as best practices but that makes the text overflow/cut from the viewport.
Two solutions exists for this either remove this part of css too or extend the search bars width more but that will make the control panel page search bar longer than others

jamalex:
@mascot6699 Great, thanks for trying that! Yes, it seems like both CSS rules could just be removed, unless there's another `input#search` somewhere in the control panel pages.

jamalex:
I think what happened is that we used to import a different version of Bootstrap on the control panel pages, for compatibility with our central, online server (which shares those templates), and these overrides were needed at that time. Now, those pages should all be using Bootstrap 3.

mascot6699:
@jamalex So what should I do now? Close this pull request and make changes in another branch and again open up a new one or there is some other method to follow. It may sound silly but I really don't know how to contribute to others project so bare with me.

jamalex:
Hey @mascot6699! Sorry, I should have given more instructions.

If you make the changes (removing those CSS rules completely, it looks like), commit it, and push that to your adjusting_searchbar branch, it will automatically update this pull request, and then we can merge it.

Thanks again!

mascot6699:
Please review it. I have checked it from my end.

jamalex:
Great, looks good -- thanks again for the contribution!

PR #2839, opened 2015-01-08
redesigned knowledge map base on d3 force layout

66eli77:
this is a fully working prototype, the UX need iterating.
(noticed: the topic.json contain receptive object under the same parent, which need to be fixed in order to run this knowledge map)


mjptak:
awesome stuff....will motivate me try and make a current branch work again
(:))....emosewA need I say it backwards

aronasorman:
@rtibbles can you review this?

rtibbles:
Yep

kalite/coachreports/api_resources.py
```diff
@@ -219,17 +219,16 @@ class TimelineReportExerciseResource(ExerciseSummaryResource):
     class Meta:
         queryset = ExerciseLog.objects.all()
         resource_name = 'TimelineReportExerciselog'
-        excludes = ['attempts_before_completion', 
-            'complete', 'counter', 'attempts', 'signed_version', 'signature', 'resource_uri', 'id', 'language', 
-            'resource_uri', 'user', 'streak_progress', 'points', 'completion_counter', 'completion_timestamp',
-            'exercise_id', 'mastered', 'struggling', 'deleted'
-            ]
+        # excludes = ['attempts_before_completion',
```
rtibbles: Reason for passing all this data to the client?

kalite/knowledgemap/api_resources.py
```diff
@@ -0,0 +1,38 @@
...
+from tastypie.exceptions import Unauthorized
+from tastypie.resources import ModelResource
+
+from kalite.shared.decorators import get_user_from_request
+
+from kalite.facility.api_resources import FacilityUserResource
+from kalite.shared.api_auth import UserObjectsOnlyAuthorization
+from kalite.main.models import ExerciseLog
+
+class ExerciseSummaryResource(ModelResource):
```
rtibbles: I am confused. Why are there changes to the coachreport endpoint above when you have another endpoint here?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,637 @@
+{% extends base_template %}
```
rtibbles: This could do with some refactoring and clean up.

First, extracting the javascript into separate javascript files, then we can sit down and discuss how to wrap this in backbone to make it more portable.

kalite/coachreports/api_resources.py
66eli77: right, cause I moved this to kalite/knowledgemap/api_resources.py

kalite/knowledgemap/api_resources.py
66eli77: you are right, it's repetitive code, I made a new commit that uses import instead.

kalite/knowledgemap/api_resources.py
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        excludes = ['attempts_before_completion', 
+            'complete', 'counter', 'attempts', 'language', 'signed_version',
+            'points', 'completion_counter',
+            'mastered', 'struggling', 'deleted'
+            ]
+        authorization = UserObjectsOnlyAuthorization()
+
+    # def get_object_list(self, request):
```
rtibbles: Could this commented out code be deleted?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
rtibbles: @66eli77 Can you put the js into a separate .js file before import?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+    <link rel='stylesheet' type='text/css' href='{% static "css/distributed/tipsy.css" %}' />
+    <link rel='stylesheet' type='text/css' href='{% static "css/knowledgemap/knowledgemap.css" %}' />
+{% endblock headcss %}
+
+{% block headjs %}{{ block.super }}
+    <script src="{% static 'js/distributed/jquery.tipsy.js' %}"></script>
+
+    <script type="text/javascript">
+        var TOPIC_DATA_URLS = {
+            Domains: '{{ data_url }}/map_domains.json',
```
rtibbles: Do we need all these data files now, or just the ALL_TOPICS_URL?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+            Topics: '{{ data_url }}/topics.json',
+            Tutorials: '{{ data_url }}/map_tutorials.json',
+            Exercises: '{{ data_url }}/exercises.json'
+        }
+        var ALL_TOPICS_URL = '{{ topics_url }}';
+    </script>
+{% endblock headjs %}
+
+{% block content %}
+<script src="http://d3js.org/d3.v3.min.js"></script>
```
rtibbles: This won't work offline. We already have d3 locally, check that it is an up to date version and change it if necessary.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+
+<script>
+window.statusModel.loaded.then(function() {
```
rtibbles: Should probably wrap this in a
```
$(function(){
     // code here
})
```
See here for more info: http://learn.jquery.com/using-jquery-core/document-ready/

This will wait for page load before executing any of this, which should reduce your chance of getting race conditions occurring.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+
+  //add map explanation
+  var graph_explain_img=document.createElement("img");
+  graph_explain_img.style.position="absolute";
+  graph_explain_img.style.top="70px";
+  graph_explain_img.style.left="10px";
+  graph_explain_img.style.width="100px";
+  graph_explain_img.setAttribute('src', "{% static 'images/distributed/map_explain_img.png' %}");
```
rtibbles: Is there no way to make this using HTML/CSS? As it is, we will not be able to translate the text strings.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+  document.body.appendChild(graph_explain_img);
+
+  var user_exercises_json;
+  var url = setGetParamDict("/api/knowledgemap/KnowledgeMapExerciselog/", {
```
rtibbles: This url should really be passed as a variable from a URL reverse tag.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+    user: statusModel.get("user_id"), 
+    completion_timestamp__gte: '2011-10-20',   //because the topic.json contain repetitive object under same parent, which cause problem for displaying knowledge map
+    completion_timestamp__lte: '2017-12-22'     //use this time filter to avoid encounter repetitive objects
```
rtibbles: How does using the time filter prevent getting repeated topic data?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+    height = window.innerHeight,
+    root;
+
+  var force = d3.layout.force()
+    // // .theta(.0001)
+    // .alpha(0.1)
+    .size([width, height]);
+    // .on("tick", tick);
+
+  $( ".kalite-navbar" ).after(function() {
```
rtibbles: Why is this a function and not just the string "<kmap></kmap>" ?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+
+  var link = svg.selectAll(".link"),
+      node = svg.selectAll(".node");
+
+  var initialized =false;
+
+  var subj_btn_top = 70;
+  var subj_btn_left = 30;
+
+  d3.json(sprintf(TOPIC_DATA_URLS.Topics, {channel_name: "khan"}), function(error, json) {
```
rtibbles: This fetch could probably happen earlier, it would not even have to wait until the statusModel is loaded.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+      }
+      update();
+  });
+
+  var loading = svg.append("text")
+      .attr("x", width / 2)
+      .attr("y", height / 2)
+      .attr("dy", ".35em")
+      .style("text-anchor", "middle")
+      .text("Simulating. One moment please…");
```
rtibbles: Not the most helpful message to display to the user.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
```diff
@@ -0,0 +1,616 @@
...
+      if(d.target.kind !== "Topic"){
+        return  d.source.children.length * 1 + d.source.title.length * 1.5; 
+      }else{
+        return  d.source.children.length * 4 + d.target.title.length * 5; 
+      }
+    })
+    .linkStrength(function(d){
+      if(d.target.kind !== "Topic"){
+        return 1;
+      }else if(d.source.kind == "Subject"){
```
rtibbles: Is this ever the case that a node has a kind of "Subject"?

rtibbles:
Some concerns about race conditions and load order that need addressing. This needs to be optimized, so I would like to merge it sooner rather than later.

@66eli77 Please address the issues here ASAP so we can merge, and then iterate on UX and optimization.

66eli77:
yeah, I will get this done by the end of the day

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: hehe, because after that time, there is no repeated topic in the json. But now we don't need to navigate via the map (now the map is static), so we can have repeated topic in the json

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: cause if I put <kmap> in the html file like this {% block content %} <kmap></kmap>{% endblock content %}, the map will show inside the content contain, which does not take all the white space. This function is so far the only way I found to put the <kmap> in the right place

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: how about "Preparing a knowledge map based on your learning data..."?

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: I guess because of js asynchronous nature, the node does have a kind of "Subject". but I'm moving this code after initial(), where some nodes get their kinds assigned to "Subject".

kalite/knowledgemap/api_resources.py
66eli77: dafdaf

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: done

kalite/knowledgemap/api_resources.py
66eli77: done

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: done

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: done

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: done

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: done

rtibbles:
Looking pretty clean now! Merging so we can start dogfooding and any further clean up.

kalite/knowledgemap/templates/knowledgemap/knowledgemap.html
66eli77: done

Issue #2843, opened 2015-01-09
Handlebars templates not being found on Windows


jamalex:
Branch: develop

The path resolution code in the handlebars template finder does something (possibly hardcoding the wrong type of slashes) that doesn't work on Windows, so no pages that use Handlebars templates work.

jamalex:
I think this is happening because of the following (regrettable) line in ka-lite/python-packages/fle_utils/handlebars/loader.py:

`template_name = filepath.split("/")[-1].split(".handlebars")[0]`

(since Windows paths use `\`, this doesn't work properly, but should hopefully be an easy fix)

jamalex:
You can test whether it's fixed by loading a path like `/handlebars/templates/exercise.js`, and checking that it contains a number of lines like:

`HB.register("exercise/test-start", "<h2> <span>{{ title }}</span> </h2> <span>{{_ \"Please click the button below when you are ready to begin the test.\" }}</span> <button class=\"btn success\" id=\"start-test\">{{_ \"Begin test\" }}{{ title }}</button>");`

`HB.register("exercise/quiz-stop", "<h2> {{_ \"Well done - you have finished the quiz!\" }} </h2> <h3> {{_ \"You got %(correct)d out of %(total_number)d questions correct.\" }} </h3>");`

PR #2845, opened 2015-01-12
Fix to use backslash character when platform is windows.

cpauya:
Hi @aronasorman - this fix is for #2843.

I have re-used the `fle_utils.platforms.is_windows()` function and did some minor PEP-8 fixes.

### DIFF
--- a/python-packages/fle_utils/handlebars/loader.py
+++ b/python-packages/fle_utils/handlebars/loader.py
@@ -13,8 +13,6 @@ from django.core.exceptions import ImproperlyConfigured
 from django.utils._os import safe_join
 from django.utils.importlib import import_module
 
-from fle_utils.platforms import is_windows
-
 
 # At compile time, cache the directories to search.
 fs_encoding = sys.getfilesystemencoding() or sys.getdefaultencoding()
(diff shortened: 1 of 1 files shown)


aronasorman:
@jtamiace can you pull this in and test if it's working?

jtamiace:
sorry, still cannot see the sidebar

jtamiace:
Error type: 

Uncaught Error: Could not find template: 'topics/sidebar'

python-packages/fle_utils/handlebars/loader.py
```diff
@@ -49,12 +53,18 @@ def get_template_module_paths(module_name):
...
     templates = []
 
+    if is_windows():
+        slash_character = "\\"
+    else:
+        slash_character = "/"
+
     for filepath in get_template_module_paths(module_name):
-        template_name = filepath.split("/")[-1].split(".handlebars")[0]
+        template_name = filepath.split(slash_character)[-1].split(".handlebars")[0]
```
jamalex: There should be much nicer `os.path` ways to do this, rather than manually splitting.

cpauya: Thanks for the tip! :smile:

cpauya:
Hi @jtamiace can you check if this works?

Btw, may I know which versions of Windows and which browser?

I am using Windows 7 and Firefox v34.0.5 on Virtualbox v3.18.x.

Issue #2530, opened 2014-10-24
Stamp availability code for videos and content is not fit for purpose


rtibbles:
Currently unused in develop, and slows code down immensely.

rtibbles:
Fixed in #2852

PR #2852, opened 2015-01-13
Availability stamping

rtibbles:
This is an MVP fix for #2530 

Summary of changes:
* Burns down old code
* Refactors front and back end code to make Videos part of Content
* Stamps content and dubbed content urls at server start
* Stamps content availability on topic tree at server start
* Creates new API endpoint for topic_tree data
* Removes a bunch of code that is no longer in use
* Invalidates the cache at video download
* Disk caching of annotated json

Not here:
* Implementing UUID based file system as per #2857 - will be addressed in a separate PR.

@jamalex @aronasorman @benjaoming would appreciate a look through - either to merge now and continue iterating after, or as a starting point for further iteration.


data/khan/content.json
```diff
@@ -1 +0,0 @@
-{"test": {"format": "mp3", "id": "test", "description": "Some Audio", "author": "somebody", "title": "AUDIO!", "kind": "audio"}}
```
rtibbles: :fire:

kalite/caching/__init__.py
```diff
@@ -50,22 +50,6 @@ def invalidate_on_video_delete(sender, **kwargs):
         invalidate_all_caches()
 
 
-def invalidate_all_pages_related_to_video(video_id=None):
```
rtibbles: :fire:

kalite/caching/__init__.py
```diff
@@ -85,23 +69,6 @@ def create_cache_entry(path=None, url_name=None, cache=None, force=False):
         logging.warn("Did not create cache entry for %s" % path)
 
 
-def regenerate_all_pages_related_to_videos(video_ids):
```
rtibbles: :fire:

kalite/contentload/management/commands/channels/base.py
```diff
@@ -215,41 +207,7 @@ def recurse_nodes_to_remove_childless_nodes(node):
             del node["children"][ci]
     recurse_nodes_to_remove_childless_nodes(topic_tree)
 
-    def recurse_nodes_to_add_position_data(node):
```
rtibbles: :fire:

No longer needed with new knowledgemap.

kalite/distributed/static/js/distributed/videoplayer.js
```diff
@@ -1,504 +0,0 @@
-    window.VideoPlayerState = {
```
rtibbles: :fire:

kalite/distributed/templates/distributed/learn.html
```diff
@@ -75,7 +75,7 @@
     {% endif %}
 
     <script type="text/javascript">
-        var ALL_TOPICS_URL = '{{ topics_url }}';
+        var ALL_TOPICS_URL = "{% url 'topic_tree' channel='%(channel_name)s' %}";
```
rtibbles: Use an api endpoint to access the dynamically annotated information.

kalite/distributed/views.py
```diff
@@ -79,113 +78,12 @@ def check_setup_status_wrapper_fn(request, *args, **kwargs):
     return check_setup_status_wrapper_fn
 
 
-def refresh_topic_cache(handler, force=False):
```
rtibbles: :fire:

kalite/i18n/__init__.py
```diff
@@ -351,21 +351,6 @@ def set_default_language(lang_code):
     Settings.set("default_language", lcode_to_ietf(lang_code))
 
 
-def get_langs_with_subtitle(youtube_id):
```
rtibbles: :fire:

A monstrously confusing function that ultimately does very little, and was only used in one place.

kalite/main/api_forms.py
```diff
@@ -1,55 +0,0 @@
-"""
```
rtibbles: :fire:

kalite/main/api_resources.py
```diff
@@ -71,110 +71,6 @@ class Meta:
         authorization = UserObjectsOnlyAuthorization()
 
 
-class Video:
```
rtibbles: :fire:

kalite/main/api_views.py
```diff
@@ -57,104 +40,6 @@ def student_log_api_wrapper_fn(request, *args, **kwargs):
         return student_log_api_wrapper_fn
 
 
-@student_log_api(logged_out_message=ugettext_lazy("Video progress not saved."))
```
rtibbles: :fire:

kalite/topic_tools/__init__.py
```diff
@@ -45,24 +41,26 @@
...
-                            if att not in props:
-                                del node[att]
-    return TOPICS
+def get_topic_tree(force=False, annotate=False, channel=settings.CHANNEL):
+    global TOPICS, TOPICS_FILEPATHS
+    if TOPICS.get(channel) is None or force:
+        TOPICS[channel] = softload_json(TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
+        validate_ancestor_ids(TOPICS[channel])  # make sure ancestor_ids are set properly
+
+    if annotate:
```
rtibbles: Loop through all the nodes in the topic tree and cross reference with the content_cache to check availability.

kalite/topic_tools/__init__.py
```diff
@@ -120,14 +109,46 @@ def get_leafed_topics(force=False):
...
 CACHE_VARS.append("CONTENT")
-def get_content_cache(force=False):
+def get_content_cache(force=False, annotate=False):
     global CONTENT, CONTENT_FILEPATH
 
     if CONTENT is None or force:
         CONTENT = softload_json(CONTENT_FILEPATH, logger=logging.debug, raises=False)
- 
+    
+    if annotate:
```
rtibbles: Loop through all content items and put thumbnail urls, content urls, and subtitle urls on the content dictionary, and list all languages that the content is available in.

kalite/topic_tools/__init__.py
```diff
@@ -338,40 +358,6 @@ def get_all_leaves(topic_node=None, leaf_type=None):
     return leaves
 
 
-def get_child_topics(topic_node):
```
rtibbles: :fire:

So many unused functions.

kalite/topic_tools/__init__.py
```diff
@@ -601,22 +445,12 @@ def get_content_data(request, content_id=None):
...
 
-    # Fallback mechanism
-    available_urls = dict([(lang, avail) for lang, avail in content["availability"].iteritems() if avail["on_disk"]])
-    if content["available"] and not available_urls:
-        content_lang = "en"
-        messages.success(request, "Got content content from %s" % content["availability"]["en"]["stream"])
-    else:
-        content_lang = i18n.select_best_available_language(request.language, available_codes=available_urls.keys())
-
     # TODO(jamalex): clean this up; we're flattening it here so handlebars can handle it more easily
```
rtibbles: This clean enough for you to remove this, @jamalex?

kalite/updates/videos.py
```diff
@@ -51,257 +47,3 @@ def get_downloaded_youtube_ids(videos_path=None, format="mp4"):
 def delete_downloaded_files(youtube_id):
     return videos.delete_downloaded_files(youtube_id, settings.CONTENT_ROOT)
 
-
```
rtibbles: :fire:

rtibbles:
Screenshot of 'unavailable' video in sidebar. @66eli77 @jtamiace please make this better.
image

data/khan/content.json
aronasorman: :+1:

kalite/caching/tests/cache_tests.py
```diff
@@ -13,35 +13,35 @@
 from .. import caching
 from fle_utils.django_utils import call_command_with_output
 from kalite.testing.base import KALiteTestCase
-from kalite.topic_tools import get_video_cache
+from kalite.topic_tools import get_content_cache
 
 
 class CachingTest(KALiteTestCase):
-    video_cache = get_video_cache()
+    content_cache = get_content_cache()
```
aronasorman: So if I remember correctly, `get_content_cache` will fetch videos too, not just mp3s and pdfs?

data/khan/content.json
jamalex: RIP :skull:

kalite/contentload/management/commands/channels/base.py
aronasorman: So this function adds the x_pos and y_pos attributes? Cool.

kalite/caching/__init__.py
jamalex: :rainbow:

kalite/caching/__init__.py
jamalex: :hatched_chick:

kalite/contentload/management/commands/channels/khan.py
```diff
@@ -106,7 +104,7 @@ def retrieve_API_data(channel=None):
 
     exercises = khan.get_exercises()
 
-    videos = khan.get_videos()
+    content = khan.get_videos()
```
aronasorman: I guess this will turn into `khan.get_content` soon?

kalite/caching/tests/cache_tests.py
```diff
@@ -49,30 +49,30 @@ def test_cache_across_clients(self):
...
 
         # Same deal, now using requests library
-        requests.get(self.live_server_url + video_path)
-        self.assertTrue(caching.has_cache_key(path=video_path), "expect: Cache key exists after requestsget")
-        caching.expire_page(path=video_path) # clean cache
-        self.assertTrue(not caching.has_cache_key(path=video_path), "expect: No cache key after expiring the page")
+        requests.get(self.live_server_url + content_path)
+        self.assertTrue(caching.has_cache_key(path=content_path), "expect: Cache key exists after requestsget")
+        caching.expire_page(path=content_path) # clean cache
+        self.assertTrue(not caching.has_cache_key(path=content_path), "expect: No cache key after expiring the page")
```
jamalex: We should probably burn down the page cache stuff as well. I think it was only being used on the homepage and on topic pages (in order to mitigate the fact that these pages were querying the disk for availability, and hence slow). Now, with the sidebar topic tree nav as a single-page app, we won't be leveraging (or needing) any of that.

Any pages we can think of that *would* currently benefit from caching?

kalite/distributed/static/js/distributed/content/models.js
```diff
@@ -113,7 +113,7 @@ window.ContentLogCollection = Backbone.Collection.extend({
     },
 
     url: function() {
-        return "/api/contentlog/?" + $.param({
+        return this.model.urlRoot + "?" + $.param({
```
aronasorman: We probably can port this to using `setGetParam`?

jamalex:
High-level question: I'm not so familiar with the in-memory data structure caching we're doing (@aronasorman?), but I recall that one challenge we had in the production environment is that invalidating stuff on one thread didn't necessarily invalidate it on other threads. Do we have something in place that's able to test these types of scenarios, or at least know how the system will behave?

kalite/contentload/management/commands/channels/base.py
jamalex: :disappointed_relieved:

kalite/i18n/__init__.py
aronasorman: What's our strategy for maintaining this feature though? I'm assuming that now we don't show the available subtitles for a video, since this is deleted.

kalite/main/api_resources.py
aronasorman: This model was only used for caching?

aronasorman:
The caching we're doing in `topic_tools.__init__.py`  is a global variable, so invalidating it in one thread should invalidate it in all threads within the same process.

aronasorman:
And invalidating it is just a simple `CACHE_VAR = None`.

kalite/main/api_urls.py
```diff
@@ -22,20 +19,16 @@
...
     url(r'^', include(ContentLogResource().urls)),
-    # Retrieve video data to render a front-end video player
-    url(r'^', include(VideoResource().urls)),
     # Retrieve exercise data to render a front-end exercise
     url(r'^', include(ExerciseResource().urls)),
     # Retrieve assessment item data to render front-end Perseus Exercises
     url(r'^', include(AssessmentItemResource().urls)),
     url(r'^', include(ContentResource().urls)),
-    url(r'^exercise_log/(?P<exercise_id>\w+)$', 'exercise_log', {}, 'exercise_log'),
-    url(r'^attempt_log/(?P<exercise_id>\w+)$', 'attempt_log', {}, 'attempt_log'),
```
aronasorman: Is this burned down because we don't need an API for it? Or it's been transferred to tastypie now?

kalite/main/api_views.py
aronasorman: Also transferred to tastypie?

kalite/main/api_views.py
```diff
@@ -261,104 +99,7 @@ def get_content_logs(request):
...
-    video = get_video_data(request, video_id)
-    if video:
-        return JsonResponse(video)
-    else:
-        return JsonResponseMessageError("Video with id %(video_id)s not found" % {"video_id": video_id}, status=404)
-
-
-@api_handle_error_with_json
-@backend_cache_page
-def knowledge_map_json(request, topic_id):
```
aronasorman: :fire:

kalite/topic_tools/__init__.py
aronasorman: So the subtitle annotation is transferred here?

kalite/distributed/static/js/distributed/content/views.js
```diff
@@ -95,7 +108,7 @@ window.ContentBaseView = BaseView.extend({
 
         this.active = false;
 
-        this.possible_points = 500;
+        this.possible_points = ds.distributed.turn_off_points_for_videos ? 0 : ds.distributed.points_per_video;
```
jamalex: Good catch. Maybe we'd want to separate these settings by content type, though? Or burn them with :fire:? Otherwise, I guess we're changing how many points a PDF is worth, etc?

kalite/topic_tools/__init__.py
```diff
@@ -120,14 +122,57 @@ def get_leafed_topics(force=False):
...
+                    content["lang_data_" + lang] = {
+                        "stream": settings.CONTENT_URL + dubmap.get(lang) + "." + format,
+                        "stream_type": "{kind}/{format}".format(kind=content.get("kind", "").lower(), format=format),
+                        "thumbnail": thumbnail,
+                    }
+            content["languages"] = languages
+            subtitle_lang_codes = [lc for lc in os.listdir(i18n.get_srt_path()) if os.path.exists(i18n.get_srt_path(lc, content.get("id")))]
+            subtitle_urls = [{
+                "code": lc,
+                "url": settings.STATIC_URL + "srt/{code}/subtitles/{id}.srt".format(code=lc, id=content.get("id")),
```
aronasorman: Yup, subtitles are now here.

kalite/caching/tests/cache_tests.py
rtibbles: Yes, videos are now part of content in general.

kalite/contentload/management/commands/channels/base.py
rtibbles: It used to. It is no more now. And unnecessary, thanks to @66eli77's new knowledgemap.

kalite/contentload/management/commands/channels/khan.py
rtibbles: No - get_videos is a method in the Khan API wrapper, so only gets videos from there.

kalite/distributed/static/js/distributed/content/models.js
rtibbles: Should be able to, yes.

kalite/distributed/static/js/distributed/content/views.js
rtibbles: Foundation for Learning Points Equality!

Yeah, no real strong feelings either way, but it seems simpler, and more transparent to end users to have the same max points for every content kind.

kalite/i18n/__init__.py
rtibbles: All logic for this is now used in the one place where this function was referenced.

kalite/main/api_resources.py
rtibbles: This model was used for the tastypie Resource endpoint for Video item data. This has now been subsumed under the very similar code for 'Content'.

kalite/main/api_urls.py
rtibbles: All tastypie now.

kalite/main/api_views.py
rtibbles: Yes - this is all deprecated.

kalite/topic_tools/__init__.py
rtibbles: Yep!

kalite/topic_tools/__init__.py
rtibbles: Indeedly.

kalite/main/api_resources.py
aronasorman: Ah, gotcha. Unification!!

jamalex:
Cool. I guess it might not properly invalidate in contexts where Apache etc are being used, though, depending on the worker model. Thoughts, @benjaoming?

kalite/caching/tests/cache_tests.py
aronasorman: None as I'm aware of, ever since we started moving stuff around to be more RESTy, nginx/apache should be able to cache the assets properly now? So now we can pass of the caching to that.

aronasorman:
Yeah, if it's using the process model then this won't invalidate properly.

kalite/distributed/static/js/distributed/videoplayer.js
jamalex: Would be nice to keep blame history when moving/splitting files. I'm still unclear on what's required to make that happen, but it sounds from this like it might be automatic? http://stackoverflow.com/questions/1728922/how-does-git-track-source-code-moved-between-files

kalite/distributed/templates/distributed/learn.html
jamalex: Ooooooh! Channel-aware client code! :sunglasses:

kalite/distributed/tests/browser_tests/distributed.py
```diff
@@ -289,32 +289,6 @@ def test_exercise_mastery(self):
...
-            logging.debug("Testing path : " + path)
-            self.browser.get(self.live_server_url + path)
-            error_list = self.browser.execute_script("return window.js_errors;")
-            if error_list:
-                logging.error("Found JS error(s) while loading path: " + path)
-                for e in error_list:
-                    logging.error(e)
-            self.assertFalse(error_list)
-
-
```
jamalex: Was this just deleted because it doesn't work anymore? Seems useful to be able to do.

kalite/distributed/views.py
jamalex: :stars:

kalite/main/api_resources.py
jamalex: Wow, cool. Where does video data get loaded from now...? (Probably going to see it below)

kalite/main/api_views.py
rtibbles: Except save_video_log for some reason. Having burned it down, I will recreate it in tastypie now...

kalite/main/tests/base.py
```diff
@@ -77,14 +77,14 @@ def get_num_cache_entries(self):
...
+    def create_random_content_file(self):
         """
-        Helper function for testing video files.
+        Helper function for testing content files.
         """
-        video_id = random.choice(get_video_cache().keys())
-        youtube_id = get_video_cache()[video_id]["youtube_id"]
-        fake_video_file = os.path.join(settings.CONTENT_ROOT, "%s.mp4" % youtube_id)
-        with open(fake_video_file, "w") as fh:
+        content_id = random.choice(get_content_cache().keys())
```
jamalex: Oooooh, randomized tests -- risky and fun! :game_die:

kalite/main/tests/base.py
```diff
@@ -77,14 +77,14 @@ def get_num_cache_entries(self):
...
+        youtube_id = get_content_cache()[content_id]["youtube_id"]
+        fake_content_file = os.path.join(settings.CONTENT_ROOT, "%s.mp4" % youtube_id)
```
jamalex: Wait... isn't it sometimes going to return a PDF or something...?

kalite/student_testing/api_views.py
```diff
@@ -1,59 +0,0 @@
...
-            context_type="exam",
-            context_id=data["title"]
-        )
-    except ValidationError as e:
-        return JsonResponse({"error": _("Could not save AttemptLog") + u": %s" % e}, status=500)
-      
-    # Special message if you've just completed.
-    #   NOTE: it's important to check this AFTER calling save() above.
-    if not previously_complete and testlog.complete:
-        return JsonResponse({"success": _("You have finished the test!")})
```
jamalex: Does the logic in this endpoint get replicated anywhere? Don't see this message, in particular, anywhere else now. But I guess we compute test completion on the client-side now, so this wasn't getting used anymore anyway?

kalite/topic_tools/__init__.py
```diff
@@ -45,24 +42,38 @@
...
-
-        # Limit the memory footprint by unloading particular values
-        if props:
-            node_cache = get_node_cache()
-            for kind, list_by_kind in node_cache.iteritems():
-                for node_list in list_by_kind.values():
-                    for node in node_list:
-                        for att in node.keys():
-                            if att not in props:
-                                del node[att]
```
jamalex: Aww, but this code was so beautiful!!!

image

kalite/topic_tools/__init__.py
jamalex: Might be good to include this as an inline comment?

kalite/topic_tools/__init__.py
```diff
@@ -45,24 +42,38 @@
...
+    if annotate:
+        if settings.HARD_CONTENT_CACHE and not force:
+            topics = softload_json(TOPICS_FILEPATHS.get(channel) + ".cache", logger=logging.debug, raises=False)
+            if topics:
+                TOPICS[channel] = topics
+                return TOPICS[channel]
+
+        content_cache = get_content_cache()
+        def recurse_nodes(node):
+            # By default this is very charitable, assuming if something has not been annotated it is available
```
jamalex: Ooh, tricksy. At first glance, I would have assumed it defaulted to `False`, but of course `[""]` is `Truthy`.

jamalex: Since it's just doing a Boolean check anyway, perhaps it could just be `True` there.

kalite/topic_tools/__init__.py
```diff
@@ -45,24 +42,38 @@
...
+            if content_cache.get(node.get("id"), {}).get("languages", [""]):
+                node["available"] = True
+            # Do the recursion
+            for child in node.get("children", []):
+                recurse_nodes(child)
+        recurse_nodes(TOPICS[channel])
+        if settings.HARD_CONTENT_CACHE:
```
jamalex: This name isn't super intuitive to me -- perhaps `DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP` or something more explicit?

What is the current mechanism for invalidating this cache?

jamalex: (e.g. I guess we would want to invalidate it when videos are finished downloading through the interface, or are deleted there?)

kalite/distributed/static/js/distributed/videoplayer.js
rtibbles: Yeah - but now you can just blame me. I did also move a bunch of stuff around, moving a lot of what was in the model into the view.

kalite/distributed/tests/browser_tests/distributed.py
rtibbles: Mostly because it is now testing code that is entirely Javascript. We should be doing this in a Javascript framework. And it usually gets skipped anyway.

kalite/main/api_resources.py
rtibbles: From the content endpoint. ONE ENDPOINT TO RULE THEM ALL!

kalite/main/tests/base.py
rtibbles: As you can see in what used to be line 84 - it was like that when I got here!

kalite/student_testing/api_views.py
rtibbles: Yeah, the logic in the endpoint is now client side, and the server side stuff is in tastypie now.

kalite/topic_tools/__init__.py
rtibbles: And unused...

kalite/topic_tools/__init__.py
rtibbles: :+1:

kalite/topic_tools/__init__.py
rtibbles: :+1:

kalite/topic_tools/__init__.py
rtibbles: It gets invalidated when force and annotate are both set to true. This happens on video download, will have to double check deletion.

rtibbles: Wasn't being invalidated on deletion, is now.

kalite/topic_tools/__init__.py
```diff
@@ -120,14 +122,57 @@ def get_leafed_topics(force=False):
...
+        if settings.HARD_CONTENT_CACHE and not force:
+            content = softload_json(CONTENT_FILEPATH + ".cache", logger=logging.debug, raises=False)
+            if content:
+                CONTENT = content
+                return CONTENT
+        for content in CONTENT.values():
+            languages = []
+            default_thumbnail = create_thumbnail_url(content.get("id"))
+            dubmap = i18n.get_id2oklang_map(content.get("id"))
+            for lang, dubbed_id in dubmap.items():
```
jamalex: Can you somehow use the variable name `dubstep` in this loop?

kalite/topic_tools/__init__.py
```diff
@@ -120,14 +122,57 @@ def get_leafed_topics(force=False):
...
+                # TODO (rtibbles): Explicitly stamp "mp4" on KA videos in contentload
+                format = content.get("format", "mp4")
+                if is_content_on_disk(dubbed_id, format):
+                    languages.append(lang)
+                    thumbnail = create_thumbnail_url(dubbed_id) or default_thumbnail
+                    content["lang_data_" + lang] = {
+                        "stream": settings.CONTENT_URL + dubmap.get(lang) + "." + format,
+                        "stream_type": "{kind}/{format}".format(kind=content.get("kind", "").lower(), format=format),
```
jamalex: Cool hack. Might be good to (at some point) have this be a lookup table so that we're not stuck with mime type naming for `Kind`? Or just go whole-hog and embrace mimetypes for `Kind` naming, I guess.

kalite/topic_tools/__init__.py
```diff
@@ -120,14 +122,57 @@ def get_leafed_topics(force=False):
...
+                        "thumbnail": thumbnail,
+                    }
+            content["languages"] = languages
+            subtitle_lang_codes = [lc for lc in os.listdir(i18n.get_srt_path()) if os.path.exists(i18n.get_srt_path(lc, content.get("id")))]
+            subtitle_urls = [{
+                "code": lc,
+                "url": settings.STATIC_URL + "srt/{code}/subtitles/{id}.srt".format(code=lc, id=content.get("id")),
+                "name": i18n.get_language_name(lc)
+                } for lc in subtitle_lang_codes if os.path.exists(i18n.get_srt_path(lc, content.get("id")))]
```
jamalex: These few lines are a bit dense. Could they be unpacked at all?

kalite/topic_tools/__init__.py
```diff
@@ -120,14 +122,57 @@ def get_leafed_topics(force=False):
...
+def get_content_cache(force=False, annotate=False):
     global CONTENT, CONTENT_FILEPATH
 
     if CONTENT is None or force:
         CONTENT = softload_json(CONTENT_FILEPATH, logger=logging.debug, raises=False)
- 
+    
+    if annotate:
+        if settings.HARD_CONTENT_CACHE and not force:
+            content = softload_json(CONTENT_FILEPATH + ".cache", logger=logging.debug, raises=False)
```
jamalex: Haven't followed the whole logic chain through, but will this only get run once per server run, or repeatedly every time the content cache is needed? The latter would seem inefficient, and I guess we would want (and may already have) it running only if the cache isn't already loaded or something?

kalite/topic_tools/__init__.py
```diff
@@ -261,44 +306,10 @@ def recurse_nodes(node):
...
-
-def get_parent(node, parent_type="Topic"):
-    return get_ancestor(node, ancestor_id=node["parent_id"], ancestor_type=parent_type)
-
-def get_videos(topic):
-    """Given a topic node, returns all video node children (non-recursively)"""
-    return filter(lambda node: node["kind"] == "Video", topic["children"])
-
-
-def get_exercises(topic):
```
jamalex: I remember writing some of these early on as helpers for manually traversing and doing stats on the topic tree data; as master burner, I leave it to your discretion, but some may be handy.

jamalex: Easy enough to re-write as needed, I guess.

kalite/topic_tools/__init__.py
```diff
@@ -401,106 +378,6 @@ def get_topic_videos(*args, **kwargs):
...
-
-
-def get_exercise_paths():
-    """This function retrieves all the exercise paths.
-    """
-    exercises = get_node_cache("Exercise").values()
-    return [n["path"] for exercise in exercises for n in exercise]
-
-
-def garbage_get_related_videos(exercises, topics=None, possible_videos=None):
```
jamalex: Now that's a name and a half!

kalite/main/tests/base.py
rtibbles: Yeah, I should fix this.

kalite/topic_tools/__init__.py
rtibbles: I *could*. I won't, however. I would be afraid that my loop would drop half way through.

kalite/topic_tools/__init__.py
rtibbles: I think embracing mimetypes for ```Kind``` naming is a good way forward.

kalite/topic_tools/__init__.py
rtibbles: Aye aye, cap'n!

kalite/topic_tools/__init__.py
rtibbles: This only gets run at server load, or when the cache gets invalidated, yes.

aronasorman:
Tests pass. Merging.

jtamiace:
I don't know if you still want feedback on the part about the "unavailable video," but if the problem is making it clear that it is a video that cannot yet be viewed, we can put a visual icon that would indicate that it isn't available, like a lock icon. Can explore other icon possibilities, and also play with different colors of the tabs themselves.

lock pic

benjaoming:
@jamalex @aronasorman 

> Cool. I guess it might not properly invalidate in contexts where Apache etc are being used, though, depending on the worker model. Thoughts, @benjaoming?

Yeah, the assumption of shared state does not hold at all. Not in Apache, nor on other application servers, too. So you can't invalidate cache by setting a static variable, because if the next hit on the cache happens in another process, then it won't see the request to renew the cache.

rtibbles:
Seems good, @jtamiace - this is live on develop, so feel free to iterate there!

rtibbles:
@benjaoming The cache variables are Global variables in Django (it's about holding topic and content metadata in memory, rather than caching pages). I think the concern was that these global variables would not be shared across processes (although they would be shared across threads). Is there a way around this?

aronasorman:
@benjaoming yeah, they won't be shared across processes. I agree that we'd like to move to a more robust python-side caching system soon. 

I remember you mentioning yesterday that we can create Django models that are backed by an in-memory SQLite DB. Maybe that's a good way to do that?

benjaoming:
@rtibbles -- Theoretical ramble coming your way :) No way around that, nope... you can have a truly shared state stored in files and databases, but sharing state across processes in variables is impossible, and across threads it's an unnecessary complexity (you'll have read/write locks on stuff) that won't solve things for a large part of the application servers out there.

Storing in files doesn't help you with race conditions, and storing in database will add db dependencies.. but that's not necessarily a bad thing.

Having multiple cache stores per process doesn't guarantee consistency across the cache, but usually that's not important anyways as long as you've got a consistent way of invalidating it.. i.e. you could store the cache in some super-fast memory based fashion, and make sure that the invalidation mechanism is consistent. So for instance, store topic and content meta data in variables and only hit the database to know their latest timestamp.

@aronasorman and @rtibbles
If we're doing caching, I think it's a good time to not invent something that Django already handles fine :) Django has a nice key/value storage for caching with really good utility functions. You choose the storage engine yourself: So for debugging or some other efficiency deal, you can choose in memory caching (the default), or you can choose a database backend mechanism based on one of the entries in `settings.CACHES` (similar to settings.DATABASES).

Choosing the right backend can be done with a cache router, so you can have different types of cache engines for different purposes / apps:

https://docs.djangoproject.com/en/1.6/topics/cache/

Btw. all this talk about SQLite, you know it's got a global lock, right? Having a separate database file for the cache store would help a bit, but not sure if all this SQLite scales much :)

rtibbles:
I would much rather use the default Django caching mechanisms. I was just working off what we had previously.

Do you have a good plan for redoing this kind of in memory caching that we are doing?

benjaoming:
As far as I can tell, you don't need to setup anything, if we go ahead and refactor without adding any additional configuration, Django will be using in-memory caching.

So this would need to happen:
https://docs.djangoproject.com/en/1.7/topics/cache/#basic-usage

Refactoring current code base, we would get:

***Before***

    NODE_CACHE = None
    CACHE_VARS.append("NODE_CACHE")
    def get_node_cache(node_type=None, force=False):
        global NODE_CACHE
        if NODE_CACHE is None or force:
            NODE_CACHE = generate_node_cache(get_topic_tree(force))
        if node_type is None:
            return NODE_CACHE
        else:
            return NODE_CACHE[node_type]

***After***

    from django.core import cache
    # ...
    def get_node_cache(node_type=None, force=False):
        """
        Get a cached version of the nodes. If it doesn't exist, it will be generated.
        
        :param force: Forces cache reload
        :param node_type: The type of node to return, don't know what this is neither :)
        """
        cache_key = 'kalite_node_cache'
        node_cache = cache.get(cache_key) if not force else None
        if node_cache is None:
            cache.set(cache_key, generate_node_cache(get_topic_tree(force)))
        if node_type is None:
            return node_cache
        else:
            return node_cache[node_type]

A quick note on cache keys: Since they are unique strings, don't store them in some constant variable. It's the same as URL names.

PR #2854, opened 2015-01-13
master to nalanda-rct3

aronasorman:


aronasorman:
This is mostly Ben's deletion of the docs/ folder. I'm skipping 0.12.10 since it's EOL soon.

Issue #2312, opened 2014-08-11
Reposition check answers and scratchpad (esp for tablet use)


mikewray:
This is a navigation /ease of use thing -  is there any chance you can move the 'check answer' button above the numpad - so either below or above the answer box. When on a tablet you enter the answer on numpad, have to go to the bottom...check answer, hit it again then scroll back up and its easy to get 'lost' and requires energy that would be better to have on maths. Also, I think it would be useful to have the 'show scratchpad' at the top rather than the bottom for exactly the same reason. Both will make the user experience more pleasant and the functionality better used , which means more points for everyone!

![check answers and scratchpad](https://cloud.githubusercontent.com/assets/7093712/3884318/9abf65fe-21ad-11e4-88af-f89683940c92.png)

PR #2861, opened 2015-01-14
Made 'show scratchpad' more prominent in exercises

Antrikshy:
This targets part of issue #2312. The "show scratchpad" button is hard to find. Since the exercises widget is not loading correctly in the develop branch, I was not able to find a better spot to put the button inside the exercises widget. So I put it at the top, and it looks fine to me.

This is what my version looks like:

screen shot 2015-01-14 at 3 38 41 pm

Here's the before:

screen shot 2015-01-14 at 3 41 58 pm

### DIFF
--- a/kalite/distributed/hbtemplates/exercise/exercise.handlebars
+++ b/kalite/distributed/hbtemplates/exercise/exercise.handlebars
@@ -9,12 +9,6 @@
     <div class="exercises-body col-sm-12">
         <div class="exercises-card current-card">
             <div class="current-card-container card-type-problem">
-                <br/>
-                <div>
-                    <button class="simple-button" id="scratchpad-show" href style>{{_ "Show scratchpad" }}</button>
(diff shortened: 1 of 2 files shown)


aronasorman:
@mikewray will be happy to see this.

jamalex:
Cool! I might recommend having this below the exercise area, so it doesn't push the exercise area off the bottom of the screen for tablet users.

Antrikshy:
Okay, I'll move it to the bottom and maybe make the button color more prominent after my class.

Antrikshy:
Whaddya guys think about this? I just found existing button classes and tweaked some spacing to get this.

screen shot 2015-01-14 at 5 53 11 pm

I'm also working on updating that disabled version of the button.

Antrikshy:
I left that 'not available' message as a `span` instead of a `button` because the JS selecting it may expect it to be a `span`. I added classes to make it a disabled button, but I can't test it until the exercises load up properly.

aronasorman:
Looks good to me.

PR #2862, opened 2015-01-15
install instructions change now that scripts are in scripts dir

benjaoming:


aronasorman:
These changes in script locations are important to document in the user manual + release. @MCGallaspy

benjaoming:
I'll try to keep things updated in the README.md, but I suggest we don't do install instructions before we're at the goal line... the `kalite start` and `kalite stop` commands are quite significant and we should put some effort into discouraging using the scripts directly.

aronasorman:
Yup, sounds good, especially while they're being programmed, and we haven't pinned down the exact behaviour yet for the `kalite` binary.

PR #2863, opened 2015-01-15
use the url template tag instead of a context variable

benjaoming:
A fairly simple change :) And maybe more a small example or madman OCD. But putting a link somewhere is template logic, not view logic :)

### DIFF
--- a/kalite/facility/templates/facility/login.html
+++ b/kalite/facility/templates/facility/login.html
@@ -25,7 +25,8 @@
         </form>
         <div id="info_block">
             {% if not restricted %}
-                {% blocktrans %}If you don't have a login, please <a href="{{ sign_up_url }}">sign up</a>.{% endblocktrans %}
+                {% url 'facility_user_signup' as signup_url %}
+                {% blocktrans %}If you don't have a login, please <a href="{{ signup_url }}">sign up</a>.{% endblocktrans %}
             {% else %}
                 {% blocktrans %}If you don't have a login, please contact your administrator.{% endblocktrans %}
             {% endif %}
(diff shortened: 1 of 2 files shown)


kalite/facility/templates/facility/login.html
```diff
@@ -25,7 +25,8 @@ <h3 class="no-margin-top">{% trans "Log in" %}</h3>
         </form>
         <div id="info_block">
             {% if not restricted %}
-                {% blocktrans %}If you don't have a login, please <a href="{{ sign_up_url }}">sign up</a>.{% endblocktrans %}
+                {% url 'facility_user_signup' as signup_url %}
+                {% blocktrans %}If you don't have a login, please <a href="{{ signup_url }}">sign up</a>.{% endblocktrans %}
```
jamalex: Where is the `signup_url` variable defined?

benjaoming: `{% url 'facility_user_signup' as signup_url %}` ;)

jamalex: Oh, totally didn't even see that line. Any reason the url blocktag can't just go right inside the href? I guess this is cleaner for i18n?

benjaoming: Yeah, it's a "feature" of blocktrans that you cannot put other block tags inside it.. just simple variable substitution. Not even template filters are allowed.

aronasorman:
Can you put this into the https://github.com/learningequality/ka-lite/wiki/Coding-guidelines-and-conventions wiki article @benjaoming?

benjaoming:
Absolutely :) I have a small notebook on "things that could improve over time" :) So I'm hoping we can have time for a Django oriented session together next week to talk about best practices, conventions, coding patterns etc...

benjaoming:
Rearranged the article a bit by putting a Django chapter at the bottom and including the existing template paragraph in that one.

I wrote the MVT chapter... let's talk more about these things soon!

PR #2870, opened 2015-01-16
Javascript testing

rtibbles:
@aronasorman This should be the basic requirements for QUnit based JavaScript unit testing.

However, this won't work without something resembling a build process in place, because we need an explicit ordering of import for the javascript files. Just wanted to put this up here for consideration and to drive it forward.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ script:
 - python kalite/manage.py syncdb --migrate --noinput --traceback
 - python kalite/manage.py test --traceback -v 2
 - grunt jshint
+- karma start --single-run --browsers PhantomJS
 notifications:
   email:
     - aron@learningequality.org
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ script:
 - python kalite/manage.py syncdb --migrate --noinput --traceback
 - python kalite/manage.py test --traceback -v 2
 - grunt jshint
-- ./node_modules/.bin/karma start --single-run --browsers PhantomJS
+- ./node_modules/karma/bin/karma start --single-run --browsers PhantomJS
 notifications:
   email:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ script:
 - python kalite/manage.py syncdb --migrate --noinput --traceback
 - python kalite/manage.py test --traceback -v 2
 - grunt jshint
+- ./node_modules/karma/bin/karma start --single-run --browsers PhantomJS
```
aronasorman: I think doing it this way means you need to install `karma` locally, not with `-g`. See how `grunt` is invoked above.

package.json
```diff
@@ -23,6 +23,13 @@
...
     "grunt-contrib-less": "~0.6.4",
-    "grunt-contrib-jshint": "^0.10.0"
+    "grunt-contrib-jshint": "^0.10.0",
+    "karma": "~0.12.31",
+    "qunit": "~0.7.5",
+    "sinon": "~1.12.2",
+    "qunitjs": "~1.16.0",
+    "karma-qunit": "~0.1.4",
+    "karma-sinon": "~1.0.4",
+    "karma-phantomjs-launcher": "~0.1.4"
```
aronasorman: So many new dependencies! It would be nice to make an inline comment for what each library does.

aronasorman:
@MCGallaspy @benjaoming might be good to be aware of this.

package.json
benjaoming: Agreed!! It's nice to have a couple of words guiding our future selves when suddenly we feel like it ad have time for optimizing the stack :)

rtibbles:
Closing this as it is stale, and we should discuss what JS testing framework to use going forwards.

PR #2875, opened 2015-01-20
Prevents videoscan.py from invalidating cache

### LINKED ISSUE #2873: Make videoscan cache invalidation less aggressive

rtibbles:
Fixes #2873 

Videoscan does not actually touch any files, so will not change the status of the cache anyway. It only modifies VideoObjects in the repo.

What this does mean, however, is that our availability stamping might show partially downloaded videos(?)

### DIFF
--- a/kalite/updates/management/commands/videoscan.py
+++ b/kalite/updates/management/commands/videoscan.py
@@ -100,6 +100,3 @@ class Command(CronCommand):
                 self.stdout.write("Deleted %d VideoFile models (because the videos didn't exist in the filesystem)\n" % len(deleted_video_ids))
             return deleted_video_ids
         touched_video_ids += delete_objects_for_missing_videos(youtube_ids_in_filesystem, videos_marked_at_all)
-
-        if len(touched_video_ids):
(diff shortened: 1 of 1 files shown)


rtibbles:
@jamalex @aronasorman not sure how the file downloading works, but is it possible that partially downloaded files will show as available from an os.path.exists check? If so, I will open a separate issue about it.

aronasorman:
The `videodownload` process for undubbed videos eventually leads to a `[urllib.urlretrieve](https://docs.python.org/2/library/urllib.html#urllib.urlretrieve)]` call. It says here that it does some caching of the download, but I'm not sure if it writes it immediately to the specified location or it places it on a temporary file first.

rtibbles:
OK - either way I am happy to leave this as an open edge case, which can be cleaned up when we do the UUID based file system, which will allow for a 'receiving' directory for in progress content that then gets shuffled into the file system when complete.

aronasorman:
Yeah, we can have that fixed when the UUID filesystem lands in develop.

PR #2882, opened 2015-01-22
use translate3d instead of CSS left

66eli77:


kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -145,12 +145,12 @@ window.SidebarView = BaseView.extend({
 
     update_sidebar_visibility: function() {
         if (this.state_model.get("open")) {
-            this.sidebar.css({left: 0});
-            this.$(".sidebar-tab").css({left: this.$(".sidebar-panel").width()}).html("&lt");
+            this.sidebar.css({'-webkit-transform': 'translate3d(0, 0, 0)'});
+            this.$(".sidebar-tab").css({'-webkit-transform': 'translate3d(' + this.$(".sidebar-panel").width() + 'px, 0, 0)'}).html("&lt");
```
aronasorman: Hmm wouldn't this lock us to webkit based browsers? I guess if this is meant as a hack for the android ports then we don't mind, but if this is meant to bubble up to develop then we'll want a more general code.

66eli77: I added support for firefox, opera, ie. for IE, it requires IE10 and above

rtibbles:
only concerns were about browser support, but this is wch only, so will merge when tests pass (seemed to be non-deterministic failure, so rerunning).

PR #2886, opened 2015-01-23
Small cleanups to develop

rtibbles:
Summary of changes:
* Removes unnecessary json files
* Changes vestigial references to 'Video' node cache
* Deletes leaflet js and css
* Fixes bug in assessment_item_id setting.

@aronasorman @jamalex Merging this now to make develop more testable. Please review post.


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -245,6 +240,22 @@ window.ExerciseView = Backbone.View.extend({
 
             } else if (framework == "perseus") {
 
+                var item_index;
```
rtibbles: This needs to happen after data_model load, otherwise "all_assessment_items" is undefined.

kalite/topic_tools/__init__.py
```diff
@@ -123,12 +124,27 @@ def get_assessment_item_cache(force=False):
 
     return ASSESSMENT_ITEMS
 
+def recurse_topic_tree_to_create_hierarchy(node, level_cache={}, hierarchy=[]):
```
rtibbles: Khan-specific topic tree hierarchy generation for coach report topics. I don't think these get used anywhere else now. Hopefully we can blow this away with the coach reports redesign.

rtibbles:
Tests pass, merging.

kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -245,6 +240,22 @@ window.ExerciseView = Backbone.View.extend({
...
+
+                var assessment_items = self.data_model.get("all_assessment_items") || [{id: ""}];
+
+                if (typeof attempts !== "undefined") {
+
+                    item_index = attempts % assessment_items.length;
+
+                } else {
+                    
+                    item_index = Math.floor(Math.random() * assessment_items.length);
```
jamalex: Hm, so does it start with a random index if there were no previous attempts? Or is attempts `0` if there were no previous attempts?

jamalex: If the former, then it seems as if it starts with a random index, then switches to the 2nd item in the list after that?

jamalex:
Cool! Just had one question in there. And a tear to shed for Leaflet.js. :cry:

kalite/distributed/static/js/distributed/exercises/views.js
rtibbles: Attempts defaults to 0 on the exerciseLog model, so it should work :)

rtibbles:
Shed the tear. But 30000 lines gone!

Issue #2884, opened 2015-01-23
Error when POSTing to /api/videolog/ while viewing a video


MCGallaspy:
branch: develop

The request has the following payload
```js
complete: false
content_kind: "Video"
content_source: ""
extra_fields: "{"video_id":"aNqG4ChKShI","last_percent":0.00037763252145081117}"
last_percent: 0.00037763252145081117
points: 0
progress: 0.0004881041597095554
time_spent: 0.25
user: "/securesync/api/user/5771031293ea5a7ab7ef102bd7ca7752/"
video_id: "aNqG4ChKShI"
views: 1
```

And the error is `{"error": ["This field cannot be blank."]}`. Now presumably this is because the content_source field is blank. I'm hoping there's a quick fix here, but I'm not too familiar with this backbone stuff yet. @rtibbles do you have a take on it?

rtibbles:
@MCGallaspy I think the error is because from main/models.py
```total_seconds_watched = models.IntegerField(default=0)```
is blank for this. This is due to my attempts to extend the ContentLog model to create the VideoLog model on the client side.

**Comment by MCGallaspy** [comment]
I tried setting that field in the POST request, got same error.

**Comment by rtibbles** [comment]
Anything else not being set from this?
```
    user = models.ForeignKey(FacilityUser, blank=True, null=True, db_index=True)
    video_id = models.CharField(max_length=100, db_index=True); video_id.minversion="0.10.3"  # unique key (per-user)
    youtube_id = models.CharField(max_length=20) # metadata only
    total_seconds_watched = models.IntegerField(default=0)
    points = models.IntegerField(default=0)
    language = models.CharField(max_length=8, blank=True, null=True); language.minversion="0.10.3"
    complete = models.BooleanField(default=False)
    completion_timestamp = models.DateTimeField(blank=True, null=True)
    completion_counter = models.IntegerField(blank=True, null=True)
```

MCGallaspy:
Looks like the culprit is the youtube_id attribute. Can we allow this to be blank, or should we populate it with something?

rtibbles:
Ah, right. Might be best just to populate it with something for now. Maybe in the VideoPlayerView just do:
```this.log_model.set("youtube_id", this.log_model.get("video_id"));```?

Issue #2881, opened 2015-01-22
Unable to download videos


MCGallaspy:
Branch: develop

To reproduce, simply launch the development server using `manage.py runserver`, login as an admin and attempt to download videos. You should then receive the opaque error message `INFO:kalite:Error in downloading NjJFJ7ge_qk: 'Video'; continuing to next video.` (If possible, I'd like to get people to try to reproduce this bug! Might shed some light.)

This issue is blocking further work on #2864 and #2858 for me.

MCGallaspy:
Was caused by an extant get_node_cache("Video") line. Should be fixed by #2874.

rtibbles:
Should be fixed in develop now.

Issue #2869, opened 2015-01-16
BrowserActionMixins need to be refactored


MCGallaspy:
Branch: develop

In kalite/testing/mixins/browser_mixins.py, the method `browser_register_user` doesn't have the desired effect. Perhaps other methods need to be refactored too?

aronasorman:
As suggested jokingly here, but needs some merit: "We need to write tests for the test helpers!"

MCGallaspy:
There wasn't an error in the mixins per se, but rather the usage wasn't immediately clear, and incorrect usage failed silently. I made failure throw an assertion error, it's in my branch for #2858. Will make a PR when I'm ready for that.

rtibbles:
Fixed in #2889

Issue #2865, opened 2015-01-15
Error when viewing exercises/videos: video_id better be set by internal code.


MCGallaspy:
Branch: develop

When accessing *some* (not all) videos or exercises, I get a 500 response. Unfortunately, it appears to happen at random (i.e., the same request may or may not produce the error). I'm including some potentially relevant request/response info below:

Remote Address:127.0.0.1:8008
Request URL:http://127.0.0.1:8008/api/videolog/
Request Method:POST
Status Code:500 INTERNAL SERVER ERROR

Request payload:
{"id_key":"9XZypM2Z3Ro","content_source":"","content_kind":"Video","user":"/securesync/api/user/fcc9e39f9bc45e9598fe797b443e2957/","complete":false,"points":0,"views":1,"progress":0,"time_spent":0,"extra_fields":"{\"id_key\":\"9XZypM2Z3Ro\"}"}

Response:
{"error_message": "video_id better be set by internal code.", "traceback": "Traceback (most recent call last):\n\n  File \"/home/gallaspy/Desktop/ka-lite/python-packages/tastypie/resources.py\", line 195, in wrapper\n    response = callback(request, *args, **kwargs)\n\n  File \"/home/gallaspy/Desktop/ka-lite/python-packages/tastypie/resources.py\", line 426, in dispatch_list\n    return self.dispatch('list', request, **kwargs)\n\n  File \"/home/gallaspy/Desktop/ka-lite/python-packages/tastypie/resources.py\", line 458, in dispatch\n    response = method(request, **kwargs)\n\n  File \"/home/gallaspy/Desktop/ka-lite/python-packages/tastypie/resources.py\", line 1320, in post_list\n    updated_bundle = self.obj_create(bundle, **self.remove_api_resource_names(kwargs))\n\n  File \"/home/gallaspy/Desktop/ka-lite/python-packages/tastypie/resources.py\", line 2084, in obj_create\n    return self.save(bundle)\n\n  File \"/home/gallaspy/Desktop/ka-lite/python-packages/tastypie/resources.py\", line 2230, in save\n    bundle.obj.save()\n\n  File \"/home/gallaspy/Desktop/ka-lite/kalite/main/models.py\", line 60, in save\n    assert kwargs.get(\"imported\", False), \"video_id better be set by internal code.\"\n\nAssertionError: video_id better be set by internal code.\n"}

[...]

PR #2889, opened 2015-01-23
Issue 2858

MCGallaspy:
Solves issues #2858, #2884, #2865, #2881, and #2869. (And possibly a few other issues I opened along the way.)


MCGallaspy:
And #2864.

kalite/distributed/static/js/distributed/topics/router.js
```diff
@@ -5,7 +5,8 @@ ChannelRouter = Backbone.Router.extend({
 
     routes: {
         "":   "navigate_default_channel",
-        ":channel/(*splat)":    "navigate_channel"
+        ":channel/(*splat)":    "navigate_channel",
+        "/(.*)/": "trigger_navigation_callback"
```
rtibbles: Might be good to comment what the purpose of this is.

kalite/main/api_resources.py
```diff
@@ -269,6 +270,7 @@ def __init__(self, lang_code="en", **kwargs):
         self.selected_language = lang_code
         if self.description == "None":
             self.description = ""
+        self.source = settings.CHANNEL
```
rtibbles: Can you add a TODO (BLOCKER) to change this for 0.14? We will need to make this happen dynamically when we have multiple channels active at once.

MCGallaspy:
Tried running the tests on my own machine, I'm also hanging on this test:
`test_user_interface_teacher (kalite.control_panel.tests.control_panel.CSVExportBrowserTests)`
I'll look in to it.

MCGallaspy:
Hanging on a test again (`test_points_update_after_backbone_navigation`), which is a bit disturbing because it doesn't hang on that test on my machine... non-deterministic browser tests?

MCGallaspy:
I'm deeply suspicious of this, as the indicated failed tests run fine on my machine, and further my own machine fails tests which are passing here. I think the deeper issue is with our testing framework, and the fact that variable server response times are causing expected DOM elements (or in my case, certain javascript objects) to be missing or not populated in the expected way. Thoughts on how to proceed?

aronasorman:
Talked with @MCGallaspy. We decided to just add `skipIf`s since the tests are failing for no reason, and it's the test suite's fault, not our code. We'll revisit this once we start refactoring the whole testing infrastructure.

rtibbles:
I agree - there is a limit to how robust we can make our client side testing by doing it through Selenium in this way. Roll on #2870!

MCGallaspy:
Okay, maybe for now if a test fails on Travis we can just get someone to check it out manually before merging if we suspect that it's a failure of the test suite and not an actual bug?

rtibbles:
Yeah, just adding that you ran the test manually, and setting it to skip if travis is probably best. If you want to do that with these failing tests, we can get this PR merged!

rtibbles:
All comments addressed. Tests pass. Merging.

Issue #2888, opened 2015-01-23
'Show Keypad' button on exercises floats around in an unsightly fashion


rtibbles:
![image](https://cloud.githubusercontent.com/assets/1680573/5881617/1bfce1d0-a2f7-11e4-8436-a911ae10ed43.png)

Could really do with this being styled in a way that it didn't fall out the bottom of its containing box.

Any takers?

aronasorman:
@Antrikshy wanna work on this? It's not Ruby.

Antrikshy:
Not Ruby and not the annoying GitHub personal access tokens? Sold! Exercises were not loading properly for me before. Is that fixed now?

aronasorman:
Should be fixed now I think for the latest develop. Although if you don't have the assessment items json file you can only load a subset of the exercises. @rtibbles can you point @Antrikshy to exercises he can test this with?

Antrikshy:
@rtibbles Where are you? None of the exercises seem to work.

rtibbles:
They just require the special sauce of ```assessmentitems.json``` @jamalex and @aronasorman are brewing a new batch right now.

Antrikshy:
Update: This will be fixed very soon. I'm just cleaning up the button to make it uniform with the rest.

PR #2897, opened 2015-01-27
Fixed issue #2888: 'Show keypad' button floating around

Antrikshy:
This should fix issue #2888. The 'show keypad' button now sits where it should and matches the hint and scratchpad buttons. Software keyboard has been tweaked slightly. The buttons are now `simple-button`s instead of the weird `green_button` class (which had no CSS styling anywhere LOL).

screen shot 2015-01-26 at 11 48 43 pm

Things are now pixel-perfect!

I have also improved the coloring on the main green buttons that I edited with the recent overhaul. They match the orange buttons better.

> I had some trouble with my local clone and had to fiddle around a bit. Ignore the "changes" to `models.js` and `views.js`. Unless I'm missing something, the changes it shows for both those files match the current versions of those files in the develop branch. Nothing should change. Sorry if it causes any trouble.

### DIFF
--- a/kalite/distributed/static/js/distributed/software-keyboard.js
+++ b/kalite/distributed/static/js/distributed/software-keyboard.js
@@ -96,8 +96,7 @@ window.SoftwareKeyboardView = Backbone.View.extend({
                 .appendTo(self.software_keyboard);
 
             jQuery.each(row, function(j, key) {
-                var keySpan = $("<div class='.col-xs-4'><button class='key simple-button " + (key === "bs" ? "key-bs" : "") + "' value='" + key + "'>" + (key === "bs" ? "Del" : key) + "</button></div>").appendTo(rowDiv);
-
+                var keySpan = $("<div class='.col-xs-4'><button class='key simple-button' value='" + key + "'>" + (key === "bs" ? "Del" : key) + "</button></div>").appendTo(rowDiv);
             });
         });
 
--- a/kalite/distributed/static/js/distributed/software-keyboard.js
+++ b/kalite/distributed/static/js/distributed/software-keyboard.js
@@ -74,7 +74,7 @@ window.SoftwareKeyboardView = Backbone.View.extend({
 
         // TODO-BLOCKER (rtibbles): 0.13 - Turn this into a handlebars template, conditionally render templates based on exercise types.
 
-        this.$el.append("<button id='show-keyboard'>" + gettext("Hide Keypad") + "</button>");
+        this.$el.append("<button class='simple-button orange' id='show-keyboard'>" + gettext("Hide Keypad") + "</button>");
 
         this.$el.append("<div class='container-fluid' id='software-keyboard'></div>");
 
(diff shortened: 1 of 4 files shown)


rtibbles:
Yeah, I think the green-button class was added pre-bootstrap and probably got culled during bootstrap integration. Will check the changes to models and views.

kalite/distributed/static/js/distributed/software-keyboard.js
```diff
@@ -96,7 +96,7 @@ window.SoftwareKeyboardView = Backbone.View.extend({
                 .appendTo(self.software_keyboard);
 
             jQuery.each(row, function(j, key) {
-                var keySpan = $("<div class='.col-xs-4'><button class='key green_button " + (key === "bs" ? "key-bs" : "") + "' value='" + key + "'>" + (key === "bs" ? "Del" : key) + "</button></div>").appendTo(rowDiv);
+                var keySpan = $("<div class='.col-xs-4'><button class='key simple-button " + (key === "bs" ? "key-bs" : "") + "' value='" + key + "'>" + (key === "bs" ? "Del" : key) + "</button></div>").appendTo(rowDiv);
```
rtibbles: One thing here - the 'Del' key is slightly misaligned on the keypad. I think it is inheriting some styling that changes the font size and makes it misaligned, can you play around with it (first, just make its styling the same as all the other keys), and see if that fixes the alignment issue too?

rtibbles:
Merges seem fine. You're just going to get blamed for code you didn't write now though :)

Antrikshy:
Hah "blamed".

I totally forgot about that Del button. It's not pixel perfect, I guess. I can't seem to find why it is misaligned. It must be because of the way it's added using JavaScript. I'll look into it today. Maybe make it an empty button somewhere in the template and only modify it using JS.

rtibbles:
In khan-lite.css there is a class called .key-bs that is being applied to the "Del" key. I think just deleting this definition (and then deleting its application in the line I annotated) should make this align properly.

MCGallaspy:
I wanted to check this out, but I can't figure out where the show keypad button is supposed to be... :(

Steps to reproduce?

Antrikshy:
@MCGallaspy I have fixed the show keypad button. The Hide Keypad button *is* the Show Keypad button. It changes contextually. The only thing left to do is to align the Del button correctly.

aronasorman:
@Antrikshy or @rtibbles: @MCGallaspy can't seem to find the "Show Keypad" button, any guidance on this?

rtibbles:
Looking at the basic addition exercise is the best test case.

Antrikshy:
screen shot 2015-01-29 at 10 15 55 am

Everything is fixed and cleaned up. Say hello to keypad 2.0 (more like 1.0.1)!

Antrikshy:
By the way, @rtibbles, when I click any of the buttons on the keypad, it says "There are still more parts of this question to answer." Is this a known issue or just something on my end because I am missing that assessmentitems file? Seems fishy to me.

MCGallaspy:
Keypad looks beautiful, but I get the following error when I click on it:
```
 Uncaught TypeError: Cannot read property 'setRangeText' of undefined software-keyboard.js:58
```

Antrikshy:
The keypad clicks are unrelated to the changes in this PR. I have the same problem with the develop branch. I have created a new issue #2907 for this and I'm looking into it. This PR is ready to be merged.

PR #2900, opened 2015-01-27
Properly render exercises when not logged in.

### LINKED ISSUE #2899: When not logged in, question_data is undefined in 'load_question' function on ExerciseView
This gives causes the problem to fail to load for non-logged in users.

rtibbles:
Fixes #2899


rtibbles:
Small change, manually tested. Tests pass. Merging.

Issue #2902, opened 2015-01-27
Cannot load exercises in develop branch at all


Antrikshy:
I have a fresh tree in Git because reasons. Now I get this when I try to view an exercise:

![screen shot 2015-01-27 at 11 50 57 am](https://cloud.githubusercontent.com/assets/6601788/5925781/e4941714-a61a-11e4-878d-95ca3497c6f4.png)

The problem may be with the `/kalite/distributed/settings.py`, particularly on this line:

    KHAN_EXERCISES_DIRPATH = os.path.join(os.path.dirname(__file__), "static", "khan-exercises")

It refers to a folder that does not exist. @rtibbles suggested that the line should point to `os.path.join(os.path.dirname(__file__), "distributed", "static", "perseus", "ke")`, but that doesn't seem to fix it. Neither does `os.path.join(os.path.dirname(__file__), "distributed", "static", "perseus")`. Still the same error.

aronasorman:
@Antrikshy should be opening a PR for this soon.

rtibbles:
Fixed in #2903

PR #2903, opened 2015-01-27
Updated khan-exercises directory paths

Antrikshy:
Fixes issue #2902 and possibly various other problems.

### DIFF
--- a/kalite/distributed/settings.py
+++ b/kalite/distributed/settings.py
@@ -98,7 +98,6 @@ SECURESYNC_PROTOCOL = getattr(local_settings, "SECURESYNC_PROTOCOL", "https" if 
 CENTRAL_SERVER_HOST   = getattr(local_settings, "CENTRAL_SERVER_HOST",   ("adhoc.%s:7007" if DEBUG else "kalite.%s") % CENTRAL_SERVER_DOMAIN)
 CENTRAL_WIKI_URL      = getattr(local_settings, "CENTRAL_WIKI_URL",      "http://kalitewiki.%s/" % CENTRAL_SERVER_DOMAIN)
 
-KHAN_EXERCISES_DIRPATH = os.path.join(os.path.dirname(__file__), "static", "perseus", "ke")
 PDFJS = getattr(local_settings, "PDFJS", True)
 
 ########################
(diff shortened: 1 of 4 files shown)


kalite/distributed/settings.py
```diff
@@ -98,7 +98,7 @@ def USER_FACING_PORT():
 CENTRAL_SERVER_HOST   = getattr(local_settings, "CENTRAL_SERVER_HOST",   ("adhoc.%s:7007" if DEBUG else "kalite.%s") % CENTRAL_SERVER_DOMAIN)
 CENTRAL_WIKI_URL      = getattr(local_settings, "CENTRAL_WIKI_URL",      "http://kalitewiki.%s/" % CENTRAL_SERVER_DOMAIN)
 
-KHAN_EXERCISES_DIRPATH = os.path.join(os.path.dirname(__file__), "static", "khan-exercises")
+KHAN_EXERCISES_DIRPATH = os.path.join(os.path.dirname(__file__), "static", "perseus", "ke")
```
aronasorman: Actually, we'll wanna move this definition out of here and into kalite/settings.py.

kalite/topic_tools/__init__.py
```diff
@@ -430,7 +430,7 @@ def get_exercise_data(request, exercise_id=None):
     if not exercise:
         return None
 
-    exercise_root = os.path.join(settings.KHAN_EXERCISES_DIRPATH, "exercises")
+    exercise_root = os.path.join(settings.KHAN_EXERCISES_DIRPATH)
```
rtibbles: I don't believe this edit is necessary. It should be looking inside the exercises folder.

kalite/settings.py
```diff
@@ -170,7 +170,6 @@ def package_selected(package_name):
 import_installed_app_settings(INSTALLED_APPS, globals())
 
 # Override
-KHAN_EXERCISES_DIRPATH = getattr(local_settings, "KHAN_EXERCISES_DIRPATH", os.path.join(STATIC_ROOT, "khan-exercises"))
```
aronasorman: Move the definition from distributed/settings.py, then wrap it in the `getattr(local_settings, "KHAN_EXERCISES_DIRPATH", ...)` call, to allow users to override it.

kalite/topic_tools/__init__.py
aronasorman: Breaks i18n. Will fix in a later PR.

rtibbles: How? The edit to i18n above will make language pack exercise templates unpack into static/perseus/ke/exercises/{ka_lang_code} folder

rtibbles:
Merging to iterate.

Issue #2472, opened 2014-10-11
Have makemessages scan handlebars files to extract translatable strings


jamalex:
We have an i81n Handlebars helper function, e.g. `{{_ "Back to topics" }}`, that uses the JS i18n mechanisms to translate strings being rendered within the templates. We need to make sure these strings are getting picked up and included in _djangojs.pot_.

The templates all:
- End with .handlebars
- Are in a subfolder of an hbtemplates folder within an app that's in `INSTALLED_APPS`.

The current Handlebars helper is:

```
Handlebars.registerHelper("_", function(i18n_key) {

    // TODO(jamalex): make sure this is getting picked up and included in the po files
    return sprintf(gettext(i18n_key), this);

});
```

PR #2905, opened 2015-01-27
Khan exercise i18n

### LINKED ISSUE #2904: Reinstate translated khan-exercises
Due to https://github.com/learningequality/ka-lite/pull/2903 to get khan-exercises working again, we're gonna have to apply a fix to make translated exercises from language packs work again.

rtibbles:
Fixes #2904, and other i18n fixes.

Summary of changes:
* Moves unpack location of translated khan-exercises templates
* Makes client side code actually look for translated template
* Generates a unique topic_tree for each language
* Translates coach report topic titles
* Translates exercise titles and descriptions for client side display
* Translates topic tree for video download screen

This is not quite ready to merge, as I have some questions about approach, @aronasorman:
Is it efficient to be storing a separate topic tree in memory for each language?

If so, it would seem sensible to also do the translation for content_cache and exercise_cache inside get_content_cache and get_exercise_cache, so that whenever they are referenced elsewhere (such as in coach reports) they can also have translated data used on the client side (which does not currently happen).


kalite/topic_tools/__init__.py
```diff
@@ -45,12 +44,15 @@
...
-def get_topic_tree(force=False, annotate=False, channel=settings.CHANNEL):
+def get_topic_tree(force=False, annotate=False, channel=settings.CHANNEL, language="en"):
+    translation.activate(language)
     global TOPICS, TOPICS_FILEPATHS
     if not TOPICS:
         TOPICS = {}
     if TOPICS.get(channel) is None:
-        TOPICS[channel] = softload_json(TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
+        TOPICS[channel] = {}
+    if TOPICS.get(channel, {}).get(language) is None:
```
rtibbles: @aronasorman Comments on this approach appreciated. If doing this for every language seems sensible, I would like to extend this approach to content_cache and exercise_cache.

rtibbles:
Screenshots:
image

image

image

kalite/topic_tools/__init__.py
aronasorman: I'm trying to think of negative consequences to this approach, the only one I can think of right now is that there might be a bit more memory bloat in storing an entire topic as part of the cache tree vs. just regenerating the topic tree on the fly (if i remember how the old one was done correctly).

kalite/topic_tools/__init__.py
```diff
@@ -58,10 +60,10 @@ def get_topic_tree(force=False, annotate=False, channel=settings.CHANNEL):
 
     if annotate:
         if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
-            topics = softload_json(TOPICS_FILEPATHS.get(channel) + ".cache", logger=logging.debug, raises=False)
+            topics = softload_json(TOPICS_FILEPATHS.get(channel) + "_" + language + ".cache", logger=logging.debug, raises=False)
```
aronasorman: So if we don't set `annotate` to `True` we won't generate the topic tree for the language?

kalite/topic_tools/__init__.py
aronasorman: Otherwise the approach seems good, and isn't too far from what we eventually want.

kalite/topic_tools/__init__.py
```diff
@@ -45,12 +44,15 @@
 # Globals that can be filled
 TOPICS          = None
 CACHE_VARS.append("TOPICS")
-def get_topic_tree(force=False, annotate=False, channel=settings.CHANNEL):
+def get_topic_tree(force=False, annotate=False, channel=settings.CHANNEL, language="en"):
+    translation.activate(language)
```
aronasorman: might be good to wrap this in `lcode_to_django_lang`, unless you can guarantee up the stack that this language code is run through the same function.

kalite/topic_tools/__init__.py
rtibbles: annotate gets set to true if the topic_tree for that language does not exist already.

kalite/topic_tools/__init__.py
rtibbles: Good call!

kalite/topic_tools/__init__.py
rtibbles: Seems a topic tree occupies about 1847232 bytes in memory.

kalite/coachreports/api_views.py
```diff
@@ -168,7 +168,7 @@ def query_logs(users, items, logtype, logdict):
     return logdict
 
 
-def compute_data(data_types, who, where):
+def compute_data(data_types, who, where, language="en"):
```
rtibbles: Allow for language specificity in all coach report output.

kalite/coachreports/templates/coachreports/tabular_view.html
```diff
@@ -86,7 +86,7 @@
             <select id="report_type">
                 <option {% if not request_report_type %}selected{% endif %}>----</option>
                 {% for report_type in report_types %}
-                    <option value="{% trans report_type %}" {% if report_type == request_report_type %}selected{% endif %}>{{ report_type }}</option>
+                    <option value="{{ report_type.value }}" {% if report_type.value == request_report_type %}selected{% endif %}>{{ report_type.name }}</option>
```
rtibbles: This allows coach reports to work in all languages - previously would break as the value was getting translated and then checked in English.

kalite/coachreports/views.py
```diff
@@ -207,7 +209,7 @@ def tabular_view(request, report_type="exercise"):
 
     context.update({
         # For translators: the following two translations are nouns
-        "report_types": (_("exercise"), _("video")),
+        "report_types": ({"value": "exercise", "name":_("exercise")}, {"value": "video", "name": _("video")}),
```
rtibbles: Concomitant with change in template above. Allows value and translated value not to be conflated.

kalite/distributed/static/js/distributed/exercises/models.js
```diff
@@ -57,7 +57,7 @@ window.ExerciseDataModel = Backbone.Model.extend({
                 "secondsPerFastProblem": this.get("seconds_per_fast_problem"),
                 "authorName": this.get("author_name"),
                 "relatedVideos": this.get("related_videos"),
-                "fileName": this.get("name") + ".html" // this.get("file_name")
+                "fileName": this.get("template")
```
rtibbles: Even when we passed the right template previously, it was not used. Now it is.

kalite/distributed/views.py
```diff
@@ -5,35 +5,23 @@
 * Administrative pages
 and more!
 """
-import copy
```
rtibbles: Lots of unused imports and code here.

kalite/i18n/__init__.py
```diff
@@ -38,7 +38,7 @@ class LanguageNotFoundError(Exception):
 
 def get_localized_exercise_dirpath(lang_code):
     ka_lang_code = lang_code.lower()
-    return os.path.join(settings.KHAN_EXERCISES_DIRPATH, "exercises", ka_lang_code)
+    return os.path.join(os.path.dirname(__file__), settings.KHAN_EXERCISES_RELPATH, "exercises", ka_lang_code)
```
rtibbles: Use KHAN_EXERCISES_RELPATH to allow language pack Khan-Exercises templates be copied into the correct directory after unpacking.

kalite/main/api_views.py
```diff
@@ -20,4 +20,4 @@ def flat_topic_tree(request, lang_code):
 @api_handle_error_with_json
 @backend_cache_page
 def topic_tree(request, channel):
-    return JsonResponse(get_topic_tree(channel=channel))
+    return JsonResponse(get_topic_tree(channel=channel, language=request.language))
```
rtibbles: Passes translated topic_tree to sidebar.

kalite/settings.py
```diff
@@ -169,7 +169,9 @@ def package_selected(package_name):
 
 import_installed_app_settings(INSTALLED_APPS, globals())
 
-KHAN_EXERCISES_DIRPATH = getattr(local_settings, "KHAN_EXERCISES_DIRPATH", os.path.join(os.path.dirname(__file__), "static", "perseus", "ke"))
+KHAN_EXERCISES_RELPATH = os.path.join("static", "perseus", "ke")
```
rtibbles: Hard coded to allow this to be referenced below, and in i18n to unpack language specific khan-exercise templates.

kalite/topic_tools/__init__.py
```diff
@@ -81,26 +82,32 @@ def recurse_nodes(node):
             else:
                 # By default this is very charitable, assuming if something has not been annotated
                 # it is available - needs to be updated for exercises.
-                if content_cache.get(node.get("id"), {}).get("languages", True):
+                if content_cache.get(node.get("id"), {}).get("available", True):
                     node["available"] = True
 
-        recurse_nodes(TOPICS[channel])
+            # Translate everything for good measure
+            translation.activate(i18n.lcode_to_django_lang(language))
```
rtibbles: Do the translation activation here so that it doesn't get disrupted by other calls.

kalite/topic_tools/__init__.py
```diff
@@ -108,12 +115,39 @@ def get_node_cache(node_type=None, force=False):
 
 EXERCISES          = None
 CACHE_VARS.append("EXERCISES")
-def get_exercise_cache(force=False):
+def get_exercise_cache(force=False, language="en"):
     global EXERCISES, EXERCISES_FILEPATH
-    if EXERCISES is None or force:
-        EXERCISES = softload_json(EXERCISES_FILEPATH, logger=logging.debug, raises=False)
+    if EXERCISES is None:
```
rtibbles: This code now follows the topic_tree logic above, and a separate exercise_cache is generated on demand for each language.

kalite/topic_tools/__init__.py
```diff
@@ -165,39 +201,43 @@ def create_thumbnail_url(thumbnail):
 
 CONTENT          = None
 CACHE_VARS.append("CONTENT")
-def get_content_cache(force=False, annotate=False):
+def get_content_cache(force=False, annotate=False, language="en"):
     global CONTENT, CONTENT_FILEPATH
```
rtibbles: Following the pattern above, content_cache is now generated on a per language basis.

rtibbles:
This should be complete now, @aronasorman - have made the same changes to get_exercise_cache, and get_content_cache.

kalite/coachreports/views.py
```diff
@@ -192,13 +192,15 @@ def tabular_view(request, report_type="exercise"):
     """Tabular view also gets data server-side."""
     # important for setting the defaults for the coach nav bar
 
+    language = request.language
```
aronasorman: We should convert the language code in here to the right format through `lc_to_django_lang`, so that all the other functions below can just assume that it's been processed already.

kalite/settings.py
aronasorman: looks like this is not meant to be replaced (no `getattr` pattern), so we should just put this on `i18n/__init__.py` instead.

kalite/topic_tools/__init__.py
aronasorman: Interesting, maybe it's time to wrap the `translation.activate/deactivate` calls into a decorator/context manager. Could you make one real quick?

kalite/topic_tools/__init__.py
```diff
@@ -81,26 +82,32 @@ def recurse_nodes(node):
...
                 logging.warn("Annotated topic cache file failed in saving with error {e}".format(e=e))
 
-    return TOPICS[channel]
+    return TOPICS[channel][language]
 
 
 NODE_CACHE = None
 CACHE_VARS.append("NODE_CACHE")
-def get_node_cache(node_type=None, force=False):
+def get_node_cache(node_type=None, force=False, language="en"):
```
aronasorman: what are the negative effects of just setting the default value of the language parameter to `settings.LANGUAGE`?

kalite/settings.py
rtibbles: Righto, will move. settings["KHAN_EXERCISES_RELPATH"] = ?

kalite/topic_tools/__init__.py
rtibbles: Hrm, not sure I understand. The problem I had is that doing this at the whole function level would cause problems, as when the individual content items get translated, the translation gets switched to the best match for them instead (which could mean that the topics translation now switches over to a different language, because the content did not have translation available).

kalite/coachreports/views.py
rtibbles: Right.

kalite/topic_tools/__init__.py
rtibbles: None, should probably change this in every instance!

kalite/settings.py
rtibbles: This seems to be contra-indicated by Django convention:

settings.DEBUG = True   # Don't do this!

kalite/topic_tools/__init__.py
aronasorman: I think it's just idiomatic code in general to wrap patterns that set up and shut down resources into a with statement. In this case, we can easily see where a `translation.activate` applies by seeing the scope of the `with` statement.

rtibbles: Right, makes sense.

kalite/i18n/__init__.py
```diff
@@ -442,3 +444,10 @@ def set_request_language(request, lang_code):
 
     request.language = lcode_to_ietf(lang_code)
     translation.activate(request.language)
+
+@contextmanager
```
rtibbles: @aronasorman Here is the context block for translation. Have used it as a with statement in all relevant places in code.

rtibbles:
@aronasorman the only open question for me on this is the upgrade path. If people already have language packs downloaded, they would need to copy the templates over to the new directory - or should we be forcing them to download the new language pack?

aronasorman:
Good question. Since a lot of our interface UI has changed it might be good to just have them redownload the language packs, so they can get the new strings as well.

kalite/i18n/__init__.py
aronasorman: :+1:

rtibbles:
Items that should now be translated:
Khan-exercises exercise templates
Sidebar topic data
Exercise metadata
Video/Content metadata

aronasorman:
Strange, Chrome is acting up when I try to view the translated sidebar.

rtibbles:
Screenshot?

aronasorman:
By acting up I mean suddenly occupying all my ram and CPU. It also takes quite a while for the sidebar to show the topics.

rtibbles:
Hrm, might be the first translated load of the topics cache file?

aronasorman:
Looking at chrome's devtools, the time for the topic tree API request's first byte to come in is 1.0+ seconds. Client-side caching turned off. Any clue on why it would take this long to respond @rtibbles?

aronasorman:
Anyway, can confirm that the topic tree sidebar is now translated.

rtibbles:
Do you get the same latency on repeated page loads?

aronasorman:
screen shot 2015-01-29 at 4 23 15 pm

topic description isn't translated though

aronasorman:
P.S. look at `Geometrie`, where the topic name is translated but the description isn't.

aronasorman:
Oh strange, it's not consistent: 
screen shot 2015-01-29 at 4 24 42 pm

aronasorman:
Yup, same latency across page loads @rtibbles

aronasorman:
Are the buttons intentionally not translated? 
screen shot 2015-01-29 at 4 26 21 pm

rtibbles:
Those are coming from the {{_ }} translation wrapper in Handlebars. Did we implement that yet?

rtibbles:
I would guess that the inconsistent translation is based on availability? If it's not available, will pass through the description.

aronasorman:
re: handlebars -- looks like we haven't implemented the actual substitution of strings yet. Will make an issue for that.

jamalex:
It *should* be translating Handlebars stuff already, if the strings are in djangojs.po or whatever, and gettext is workng, given: https://github.com/learningequality/ka-lite/blob/develop/python-packages/fle_utils/handlebars/static/js/handlebars/handlebars-helpers.js#L40

jamalex:
We're probably just not (as the comment in there indicates) scraping handlebars strings to djangojs.po yet?

jamalex:
See https://github.com/learningequality/ka-lite/issues/2472

aronasorman:
I think we do pick them up already when we run `makemessages`. So it might be in this case that they're not just translated on crowdin..?

cpauya:
During `makemessages`, we include processing the `*.handlebars` templates and put them on the `djangojs` domain (that's why we made those changes on the `makemessages` codes).  It may be that the strings are not yet translated on crowdin.

alainschaefer:
Dear ka-light devs,

I am delighted you are using de for testing I18N! I do not know where you
are pulling the translations from but those strings in the topic tree are
translated. I just checked on crowdin. Are you loading all translations or
just approved translations ? KA switched recently to only exporting
approved translations. We have only a few weeks ago been given approval
permission from KA. So we are behind with the approval as we are using this
for a careful review of translation quality.

Of course with you guys making so great progress with 18N this will be a
tremendous motivation to immediately focus string approval :-)

Is kitchen updated and working again ? Or shall I switch my own
installation to develop branch to test these exciting developments ?

Alain

aronasorman:
Ah, it might be that some of those strings aren't included in the language pack as we had an old version of the topic tree before, and thus filtered strings that aren't part of that topic tree. I'll look into expanding the strings included.

aronasorman:
@alainschaefer I'm not particularly sure, but I think CrowdIn gives us the approved translations. So I think we'll eventually get the translations, once they're approved and the strings included in the language pack.

As for the kitchen server, it's still open, although it hasn't been updated recently yet: http://kitchen.learningequality.org:1187/

aronasorman:
@rtibbles so we don't have to make the handlebars translation as a blocker for this PR. I think we just need to resolve why it's taking 1+ seconds to return the topic tree, and then we can merge this.

rtibbles:
And also fix the merge conflict.

Issue #2872, opened 2015-01-16
Redirect Loop Issue


kkmach:
Trying to access video downloads page

http://127.0.0.1:8008/securesync/register/?next=%2Fupdate%2Fvideos%2F

More details:

```The webpage at http://127.0.0.1:8008/securesync/register/?next=%2Fupdate%2Fvideos%2F has resulted in too many redirects. Clearing your cookies for this site or allowing third-party cookies may fix the problem. If not, it is possibly a server configuration issue and not a problem with your computer. ```

kkmach:
https://s3.amazonaws.com/uploads.hipchat.com/86198/1615667/xcvNvUwpK64kZ2w/data.sqlite

MCGallaspy:
@kkmach Is this on the develop branch?

kkmach:
no I think Jamie and I found it on the master or central master.

MCGallaspy:
Managed to recreate this issue by judiciously deleting database rows in both the distributed server and the central server. Check out the file:
`/ka-lite-central/ka-lite-submodule/python-packages/securesync/devices/api_views.py`, starting around line 73.

If a device does not have an associated zone in the central server, instead of claiming the device has already been registered let's see if we can fix up the database.

PR #2906, opened 2015-01-28
Fix redirect loop issue in one case

MCGallaspy:
Minimum to reproduce seems to be:
An entry in `securesync_device` but not `securesync_devicezone` corresponding to the client_device (on the central server)
No `securesync_zone` or `securesync_devicezone` in the client database
But so what? If the registration object exists, then we'll just get the zone there and create a new DeviceZone.
There's another redirect loop possible, see comment in api_view.py, however I've been assured
it shouldn't be reachable without some serious database deletions. Disregarding for now to avoid further time sink.

### DIFF
--- a/python-packages/securesync/__init__.py
+++ b/python-packages/securesync/__init__.py
@@ -13,3 +13,13 @@ except:
 
 from .devices.__init__ import *
 from .engine.__init__ import *
+
+# JsonResponseMessageError codes
+class ERROR_CODES:
+    CLIENT_DEVICE_CORRUPTED = "client_device_corrupted"
+    CLIENT_DEVICE_NOT_DEVICE = "client_device_not_device"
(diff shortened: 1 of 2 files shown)


MCGallaspy:
Fixes #2872.

MCGallaspy:
> Is there a toplevel variable somewhere where the values for the code parameter is stored? We should move it to such a variable.

So my first thought is to put the various codes in `pythonpackages/securesync/settings.py`, and then replace uses of raw strings in the securesync package with something like `settings.CODENAME`. It might be difficult to hunt down uses outside of the securesync package. Thoughts?

aronasorman:
I think for now we can focus on just the `securesync` package first, and then migrate code later (although an easy way to hunt these down is just to use `git grep`). Also, it might be good to place these on the securesync `__init__.py` package instead, since these aren't meant to be configured by the user.

PR #2910, opened 2015-01-29
Pressing enter in perseus exercise submits answer

### LINKED ISSUE #2833: Allow Enter key to be used to submit answers for inline Perseus exercises
In Perseus exercises, there is often a text input box in the exercise area (not the answer area). Pressing enter didn't seem to submit the answer.

MCGallaspy:
Add an event listener to elements with the class "perseus-input"
so that it triggers the click event on the check answer button.

Resolves #2833.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -116,9 +116,17 @@ window.ExerciseView = Backbone.View.extend({
     },
 
     events: {
-        "submit .answer-form": "answer_form_submitted"
+        "submit .answer-form": "answer_form_submitted",
+        "keyup .perseus-input": "click_check_answer_button",
+        "keyup #solutionarea>input": "click_check_answer_button"
     },
 
+    click_check_answer_button: function(e) {
+        if(e.keyCode == $.ui.keyCode.ENTER) {
+            $("#check-answer-button").trigger("click");
+        }
+    },
+    
     render: function() {
 
         var data = $.extend(this.data_model.attributes, {test_id: this.options.test_id});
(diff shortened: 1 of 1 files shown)


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -138,6 +138,14 @@ window.ExerciseView = Backbone.View.extend({
 
         // Catch the "next question" button click event -- needs to be explicit (not in "events")
         this.$("#next-question-button").click(this.next_question_clicked);
+        
+        // Workaround so that pressing enter on perseus exercises submits the answer
```
rtibbles: Would be better to trigger this using Backbone event binding:
http://backbonejs.org/#View

(This way it will get cleaned up properly by backbone and we won't have zomebie listeners).

kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -138,6 +138,14 @@ window.ExerciseView = Backbone.View.extend({
 
         // Catch the "next question" button click event -- needs to be explicit (not in "events")
         this.$("#next-question-button").click(this.next_question_clicked);
+        
+        // Workaround so that pressing enter on perseus exercises submits the answer 
+        var selector = $(".perseus-input");
```
rtibbles: Also, I don't think enter is working on non-perseus exercises too, so may be worth double checking this on addition_1 too.

kalite/distributed/static/js/distributed/exercises/views.js
MCGallaspy: Maybe I'm missing something, but there doesn't appear to be a Backbone event corresponding to "keydown": http://backbonejs.org/#Events-catalog

benjaoming: Aren't such events more of a jquery thing?

rtibbles: That's on the Backbone Events object - on a Backbone View you can define an events object that binds certain DOM events to callbacks:
```
  events: {
    "click .icon":          "open",
    "click .button.edit":   "openEditDialog",
    "click .button.delete": "destroy"
  },
```
So you could just add:
```
events: {
     "keydown .perseus-input": "methodnameforprocessingkeypress",
     "keydown #solution-area": "methodnameforprocessingkeypress"
},
```
(Hopefully this second line would then also make it work for non-perseus exercises.)

kalite/distributed/static/js/distributed/exercises/views.js
MCGallaspy: It doesn't working on that now too.

rtibbles:
Looks good. Will manually test.

rtibbles:
Works. Merging.

PR #2915, opened 2015-01-30
Replace time.wait with more robust WebDriverWait

MCGallaspy:
Which will throw a TimeoutException instead of quietly
failing. Also generously increased the wait time.


MCGallaspy:
This PR should fix some failing tests in the central server repo, or at least unambiguously identify a timeout as the issue.

PR #2917, opened 2015-01-30
Software keyboard fix

### LINKED ISSUE #2907: Software keyboard submits answer when keys are clicked
The software keyboard keys in exercises seem to do the same thing as the check answer button when clicked. @MCGallaspy [gets a `TypeError`](https://github.com/learningequality/ka-lite/pull/2897#issuecomment-72080753) on click. I am guessing this is related.

I can give this one a shot. I'll have to dig through and see where the clicks are handled.

EDIT: I also get the same error. I didn't notice it earlier. It does seem to submit the current answer.

rtibbles:
This fixes #2907, but only for Khan-Exercises.

Now also fixed for Perseus.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -176,7 +176,7 @@ window.ExerciseView = Backbone.View.extend({
                         el: self.$("#software-keyboard-container")
                     });
                 }
-
+                self.software_keyboard_view.set_input("#solutionarea");
                 self.software_keyboard_view.show();
             } else if (typeof self.software_keyboard_view !== "undefined") {
                 self.software_keyboard_view.hide();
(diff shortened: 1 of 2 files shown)


MCGallaspy:
Just testing it out, seems like pressing enter or backspace has unexpected results when the keypad is activated both in Khan exercises and in Perseus exercises. When I tried the keypad in http://127.0.0.1:8008/learn/khan/math/early-math/cc-early-math-counting-topic/cc-early-math-counting/count-to-100/, after entering the answer and clicking "Check Answer" button I get the prompt "There are still more parts of this question to answer." The button works fine for Khan exercises.

rtibbles:
Dammit, that was the behaviour I had spent the time debugging to fix. Will look at it again.

rtibbles:
For 'enter' interactions - I can do a catch on all the buttons to trigger a submit answer, sound good?

MCGallaspy:
Just kidding, I hadn't pulled your latest branch... Sorry.

rtibbles:
Oh thank god...

MCGallaspy:
But regarding enter/backspace interactions, that's still extant.

MCGallaspy:
So maybe if you just add an event listener for "keyup" to the SoftwareKeyboardView to catch enter/backspace... I'm just not sure which element the event will be attached to.

rtibbles:
This should catch enter and pass it through to 'check answer'.

PR #2934, opened 2015-02-03
Replace user-facing uses of teacher/student with coach/learner

MCGallaspy:
Only changed user-facing strings in wrapped in gettext calls and django templte trans blocks. So, for example, logs and tests still use the teacher/student terminology.


MCGallaspy:
Someone else wanna pull this and run the tests for me? Just for grins and giggles?

aronasorman:
Tests pass. Let it be merged!

PR #2936, opened 2015-02-03
Resolves windows path to kalite command requires Windows XP SP2 or above...

EdDixon:
....


MCGallaspy:
@EdDixon I'm just going to test this out on my own windows machine real quick. If it puts kalite.bat into my path, and persists it for any given instance of cmd.exe (and after a reboot), then I think we can merge this. Tests are failing for what I would say are clearly unrelated issues.

MCGallaspy:
@EdDixon That is of course if you can vouch for it on other versions of windows -- presumably XP SP2 and up, right?

MCGallaspy:
Path is set correctly, but I don't have autocompletion for the `kalite` command in cmd.exe... any thoughts on that @ed @ruimalheiro?

PR #2938, opened 2015-02-03
Topic tree fixes

### LINKED ISSUE #2930: Issues with JSON data after latest contentload (khan-api?)
- The Perseus exercises no longer show up in the topic tree (e.g. `grep brunelleschi-and-ghiberti-s-sacrifice-of-isaac-quiz data/khan/topics.json`)
- The assessmentitem data has all been denormed into exercises.json (e.g. `grep content data/khan/exercises.json`). Doesn't seem to be coming from KA's api itself: http://api.khanacademy.org/api/v1/exercises/brunelleschi-and-ghiberti-s-sacrifice-of-isaac-quiz

rtibbles:
This fixes #2930.

Summary of changes:
* Stupidly inflated exercises.json is fixed by fixing the toJSON method in Khan-API-PY
* Perseus exercises missing from topictree is inexplicable, but they were there when I ran contentload again.

### DIFF
--- a/python-packages/khan_api_python/api_models.py
+++ b/python-packages/khan_api_python/api_models.py
@@ -164,27 +164,25 @@ class APIModel(AttrDict):
         self._loaded = True
 
     def toJSON(self):
-        output = copy.copy(self)
-        for key in output._related_field_types.keys() + output._lazy_related_field_types.keys():
-            if output.get(key, None):
-                if isinstance(output[key], APIModel):
-                    output[key] = output[key].toJSON()
-                elif isinstance(output[key], dict):
-                    output[key] = json.dumps(output[key])
-                elif isinstance(output[key], list):
-                    for i, item in enumerate(output[key]):
-                        if isinstance(output[key][i], APIModel):
-                            output[key][i] = output[key][i].toJSON()
-                        elif isinstance(output[key][i], dict):
-                            output[key][i] = json.dumps(output[key][i])
-        deletekeys = []
-        for key in output:
-            if key.startswith("_"):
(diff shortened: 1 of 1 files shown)


python-packages/khan_api_python/api_models.py
```diff
@@ -164,28 +164,26 @@ def fetch(self):
...
+                    output[key] = self[key].toJSON()
+                elif isinstance(self[key], dict):
+                    output[key] = json.dumps(self[key])
+                elif isinstance(self[key], list):
+                    output[key] = []
+                    for i, item in enumerate(self[key]):
+                        if isinstance(self[key][i], APIModel):
+                            output[key].append(self[key][i].toJSON())
+                        elif isinstance(self[key][i], dict):
+                            output[key].append(json.dumps(self[key][i]))
```
jamalex: Any way this could be made recursive, or simpler -- feels like lines 170-173 and 177-180 may be covering similar ground?

jamalex:
Woohoo! Tests finally passed.

Added a comment about potential simplification, but I'm going to merge and run this now. Thanks!

python-packages/khan_api_python/api_models.py
rtibbles: Yeah, it does feel very similar. The only way I could imagine making it less repetitive would be to put this code block:
```
                if isinstance(self[key], APIModel):
                    output[key] = self[key].toJSON()
                elif isinstance(self[key], dict):
                    output[key] = json.dumps(self[key])
```
in another function and then call it, returning none if neither conditions pertained.

PR #2941, opened 2015-02-05
Allow translation of Khan channel data

rtibbles:
Allows translation of the channel data that populates the footer and home page.

@aronasorman

### DIFF
--- a/kalite/distributed/templates/distributed/partials/khan_dummy_channel_translation.html
+++ b/kalite/distributed/templates/distributed/partials/khan_dummy_channel_translation.html
@@ -3,5 +3,4 @@
 {% trans "KA Lite" %}
 {% trans "A free world-class education for anyone anywhere." %}
 {% trans "KA Lite is a lightweight web server for viewing and interacting with core Khan Academy content (videos and exercises) without needing an Internet connection." %}
-{% trans "CC-BY-NC-SA" %}
 {% trans "Videos and Exercises © 2014 Khan Academy" %}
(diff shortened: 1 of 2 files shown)


kalite/distributed/templates/distributed/partials/khan_dummy_channel_translation.html
```diff
@@ -0,0 +1,7 @@
+<!-- TODO BLOCKER: 0.14 (rtibbles) this is dummy text to pull in Khan Academy channel data for translation. This is not ideal !-->
+
+{% trans "KA Lite" %}
+{% trans "A free world-class education for anyone anywhere." %}
+{% trans "KA Lite is a lightweight web server for viewing and interacting with core Khan Academy content (videos and exercises) without needing an Internet connection." %}
+{% trans "CC-BY-NC-SA" %}
```
aronasorman: Do we really want this translated? I'd imagine this set of strings to indicate the type of CC license is standard everywhere, and it would just be confusing if it's translated.

rtibbles: Had thought that it might get translated, but judging by the Farsi version of the CC-BY-NC-SA page, I guess not!
https://creativecommons.org/licenses/by-nc-nd/3.0/us/deed.fa

rtibbles:
Hopefully with :strawberry: we will include the license descriptions, and have the translated versions in the language packs, so that we can just reference those licenses.

PR #2945, opened 2015-02-05
Checkbox Autocomplete Off

### LINKED ISSUE #2929: Students section not highlighting properly after attempting to change groups
*Expected Behavior*: When a student is selected, a checkmark should appear in the box, and the student should be highlighted purple.

*Current Behavior*: A checkmark appears, but the highlighting is not in sync with the checkmark. Example: if a student was checked and then unchecked, the student will be highlighted, but there will be no checkmark in the box.

Steps to reproduce:
1. Select any student(s), and then attempt to move them into another group.
2. The page will refresh.. select/deselect the students and then you will see the issue.

rtibbles:
Fixes #2929 

Summary of changes:
* Adds 'autocomplete="off"' to checkbox type inputs to prevent students from being checked (but not highlighted) on page refresh.
* Regression test for bug.

### DIFF
--- a/kalite/control_panel/tests/control_panel.py
+++ b/kalite/control_panel/tests/control_panel.py
@@ -108,6 +108,24 @@ class GroupControlTests(FacilityMixins,
         with self.assertRaises(NoSuchElementException):
             self.browser.find_element_by_xpath('//button[@id="delete-coaches"]')
 
+    def test_checkbox_not_autocompleted(self):
+        # This is a regression test for issue #2929
+        # Firefox aggressively autocompletes form elements, meaning if a checkbox is checked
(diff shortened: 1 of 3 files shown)


kalite/control_panel/tests/control_panel.py
```diff
@@ -108,6 +108,24 @@ def test_teachers_have_no_delete_coaches_button(self):
         with self.assertRaises(NoSuchElementException):
             self.browser.find_element_by_xpath('//button[@id="delete-coaches"]')
 
+    def test_checkbox_not_autocompleted(self):
```
aronasorman: :+1:

PR #3009, opened 2015-02-10
Coverage.io first stab, will it work?

benjaoming:
Sorry if it doesn't work the first time.. I would kind of expect that.


aronasorman:
Why is it passing though? Does this mean we have 100% code coverage??

benjaoming:
Yes, it means we have 100% coverage. What else were you expecting? This Coverage is just to get the shiny badge, didn't actually wanna run it :)

benjaoming:
Closing because now it looks pretty messed up

Issue #2919, opened 2015-01-30
`kalite` command remaining issues


benjaoming:
Recall the task list from #2893 

- [ ] Testing on Win
- [x] Testing on OSX
- [ ] Testing on Android
- [x] Testing on Linux (Ubuntu)
- [x] Test case for .travis
- [x] Bat script for Windows, currently do: `python.exe bin/kalite`
- [ ] Resolve how implementation should work with service / task / rc.d etc systems.
- [x] Document SHELL environment variables
- [ ] Add a `kalite reload` command (not gonna call it setup, because then there's 4 commands beginning with *S*?)
- [ ] Add deprecation messages when running deprecated commands
- [ ] Update documentation
- [ ] Add a --port option to start and status
- [x] Do not fail stop or restart with "Not Responding", just kill process
- [x] Spontaneous occurrences of 'unknown command kaserve' and 'unknown command cronserver'

benjaoming:
@aronasorman this does not warrant the ready label yet :)

benjaoming:
I'm having an issue that the `kaserve` and `cronserver` commands are randomly missing. Traceback can look like this:

    ➜  ka-lite git:(kalite-command-fixes) ✗ bin/kalite restart
    kalite stopped
    Unknown command: 'kaserve'
    Type 'kalite manage help' for usage.

benjaoming:
It may have something to do with Django not supporting true concurrency in django.core.management.get_commands() and thus we need to force a call to this function to build the list of commands before trying to execute any. Otherwise the first process will start building the list of commands, and the others will access it before it's done being built.

benjaoming:
Yup, it was a problem of running two Django commands concurrently without a call to get_commands() first. Fixed in aede7dd

aronasorman:
Oh, that was placed automatically by waffle.io whenever I move something to "this week" :).

jamalex:
Love what I'm seeing!

Originally, I had been thinking the `kalite status` command would show more detail, but I can see the utility of keeping it simple like it is, and putting these other details into a separate command, like `kalite info`. This would include things like:

- [] The IP addresses at which KA Lite (if running) is listening (this could then be used for the GUI; see https://github.com/learningequality/installers/issues/6#issuecomment-73297876)
- [] The current version (number and hash) of KA Lite
- [] The device name and ID, and registration status (including Zone info)
- [] How much disk space is being used by KA Lite (and, when possible, how much is free)

[...]

PR #3020, opened 2015-02-10
Fixes for #2919 Chapter 3

benjaoming:
#2919

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -219,7 +219,7 @@ def get_pid():
     listen_port = getattr(settings, "CHERRYPY_PORT", LISTEN_PORT)
 
     # Timeout is 1 second, we don't want the status command to be slow
-    conn = httplib.HTTPConnection(LISTEN_ADDRESS, listen_port, timeout=1)
+    conn = httplib.HTTPConnection(LISTEN_ADDRESS, listen_port, timeout=3)
     try:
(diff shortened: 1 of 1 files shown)


kalitectl.py
```diff
@@ -219,7 +219,7 @@ def get_pid():
     listen_port = getattr(settings, "CHERRYPY_PORT", LISTEN_PORT)
 
     # Timeout is 1 second, we don't want the status command to be slow
-    conn = httplib.HTTPConnection(LISTEN_ADDRESS, listen_port, timeout=1)
+    conn = httplib.HTTPConnection(LISTEN_ADDRESS, listen_port, timeout=3)
```
jamalex: The first time the server starts up, on Pi's and similar servers, it can take > 15 seconds to load the first page, then it's fast(er) after that. But I think it may be possible to connect to the port well before that, so hopefully this will be sufficient.

benjaoming: I don't know how the view that serves the PID request (PING_URL below) goes about its things, but it would be worth trying to optimize this view in all cases.... however, having a sequence like this is not terrible:

Status: Starting up
Status: Not responding
Status: Running (OK)

I mean, the not responding status is mainly for humans to make sense of, so they can hopefully bear with the Pi being a bit slow about things :)

PR #3021, opened 2015-02-10
Put a default title on main landing page and remove invalid line breaks

benjaoming:


kalite/distributed/templates/distributed/base.html
```diff
@@ -13,7 +13,7 @@
         <meta name="wot-verification" content="075abe861486e6bb3ab9"/>
         <meta name="viewport" content="width=device-width, initial-scale=1">
         {% block titletag %}
-        <title>{% block title %}{% trans title %}{% endblock title %} | KA Lite</title>
+        <title>{% block title %}{% trans "Welcome" %}{% trans title %}{% endblock title %} | KA Lite</title>
```
MCGallaspy: Why did you add the Welcome here? Does that create a grammatically correct title?

benjaoming: It's a default so that pages without a title have something rather than " | KA Lite"

jamalex: IT'S @BENJAOMING! HE SURVIVED HIS TRIP!!!!!

MCGallaspy:
Oh I see, it's overridden by templates that inherit from this.

kalite/distributed/templates/distributed/base.html
benjaoming: No no, this is the us government, we just maintain his online identity while he's gone... it's a new service :)

aronasorman: Cool, is this new service webscale?

Issue #3000, opened 2015-02-09
Coach has no logout option in username tab


MCGallaspy:
On `develop`, logged in as a coach, I don't see a logout option... :(
I'm trapped:
![ss](https://cloud.githubusercontent.com/assets/8888020/6118130/4d178d6c-b06f-11e4-9e20-31a51dd6879e.png)

MCGallaspy:
Weirdly, this issue seems to be due to `ka-lite/kalite/distributed/static/js/distributed/distributed-server.js`, [line 89](https://github.com/learningequality/ka-lite/blob/develop/kalite/distributed/static/js/distributed/distributed-server.js#L89) or thereabouts. This callback function is never (or rarely?) executed for coaches. The weird part is that the get request made as a result of that line receives a response (200) just fine when logged in as coach, it's just the callback that's not executed. I even tried the more explicit:
```js
        this.loaded = this.fetch({ success: this.after_loading,
                                   error: function(){console.log("banana slugs")}  });
```
in which case neither callback is executed when logged in as a coach. Any thoughts @rtibbles?

PR #3024, opened 2015-02-11
Set statusModel points attributes default for coaches

MCGallaspy:
After `fetch`ing `window.statusModel` returns among other
things a value of `null` for `points` when logged in as a
coach. In some sort of tangled javascript fashion, this prevented
the `success` callback of the `StausModel.fetch` method from being
executed.

Solves #3000.

### DIFF
--- a/kalite/distributed/tests/browser_tests/distributed.py
+++ b/kalite/distributed/tests/browser_tests/distributed.py
@@ -453,12 +453,10 @@ class PointsDisplayUpdatesCorrectlyTest(KALiteBrowserTestCase, BrowserActionMixi
         self.browser.execute_script(log_model_object + ".set(\"points\", 9000);" )
         self.browser.execute_script(log_model_object + ".saveNow();" )
 
-class CoachHasLogoutLinkTest(KALiteBrowserTestCase, BrowserActionMixins, CreateAdminMixin, FacilityMixins):
+class CoachHasLogoutLinkTest(BrowserActionMixins, CreateAdminMixin, FacilityMixins, KALiteBrowserTestCase):
     """
     A regression test for issue 3000. Note the judicious use of waits and expected conditions to account for
-    various browser sizes and potential server hiccups. Even though no TimeoutException or NoSuchElementException
-    is expected, I still catch it so I can have my say, dangit, given that the test server likes to throw these
-    for apparently nondeterministic reasons sometimes.
+    various browser sizes and potential server hiccups. 
     """
 
     def setUp(self):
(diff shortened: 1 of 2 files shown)


MCGallaspy:
I'm gonna take my own medicine and write a test for this, so don't merge it yet folks.

MCGallaspy:
Someone wanna review this for me? @aronasorman?

aronasorman:
Will review today.

kalite/distributed/tests/browser_tests/distributed.py
```diff
@@ -453,3 +453,43 @@ def _play_video(self):
         self.browser.execute_script(log_model_object + ".set(\"points\", 9000);" )
         self.browser.execute_script(log_model_object + ".saveNow();" )
 
+class CoachHasLogoutLinkTest(KALiteBrowserTestCase, BrowserActionMixins, CreateAdminMixin, FacilityMixins):
```
aronasorman: BrowserTestCase should go last in the inheritance hierarchy.

kalite/distributed/tests/browser_tests/distributed.py
```diff
@@ -453,3 +453,43 @@ def _play_video(self):
...
+                expand_menus_button = self.browser.find_element_by_xpath("//*[@id=\"wrapper\"]/div[1]/div/div/div[1]/button")
+                self.browser_activate_element(elem=expand_menus_button)
+                # Wait for the animation to finish
+                WebDriverWait(self.browser, 3).until(
+                    expected_conditions.visibility_of(dropdown_menu)
+                )
+                self.browser_activate_element(elem=dropdown_menu)
+            self.assertTrue(nav_logout.is_displayed(), "The dropdown menu logout item is not displayed!")
+        except NoSuchElementException as e:
+            self.assertTrue(False, "Test raised a NoSuchElementException... probably an issue with `setUp` method? Exception: %s" % repr(e))
```
aronasorman: Strange idiom going on here. You're gonna raise an exception all the time with the `self.assertTrue(False)` call, yet you're reraising the error again with the `raise` call? What's the intent with these two except clauses?

aronasorman:
If I understand how you wrote your tests correctly, you're trying to make it more robust by catching unrelated exceptions and then trying to fix them. Is that correct?

kalite/distributed/tests/browser_tests/distributed.py
MCGallaspy: Indeed, that's a questionable decision, which is why we have code reviews I suppose. My intent was to offer some helpful advice on why the test might be throwing unexpected exceptions, but I guess it's better just to let them get thrown normally. The trailing `raise` was just an oversight and really by `self.assertTrue(False,...` I should have used `self.fail(...`, but I guess I had a brain fart or something.

MCGallaspy:
I think it's best if I just remove the outermost `try` block and don't attempt to catch unexpected exceptions.

aronasorman:
Nitpicks picked. Tested. Working. Merging.

Issue #3002, opened 2015-02-09
After registering, the "Register now" buttons etc still show up, until next login


jamalex:
Expected: After registering, the interface should no longer show links/instructions to "register now".

Observed: 
After registering the Manage tab still looks like:
![image](https://cloud.githubusercontent.com/assets/618416/6118363/9a8a3b7e-b071-11e4-96e2-9da19afde67d.png)

But if you click the register now button, it shows:
![image](https://cloud.githubusercontent.com/assets/618416/6118366/b4e18720-b071-11e4-964c-d36f4e9a6078.png)

Note that once you logout and login again, the registration status is properly indicated.

Steps to replicate:
- Register
- Go to Manage tab

arceduardvincent:
Hi @jamalex and @aronasorman .
As I examine and replicate this issue I found out this is not a bug but it is a feature that @anuragkanungo made.

Details:
If the boolean statement returns true `is_config_package_nalanda` which declared in `distributed/custom_contex_processor.py` and the `CONFIG_PACKAGE` use is 'Nalanda' the button will appear.

Conclusion:
 @anuragkanungo specify if the device used for Nalanda package only and hide update software for non Nalanda. 

Question:
Does this feature is a bug?
Does we allow the non nalanda user to update software?

Thanks.

arceduardvincent:
I use a `Nalanda` package see attach image below.
![screen shot 2015-02-10 at 11 31 42 am](https://cloud.githubusercontent.com/assets/4099119/6121133/9d9a18c8-b118-11e4-9207-92d9b6e57f57.png)

jamalex:
This issue doesn't have anything to do with Nalanda in particular. It's that the "Register" button should not appear anymore after the server has already been registered. It's an error in the code that checks the registration status, to know whether or not to show the button.

The "Update" button (unrelated to this issue) should only show up for Nalanda, yeah.

arceduardvincent:
hi @jamalex 
thanks for the fast response,

arceduardvincent:
@jamalex I think this issue is done. The fix is merged #3026 .

PR #3026, opened 2015-02-11
Hide button when the device is registered.

arceduardvincent:
Hi @aronasorman this fix this issue #3002

Fix: If the device is registered hide the register button else show the button for unregistered device.
Reference: you can identify if the device is registered or not by device.is_registered() method. 

you can see the image below the button is not showing if the device is already registered.
screen shot 2015-02-11 at 12 11 31 pm

### DIFF
--- a/kalite/control_panel/tests/control_panel.py
+++ b/kalite/control_panel/tests/control_panel.py
@@ -8,6 +8,12 @@ from selenium.common.exceptions import NoSuchElementException
 from kalite.testing.base import KALiteBrowserTestCase, KALiteClientTestCase, KALiteTestCase
 from kalite.testing.mixins import BrowserActionMixins, FacilityMixins, CreateZoneMixin, StudentProgressMixin, CreateAdminMixin, StoreMixins
 
+from selenium.webdriver.common.by import By
+from selenium.common.exceptions import TimeoutException
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+import time
+
 logging = settings.LOG
 
 
@@ -34,10 +40,12 @@ class DeviceRegistrationTests(FacilityMixins,
         self.fac = self.create_facility(name=facility_name)
         self.browser_login_admin(**self.admin_data)
         self.browse_to(self.reverse('zone_redirect'))  # zone_redirect so it will bring us to the right zone
-
-        device_status = self.browser.find_elements_by_class_name("not-registered-only")
-        expected_result = "Register Now!"
-        self.assertEqual(device_status[0].text, expected_result)
+        element = self.browser.find_element_by_id('not-registered')
+        try:
+             WebDriverWait(self.browser, 0.7).until(EC.visibility_of(element))
+        except TimeoutException:
+             pass
+        self.assertTrue(element.is_displayed())
 
 
     def test_device_already_register(self):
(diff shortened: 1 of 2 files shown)


MCGallaspy:
I think it would be good to write a test for this one.

arceduardvincent:
Ok I will make a test. Thanks for the remind

arceduardvincent:
Hello @MCGallaspy 
I have already done the test.
The test cover the possible scenario below.

Scenario:
> 1 If the device is not yet registered the button `Register Now!` must show else the test will fail.
> 1 if the device is registered the link `Sync Now!` must show else the test will fail.

kalite/control_panel/tests/control_panel.py
```diff
@@ -11,6 +11,51 @@
...
+        The Registration button must appear else the test will fail.
+        """
+        facility_name = 'default'
+        self.fac = self.create_facility(name=facility_name)
+        self.browser_login_admin(**self.admin_data)
+        self.browse_to(self.reverse('zone_redirect'))  # zone_redirect so it will bring us to the right zone
+
+        device_status = self.browser.find_elements_by_class_name("not-registered-only")
+        expected_result = "Register Now!"
+        self.assertEqual(device_status[0].text, expected_result)
```
MCGallaspy: Looks like both elements (meaning those with class `not-registered-only` and `registered-only` are included in the page, it's just the visibility that's set depending on the registration status. So I'd expect the assertions in this test and the next one to be true independent of whether the device is registered or not, e.g. the element will be in the DOM and will have that text, it just won't be visible. Further, my experience so far is leading me to believe that `find` methods in Selenium without an implicit or explicit wait can cause trouble.

Therefore, I suggest using [WebDriverWait.until]() with the [visibilit_of](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.support.expected_conditions.visibility_of) expected condition, along with a `try` block, something like this:

```python
try:
    WebDriverWait(self.browser, some_timeout).until(EC.visibility_of(some_element))
except TimeoutException:    # until will throw this exception if the element is not visible
    # Fail the test
# Otherwise pass the test.
```

kalite/control_panel/tests/control_panel.py
```diff
@@ -11,6 +11,51 @@
...
+        This test will simulate the device that has already registered and the only option is available is update
+        the software else the test will fail.
+        """
+        facility_name = 'default'
+        self.zone = self.create_zone()
+        self.device_zone = self.create_device_zone(self.zone)
+        self.fac = self.create_facility(name=facility_name)
+        self.browser_login_admin(**self.admin_data)
+        self.browse_to(self.reverse('zone_redirect'))  # zone_redirect so it will bring us to the right zone
+        device_status = self.browser.find_elements_by_class_name("registered-only")
```
MCGallaspy: Above comment applies here, too.

arceduardvincent: Hi @MCGallaspy .
Thanks for the wonder tips you give and I will take note that for my TDD.
appreciate it alot.

arceduardvincent:
Hi @MCGallaspy 

I update the test that check the visibility of the web element `Register Now!` and `Sync Now!`

MCGallaspy:
Since I can't merge this automatically anyway, I'll just remove the unused import myself.

arceduardvincent:
Thank @MCGallaspy

PR #3027, opened 2015-02-11
add html div tag at register devise button to align it on center.

### LINKED ISSUE #2972: Misaligned "Register device" button on admin UI
screen shot 2015-02-09 at 2 24 36 pm

I have attempted to fix but never could. Time to open an issue!

amodia:
Hi @aronasorman this fixed #2972 

 I attached screenshot.
screen shot 2015-02-11 at 11 15 35 am

### DIFF
--- a/kalite/control_panel/templates/control_panel/partials/_device_table.html
+++ b/kalite/control_panel/templates/control_panel/partials/_device_table.html
@@ -63,6 +63,7 @@
                                     {% endif %}
                                 </td>
                                 <td>
+                                    <div style="margin-top: -24px;">
                                     {% if device.is_own_device %}
                                         <a href="#" id="force-sync" class="registered-only">
                                             {% trans "Sync Now!" %}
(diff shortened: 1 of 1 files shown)


kalite/control_panel/templates/control_panel/partials/_device_table.html
```diff
@@ -63,6 +63,7 @@
                                     {% endif %}
                                 </td>
                                 <td>
+                                    <div style="margin-top: -24px;">
```
MCGallaspy: We should avoid inline styling like this. Can you put this in a stylesheet somewhere?

amodia:
Hi @MCGallaspy  i realize  the `div tag` is not necessary there because the one causing the space at the top of the register button is the `html br tag `so i remove it and its fixed the issue. 

i attached a screenshot. 
screen shot 2015-02-12 at 12 23 23 pm

MCGallaspy:
Looks good to me.

Issue #2987, opened 2015-02-09
Inconsistent behaviour of dropdown item selection on coach report pages


jamalex:
Expected: Unified, consistent experience.

Observed:

The top dropdowns require a button press in order to have an effect:
![image](https://cloud.githubusercontent.com/assets/618416/6117638/cdde8b4e-b06a-11e4-8c0a-4dd4b5f03d6e.png)

Whereas the bottom dropdowns trigger a reload immediately upon item selection:
![image](https://cloud.githubusercontent.com/assets/618416/6117658/f4f84170-b06a-11e4-9454-44c57cc3c8f8.png)

Recommended:
Have all dropdowns require a button before having an effect. Side effects of selecting items (especially when those side-effects may be expensive/slow) isn't ideal.

jesumer:
Hi, 

I have already fixed this issue.
I've added a button that refreshing the browser according to to the selected item. Heres the screenshot
![screen shot 2015-02-11 at 2 57 35 pm](https://cloud.githubusercontent.com/assets/2810009/6143138/05eaf5f6-b1fe-11e4-8c9b-87feb38fa4e0.png)

PR #3030, opened 2015-02-11
Inconsistent behavior of dropdown item selection on coach report pages f...

jesumer:
Hi @aronasorman , 

This is for issue #2987.
I've added a button that refreshing the browser according to to the selected item. Heres the screenshot
screen shot 2015-02-11 at 2 57 35 pm

/review: @arceduardvincent

### DIFF
--- a/kalite/coachreports/static/css/coachreports/base.css
+++ b/kalite/coachreports/static/css/coachreports/base.css
@@ -7,3 +7,7 @@
 .container-margin-bottom {
     margin-bottom: 50px;
 }
+
+#tabularGo {
+    margin-top: 12px;
+}
--- a/kalite/coachreports/templates/coachreports/tabular_view.html
+++ b/kalite/coachreports/templates/coachreports/tabular_view.html
@@ -23,7 +23,8 @@
             // navigation event in order to produce the new report.
 
             $("#report_type").change(function(){
-                window.location.href="{% url 'tabular_view' %}"+$("#report_type option:selected").val()+ "/" + window.location.search;
+                var url ="{% url 'tabular_view' %}"+$("#report_type option:selected").val()+ "/" + window.location.search;
+                $("#tabularGo").attr("href", url);
             });
 
             $("#student").change(function(){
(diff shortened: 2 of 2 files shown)


arceduardvincent:
Hi @jesumer .
as I inspect the the behaviour that @jamalex talk about, your fix comply the recommendation said.
however could you rename the button id `tabularGo`.

I think it is beneficial that we use generic terms like `display-topic-report`something like that. hence our objective is to display the report under topic selected.

MCGallaspy:
This is problematic, in that the order in which the topic and report dropdowns are selected can cause the `href` value of the `#tabularGo` button to be invalid. For example, selecting a topic and then a report type will result in an invalid URL. Check out how it's handled in handlebars in `ka-lite/kalite/coachreports/static/js/coachreports/base.js` in the `NavigationContainerView` and write a new view there that handles it similarly. Then create a new handlebars template and include it in `tabular_view.html` by attaching your view to an element in there. You can check out how it's done in `ka-lite/kalite/coachreports/templates/coachreports/base.html` for an example.

I also agree @arceduardvincent about the id name.

jesumer:
Hi @MCGallaspy 

My findings was you are right about the invalid url it will resetting to default after selecting first the topic before the report  type. Now I was trying your suggestion to fix this and trying the best way of fixing this. I'm currently working on this.

jesumer:
HI @MCGallaspy ,

My fix on this was as is and changed the id of the button except someone will help and can make your suggestion because it will take me sometime to do that. I need to move into other priority issues.

jesumer:
There's no error here and just a url reset if the user selects the topics first like:
screen shot 2015-02-13 at 11 02 31 am
the user will redirect to this page

screen shot 2015-02-13 at 10 59 33 am
 So if the report type will be the last to select the url will reset as of what is the default url of the report type.

arceduardvincent:
Hi @MCGallaspy 
I would like to work on this issue.

MCGallaspy:
@jesumer you're right, my suggested changes are more involved than I thought. Probably best to just make some quick changes to the template js in this case. Also, did you forget to push some changes that you made here? Seems like something is missing.

MCGallaspy:
If [this](https://github.com/mrpau/ka-lite/pull/3) is merged, it'll automatically be reflected here.

MCGallaspy:
And we can get this PR rolling.

Issue #3004, opened 2015-02-09
Messages never go away when navigating through content in the sidebar


jamalex:
Context: Navigating through content in the sidebar happens entirely in the client-side, with no page reloads.

Expected behaviour: Navigating to a new content item should clear and no longer display messages that came from interactions with other content items, or from initially logging in, e.g.:
![image](https://cloud.githubusercontent.com/assets/618416/6118514/296e5950-b073-11e4-8695-5d57054d061b.png)

Possible solution: Clear all messages prior to loading a new content item.

rtibbles:
Fixed in #3056

PR #3031, opened 2015-02-11
Hotfix for issue: 3004 messages never go away when navigating through content in the sidebar

djallado:
@aronasorman.

This fixes the issue: https://github.com/learningequality/ka-lite/issues/3004, Messages never go away when navigating through content in the sidebars.

/review: @jesumer.

### DIFF
--- a/static-libraries/js/khan-lite.js
+++ b/static-libraries/js/khan-lite.js
@@ -251,11 +251,6 @@ function clear_messages(msg_type) {
     return $("#message_container");
 }
 
-// To clear all messages including django and ajax.
-function clearAllMessages(msg_type) {
-    $(".alert").remove();
-}
-
 function get_message(msg_id) {
     return $("#" + msg_id).text();
 }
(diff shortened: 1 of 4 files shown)


static-libraries/js/khan-lite.js
```diff
@@ -251,6 +251,11 @@ function clear_messages(msg_type) {
     return $("#message_container");
 }
 
+// To clear all messages including django and ajax.
+function clearAllMessages(msg_type) {
```
MCGallaspy: Why does this function take a parameter that you don't use?

MCGallaspy:
Seems fine. Looks like you incorporated changes from a commit that I made other comments on, so you should address those too in this PR.

Otherwise, just make a test... you can simulate client side navigation with selenium by firing off the appropriate js event with `execute_script`. A minimal test would be something like
``` python
# Login and let yourself be redirected so you have the "login" alert visible
# Assert the alert is visible
self.browser.execute_script("channel_router.control_view.topic_node_view.content_view.show_view()") # This function is called by a client-side navigation event
# Assert the alert is no longer visible on the page.
```

Issue #3001, opened 2015-02-09
Useless FAQ link on videos update UI


Antrikshy:
1. Go to `/update/videos` as admin. 
2. See link "Q: How do I bulk-download videos?" under "Help for downloading videos".
3. Click it.
4. Notice that there is no such question on that page (as far as I can see).
5. ?????
6. PROFIT

cpauya:
Let's add something like "Q: How do I bulk-download videos or download videos via Torrent?".  

I think that'll make sense than the original text.

Antrikshy:
Or simply "Frequently Asked Questions"...

jesumer:
Hi @aronasorman ,

Please close this issue. This issue has a PR that was merged and closed by @MCGallaspy.

PR #3032, opened 2015-02-11
Text issue was fixed by changing into "How do I bulk-download videos or ...

jesumer:
Hi @aronasorman,

Fix issue #3001.
Changing text to "How do I bulk-download videos or download videos via Torrent?".

/review: @amodia


MCGallaspy:
I'm okay with this, especially seeing as how the link takes you directly to a relevant question (even if they're worded somewhat differently). End user would have to be somewhat dense to misconstrue this.

Issue #2984, opened 2015-02-09
Inconsistent placement of "No learner accounts in this group have been created." message


jamalex:
Expected: The coach reports have been unified, with switching between them done via a single dropdown. The experience should be as consistent across them as possible.

Observed:

On the tabular coach reports, empty groups are indicated in the lower area of the report:
![image](https://cloud.githubusercontent.com/assets/618416/6117550/2bdeda10-b06a-11e4-80f1-34be2a7a2387.png)

On the other reports, however, this message is displayed as a bubble error message, above everything:
![image](https://cloud.githubusercontent.com/assets/618416/6117565/49d55d8c-b06a-11e4-82d4-384958ed73d8.png)

rtibbles:
Fixed in #3036

PR #3036, opened 2015-02-12
Fix to show error messages as bubble on tabular coach reports.

cpauya:
Hi @aronasorman - this is for #2984.

It now shows the error message at the top as bubble, see screenshot.

screenshot 2015-02-12 15 11 58


MCGallaspy:
The tests are failing because they're looking for the error message you removed (by id `#error_message`), and one because `global name fail is not defined` which was probably meant to be `self.fail`. Modify the tests so that they assert the bubble error is visible instead (in `/ka-lite/kalite/distributed/tests/browser_tests/coachreports.py`). You'll probably want to assert the visibility of an element with class `alert-danger` and with the expected error text. Be sure to use appropriate Selenium expected conditions like [text_to_be_present_in_element](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.support.expected_conditions.text_to_be_present_in_element) in conjunction with a [WebDriverWait.until](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.support.wait.WebDriverWait.until). (And replace those instances of `fail` with `self.fail` while you're in there...)

cpauya:
Thanks for the pointers, working on these now. :smile_cat:

cpauya:
Yey - tests pass!  Please review and merge @aronasorman / @MCGallaspy.

kalite/distributed/tests/browser_tests/coachreports.py
```diff
@@ -23,71 +21,100 @@ def setUp(self):
...
+
     def test_no_groups_no_topic_selected(self):
         self.browser_login_admin(**self.admin_data)
-        self.browse_to(self.reverse("tabular_view") + "?topic=addition-subtraction")
-        self.browser.find_element_by_css_selector('#error_message')
-        self.assertEqual(self.browser.find_element_by_css_selector('#error_message').text, _("No learner accounts have been created."), "Error message with no users available, no topic selected.")
+        url = "%s?topic=%s" % (self.url, self.topic,)
+        self.browse_to(url)
+        try:
+            elem = self.browser.find_element_by_css_selector('.alert-danger')
```
MCGallaspy: I think it's best to use a `WebDriverWait` here, just to hedge our bets against a race condition. For example, this line might throw a `NoSuchElementException` if the page hasn't finished loading. You can replace it with
```python
elem = WebDriverWait(self.browser, 3).until( EC.presence_of_element_located((By.CLASS_NAME, "alert-danger"))
```

And then change the `except NoSuchElementException` to `except TimeoutException` on line 38, since that's what will be thrown if the element is not found.

MCGallaspy: This applies to several other tests here, too.

Issue #2995, opened 2015-02-09
Learn tab not active when it should be


Antrikshy:
![screen shot 2015-02-09 at 3 08 17 pm](https://cloud.githubusercontent.com/assets/6601788/6117946/8b434560-b06d-11e4-8ad0-c4b7cd853e66.png)

I find myself clicking it multiple times before realizing I'm already there and need to click the "Learn!" box. It's active in the actual "Learn!" area though.

MCGallaspy:
@Antrikshy can you please provide more information about how to reproduce this error? Like what kind of user you're logged in as (learner/coach/admin) and what relative url you navigate to to get this error?

MCGallaspy:
Okay, I think I figured this out. You're at the base url logged in as an admin? Can you reach this page as any other kind of user?

MCGallaspy:
Doesn't look like it.

Antrikshy:
Sorry I couldn't reply earlier. Been really busy this week. I got it logged in as admin. Base URL, yes.

PR #3037, opened 2015-02-12
Hotfix for issue #2995 change learn tab url to learn instead of home.

djallado:
@aronasorman.

This will fix the issue: https://github.com/learningequality/ka-lite/issues/2995 - Learn tab not active when it should be.

Maybe the `KA-Lite Logo` would link to `homepage`, for now this is link to `learn`.


MCGallaspy:
I think this change is trivial enough that we don't need a test for it. If anyone disagrees, I'll write the test myself.

PR #3038, opened 2015-02-12
Don't error out in case we don't recognize the OS

### LINKED ISSUE #3023: get_ip_addresses not working on FreeBSD
branch: develop

See this snippet:
```
    if system.lower() in ["linux", "darwin", "macosx"]:
        # on Linux and OSX, use the ifcfg library to wrap ifconfig
        ips = [iface.get("inet") for iface in ifcfg.interfaces().values()]
    elif system.lower() == "windows":
        # on Windows, run ipconfig and parse the output
        ipconfig = os.popen("ipconfig /all").read()
        ips = [match[1] for match in re.findall("IP(v4)? Address[\.\: ]+([\d\.]+)", ipconfig)]

    # remove empty values for adapters without an IP
    ips = set(ips) - set([None, ""])

```

The `ips` variable won't be defined if `system.lower() not in ['linux', 'darwin', 'macosx', 'windows']`. So OSes like Solaris, *BSD, and others will error out whenever you run `kaserve`.

aronasorman:
Fixes #3023. We error out whenever we try to retrieve the external IP address of the computer, since we hardcode the OS values we're looking for. This just sets a default so we don't error out.

A proper fix for the given issue would be to make `ifcfg` work for *BSD.

### DIFF
--- a/python-packages/fle_utils/internet/functions.py
+++ b/python-packages/fle_utils/internet/functions.py
@@ -96,6 +96,8 @@ def get_ip_addresses(include_loopback=True):
         # on Windows, run ipconfig and parse the output
         ipconfig = os.popen("ipconfig /all").read()
         ips = [match[1] for match in re.findall("IP(v4)? Address[\.\: ]+([\d\.]+)", ipconfig)]
+    else:
+        ips = []
 
     # remove empty values for adapters without an IP
     ips = set(ips) - set([None, ""])
@@ -106,4 +108,3 @@ def get_ip_addresses(include_loopback=True):
         ips = ips - set(["127.0.0.1"])
 
     return list(ips)
-


MCGallaspy:
Unsure if related to your changes, but posting since it's suspicious...
```
FAIL: test_facility_user_logout_after_interval (kalite.distributed.tests.browser_tests.distributed.TestSessionTimeout)
Students should be auto-logged out
  File "/home/travis/build/learningequality/ka-lite/kalite/distributed/tests/browser_tests/distributed.py", line 349, in test_facility_user_logout_after_interval
    self.browser_check_django_message(message_type="error", contains="Your session has been timed out")
  File "/home/travis/build/learningequality/ka-lite/kalite/testing/mixins/browser_mixins.py", line 108, in browser_check_django_message
    self.assertIn(contains, message.text, "Make sure message contains '%s'" % contains)
AssertionError: 'Your session has been timed out' not found in u'\xd7\nUnexpected error; contact the FLE with the following information: 401\n\n[object Object]' : Make sure message contains 'Your session has been timed out'
```

MCGallaspy:
Tests passing, but I think I'm gonna take a look at the intermittent failures of TestSessionTimeout... Anyway, that seems to be unrelated.

MCGallaspy:
Is it actually ok to return an empty list here? Will it cause errors down the line? If so, perhaps it's preferable to raise a more informative error here (i.e. `NotImplementedError`) rather than just letting it pass by unnoticed.
> A proper fix for the given issue would be to make ifcfg work for *BSD.
Not sure if you're serious, but if so then that might be preferable to a quick fix here.

aronasorman:
Bam! Tests pass.

aronasorman:
Well this only generates the list of IP addresses for people to connect. So there's no harm on making an empty list. Doing so would just show `127.0.0.1:8008` as the IP address they can connect to, to open KA Lite. And since this function is only used for displaying the IP addresses, I don't foresee any errors popping up.

aronasorman:
However, I should write a test for this.

MCGallaspy:
If it's just for display/information purposes it's probably okay... Maybe just print a warning to the user in that case, like "Not able to determine available ip addresses" or something. And I'm not sure what kind of test you are proposing here? I'm okay with merging if the offending function is not used elsewhere.

aronasorman:
Just so I won't forget, here's the test I tried to write, but failed because of import hell caused by the automagical import mechanism:
```
class GetIPAddressesTests(TestCase):

    @patch.object(platform, "system")
    def test_returns_one_element_list_when_unknown_os(self, system_method):
        system_method.return_value = "Unknown OS"

        from internet.functions import get_ip_addresses
        ret = get_ip_addresses()

        self.assertEqual(ret, ['127.0.0.1'])

```

PR #3052, opened 2015-02-14
Fix Nalanda package checks, and reintegrate Nalanda-specific coach reports

jamalex:
Even though 0.13 won't be used for Nalanda for quite some time, they will be using the updated Hub server, so we need these fixes to avoid breaking functionality for them on there.


aronasorman:
Merge conflicts.

jamalex:
Hadn't realized this wasn't merged yet. Resolved this merge conflict already, in #3080.

PR #3028, opened 2015-02-11
Fixed the issue, "This content was not found!" message only displayed after navigation to new page #2978

djallado:
@aronasorman!

This PR fixed the issue: https://github.com/learningequality/ka-lite/issues/2978, "This content was not found!" message only displayed after navigation to new page.

PR #3053, opened 2015-02-16
Hotfix/2978 this content was not found message only displayed after navigation to new page

MCGallaspy:
Replaces #3028.

### DIFF
--- a/kalite/distributed/api_views.py
+++ b/kalite/distributed/api_views.py
@@ -91,9 +91,9 @@ def compute_total_points(user):
 
 def get_messages_for_api_calls(request):
     """
-    Re-usable function that returns a dictionary of messages to be used by API calls.
+    Re-usable function that returns a list of messages to be used by API calls.
     """
-    message_dicts = []
+    message_lists = []
     for message in get_messages(request):
         # Make sure to escape strings not marked as safe.
         # Note: this duplicates a bit of Django template logic.
@@ -101,8 +101,8 @@ def get_messages_for_api_calls(request):
         if not (isinstance(msg_txt, SafeString) or isinstance(msg_txt, SafeUnicode)):
             msg_txt = cgi.escape(unicode(msg_txt))
         msg_type = message.tags
-        message_dicts.append({msg_type: msg_txt})
-    return message_dicts
+        message_lists.append({msg_type: msg_txt})
+    return message_lists
 
 
 # On pages with no forms, we want to ensure that the CSRF cookie is set, so that AJAX POST
--- a/kalite/distributed/api_views.py
+++ b/kalite/distributed/api_views.py
@@ -33,7 +33,8 @@ from django.views.decorators.gzip import gzip_page
 from .api_forms import DateTimeForm
 from fle_utils.config.models import Settings
 from fle_utils.general import break_into_chunks
-from fle_utils.internet import api_handle_error_with_json, JsonResponse, JsonResponseMessage, JsonResponseMessageError, JsonResponseMessageWarning
+from fle_utils.internet import api_handle_error_with_json, JsonResponse, JsonResponseMessage, \
+    JsonResponseMessageError, JsonResponseMessageWarning
 from fle_utils.orderedset import OrderedSet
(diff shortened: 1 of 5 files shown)


MCGallaspy:
@jamalex @rtibbles someone wanna check this one out?

rtibbles:
Taking a peek.

kalite/distributed/api_views.py
```diff
@@ -67,7 +68,8 @@ def time_set(request):
             raise PermissionDenied
 
     except PermissionDenied as e:
-        return JsonResponseMessageError(_("System permissions prevented time setting, please run with root permissions."))
+        return JsonResponseMessageError(
+            _("System permissions prevented time setting, please run with root permissions."))
```
rtibbles: Much linting.

kalite/distributed/api_views.py
```diff
@@ -82,9 +84,25 @@ def compute_total_points(user):
         return None
     else:
         return (VideoLog.get_points_for_user(user) +
-           ExerciseLog.get_points_for_user(user) +
-           StoreTransactionLog.get_points_for_user(user) +
-           ContentLog.get_points_for_user(user))
+                ExerciseLog.get_points_for_user(user) +
```
rtibbles: Such indent.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -20,7 +20,6 @@ window.ContentAreaView = BaseView.extend({
     show_view: function(view) {
 
         // hide any messages being shown for the old view
```
rtibbles: What is the rationale for not clearing messages when we show a new content view?

rtibbles: Also, if this is appropriate, shouldn't we delete the comment as well?

rtibbles:
Does work, but as I feared in my comment above, the messages now persist when navigating to new content pages. See below:
image

kalite/distributed/api_views.py
MCGallaspy: I'm unclear... are you saying we should prefer a single tabwidth, instead of aligning with the start of the first variable name?

kalite/distributed/static/js/distributed/topics/views.js
MCGallaspy: I'm just gonna pop that clear_messages back in, should fix up the persistent message when navigating to a content page.

kalite/distributed/api_views.py
rtibbles: Sorry, I have a tendency to make comments for the sake of it. If there is no meaning evident (or I am speaking in doge speak), I am just being approving.

kalite/distributed/static/js/distributed/topics/views.js
rtibbles: Yeah, in a parallel PR this got replaced by the weird, and what I think is unnecessary clearAllMessages function (which appears to just do the same as clear_messages, but without the capacity to set an argument for message type).

kalite/distributed/tests/ajax_messages_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class TestShowAjaxMessages(BrowserActionMixins, FacilityMixins, KALiteBrowserTestCase):
+
+    def test_ajax_messages(self):
+
+        # Navigate to video that is not downloaded
+
+        self.browse_to(
+            self.reverse("learn") + "khan/math/algebra/introduction-to-algebra/overview_hist_alg/origins-of-algebra/")
```
MCGallaspy: By the way, this seems to work on the Travis server, but (as I noted in one of my commit messages) there's no reason to assume that this particular video is not downloaded, in which case the test will fail. We'll need to specifically find some content item that is not available.

MCGallaspy: @rtibbles this note touches on the lack of proper testing fixtures as well... what do you think about merging this PR?

rtibbles:
Let's do this!

PR #3054, opened 2015-02-16
Hotfix/2987 inconsistent behaviour of dropdown item selection on coach report pages

MCGallaspy:
Replaces #3030.

### DIFF
--- a/kalite/coachreports/static/css/coachreports/base.css
+++ b/kalite/coachreports/static/css/coachreports/base.css
@@ -8,6 +8,6 @@
     margin-bottom: 50px;
 }
 
-#tabularGo {
+#display-topic-report {
     margin-top: 12px;
 }
--- a/kalite/coachreports/static/css/coachreports/base.css
+++ b/kalite/coachreports/static/css/coachreports/base.css
@@ -7,3 +7,7 @@
 .container-margin-bottom {
     margin-bottom: 50px;
 }
+
+#tabularGo {
+    margin-top: 12px;
+}
--- a/kalite/coachreports/static/js/coachreports/base.js
+++ b/kalite/coachreports/static/js/coachreports/base.js
@@ -142,7 +142,7 @@ var NavigationContainerView = Backbone.View.extend({
         // attached as a subview here, but that would require making a
         // new API endpoint instead of passing info as context variables
         // to the template.
-        var params = {}
+        var params = {};
         if( url === TABULAR_REPORT_URL ) {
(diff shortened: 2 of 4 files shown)


MCGallaspy:
Nontrivial failing tests... gotta take a look at it.

rtibbles:
Failed only on 2.6. Restarting in case it was non-deterministic, but it may be a matter of not sleeping long enough?

rtibbles:
One issue here is that the two different "Go" buttons have different behaviour.

Although the bottom go button only appears on this page, we should make both the 'Go' buttons produce the same result on this page (i.e. for the top button including topic info and exercise/video in the get parameters, and for the bottom button changing 'report type', 'facility', and 'group'.)

In addition, the exercise/video selector is no longer being picked up, so it is no longer possible to view tabular coach reports for videos. Not sure when this might have vanished, but assume it was during this process.

MCGallaspy:
> One issue here is that the two different "Go" buttons have different behaviour.
@rtibbles I agree that this is problematic. Seems like a cleaner way to handle the issue would be to attach the content type/topic selectors as a subview to the report type/facility/group menu. I was trying to avoid that as it would require writing a new API endpoint (the information used by the type/topic selectors is currently passed as a context variable to the Django template), but maybe I should drink my own Kool-Aid and get rid of this jQuery spaghetti? The disparity of handling half the UI with Backbone and half without is increasing the complexity here.

rtibbles:
I think a simple hack would be fine here that just makes both the 'Go' buttons do the same thing. We can leave the spaghetti for now, unless it hinders your fixing of this.

MCGallaspy:
I guess my willingness to hack this vs. port it all to Backbone depends on the expected lifetime of these templates/views... Are there specific plans to scrap them and start from scratch? If not I'd rather just take the time right now to fix 'em up all purty.

rtibbles:
The teacher dashboards project will hopefully be burning this down by 0.14.

MCGallaspy:
Okay, hack it is, but if I find myself revisiting this module again then I'm just gonna do it the right way

MCGallaspy:
Maybe a python 2.6 specific issue? The offending test passes on my machine with python 2.7, but I'll take a closer look at it.

rtibbles:
Either a Python 2.6 issue, or it's to do with not waiting long enough for the client side content to change. Which is more likely on Python 2.6 as it is generally slower (the same reason we get more non-deterministic test failures on 2.6).

PR #2946, opened 2015-02-05
Mobile optimization: Admin pages cleanup

Antrikshy:
This PR comes with a ton of front-end changes to the admin pages. I have significantly cleaned up the messy reports page and other things. These are simple changes to templates and CSS and nothing else. 

I wonder why there are conflicts...

Here are a few screenshots:

#### Facilities and Reports pages (tables cleaned up):
screen shot 2015-02-05 at 10 59 20 am

---

screen shot 2015-02-05 at 11 03 19 am

---

screen shot 2015-02-05 at 11 05 10 am

---

#### Reports page on smaller width viewports:
screen shot 2015-02-05 at 11 24 43 am

---

screen shot 2015-02-05 at 11 04 42 am

Some tables still have green headers, conflicting with @benjaoming's blue colors from PR #2924. I'll fix those in a separate PR very soon because I plan to do some CSS refactoring related to the poorly named `.faded-blue-bg` class that we use across the site.

PR #3056, opened 2015-02-16
Hotfix/3004 messages never go away when navigating through content in the sidebar

MCGallaspy:
Replaces #3031.


rtibbles:
Seems to be a genuine test failure: "NameError: global name 'WebDriverException' is not defined"

An error in the test, no less!

MCGallaspy:
Oh? How did that one slip in there?

static-libraries/js/khan-lite.js
```diff
@@ -251,6 +251,11 @@ function clear_messages(msg_type) {
     return $("#message_container");
 }
 
+// To clear all messages including django and ajax.
+function clearAllMessages(msg_type) {
```
rtibbles: What is the purpose of the argument here, and what does this function do that clear_messages does not? If messages are appearing outside of #message_container that is a problem that should be addressed, not the addition of an almost identical function to the one directly above here.

rtibbles: @djallado ^

djallado:
@rtibbles. 
Sorry. Yah, the argument has no purpose at all. This clearAllMessages() should clear django message in #message_container.

static-libraries/js/khan-lite.js
djallado: The argument has no purpose at all. The function should clear django messages or normal messages loaded in #message_container.

MCGallaspy:
>If messages are appearing outside of #message_container that is a problem that should be addressed, not the addition of an almost identical function to the one directly above here.

That is indeed the issue here.

MCGallaspy:
I mean messages appearing outside the #message_container.

MCGallaspy:
@rtibbles I'm convinced the 2.6 tests are failing just because the testing server is too pokey right now. Further, the inability to automatically merge is because of some front-end changes from @Antrikshy that I just merged (#2946), which won't affect the tests. So if you (or someone else) signs off on this I'm going to merge it.

rtibbles:
Our test suite seems to be broken as far as 2.6 is concerned. Merge it, and we'll suffer the consequences.

Issue #2975, opened 2015-02-09
Ka-Lite Linux Installer


EdDixon:
Confusing documentation regarding starting the installation script. The INSTALLER.MD tells the user to follow the older method of manually running the setup_unix.py script where as the prompt informs the user this method You should not run kalite/manage.py, as it is deprecated and will be gone in KA Lite 0.14. Instead, use the 'kalite' command. However the command is not available from the directed folder of the underlying kalite folders.

Branch: develop
Expected Behavior: Installation should proceed as documented in the INSTALLER.MD without warning or information about future proposed changes potentially confusing a new user installing the program for the first time.
Current Behavior: Installation will proceed as documented after presenting these warnings.

Steps to reproduce:
1. Open INSTALLER.md
2. Follow steps 1-4
3. <Break> CTRL C out and attempt to run kalite command as instructed.

![screenshot from 2015-02-09 15 12 43](https://cloud.githubusercontent.com/assets/4349806/6117235/63dc29bc-b070-11e4-9e2a-5d339ca4c984.png)

PR #3063, opened 2015-02-19
Use kalite binary for setup_unix script instead of deprecated manage.py

MCGallaspy:
Address #2975. Just changes the command used by the setup script, the instructions in INSTALL.md are fine, and will be removed once we merge in the new docs and/or get added to package repositories.


rtibbles:
Tests pass. Merging.

PR #3064, opened 2015-02-19
Update footer

### LINKED ISSUE #2971: Copyright notices in footer need to be updated
Observed:
image

Expected:
Khan Academy's Perseus exercises are no longer CC-licensed, so we need to divide up the copyright section to indicate that while the videos are, the exercises are not.

We should also display license information on the video/exercise pages themselves.

MCGallaspy:
How's this:
ss

Resolves #2971.


rtibbles:
Doesn't address last part of issue: "We should also display license information on the video/exercise pages themselves."

But I think we can punt that to 0.14.

rtibbles:
Additional part put in #3065

rtibbles:
Tests pass. Merging.

PR #3046, opened 2015-02-13
Fixes for Exercises

rtibbles:
Fixes #2994, resolves #2996, and should close #3005

Summary of changes:
* Cleans up residual assessmentitems.json and properly adds it to gitignore.
* Adds availability flagging for Khan Exercises (by checking template files on disk)
* Adds availability flagging for Perseus Exercises (by checking existence of assessmentitems in cache)
* Only returns all_assessment_items for Perseus Exercises for ones that exist
* Loads exercise cache on startup
* Fixes Exercise and Content Resource API endpoints to return the right properties (some were missing)
* Reduces the amount of cruft in Exercise Resource class by just importing all properties.
* Try to get another assessmentitem if the requested one 404s.
* Warn users and don't try to render an exercise if the exercise is unavailable or is a perseus exercise and has no assessment items.
* Style acceptable formats tooltips to prompt users
* Focus highlights input boxes

PR #3066, opened 2015-02-19
Warn, without error, when assessmentitems are missing (Fixes typo in #3046)

jamalex:


rtibbles:
Woops...

rtibbles:
Will merge when tests pass.

Issue #3006, opened 2015-02-10
Viewing one (or all) of the hints for an exercise doesn't mark it as incorrect


MCGallaspy:
On `develop`, at `/learn/khan/math/pre-algebra/fractions-pre-alg/equivalent-fractions-pre-alg/equivalent_fractions/` (but probably some other exercises as well), my progress for the exercise increases no matter how many hints I take.

MCGallaspy:
I wonder if this is related to #2977?

rtibbles:
Fixed in #3069.

PR #3069, opened 2015-02-20
Add proxy listener to hintUsed event

MCGallaspy:
Solves #3006. Everything was in order, just a matter of firing off a Backbone event instead of a jQuery event.


rtibbles:
I approve of this code.

PR #3080, opened 2015-02-21
Fixes made in nalanda-rct3 and then brought into the central server submodule

jamalex:
Merging back to develop, unless anything looks dangerous, so that we avoid future merge conflicts/divergence.


jamalex:
(This doesn't include the latest nalanda changes, just the ones as of a couple of weeks ago, which were needed in order to support Nalanda reports etc on the central server)

jamalex:
@rtibbles Would you be able to have a peek?

kalite/settings.py
```diff
@@ -169,10 +169,6 @@ def package_selected(package_name):
 
 import_installed_app_settings(INSTALLED_APPS, globals())
 
-KHAN_EXERCISES_RELPATH = os.path.join("static", "perseus", "ke")
```
rtibbles: Where did these go?

kalite/topic_tools/settings.py
```diff
@@ -11,4 +12,8 @@
 #######################
 # Set module settings
 #######################
-DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = getattr(local_settings, "DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP", False)
\ No newline at end of file
+DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = getattr(local_settings, "DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP", False)
+
+KHAN_EXERCISES_RELPATH = os.path.join("static", "perseus", "ke")
```
rtibbles: Ah, here it is.

rtibbles:
Nothing jumps out. Tests pass. Merging.

jamalex:
Yayyy, thanks!

PR #3117, opened 2015-02-24
Remove redundant call to khan_loaded function

### LINKED ISSUE #3096: Show scratchpad button does not work sometimes
1. Go to an exercise.
2. Open scratchpad to confirm it works.
3. Go to another exercise.
4. Attempt to open scratchpad and be surprised.

Reload page to make it work again.

Related: After switching exercises, the show hints button seems to misbehave (even goes to -1 hints).

MCGallaspy:
Resolves #3096. I'm not writing a test for this one, because I'm taking a test hiatus while we discuss testing best practices. If that's a lame excuse, call me out and I'll write a test.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -158,7 +158,11 @@ window.ExerciseView = Backbone.View.extend({
         var self = this;
 
         // TODO-BLOCKER(jamalex): does this need to wait on something, to avoid race conditions?
-        _.defer(this.khan_loaded);
+        if(Khan.loaded) {
+            Khan.loaded.then(this.khan_loaded);
+        } else {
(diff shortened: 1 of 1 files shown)


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -159,10 +159,6 @@ window.ExerciseView = Backbone.View.extend({
 
         // TODO-BLOCKER(jamalex): does this need to wait on something, to avoid race conditions?
         _.defer(this.khan_loaded);
-        if (Khan.loaded) {
```
rtibbles: Might prefer to do this as an if/else?

If Khan.loaded exists, we should use that to avoid race conditions (this only happens for Khan Exercises, not Perseus), otherwise, we can just do the _.defer.

rtibbles:
Bam.

PR #3121, opened 2015-02-25
Fix for #3094. Drawing on scratchpad sometimes selects page elements.

### LINKED ISSUE #3094: Drawing on scratchpad sometimes selects page elements
This one is a bit hard to describe. I encountered this in the radio-button style exercise as pictured.

screen shot 2015-02-23 at 1 35 45 pm

It happens when the user starts dragging from within the `label` element highlighted below.

screen shot 2015-02-23 at 1 37 14 pm

Just a minor issue caused by having the scratchpad under the exercise.

EDIT: Also that scratchpad border overlaps another element's border. Kinda unsightly.

djallado:
Hi @aronasorman this fixed #3094 - Drawing on scratchpad sometimes selects page elements.

I reduced the width of `perseus-widget-radio.above-scratchpad.blank-background` class to 40% so that the remaining space will be used for scratchpad.

I attached screenshot below:
screen shot drawing

/review: @amodia, @arceduardvincent.

### DIFF
--- a/kalite/distributed/static/perseus/ke/css/khan-exercise.css
+++ b/kalite/distributed/static/perseus/ke/css/khan-exercise.css
@@ -1274,3 +1274,7 @@ body.mobile ul.inequalities-one-line-radios > li {
 .inequalities-padding#grid {
     margin-bottom: 2em;
 }
+
+.perseus-widget-radio.above-scratchpad.blank-background {
+    width: 40%;
+}


MCGallaspy:
I'm hesitant to change these styles, because it begs the question of whether we're breaking the display for other exercises. Plus this change doesn't affect the exercise from the issue, where a transparent `label` tag is above the scratchpad. (This exercise: `/learn/khan/math/pre-algebra/applying-math-reasoning-topic/greater-than-less-than/comparing_whole_numbers/`)

djallado:
Hi. @MCGallaspy. I think this is not a high priority issue. For some deployments, they used notebook instead of our scratchpad. 
I will look for some high priority issue and then I will fix this some time this week.

rtibbles:
Closing this PR, doesn't quite seem the right approach.

PR #3123, opened 2015-02-25
Add clear message functon at the top of the video error message.

### LINKED ISSUE #3122: Language pack error message should not appear at video tab.
Branch: `Develop`
Expected behavior: Language pack error message should not appear when video tab is active.
Steps to reproduce:
 1. make sure the distributed is not connected to central server. 
 2.  log in as admin.
 3. navigate to video and language tab.
 
ka-lite-a

amodia:
Hi @MCGallaspy this fixed #3122 

ka-lite-a

### DIFF
--- a/kalite/updates/static/js/updates/update_videos.js
+++ b/kalite/updates/static/js/updates/update_videos.js
@@ -151,6 +151,7 @@ $(function() {
                             $(".enable-when-server-online").removeAttr("disabled");
                             updatesStart("videodownload", 5000, video_callbacks);
                         } else {
+                            clear_messages();
                             show_message("error", gettext("The server does not have internet access; videos cannot be downloaded at this time."));
                         }
                     });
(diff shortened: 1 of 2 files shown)


MCGallaspy:
This isn't a backbone app? :(

MCGallaspy:
I don't think this has the intended behavior... the language packs error message will only disappear when the video packs message error appears. I think a better solution would just be to clear messages after a navigation event, otherwise when navigating back to videos from the language page, the language pack error message will hang around for a few seconds until the video error message appears. Does that make sense?

amodia:
The current behavior is when the video message error appears the languages pack error is not disappear. 

screen shot 2015-02-23 at 4 08 16 pm

MCGallaspy:
Right, so if we could just call `clear_messages` to dismiss outstanding messages when the page loads... kind of like what happens with the Backbone app for  `/learn/`, except different since this isn't Backbone.

amodia:
Ah ok i need to find how this languages pack error message execute.

amodia:
Hi @MCGallaspy i just remove `update_language.js` at update videos html to prevent language error message execute.

rtibbles:
This is the right fix. No reason I can think of for update_languages.js to be in the video updating page.

Issue #3119, opened 2015-02-25
Hover on tabular coach report obscures user data


rtibbles:
@Antrikshy this looks like it came from your addition of .table-hover class on the tabular coach report. Care to fix?

![image](https://cloud.githubusercontent.com/assets/1680573/6364902/67ecb334-bc62-11e4-8092-fdaa1e952300.png)

Antrikshy:
I'll take a look at it as soon as possible.

Antrikshy:
There you go: #3125.

From what I gather from the screenshot, removing .table-hover should fix it, correct?

cpauya:
PR fix #3125, closing.

PR #3125, opened 2015-02-25
Coach reports table hover hotfix

Antrikshy:
I removed that Bootstrap .table-hover from the table mentioned in #3119. Also made a quick change in a column on the same page to make the inputs flow slightly better on smaller devices.

I didn't know the table cell backgrounds were used to show information. Should be all good now.


rtibbles:
Works. Merging.

PR #3127, opened 2015-02-26
Add student-only css class to certain dropdown menu elements

### LINKED ISSUE #2912: Superuser menu is empty
Looks weird

superuser

MCGallaspy:
Seems like they present information that's only applicable to
students, so I just added the student-only class to set their
display to none.

Resolves #2912.


rtibbles:
Yep.

PR #3136, opened 2015-02-26
Fix for #3130-Tab for Update Software is missing.

### LINKED ISSUE #3130: Tab for Update Software is missing.
Branch **develop**

Two issues for this:

1. There is no link to update the software on the Manage tabs.
2. The actual link points to http://127.0.0.1:8008/update// - notice that it has two `/` at the end - so we need to fix this on `/kalite/updates/urls.py` too.

Possible solutions:

1. Add a tab under the Manage nav along with the `Videos` / `Languages` tabs.  The tab content should look like the attached screenshot.
2. Fix the url issue and remove the extra `/`.

The update software page should look like this:
screenshot 2015-02-26 12 03 24

@jtamiace I found the `/kalite/distributed/static/images/distributed/update-*-icon.png` images - I think we will use these for the tab icons?

Notes:

* On the `updates/urls.py` - notice that we check if we are using `nalanda` package - we need to check if this works too when making a fix for this issue.
* This link must show and work only if user is an admin.

djallado:
Hi @aronasorman and @cpauya. This fixed #3130 - Tab for Update Software is missing.

Here's the screenshot:
screen shot 2015-02-26 at 6 10 05 pm

### DIFF
--- a/kalite/distributed/static/css/distributed/khan-lite.css
+++ b/kalite/distributed/static/css/distributed/khan-lite.css
@@ -354,6 +354,16 @@ a:hover {
     background-position: center 7px;
     height: 87px;
 }
+.update {
+    background: url("../../images/distributed/update-icon.png") no-repeat;
+    background-position: center 7px;
+    height: 87px;
+}
+.update:hover, .update-active {
+    background: url("../../images/distributed/update-active-icon.png") no-repeat;
+    background-position: center 7px;
+    height: 87px;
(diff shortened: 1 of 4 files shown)


rtibbles:
Ah, one problem - after discussion with @jamalex this is actually deliberately hidden because the Update feature is not production ready.

Not sure when this will be ready, so going to close the PR.

PR #3138, opened 2015-02-26
Prevents user text selection on work area where exercises are rendered

### LINKED ISSUE #3094: Drawing on scratchpad sometimes selects page elements

rtibbles:
Fixes #3094 without messing with any other styles.

Summary of changes:
* Prevent text selection in the work area (note, that this still allows for clicking of radio buttons when scratchpad is open, it just prevents inadvertent text selection).

@Antrikshy @MCGallaspy

### DIFF
--- a/kalite/distributed/static/css/distributed/khan-lite.css
+++ b/kalite/distributed/static/css/distributed/khan-lite.css
@@ -1008,3 +1008,12 @@ div#fade {
   border-color: rgba(81, 203, 238, 1);
   border-style: solid;
 }
+
+#workarea {
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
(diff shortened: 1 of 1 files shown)


MCGallaspy:
You're a wizard, @rtibbles.

Antrikshy:
>`user-select: none`

Genius. I'd never have thought of that.

PR #3139, opened 2015-02-26
(Re)Introducing the back button: Sidebar now pushed off screen on narrower viewports

### LINKED ISSUE #2859: Sidebar can sometimes cover the entire screen
Screen covered!

Horizontal scrolling does not cover the entire interface. It only appears when the viewport becomes smaller than the `html` width or something like that. One solution would be to introduce horizontal scrolling within the sidebar when the number of submenus exceeds a threshold. 

I might dig into some of the templates and CSS myself, but that will take me some time. If this is an easy fix, feel free to go ahead and do it.

Antrikshy:
This is a fix for one of the most important UI issues. The notorious #2859. I have made some heavy changes to the sidebar. It moves out the left side of the screen when the viewport is too narrow. Depending on screen size, 3, 2 or 1 panels are shown at once. On wider screens, the behavior is unchanged.

I added a back button to one of the sidebar templates, complete with an icon from IcoMoon. `SidebarView`'s `resize` function is now split into two. One contains the unchanged behavior for wider screens and the other manages the job of hiding it on smaller devices. Sidebar behavior is calculated on every interaction and the hiding is based on `$(window).width()`.

This is mostly untested on smaller devices, but it works really well when I narrow my browser window.

As far as the performance is concerned, I have added a jQuery `animate` call when the sidebar is being pushed off the screen. This is really necessary because without it the behavior was really disorienting (unless someone has a better idea).

screen shot 2015-02-26 at 11 39 53 am

screen shot 2015-02-26 at 11 36 31 am

screen shot 2015-02-26 at 11 36 53 am

Back button color is provisional and will change when the new color palette is implemented. I'm open to suggestions!

Moving forward, I plan to make the sidebar even more mobile friendly by using responsive typography. In this PR, I have increased the use of ems for sizing, but they hardly do anything right now. It is a big enough change to warrant its own PR in the coming weeks.

Fixes #2859


Antrikshy:
Oh noes, my very first Travis error!

kalite/distributed/static/css/distributed/sidebar.css
```diff
@@ -134,7 +131,20 @@ div.fade {
...
+    text-align: center;
+    color: #fff;
+    background-color: #467471;
+    font-size: 24px;
+    border-right: 1px solid rgba(0, 0, 0, 0.3);
+    position: absolute;
+    z-index: 10000;
+}
+
+/*.back-to-parent {
```
MCGallaspy: Is it safe to remove these styles? Did you remove the elements that were styled by this? Better to delete instead of cluttering up the file with commented-out code.

kalite/distributed/static/css/distributed/sidebar.css
```diff
@@ -142,13 +152,14 @@ div.fade {
...
     padding-right: 14px;
-}
+}*/
 
 /* Sidebar panel & entry styling */
 
 .topic-container-inner {
     float: left;
     width: 200px;
+    /*max-width: 25%;*/
```
MCGallaspy: See above. Delete instead of commenting.

kalite/distributed/static/fonts/icomoon.svg
```diff
@@ -21,6 +21,7 @@
...
 <glyph unicode="&#xe614;" d="M892.118 771.882l-120.234 120.236c-37.338 37.336-111.084 67.882-163.884 67.882h-448c-52.8 0-96-43.2-96-96v-832c0-52.8 43.2-96 96-96h704c52.8 0 96 43.2 96 96v576c0 52.8-30.546 126.546-67.882 163.882zM640 824.438c2.196-0.804 4.452-1.68 6.758-2.636 18.060-7.482 30.598-16.176 34.616-20.194l120.236-120.238c4.018-4.018 12.712-16.554 20.194-34.614 0.956-2.306 1.832-4.562 2.636-6.756h-184.44v184.438zM832 64h-640v768h384v-256h256v-512zM256 448h512v-64h-512zM256 320h512v-64h-512zM256 192h512v-64h-512z" />
+<glyph unicode="&#xea40;" d="M402.746 82.746l-320 320c-24.994 24.992-24.994 65.516 0 90.51l320 320c24.994 24.992 65.516 24.992 90.51 0 24.994-24.994 24.994-65.516 0-90.51l-210.746-210.746h613.49c35.346 0 64-28.654 64-64s-28.654-64-64-64h-613.49l210.746-210.746c12.496-12.496 18.744-28.876 18.744-45.254s-6.248-32.758-18.744-45.254c-24.994-24.994-65.516-24.994-90.51 0z" />
```
MCGallaspy: Ah, I see...

kalite/distributed/static/css/distributed/sidebar.css
Antrikshy: `.back-to-parent`? They have been in there for a while. There was never a back button implemented. They can be removed now that I have a working back button. I'll clean that stuff up. It's used in some JS files too.

kalite/distributed/static/js/distributed/topics/router.js
```diff
@@ -41,7 +41,7 @@ ChannelRouter = Backbone.Router.extend({
         var current_url = Backbone.history.getFragment();
         var fragments = current_url.split("/").slice(0,-1);
         if (fragments.length > 0) {
-            this.navigate(fragments.slice(0,-1).join("/") + "/");
+            this.navigate(fragments.slice(0,-1).join("/") + "/", {trigger: true});
```
MCGallaspy: I remember when we added this, so that your back button (which uses this function as a callback) works as intended... Does this change the behavior of existing uses of `url_back`?

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -103,27 +105,68 @@ window.SidebarView = BaseView.extend({
...
+        var current_level = this.state_model.get("current_level");
+        var column_width = this.$(".topic-container-inner").width();
+        var last_column_width = 400;
+        
         this.width = (current_level-1) * column_width + last_column_width + 10;
         this.$(".sidebar-panel").width(this.width);
         this.$(".sidebar-tab").css({left: this.width});
         this.update_sidebar_visibility();
-
-        // TODO(dylanjbarth): Resize sidebar to not cover top nav
```
MCGallaspy: So we've given up on resizing the sidebar to not cover the top nav?

kalite/distributed/static/css/distributed/sidebar.css
Antrikshy: I'm not sure why that huge block of commented CSS exists in sidebar.css so I have always scrolled past it.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -103,27 +105,68 @@ window.SidebarView = BaseView.extend({
...
+
+        else
+            this.resize_for_wide();
+    },
+
+    resize_for_narrow: _.debounce(function() {
         var current_level = this.state_model.get("current_level");
         // TODO(jamalex): have this calculated dynamically
         var column_width = 200; // this.$(".topic-container-inner").width();
         var last_column_width = 400;
```
MCGallaspy: Right, so even the one column can still overflow really small screens. Perhaps we can set the `.topic-container-inner` width property using css media rules and, below a certain viewport threshold, simply have the column take up the full width of the screen?

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -103,27 +105,68 @@ window.SidebarView = BaseView.extend({
...
         this.$('.sidebar-content').append(this.topic_node_view.el);
 
+        this.set_sidebar_back();
+
         return this;
     },
 
-    resize_sidebar: _.debounce(function() {
+    resize_sidebar: function() {
+        if ($(window).width() < 1260)
```
MCGallaspy: How did you come to this number?

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -145,14 +187,40 @@ window.SidebarView = BaseView.extend({
...
-        } else {
+        } 
+        else {
             this.sidebar.css({left: - this.width});
-            this.$(".sidebar-tab").css({left: 0}).html('<span class="icon-circle-right"></span>');
+            this.sidebarTab.css({left: 0}).html('<span class="icon-circle-right"></span>');
             this.$(".fade").hide();
         }
+
+        this.set_sidebar_back();
```
MCGallaspy: This function name suggested to me that the sidebar element was being set back, instead of setting the position of the sidebarBack element. Can the name be made clearer?

kalite/distributed/static/js/distributed/topics/router.js
rtibbles: Fortunately, this code is left over from the last time we had a back button in the sidebar (plus ça change), and is not called by any other functions, so should be fine to modify.

rtibbles: Well, it's called by 'move_back' which is itself never called.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -56,7 +56,8 @@ window.SidebarView = BaseView.extend({
 
     events: {
         "click .sidebar-tab": "toggle_sidebar",
-        "click .fade": "check_external_click"
+        "click .fade": "check_external_click",
+        "click .sidebar-back": "sidebar_back_one_level"
```
rtibbles: Would be nice to remove the 'move_back' method if this is no longer being used for this functionality.

kalite/distributed/static/js/distributed/topics/views.js
rtibbles: I would also note that 1260 pixel width is pretty wide.

rtibbles:
@Antrikshy other feedback from @cpauya https://github.com/learningequality/ka-lite/issues/2859#issuecomment-76317907

cpauya:
This PR fixes #2859 - making this comment so Waffle attaches this PR on the issue card.

rtibbles:
:+1:

kalite/distributed/static/js/distributed/topics/views.js
Antrikshy: It was part of `TopicInnerContainer` so I ignored it. Never noticed that it was not being used anywhere.

kalite/distributed/static/js/distributed/topics/views.js
Antrikshy: I wonder if there is a better way to do this instead of pixels. 1260 was approximately the width of the fully extended sidebar with the tab on my screen at least. I'm assuming the sidebar is somehow scaled down on smaller devices? Curious to see how it works on a Nexus tablet.

kalite/distributed/static/js/distributed/topics/views.js
Antrikshy: I'll look into that next, as part of the responsive typography update using ems and media queries.

kalite/distributed/static/js/distributed/topics/views.js
Antrikshy: I should make it clearer. `set_sidebar_back()` actually makes sure the `.sidebar-back` button is in whatever state it is supposed to be in. I should call it `set_sidebar_back_btn()` instead. I just decided to go with the shortened version.

MCGallaspy:
It seems like some of the suggested changes (like responsive column sizes and adding breadcrumbs) extend the scope. I favor removing some cruft (like the unused `move_back` function) from this PR, merging, and continuing to work on the sidebar in a different branch. Then the app will at least be usable on small screens, even though its not perfect. Thoughts?

Antrikshy:
@MCGallaspy That's what I was leaning towards. I already had a branch for the responsive sidebar anyway.

I have removed some cruft and this branch is ready to be merged. There is one known issue: when user is on some content (exercise, video etc), the back button has to be pressed twice to go back (first time just deselects the content). Since it will require some tinkering and because it's not that huge of a deal, I will do it as part of the responsiveness + breadcrumbs + other enhancements update.

rtibbles:
This works. Further iteration can happen with the sidebar redesign stuff that our awesome UX interns are working on.

PR #3140, opened 2015-02-26
Move footer tag and set overflow for #wrapper

### LINKED ISSUE #3067: Tooltip on accepted answer formats can go off bottom of screen
This is on:
http://127.0.0.1:8008/learn/khan/math/algebra/solving-linear-equations-and-inequalities/super-yoga/equations-in-one-variable-1/

image

(note how the footer doesn't expand down to stay at the bottom)

We should also see how big tooltips like this display on small tablet screens.

MCGallaspy:
Now the footer is positioned statically after the #wrapper element
If content overflows the #wrapper now instead of extending past
the footer, #wrapper will become scrollable.

Fixes #3067.

### DIFF
--- a/kalite/distributed/static/css/distributed/bootstrap-overrides.css
+++ b/kalite/distributed/static/css/distributed/bootstrap-overrides.css
@@ -12,7 +12,7 @@ html, body {
     min-height: 100%;
     position: relative;
     display: block;
-    overflow: auto;
+    overflow: visible;
 }
 #content-container {
     padding-top: 20px;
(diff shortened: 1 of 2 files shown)


kalite/distributed/static/css/distributed/bootstrap-overrides.css
```diff
@@ -11,6 +11,8 @@ html, body {
 #wrapper {
     min-height: 100%;
     position: relative;
+    display: block;
+    overflow: auto;
```
rtibbles: Could we change this to overflow: visible? The auto behaviour gives some ugly scrollbars that go kinda screwy when the tooltip disappears again.

With visible rather than auto, it looks like this:
image

PR #3148, opened 2015-02-27
Load testing suite

rtibbles:
We previously had a load testing suite using an iframe to automatically interact with the website to enable simulated user interactions on a large number of devices.

This reinstates that capability for 0.13.


MCGallaspy:
Can you make some comments about the intended use somewhere?

rtibbles:
Will do.

rtibbles:
Added comment in the associated Django view function.

PR #3150, opened 2015-02-27
runserver is blazingly fast after removing prefetched variables

MCGallaspy:
Because of import_all_child_modules nonsense, this file is loaded and
these functions are called long before they are ever used. Why don't we just
wait until they're first called to be used? If it's just too intolerably slow
and should be called when the server starts anyway, let's do it explicitly
somewhere else so that innocently importing this file won't take forever.


rtibbles:
I approve this message.

PR #3151, opened 2015-02-27
Fixes the alignment bug on the popup for students completing a streak in exercises

jtamiace:


kalite/distributed/static/css/distributed/khan-lite.css
```diff
@@ -924,6 +924,7 @@ to items you want to have the icon appear with.
 div#modal {
     position: absolute; /* makes the div go into a position that’s absolute to the browser viewing area */
     margin: auto;
+    padding-bottom: 50px;
```
MCGallaspy: I think this much padding is not appropriate in all cases:
ss

MCGallaspy:
Is it possible to just grow the modal element's height with the content (instead of having a fixed height)?

rtibbles:
I would like to suggest a different strategy here. KILL THE MODAL.

Seriously, we only added the modal for a very specific request in the Nalanda RCT, so, I would rather just watch it die, and use our normal messaging system instead.

rtibbles:
(Also, the modal performance is terrible on low end tablets, with some students pressing multiple times to try to dismiss it, and all sorts of weird non-captured click event artifacts happening as a result).

jamalex:
> KILL THE MODAL.

I agree. So much :crying_cat_face: here in watching students interact with the modal. Make it :game_die:!

rtibbles:
Closing in favour of #3159

Issue #2891, opened 2015-01-25
Replace fle_utils.cronograph with a third party (btw I hear that celery contains a lot of vitamins)


benjaoming:
So instead of having a buggy and inferior job scheduler, let's use a truly great third-party app for this!!

I'm thinking about [Celery](http://celery.readthedocs.org/en/latest/) of course :) KA Lite seems to have a lot of background jobs, and this is a fundamental part of the platform that cannot be refactored.

For motivation on why our own implementation should be removed, have a look at the issues it has caused over time:

https://github.com/learningequality/ka-lite/issues?q=cron

How you want to run Celery is configurable. For instance, for dev and testing, there is the `CELERY_ALWAYS_EAGER` setting.

I would recommend kicking things off with having a pure-Django broker. The amount of jobs currently does not warrant for any super-fast compiled OS-specific broker.

Yet, the Celery worker process has to be started in an OS-specific fashion.

Running on OSX:
https://github.com/celery/celery/tree/3.1/extra/osx/

Running on Windows:
https://www.calazan.com/windows-tip-run-applications-in-the-background-using-task-scheduler/

jamalex:
Interesting idea!

> KA Lite seems to have a lot of background jobs

The original version did have a number of things run via background jobs (e.g. the video downloading was queued by the main process and then picked up and run by the cron), but most of this was removed over a year ago, and turned into immediate direct thread spawning. The only thing still being handled by chronograph is the automatic background hourly data sync.

> have a look at the issues it has caused over time

I didn't recall us having too many issues with chronograph, at least since very early on, so I did a bit of a survey.

Most of those issues just mention cron in passing (e.g. quoting the log output with `Starting the cron server in the background`, or comments mentioning/recommending the use of cron), not citing cron as a cause of a problem (#15, #101, #246, #263, #271, #278, #308, #320, #529, #652, #659, #688, #875, #876, #908, #1653, #2025, #2278, #2434).

Most of the others were either related to the old way of doing things, or were about small optimizations to improve the efficiency of what is already a very lightweight library (it's basically just two tables -- jobs and logs -- with a few helper functions around it for scheduling, and tracking status).

I've wanted to use Celery in projects for a long time now, and our central server could definitely be a great place to do that (since we have full control over the deployment environment, and we will be doing more and more job-processing stuff there). For the ka-lite app itself, I hadn't considered celery as an option, since:

[...]

PR #3152, opened 2015-02-27
It was the cronserver mgmt command all along

MCGallaspy:
cronserver gunks up the works for the other kaserve command.
Not clear to me why. I'm calling not executing it a workaround.
Does it actually do anything?
Also, fixed a bug in the 'status' command. Windows doesn't like
an HTTPConnection with "0.0.0.0" address.

This provisionally addressed #2919, but I think we need to look more closely at why the cronserver management command is causing these issues on Windows. `kalite.bat start/stop/status` all work in this state, though. Sad me is getting slightly less sad. :|


jamalex:
Excellent debugging!

We can't just skip starting up the cronserver, though -- it's used to auto-sync data in the background every hour, which is important. Do you think it's an issue with spawning the thread? Where exactly does it freeze up?

Possible alternative solution, if we can't get the thread spawning working, is to modify the Windows GUI to:

- pass `--skip-job-scheduler` when running kaserve
- separately run the cronserver command

MCGallaspy:
> Where exactly does it freeze up?

It doesn't freeze up, so much as a second thread with the exact same command is run (so for example, if I execute `python.exe kalite start` then another thread also trying to run that command is started) which upon finding the startup lock file exists, dies and kills the original thread. Again, it's not yet clear to me why this is happening, I'll have to take a closer look at the `cronserver` command.

MCGallaspy:
I'm not really any closer to understanding this issue. I'm not throwing in the towel, but I'm making an open call for help/insight.

MCGallaspy:
My commit message, for posterity:
For some reason, the thread running the cronserver management command
(in Windows only) barfs and kills the whole process. This prevents the
server from starting. Why does it get sick and die? Who knows?
A better solution on windows (better than trying to UNIX-ly daemonize programs)
seems to be to create windows services. Then kalitectl.py can simply start
and stop those services. Just to be clear, there's no issue on windows with
running the cronserver and kaserve command in separate processes (say, in different
instances of cmd.exe or using subprocess.Popen), there's only an issue when they
are run in different threads belonging to the same process.

kalitectl.py
```diff
@@ -290,8 +293,11 @@ def manage(command, args=[], in_background=False):
         utility.prog_name = 'kalite manage'
         utility.execute()
     else:
-        thread = ManageThread(command, args=args)
-        thread.start()
+        if os.name != "nt":
+            thread = ManageThread(command, args=args, name=" ".join([command]+args))
+            thread.start()
+        else:
+            Popen([sys.executable, os.path.abspath(sys.argv[0]), "manage", command] + args, creationflags=CREATE_NEW_PROCESS_GROUP)
```
aronasorman: Hmm I wonder if we can just have them all spawn another process. @benjaoming or @jamalex any guidance on why the cronserver was originally programmed to spawn within the same process?

aronasorman: Otherwise, we can leave this as is for 0.13. @MCGallaspy can you put a comment to try out having all platforms spawn another process, for 0.14?

MCGallaspy:
@EdDixon I think we can merge this if you tell us that the `bin/windows/kalite.bat start/stop` commands work as expected in Windows 7.

aronasorman:
Merging this. @EdDixon please test this tomorrow in the bug bash.

kalitectl.py
```diff
@@ -219,7 +222,7 @@ def get_pid():
     listen_port = getattr(settings, "CHERRYPY_PORT", LISTEN_PORT)
 
     # Timeout is 1 second, we don't want the status command to be slow
-    conn = httplib.HTTPConnection(LISTEN_ADDRESS, listen_port, timeout=3)
+    conn = httplib.HTTPConnection("127.0.0.1", listen_port, timeout=3)
```
benjaoming: Should not be hard coded?

benjaoming:
Nice getting this fixed!! I think it's problematic that kalite is using "0.0.0.0" and not a deliberate IP-address. It isn't normal precisely because of the reason we've just seen.

It would be more "normal" (as per my personal experience with linux apps) to default to "127.0.0.1", furthermore to allow binding to multiple addresses, including "0.0.0.0" which could be deliberately set on a system that supports it.

Futher... it reminds of the need to get Celery #2891

kalitectl.py
MCGallaspy: Agreed, so we need to grab a valid address (meaning not "0.0.0.0") that we can listen to, regardless of whatever options the user has specified.

PR #3155, opened 2015-02-28
Returns keypad to its former stylistic glory

### LINKED ISSUE #3153: Keypad so ugly
image

The .simple-button style got removed a while back without considering the unintended consequences.

rtibbles:
Fixes #3153.

Summary of changes:
* Refactor software keyboard to use handlebars template
* Restyle software keyboard to properly leverage Bootstrap styling

image

### DIFF
--- a/kalite/distributed/hbtemplates/exercise/software-keyboard.handlebars
+++ b/kalite/distributed/hbtemplates/exercise/software-keyboard.handlebars
@@ -5,7 +5,7 @@
         <div class="row">
             <div class="col-xs-12">
                 {{#each this}}
-                    <button class='btn btn-success key' value='{{this}}'>{{this}}</button>
+                    <button class='btn btn-success key' value='{{this.value}}'>{{{this.display}}}</button>
                 {{/each}}
             </div>
         </div>
--- a/kalite/distributed/hbtemplates/exercise/software-keyboard.handlebars
+++ b/kalite/distributed/hbtemplates/exercise/software-keyboard.handlebars
@@ -1,0 +1,13 @@
+<button class='simple-button orange' id='show-keyboard'> {{_ "Hide Keypad" }}</button>
+
+<div class='container-fluid' id='software-keyboard'>
+    {{#each keys }}
+        <div class="row">
+            <div class="col-xs-12">
+                {{#each this}}
+                    <button class='btn btn-success key' value='{{this}}'>{{this}}</button>
+                {{/each}}
+            </div>
+        </div>
+    {{/each}}
+</div>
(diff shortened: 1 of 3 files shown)


kalite/distributed/hbtemplates/exercise/software-keyboard.handlebars
```diff
@@ -0,0 +1,13 @@
+<button class='simple-button orange' id='show-keyboard'> {{_ "Hide Keypad" }}</button>
+
+<div class='container-fluid' id='software-keyboard'>
+    {{#each keys }}
+        <div class="row">
+            <div class="col-xs-12">
```
MCGallaspy: Why not put each button in a `col-xs-4` container?

kalite/distributed/static/css/distributed/khan-lite.css
```diff
@@ -820,50 +795,24 @@ a.disabled {
 }
 
 #software-keyboard div {
```
MCGallaspy: Could it be `#software-keyboard .row {`? Small change, I know, but more semantic.

kalite/distributed/hbtemplates/exercise/software-keyboard.handlebars
rtibbles: Tried it. Got this:
image

rtibbles: Bootstrap buttons are designed to be displayed inline, so it makes sense to put them in the same div container.

rtibbles:
@MCGallaspy I think this is the right fix as it stands.

rtibbles:
Now fixes #3174 too.

rtibbles:
image

Added ( and ) keys too.

rtibbles:
N.B. - used character for space rather than 'Space' to avoid adding additional strings before release @aronasorman. @jtamiace if you can think of better ways to label it going forward, I am all ears, but no strings for now!

MCGallaspy:
@rtibbles you don't think a key with whitespace will be unambiguous enough? :P
Looks good to me. Does @jtamiace have merging privileges?

MCGallaspy:
Though I do wonder if that character will display properly on all systems/fonts that could potentially be used. Is that a valid concern? Should it be an image instead?

rtibbles:
It's Unicode, so I think it will be OK - but I am not wedded to it, if you and @jtamiace think just a blank box is a better indicator, happy to go with that.

MCGallaspy:
@rtibbles no, I agree that having something is better than just a blank button. (Much like doge speak, :P should be taken to mean the preceding comments isn't particularly serious.) If you don't anticipate any display issues I'm happy with this.

rtibbles:
Sure, let's merge.

PR #3156, opened 2015-02-28
Develop into central-develop

jamalex:


MCGallaspy:
This seems like an appropriate place to ask what the advantage for having a separate `develop` and `central-develop` branch. Can we just have the included submodule in the central server be `develop`? To what extent does the central server depend on the implementation details of the distributed server (including our own packages in `python-packages` like `securesync`)?

jamalex:
> This seems like an appropriate place to ask what the advantage is to having a separate develop and central-develop branch.

In an ideal world, these could be the same. But we've often needed to do extra stuff on the central-* branches that shouldn't (yet/ever) be in the master/develop branches. For example, the nalanda branch added some models (e.g. AttemptLog) that weren't part of 0.12, but since the nalanda code was being used in production, and needing to sync against the central server, the nalanda branch needed to be merged to central-master (e.g. see https://github.com/learningequality/ka-lite/compare/master...central-master)

Longer-term, we want to have:

- all projects/partnerships to be working on the mainline release cycle, which would reduce the need for this
- every large project to have its "own" central server, if it needs to have project-specific stuff enabled (this may only be feasible after we have Zone-specific databases/certificates)

> To what extent does the central server depend on the implementation details of the distributed server (including our own packages in python-packages like securesync)

Not sure whether this is exactly what you're asking, but yeah, the stuff inside ka-lite is used extensively within ka-lite-central, in particular:

- securesync itself
- all the `SyncedModel`s
- the coach reports
- the user management stuff

PR #3157, opened 2015-02-28
Burn import child modules down a little

### LINKED ISSUE #3042: Burn down get_all_import_child_modules
This function right here imports everything under a module. This is nice for convenience, but it's annoying as hell when writing tests and you don't want to import an entire module.

I also think that it's more idiomatic python to explicitly import the modules we need, and not hide it under an automatigical import statement.

rtibbles:
This fixes #3042, and as a bonus removes a bunch of unused imports from files across KA Lite.

Summary of changes:
* Stops using import_all_child_modules on all modules.
* Changes all instances of imports from those modules to explicit imports from relevant submodules.
* Removes unused imports from all modules.
* Explicitly imports all tests from inside a test dir (when we change to Django 1.7 this will be unnecessary).
* Cleans up some HTTP response codes in securesync and i18n.
* Adds a test to run through all URLs and ensure no 500s are thrown - should make sure no template  import errors are happening.

@MCGallaspy was this the kind of thing you had in mind?


rtibbles:
Also seems to speed up the build by a couple of minutes or so.

rtibbles:
@MCGallaspy After conversation with @aronasorman we want to merge this to develop after release-0.13, so I will leave this up as a WIP while I remove any uses of the function that shall not be named from tests as well.

MCGallaspy:
@aronasorman why do you want to wait to merge this?

rtibbles:
Note: this now completely fixes #3042.

aronasorman:
@MCGallaspy this is a big refactoring for me, and now that we're in the homestretch for 0.13 + the lack of test coverage, I'm not gonna be confident with our codebase unless we have at least two more bug bashes.

benjaoming:
@aronasorman agreed that a lot of testing has happened already, the best thing would be to keep releasing 0.13 with changes that warrant new tests and then merge this in the 0.14 release which is scheduled to go out soon, anyways.

But HURRAY!!! So many benefits from this cleanup!!!

rtibbles:
Added test to read all URLs to make sure no template import errors.

rtibbles:
Will throw an error, rather than a failure, so can be improved a little.

MCGallaspy:
@rtibbles such effort, much applause. @aronasorman okay, makes sense to stress test these changes during the development cycle instead of releasing them with 0.13 in the very near future.

rtibbles:
Rebased to resolve merge conflicts.

PR #3159, opened 2015-02-28
Kill the modal

rtibbles:
The modal was added for not good design reasons.

It is poorly implemented, and gives a terrible user experience, especially on low end tablets.
(Note: I implemented it).

Hence I am just removing it, and pushing all the messages through our messaging system.

I recommend merging this instead of #3151 @jtamiace and @MCGallaspy - sorry I didn't spot you working on this earlier.

### DIFF
--- a/kalite/distributed/static/js/distributed/distributed-server.js
+++ b/kalite/distributed/static/js/distributed/distributed-server.js
@@ -35,23 +35,6 @@ function show_api_messages(messages) {
     }
 }
 
-function show_modal(msg_class, msg_text) {
-    clear_modal();
-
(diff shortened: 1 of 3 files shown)


kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -35,23 +35,6 @@ function show_api_messages(messages) {
     }
 }
 
-function show_modal(msg_class, msg_text) {
-    clear_modal();
-
-    var msg_html = sprintf("<div class='alert alert-%1$s' id='modal'><a class='close' data-dismiss='alert' href='#''>&times;</a>%2$s</div><div id='fade'></div>", msg_class, msg_text);
```
aronasorman: KILL

aronasorman:
Tested by mastering 1 exercise. Can confirm that the modal is now dead.

PR #3160, opened 2015-02-28
No cache invalidation on videoscan

rtibbles:
Currently cache invalidation gets called repeatedly during videoscan. This prevents that, and makes it only run if the 'auto_cache' option is passed to it.

This has the result of only showing the language dub mappings once on startup, and improving start up times (following on from @MCGallaspy's lead in #3150).


rtibbles:
Reduced total build time by about 7 minutes.

PR #3166, opened 2015-03-01
Fixes and simplification to loadtesting app, for develop branch

jamalex:
This is just a cherry-pick of the main part of #3165.

### DIFF
--- a/kalite/distributed/templates/distributed/loadtesting/load_test.html
+++ b/kalite/distributed/templates/distributed/loadtesting/load_test.html
@@ -293,7 +293,7 @@
 
             var login_sequence = [
                 function(input, callback) {
-                    this.goto("/securesync/login/", callback);
+                    this.goto("/securesync/login/?next=/", callback);
                 },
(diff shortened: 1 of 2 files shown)


kalite/distributed/templates/distributed/loadtesting/load_test.html
```diff
@@ -268,16 +268,19 @@
                     this.goto("/learn/khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/", callback);
                 },
                 function(input, callback) {
-                    var numbers = this.$("span[class=mord][style]")
-                    var answer = 0
+                    var numbers = this.$("span.mn");
```
rtibbles: This won't work on new exercises, needs to be span.mord, or as above before it was changed.

rtibbles:
Seems fine, just needs the one correction on selectors for answering exercises rendered using the latest KE/Perseus.

jamalex:
Updated! And added in robustness code.

PR #3172, opened 2015-03-03
Remove the css classes mistakenly translated.

### LINKED ISSUE #3075: Running i18n-ize templates script translates css classes
Some css classes are defined in template blocks used, for example, to highlight an active tab. The i18n-ize script wraps these in translate tags. See [here](https://github.com/learningequality/ka-lite/blob/develop/kalite/control_panel/templates/control_panel/device_management.html#L8) for an example. Can we add some blocks (like `{% block users_active %}`) to a blacklist for the purposes of this script?

arceduardvincent:
Hi @aronasorman and @MCGallaspy -this fixes #3075

> I remove the css classes mistakenly translated and revert it to the intended purpose. 
> I wrapped the css classes in a maker `block` that can mark this this blocks ` {% block i18n_do_not_translate %}` as not to be translated.
> As long as this `css classes` are mark within this block it will not be translated.

See the implementation and the fix here https://github.com/fle-internal/i18nize_templates/pull/3

/review: @jesumer

### DIFF
--- a/kalite/control_panel/templates/control_panel/account_management.html
+++ b/kalite/control_panel/templates/control_panel/account_management.html
@@ -4,11 +4,13 @@
 {% load my_filters %}
 {% load include_block %}
 
-{% block username_active %}{% trans "active" %}{% endblock username_active %}
-{% block control_panel_active %}{% endblock control_panel_active %}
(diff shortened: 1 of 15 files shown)


rtibbles:
All the fixes here look good.

PR #3173, opened 2015-03-03
fix for #2980 - Answer box highlighting is missing left border.

### LINKED ISSUE #2980: Answer box highlighting is missing left border
Again on `develop`, again at `/learn/khan/math/early-math/cc-early-math-geometry-topic/cc-early-math-composing-shapes/compose-shapes/`, I clicked the check answer button without first clicking any of the radio buttons for the answer. It helpfully highlighted the answer area by creating a border, but the left side of the border was cut off. Screenshot included:

ss

djallado:
Hi. @MCGallaspy. This fixed #2980 - Answer box highlighting is missing left border.

Milestone: 0.14.x
No test for now because this is a simple box-shawdow in css.

Screenshot below:
screen shot high-light

### DIFF
--- a/kalite/distributed/static/css/distributed/khan-lite.css
+++ b/kalite/distributed/static/css/distributed/khan-lite.css
@@ -987,3 +987,7 @@ to items you want to have the icon appear with.
     -ms-user-select: none;
     user-select: none;
 }
+
+#workarea .perseus-widget-container.widget-highlight {
+    box-shadow: -4px 0 9px 5px #ffa500;
+}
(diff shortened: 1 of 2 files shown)


rtibbles:
@djallado This looks like the right fix, but would prefer to avoid inline style definitions. Could you put this in khan-lite.css instead?

djallado:
@rtibbles. `khan-lite.css` is overridden by `/exercise-content-package/perseus.css`. As much as possible, I don't want to pollute the `perseus.css`. So I put inline style definitions just to override the `perseus.css`.

rtibbles:
Two options then - we add an exercise.css that we import after perseus.css to overwrite its styles, or we make the selector more specific than the perseus one.

For example:
```
#content-area .perseus-widget-container.widget-highlight {
    box-shadow: -4px 0 9px 5px #ffa500;
}
```
Should provide a more specific selector than that provided in perseus.css and hence should override.
Either way, a solution without inline styles is the preferred one.

rtibbles:
Does the trick. Thanks for fixing up nicely, @djallado!

PR #3180, opened 2015-03-03
docs - screenshot directive (and management command) can take screenshot as though the device was registered

MCGallaspy:
Because previously it couldn't. Does this by creating an associated Zone and DeviceZone for the Device, so that `Device.is_registered()` returns `True`.

Also touched up `sphinx-docs/screenshot.py` a little bit while I was in there, and added the appropriate option to a couple of screenshots.

### DIFF
--- a/sphinx-docs/screenshot.py
+++ b/sphinx-docs/screenshot.py
@@ -203,6 +203,7 @@ class Screenshot(Image):
     option_spec['user-role'] = _parse_user_role
     option_spec['navigation-steps'] = _parse_nav_steps
     option_spec['focus'] = _parse_focus
+    option_spec['registered'] = lambda x: directives.choice(x, ("true", "false"))
 
     def run(self):
         """ During the build process directives are parsed as nodes, and then
@@ -222,6 +223,11 @@ class Screenshot(Image):
         return_nodes = []
         if not hasattr(self.env, 'screenshot_all_screenshots'):
             self.env.screenshot_all_screenshots = []
+
+        if not 'registered' in self.options:
+            self.options['registered'] = False
+        else:
+            self.options['registered'] = True if self.options['registered'] == "true" else False
 
         if 'focus' in self.options:
             self.focus_selector = self.options['focus']['id']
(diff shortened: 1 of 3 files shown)


sphinx-docs/screenshot.py
```diff
@@ -283,21 +281,33 @@ def _command_handler(self, commands):
...
+            new_arg["focus"]["selector"] = self.focus_selector
+            new_arg["focus"]["styles"] = FOCUS_CSS_STYLES
+            new_arg["notes"] = self.focus_annotation if hasattr(self, "focus_annotation") else ""
+        new_arg["registered"] = self.options["registered"]
+        return new_arg
+
+    def _make_image_node(self):
+        """Make an image node by safely calling Image.run (i.e. ensure the file exists)."""
         self.arguments.append(os.path.join("/", SS_DUMP_DIR, self.filename+".png"))
         open(os.path.join(OUTPUT_PATH, self.filename+".png"), 'w').close()
```
aronasorman: Hmm strange, we open the file and close it immediately? Is this a `touch`? Think we have an idiom for that already from the previous review.

aronasorman: Also, spaces around the `+`.

sphinx-docs/screenshot.py
```diff
@@ -203,6 +203,7 @@ class Screenshot(Image):
     option_spec['user-role'] = _parse_user_role
     option_spec['navigation-steps'] = _parse_nav_steps
     option_spec['focus'] = _parse_focus
+    option_spec['registered'] = lambda x: directives.choice(x, ("true", "false"))
```
aronasorman: Interesting, so this is enforced by sphinx?

sphinx-docs/screenshot.py
MCGallaspy: It is indeed a `touch`. `utime` doesn't create files (just updates access time). This seems to be a standard way to do it.

sphinx-docs/screenshot.py
MCGallaspy: Do you mean are the values "true" and "false" enforced by Sphinx? Yes. It'll throw an error (without affecting other screenshots) if the option is specified incorrectly.

MCGallaspy: However, failure to specify will default to `False`, which was the behavior before these changes -- screenshots were taken as though the device was not registered. In other words, these changes don't backwardly break anything.

PR #3183, opened 2015-03-05
Add FAQ to the docs.

MCGallaspy:
Does what it says. FAQ from [here](https://learningequality.org/ka-lite/faq/).

### DIFF
--- a/sphinx-docs/faq.rst
+++ b/sphinx-docs/faq.rst
@@ -74,6 +74,7 @@ What are some possible hardware configurations for deploying KA Lite?
 ---------------------------------------------------------------------
 
 You will need:
+
 1. A computer that is running the KA Lite software (e.g. a desktop computer, laptop, or Raspberry Pi).
 2. One or more client devices that have web browsers (laptops, tablets, desktop computers, etc)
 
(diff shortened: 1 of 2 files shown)


sphinx-docs/faq.rst
```diff
@@ -0,0 +1,191 @@
...
+
+How do I install KA Lite?
+-------------------------
+
+Information on how to install KA Lite is available via our :doc:`user guides </usermanual/userman_main>`.
+
+How much does KA Lite cost to install?
+--------------------------------------
+
+It is **FREE**.
```
aronasorman: Might want to add an additional short sentence about us being free (gratis -- 0 USD) and free (open source, free to modify etc.) and that most of our content is open-licensed too.

aronasorman:
Just one suggestion. Once that's fixed, I'll leave it to you @MCGallaspy to merge.

mjptak:
perhaps a hook like this as well.   (from the mothership site.

MCGallaspy:
@mjptak not clear what you mean... should I add an item to the FAQ, or are you suggesting the hook goes on all the pages in the docs?

mjptak:
perhaps both but I think that little piece in the code has been pretty
effective for when I have tried to turn folk (with skills) onto Khan
Academy.  (gets back to the  "when they hit view source button we have
already won (:))".  Keep rockin it.

sphinx-docs/faq.rst
```diff
@@ -0,0 +1,191 @@
...
+
+How do I install KA Lite?
+-------------------------
+
+Information on how to install KA Lite is available via our :doc:`user guides </usermanual/userman_main>`.
+
+How much does KA Lite cost to install?
+--------------------------------------
+
+It is **FREE** -- both free as in "free speech" and free as in "free beer"!
```
aronasorman: Can you add this link too? http://www.fsf.org/about/what-is-free-software

PR #3185, opened 2015-03-05
fixed points not shown when logged in as learner. note: html id cannot c...

66eli77:
...ontain space

### DIFF
--- a/kalite/distributed/templates/distributed/base.html
+++ b/kalite/distributed/templates/distributed/base.html
@@ -276,8 +276,8 @@
                                 <span class="caret"></span>
                               </a>
                               <ul class="dropdown-menu" role="menu">
-                                <li role="presentation student-only">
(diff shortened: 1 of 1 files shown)


kalite/distributed/templates/distributed/base.html
```diff
@@ -277,7 +277,7 @@
                               </a>
                               <ul class="dropdown-menu" role="menu">
                                 <li role="presentation student-only">
-                                  <span id="points student-only">
```
MCGallaspy: This error is an oversight from a PR that I made (#3127). The intention was that `student-only` be a class. Can you change this span to have the class `student-only`, and fix the role of the `li` element on line 278? It should also have class `student-only` and the role should just be `presentation`.

66eli77: done

PR #3188, opened 2015-03-06
Software keypad only able to select one input

### LINKED ISSUE #2944: Software keypad only able to select one input
On the `develop` branch, if multiple inputs are available, then the software keypad only writes to the "first" one. For example, [this exercise](http://127.0.0.1:8008/learn/khan/math/early-math/cc-early-math-counting-topic/cc-early-math-counting/count-from-any-number/) could have multiple elements with `.perseus-input` class (which is how the sw keyboard selects its input, I believe). This is potentially an issue with Khan exercises too, though in that case I think the sw keypad selects the input element by ID(?) which might suggest it's meant to be unique.

amodia:
Hi @MCGallaspy  this fixes #2944 

default

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -118,7 +118,18 @@ window.ExerciseView = Backbone.View.extend({
     events: {
         "submit .answer-form": "answer_form_submitted",
         "keyup .perseus-input": "click_check_answer_button",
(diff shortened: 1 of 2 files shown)


cpauya:
@amodia Did you check what happens when the user submits the answers with your inputs?

amodia:
yeah @cpauya  i checked it and it will submit curectly.

MCGallaspy:
Looks good to me!

PR #3190, opened 2015-03-06
IE8 compatibility fixes

### LINKED ISSUE #3171: Cannot input into answer field using Internet Explorer
When using IE I am not able to enter a value into the answer fields while doing an exercise.

MCGallaspy:
Begins to address #3171. Conditional inclusion of pdfjs stuff and inclusion of `es5-sham.js` fixes (or at least suppresses) some errors, but (from #3171):
> Nope, KA doesn't support IE8. Given that we include the KA perseus stuff, is it still reasonable to expect that we should support IE8 too? Presumably we'll have to modify the inclusions, which isn't necessarily the best idea if we want to update them in sync with KA. Do we need to rethink how we include KA assessment items?

Fixes #3171.

### DIFF
--- a/kalite/distributed/templates/distributed/learn.html
+++ b/kalite/distributed/templates/distributed/learn.html
@@ -52,6 +52,7 @@
     {% if load_perseus_assets %}
         <!-- START PERSEUS STUFF -->
         <script src="{% static "perseus/lib/es5-shim.js" %}"></script>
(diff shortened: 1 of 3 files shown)


aronasorman:
Is this necessary now that we're dropping IE8 support?

MCGallaspy:
Suppose not.

PR #3044, opened 2015-02-13
fixing bug 2993 'no login time units'

sethachoi:
fixes #2993 
Fixed the bug with the facilities management tab not displaying a unit of time under "login time" for each user type (coach/group/student).
Went from login time : 0 
to login time : 0 hour(s)

PR #3196, opened 2015-03-08
fixing bug #2993-Adding hours as a time unit next to column name login time 

### LINKED ISSUE #2993: No "Login Time" units
On `develop`, at a facility page (click on one of the manage tabs, and then click on a facility), several of the tables there have a column called "Login Time", but it's not clear what units that value is measured in (seconds? days? microfortnights?):

ss

tarunsharma1:
Bug Fix for 
No "Login Time" units #2993
Instead of appending hrs to the time and thereby violating the string freeze, i have added the words 'in hrs' to the column name 'Login time' itself. This makes the units used clear and does not violate string freeze.
I have changed the string 'Login time' to 'Login time(in hrs)' in the  files _coaches_table.html, _groups_table.html, _group_summary.html,_students_table.html located in ka-lite/control_panel/templates/control_panel/partials
screenshot from 2015-03-08 06 02 17
screenshot from 2015-03-08 07 47 07
I have tested it on my local machine and the result is shown in the screenshot.

Fixes #2993

### DIFF
--- a/kalite/control_panel/templates/control_panel/partials/_coaches_table.html
+++ b/kalite/control_panel/templates/control_panel/partials/_coaches_table.html
@@ -40,7 +40,7 @@
                                 <th>{% trans "Coach" %}</th>
                                 <th>{% trans "Edit" %}</th>
                                 <th>{% trans "Logins" %}</th>
-                                <th>{% trans "Login Time" %}</th>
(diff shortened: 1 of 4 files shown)


MCGallaspy:
@aronasorman does this truly not violate the string freeze? Should we merge this given that we also have #3044?

PR #3201, opened 2015-03-09
This fixes #2926 to start KA Lite on Windows boot.

### LINKED ISSUE #2926: KA Lite start at boot does not work on Windows.
I've installed KA Lite on Windows XP and Win 7 and never been able to get KA Lite to start automatically on boot. There are frequent power outages and this function is really important to us at VIrginia DOC.

In our pilot, I've installed servers running both WIndows and Linux. Running Linux this works perfectly. Is this a known issue?

cpauya:
Hi @MCGallaspy / @aronasorman - this fixes #2926 initially.

I wrapped the batch file filename in quotes to successfully create the task in Windows.  See screenshot .gif.

With this fix, if the user selects "Options -> Run KA Lite at system startup." - this creates the task with the right path and thus Windows is able to run the scheduled task on system start.  One caveat, if the user is required to input password to login, ~~this will not work yet~~ this already works (see ToDos below).

Remaining ToDos:

- [x] Change the user account to run the task into SYSTEM (currently it is the user who installed the app).
- [ ] If auto-logon is enabled for the Windows user, the KA Lite app should also run automatically when the "Options -> Run KA Lite at system startup" is enabled.

Question/s:  The options are confusing me:

1. "Options -> Run KA Lite when the user logs in." ==> The KA Lite app is run upon user login, that's simple.
2. "Options -> Run KA Lite at system startup." ==> Assuming that auto-login is not enabled for the Windows box and that the user has to input a password, then this is only possible if KA Lite is a windows service?  Which upon checking, it is not.
3. "Options -> Auto-start server when KA Lite is run." ==> If users are required to input password to login, this is only possible if we run the scheduled task based on `add_systemstart_task.bat` as a `SYSTEM` user.  Else, maybe we can leave the current setting which runs the KA Lite app using the credentials of the user who installed it.

I think the only possible options would be:

1. Run KA Lite when the user logs in. ==> Only the UI/window with systray icon is run, but to run the server automatically, user must enable the second option below.
2. Auto-start server at system startup.  ==> The KA LIte server is run using the credentials of the SYSTEM user?  So it can run even when no user is logged-in.

I don't know if these comments must be put on another issue?  Please enlighten me.

Screenshot .gif running on [IE8 on Windows 7 Enterprise](https://www.modern.ie/en-us/virtualization-tools).
ka-lite

/cc: @ruimalheiro @jamalex @EdDixon

### DIFF
--- a/scripts/add_systemstart_task.bat
+++ b/scripts/add_systemstart_task.bat
@@ -2,4 +2,4 @@
 set "SCRIPT_DIR=%~dp0"
 
 TITLE Adding task to start KA Lite at system start
-schtasks /create /tn "KALite" /tr "%SCRIPT_DIR%\start.bat" /sc onstart & pause
+schtasks /create /tn "KALite" /tr "'%SCRIPT_DIR%\start.bat'" /sc onstart & pause
--- a/scripts/add_systemstart_task.bat
+++ b/scripts/add_systemstart_task.bat
@@ -2,4 +2,4 @@
 set "SCRIPT_DIR=%~dp0"
 
 TITLE Adding task to start KA Lite at system start
-schtasks /create /tn "KALite" /tr "'%SCRIPT_DIR%\start.bat'" /sc onstart & pause
+schtasks /create /tn "KALite" /tr "'%SCRIPT_DIR%\start.bat'" /sc onstart /ru "SYSTEM" & pause


cpauya:
Hmm... one issue with running it under the `SYSTEM` user account is that the "Stop Server." entry from the systray app does not work anymore... doing further investigation.

aronasorman:
Marked as high priority, as we don't want any regressions.

MCGallaspy:
Win7, trying to run the script from cmd.exe, I get `ERROR: Access is denied.`

MCGallaspy:
@cpauya I think I agree that the best approach for start-at-boot is to make this a windows service. Scheduled tasks seems to be kind of brittle.

ruimalheiro:
@MCGallaspy the reason you get "Access is denied" is because you are not running cmd as administrator.

MCGallaspy:
@rui @cpauya works on Windows 7.

MCGallaspy:
@cpauya regarding your to-do, the server seems to start fine with auto-login for me. Has anyone tested this on XP or 8?

MCGallaspy:
Works in 7 & 8, but still having issues with XP. The service is created just fine, but it fails to start.

MCGallaspy:
Schedule tasks log when I unsuccessfully try to tun the KALite scheduled task in XP:
```
"KALite.job" (start.bat') 3/10/2015 12:57:41 PM ** WARNING **
	Invalid working directory..
	The specific error is:
	0x0000007b: The filename, directory name, or volume label syntax is incorrect.
	Verify that the directory exists and try again.
"KALite.job" (start.bat') 3/10/2015 12:57:41 PM ** ERROR **
	Unable to start task.
	The specific error is:
	0x80070002: The system cannot find the file specified.
	Try using the Task page Browse button to locate the application.
"KALite.job" (start.bat') 3/10/2015 12:57:47 PM ** WARNING **
	Invalid working directory..
	The specific error is:
	0x0000007b: The filename, directory name, or volume label syntax is incorrect.
	Verify that the directory exists and try again.
"KALite.job" (start.bat') 3/10/2015 12:57:47 PM ** ERROR **
	Unable to start task.
	The specific error is:
	0x80070002: The system cannot find the file specified.
	Try using the Task page Browse button to locate the application.
```

So it looks like an error in `start.bat`?

MCGallaspy:
Given that these scripts are being deprecated in favor of the `bin/kalite` and `bin/kalite/kalite.bat` commands, can we simply edit `add_systemstart_task.bat` to use the latter command?

aronasorman:
Superseded by https://github.com/learningequality/ka-lite/issues/3274

PR #3207, opened 2015-03-09
Remove unused javascript and css

rtibbles:
Minor refactor. Summary of changes:

* Removes javascript and css files with no extant import reference in any html template, or loaded dynamically via javascript.
* Moves jquery tipsy js and css files into static-libraries.


MCGallaspy:
You removed `static-libraries/js/jquery.js`? Is jquery really unused?

MCGallaspy:
Disregard, I see it's being included under a different name.

PR #3209, opened 2015-03-09
Sidebar now responsive as ever

Antrikshy:
I have made some fancy CSS and Backbone changes that make the whole thing embrace various screen sizes more than ever before. I had never thought I'd see this happen anytime soon, but the sidebar now literally behaves like a Bootstrap component, resizing live with the window size. While I feel there is more optimization to do going forward, this is a stable enough update to be merged.

This update builds on top of #3139 and adds the following:

- Complete auto-resizing of the sidebar when window width changes (in 100px steps, too!), because users will inevitably play around with the glorious responsiveness once they discover it. ;)
- Media queries to change the text size based on screen width. I read somewhere that when implementing responsive typography, text sizes are lowered on smaller screens because they are held closer to the face.
- Probably other stuff that I have forgotten.

#### Largest size (> 1300px)
screen shot 2015-03-09 at 1 32 12 pm

#### Smallest size (504px)
screen shot 2015-03-09 at 1 32 17 pm

Suggestions for tweaks and improvements welcome (are the font sizes okay?).

>Note: I have not yet implemented the new color scheme because I can't make it look good enough on the huge sidebar. It's too saturated for that. Anyways, it will be a significant enough change to warrant for its own PR anyway because it will need several changes across the entire website.


Antrikshy:
Aaaaand I found a bug just after opening this PR. Resizing window with the sidebar closed makes it appear without actually *opening* it. Should be easy enough to fix actually.

Antrikshy:
Aaaaand fixed! That was easy.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -104,28 +115,33 @@ window.SidebarView = BaseView.extend({
     },
 
     resize_sidebar: function() {
-        if ($(window).width() < 1260)
-            this.resize_for_narrow();
+        if (this.state_model.get("open")) {
+            if ($(window).width() < 1260)
+                this.resize_for_narrow();
```
MCGallaspy: Noooo! Use the curly brackets! I don't know why this passed the linter. (If you truly must omit curly brackets, then at least don't put the statement to be executed on a new line... it suggests that more lines could be added to this `if` block, but really only the first statement will be conditionally executed and all subsequent ones will be unconditionally executed).

Antrikshy: Ooh I didn't know we were doing that. I always add them when needed and ignore when I can. What linter? Is it something on Travis?

MCGallaspy: Yeah, Travis has a linter. My immediate neighbors in the office agreed that braces were stylistically preferable even though not necessary here.

MCGallaspy:
We're antsy to merge this, so I'll just add the brackets.

Antrikshy:
@MCGallaspy You guys dig it so much, eh?

PR #3212, opened 2015-03-10
Properly update statusModel points

### LINKED ISSUE #3186: Learner's points get incorrectly increased when doing exercise
Branch: develop
Expected Behavior: Doing a exercise, when correctly answer a question, earned points for only this question get added to total points.
Current Behavior: Doing a exercise accumulated earned points including from all previous questions get added to total points.

Steps to reproduce:
1. log in as learner
2. choose a exercise that contain multiple questions
3. answer questions under that exercise and check for points you earn and your total points.

screen shot 2015-03-05 at 3 21 55 pm
screen shot 2015-03-05 at 3 22 17 pm
screen shot 2015-03-05 at 3 23 03 pm

rtibbles:
Fixes #3186 

Summary of changes:
* Track how many points have been piled onto statusModel points
* Directly increase 'points' on statusModel

This seems to be a robust way of doing it (I thought it might be problematic when switching to and from different exercises, but this way of tracking it accounts for that).

### DIFF
--- a/kalite/distributed/static/js/distributed/content/views.js
+++ b/kalite/distributed/static/js/distributed/content/views.js
@@ -199,7 +199,7 @@ window.ContentPointsView = BaseView.extend({
 
     render: function() {
         this.$el.html(this.template(this.model.attributes));
-        statusModel.set("newpoints", this.model.get("points") - this.starting_points);
+        statusModel.update_total_points(this.model.get("points") - this.starting_points);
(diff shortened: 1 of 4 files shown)


MCGallaspy:
I tested this out by watching some videos and doing some exercises. Works as expected. I would really like to see a test for this, but since the one we have for the similar earlier points issues (#2889) doesn't run on Travis I guess we'll skip it for now.

PR #3213, opened 2015-03-10
0.13.0 Release Candidate

aronasorman:
This getting merged to the 0.13.x signifies the start of the beta process.. After this, only critical bugfixes and documentation tweaks shall go into 0.13. We will also release a final beta to our users before the grand announcement.


aronasorman:
There are still some release blockers for 0.13, so we'll hold off on mergin this 'til all of those are closed or punted.

PR #3216, opened 2015-03-10
slight changes to css styling in top nav

jtamiace:
There was a slight css bug in the top nav that this fixes


MCGallaspy:
@jtamiace what does it change exactly? Can you show a before/after?

MCGallaspy:
Actually, I'm okay with these changes. Just checked them out myself.

Issue #2961, opened 2015-02-09
Unify the color on table headers in admin UI


Antrikshy:
Someone add the beginner tag. I don't mind being assigned to this either. I'll work on it once I've got that sidebar width issue fixed (#2949). Just putting it here so I don't forget.

In case someone else wants to fix this, try the following:
1. Add `.faded-green-bg` class to the table headers.
2. Change the name of this class across the entire codebase (since now it's blue), so that the colors update automatically whenever we change the color for the main website header and footer.
3. *Bonus:* Add the `.table-hover`, `.table-striped` and `.table-bordered` Bootstrap classes to the tables to make them look like the new admin table design I have in a PR.

Screenshot:
![screen shot 2015-02-09 at 1 20 18 pm](https://cloud.githubusercontent.com/assets/6601788/6116050/ad0bc21c-b05e-11e4-932e-7a631b2cd3dd.png)

PR #3219, opened 2015-03-10
Hotfix/2961 - Refactored some template classes, fixed pesky .faded-green-bg

Antrikshy:
#2961 can finally be closed. This PR contains some minor changes to the CSS. The light green table headers in the Manage tab now match the light blue header and footer. Took me a while to understand that the green color came from Bootstrap's CSS.

Now, changing the light blue in one CSS file will change the light blues across the entire site. Be sure to use the `.header-footer-bg` class wherever you want a background that matches the site header and footer. That's all it adds.

### DIFF
--- a/kalite/control_panel/templates/control_panel/device_management.html
+++ b/kalite/control_panel/templates/control_panel/device_management.html
@@ -132,7 +132,7 @@
 
                 <table class="table">
                     <thead>
-                        <tr class="success">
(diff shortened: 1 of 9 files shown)


MCGallaspy:
Just a semantic change? Sounds good.

Antrikshy:
@MCGallaspy For the most part. I've been meaning to do these things for a while.

PR #3246, opened 2015-03-10
[+] Added exception handling for _cpmodpy.py. If the module doesn't exis...

jaredestroud:
...t, it will display an error message to the user and gracefully exit rather than just crash.


aronasorman:
Hey @jaredestroud! We don't bugfix and feature PRs directly to master. Can you make another PR targeting the develop branch?

PR #3250, opened 2015-03-11
corrected the path and deleted unnecessary file modification in kalite Docs

66eli77:
the file modifications actually cause trouble on Raspberry Pi


aronasorman:
I trust that this has been tested in the RPi.

Issue #3249, opened 2015-03-10
Remove ugly hash from beside logged-in username


jamalex:
We added this 8-digit hash just for paper survey tracking purposes for Nalanda. Probably no use for it in general, and it's ugly/distracting, so let's ditch it.

![image](https://cloud.githubusercontent.com/assets/618416/6587761/b7904e30-c745-11e4-84b1-4e34c9820661.png)

jamalex:
Could you take a look, @anuragkanungo?

66eli77:
@jamalex  @anuragkanungo  you can assign to me this issue, I fixed this on my personal branch yesterday.

anuragkanungo:
@66eli77 , If you have fixed it then please go ahead take this issue and open the PR.
Thanks.

66eli77:
@aronasorman  this has been fixed at PR #3251, please close the issue

PR #3251, opened 2015-03-11
update RPi installation Docs and fixed don't display user-id issue for #3249

66eli77:


aronasorman:
Strange, some of the changes here are repeated in https://github.com/learningequality/ka-lite/pull/3250. Oh well.

PR #3252, opened 2015-03-11
Fix coachreport import error

### LINKED ISSUE #3224: ImportError on coachreports
Browser: Firefox 35
Branch: develop

Giving these specific parameters for a coachreport:
screen shot 2015-03-10 at 1 22 26 pm

Gives me this error:
screen shot 2015-03-10 at 1 24 21 pm

MCGallaspy:
Fixes #3224, which was caused by an unprefaced import statement & subsequent name error.
Additionally, several inline imports were, I think, added as a sneaky way to circumvent
a circular import... that is coachreports.views was importing control_panel.api_resources,
and vice-versa. So I just moved the offending function out of coachreports.views into
the facility package, since that was its only dependency.


MCGallaspy:
@aronasorman can you try to reproduce your error using this?

aronasorman:
Can confirm that the error is no longer there.

Issue #3208, opened 2015-03-09
Installation requirements in the installation guide


aronasorman:
We need:
1. server requirements
2. client requirements (i.e. can't be using  < IE 8)

PR #3262, opened 2015-03-11
Note that IE8 is not supported

MCGallaspy:
Addresses #3208.


aronasorman:
We'll also add another note in the release notes section regarding IE 8: https://github.com/learningequality/ka-lite/issues/3263

PR #3265, opened 2015-03-12
Refactor so convert_urls function can be used elsewhere

MCGallaspy:
Needed to resolve [an issue in the central server](https://github.com/fle-internal/ka-lite-central/issues/238).


aronasorman:
Can you write a simple test for the `convert_urls` function?

jamalex:
Does this also download these other formats? Or only convert them?

aronasorman:
No downloading going on here @jamalex. It was just refactored out so that the conversion can be used in the central server.

MCGallaspy:
@aronasorman Holding me accountable

aronasorman:
nondeterministic test failure. Merging.

Issue #3263, opened 2015-03-11
Release notes in the manual


aronasorman:
Add:

1. people need to run setup again when updating through Git
2. Windows installer now requires admin privileges
3. We also don't support IE 8 anymore

aronasorman:
Fixed in #3266.

PR #3266, opened 2015-03-12
Add release notes

MCGallaspy:
Addresses #3263.


aronasorman:
We also want to add that `kalite/manage.py` has been deprecated, and they should now use `bin/kalite`.

PR #3015, opened 2015-02-10
Coverage.io third or fourth stab, will it work?

benjaoming:

PR #3267, opened 2015-03-12
Run tests under coverage, add badge

MCGallaspy:
Replaces #3015.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_install:
 install:
 - pushd $TRAVIS_BUILD_DIR; npm install; popd
 - pushd $TRAVIS_BUILD_DIR; npm install -g grunt-cli; popd
+- pip install coveralls
 script:
 - cd $TRAVIS_BUILD_DIR
 - bin/kalite start --traceback -v 2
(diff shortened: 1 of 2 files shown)


aronasorman:
yey

aronasorman:
Does coverage error out at some point due to insufficient test coverage?

aronasorman:
oh, apparently not.

aronasorman:
So it seems like it dumps it to some file, then uploads it to coveralls.io.

Here is the run for the last commit: https://coveralls.io/jobs/4864186

68% coverage is higher than I expected. Although based on the stats, it only covers the actual `bin/kalite` file, and not all of our source files.

aronasorman:
So we may need to find a way to make `coveralls` cover all our source files. Maybe drop straight to `kalitectl.py`?

rtibbles:
Or we could add a coverage config to tell it where the source files we are testing are located?
https://pypi.python.org/pypi/coveralls/1.0a2

MCGallaspy:
Also, it's not gathering coverage data about any js, since we're not running any js tests. Just python. Could artificially inflate the coverage %.

benjaoming:
According to my stats, it's sitting at 23%... run `coverage report` from command line to get the output.

benjaoming:
@MCGallaspy 

These things, I've already done... both added the badge and the --source option.. not sure why you're replacing the PR? Is it okay that I take back this one? ;)

MCGallaspy:
@benjaoming I wanted it to get merged. :( But it's still languishing.

MCGallaspy:
@benjaoming feel free to take it back

rtibbles:
I think this was fixed in a @benjaoming PR eventually. So this is stale.

PR #3274, opened 2015-03-12
Escape spaces in variable, use supported start command

MCGallaspy:
If the SCRIPT_DIR variable has spaces in it, then it needs to be escaped. Tested in Windows 7, scheduled task runs fine. This issue exists in 0.12 too.

### DIFF
--- a/scripts/add_systemstart_task.bat
+++ b/scripts/add_systemstart_task.bat
@@ -2,4 +2,4 @@
 set "SCRIPT_DIR=%~dp0"
 
 TITLE Adding task to start KA Lite at system start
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Fixes #2926.

MCGallaspy:
Also we can close #3201 if this is merged, as we're both attempting to fix the same issue (but that one uses a deprecated start script, and I'm not sure if it'll be escaped correctly still).

cpauya:
:+1: nice one @MCGallaspy

aronasorman:
Tests pass. I'll merge this now, then @EdDixon can you test this out?

PR #3282, opened 2015-03-13
Develop to central-develop

aronasorman:


aronasorman:
Some test failures. Could you try fixing the tests @MCGallaspy?

MCGallaspy:
Regarding the js linting error, that file `kalite/coachreports/static/js/coachreports/exercise_mastery_view.js` exists on `central-develop` but not `develop`... Meaning these branches diverged at some point? `develop` is not simply a fast-forwarded version of `central-develop`? @aronasorman if you initiated this merge through the web interface then I think github attempts a --no-ff merge, which is probably not what we want here.

MCGallaspy:
We should try to merge `central-develop` with `develop` locally, and then submit that as a PR.

aronasorman:
Nondeterministic test failures (again). Merging this manually.

PR #3289, opened 2015-03-13
Destroys any existing tooltips when exercise view is closed.

### LINKED ISSUE #3286: Navigating to a new exercise while the input hint tooltip is open causes it to float to a weird position on the page
Branch: develop
Expected behaviour: tooltip will disappear
Current behaviour: tooltip persists, moves to top left corner, and inhibits appearance of new tooltips

image

rtibbles:
Fixes #3286

Destroys any existing tooltips when an exercise_view is closed, and hence prevents phantom tooltips from floating around the screen.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -422,6 +422,7 @@ window.ExerciseView = Backbone.View.extend({
         if (this.related_video_view) {
             this.related_video_view.remove();
         }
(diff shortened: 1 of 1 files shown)


MCGallaspy:
How did you manage to display a tooltip and then navigate to a different page without the tooltip vanishing? I'm trying to reproduce so that I can test this solution.

rtibbles:
Using the back/forward buttons is easiest (navigate to one exercise, and then directly to another).

I believe also pressing enter repeatedly to submit an answer and then continue to the next question should prompt this behaviour also.

PR #3300, opened 2015-03-14
Div overlap when scratchpad displayed, improved menu items Contrast ratio and added search form label for accessibility

natnael2012:
When scratchpad is displayed, the border overlaps the container.

 

Improved menu items Contrast Ratio for web accessibility
contrast ratio


rtibbles:
Adds new strings so won't go in for 0.13.

kalite/distributed/templates/distributed/base.html
```diff
@@ -181,9 +182,10 @@
                             <li class="hidden-xs">
                               <form class="navbar-form" action="{% url 'search' %}" method="get" id="search-box" role="search">
                                 {% comment %}translators: this will appear in the navigation bar. please try to keep it as short as possible.{% endcomment %}
-                                <div class="input-group">
+								<p class="search-label pull-left"><label for="search">Search</label></p>
+								<div class="input-group">
                                   <input type="text" name="query" id="search" class="form-control" placeholder="{% trans 'Topic, video, exercise...' %}" />
-                                  <button id="search-button" class="btn btn-default" type="submit"><i class="glyphicon glyphicon-search"></i></button>
+								  <button id="search-button" class="btn btn-default" type="submit"><i class="glyphicon glyphicon-search"></i></button>
```
rtibbles: Could you wrap new strings in Django {% trans '' } tags, please?

kalite/distributed/templates/distributed/base.html
```diff
@@ -168,9 +168,10 @@
                           <ul class="nav navbar-nav navbar-right nav-tabs top-nav" role="tablist">
                               <div class="visible-xs">
                                   <li id="search_input">
-                                      <form class="navbar-form " action="{% url 'search' %}" method="get"  role="search">
+									  <form class="navbar-form " action="{% url 'search' %}" method="get"  role="search">
```
rtibbles: Any particular reason for this huge indent?

kalite/distributed/templates/distributed/base.html
```diff
@@ -168,9 +168,10 @@
...
                                           {% comment %}translators: this will appear in the navigation bar. please try to keep it as short as possible.{% endcomment %}
-                                          <input type="text" name="query" id="search_box" class="form-control pull-left" placeholder="{% trans 'Topic, video, exercise...' %}" />
+                                          <p class="search-label search-label-mini"><label for="search_box">Search</label></p>
```
rtibbles: I am concerned how the addition of text will affect layout, could you post a screenshot of how this renders?

Also include screenshots for smaller form factors - if you are unfamiliar with how to do this, here are instructions for doing so in Chrome: https://developer.chrome.com/devtools/docs/device-mode

rtibbles:
@jtamiace @Antrikshy @sethachoi Please see the colour scheme for the sidebar here - I am not sure it is what we are aiming for, but the contrast is helpful for accessibility. Thoughts?

kalite/distributed/templates/distributed/base.html
natnael2012: Oh, I didnt know the indent was altered. :)

kalite/distributed/templates/distributed/base.html
natnael2012: Yes It affects the layout, But such measure should be taken to make it accessible well. I have also inspected using screen reader to look at how it affects users who use screen readers. I think this is a good solution.

Antrikshy:
@natnael2012 @rtibbles I think the green is too dark. We are actually working on making the sidebar match the new color palette we voted on, but that will require some _very_ careful color selection that falls somewhere in between the current color, which looks pretty nice, and the new greens. This is tricky work, so it may take a bit. Such drastic changes require a lot of thought.

The contrast may make the text more readable, but personally, I'm not a fan of this new green across the site and the sidebar. It looks a bit out of place and not as fresh as the new branding.

natnael2012:
@Antrikshy @rtibbles You are right. It requires very careful selection of colors. What is the new color you selected? I can work on inspecting the colors for Contrast Ratio since its big influence for Web Accessibility.

66eli77:
the color contrast definitely highlight the chosen item text but making other item text even less readable than the standard version. I think it's important to make all text more readable.

natnael2012:
@66eli77 The less readable text indicate missing content. Here is how it looks like at the standard version. 

untitled

When it comes to accessibility, according to my inspection, the text elements are hardly readable by users with low vision. The menu Items has Contrast Ratio: 2.91:1, (background color #5aa685 and foreground color #fff). This is very low contrast between foreground and background colors.

https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#-ax_color_01--text-elements-should-have-a-reasonable-contrast-ratio

66eli77:
ha, right.

PR #3253, opened 2015-03-11
Fixes #3232 (hopefully): handle metadata check when Device is imported.

jamalex:

PR #3290, opened 2015-03-13
Hotfix for #3171 - fixes breaking error in software-keypad for IE8

rtibbles:
Fixes #3171.

@MCGallaspy I think this is the minimal fix - but needs some testing to make sure that the keypad is still functional in IE8.

PR #3305, opened 2015-03-16
Hotfix (against master), derived from #3253, to fix key retrieval for an unsaved Device

jamalex:
@aronasorman This would probably be good to merge as a patch release for 0.12, since it will affect anybody trying to use the distributed server with MySQL (or, likely, any other non-sqlite DB).


aronasorman:
I'll combine this with #3290, as I'll need to put in the version.py there as well.

aronasorman:
Superseded by https://github.com/learningequality/ka-lite/pull/3290

PR #3215, opened 2015-03-10
Hotfix/2729 fetch translated exercises part 2

cpauya:
Hi @aronasorman - this fixes #2729 with these notes from @jesumer 

**Central Server Requirements**

1. [Include the .po files from Khan Academy's Crowdin account during "update_language_packs".](https://github.com/fle-internal/ka-lite-central/pull/226)
1. This is required to get the latest translations but the above should do for our test
    * [Issue fix by adding --rebuild make option into update_language_packs command](https://github.com/fle-internal/ka-lite-central/pull/230)
1. make sure it's the same version as distributed

**Distributed Server Requirements**

1. make sure it's the same version as central

**To test**

At Central server:

1. Run `python manage.py update_language_packs -l de --no-srts --no-dubbed --no-kalite-trans --rebuild` command.
1. Check if this will get the proper updated mo.files from CrowdIn.(we can check this out to centralserver/_crowdin_cache zip file).

At Distributed:

1. If it is successful then go to the distributed browser.(NOTE: Make sure you've done the "setup" already).
1. Run `python manage.py languagepackdownload -l de --commandline`.
1. Go to the language page and set the language code to `de`.
1. Go to the exercises and try to test on some of the exercises.
1. Check the translated words.

**To test thru the Python shell** 
This is taken from https://github.com/fle-internal/ka-lite-central/pull/226#issuecomment-75382550

```python
from django.utils.translation import activate, ugettext as _

activate("de")
s = "What is <code>X</code>?"
_(s)
```

Which should display this:

```python
>>> from django.utils.translation import activate, ugettext as _
>>>
>>> activate("de")
>>> s = "What is <code>X</code>?"
>>> _(s)
'Was ist <code>X</code>?'
>>>
```

References:

1. [Official Khan Academy German translations on CrowdIn](https://crowdin.com/project/khanacademy/de)
2. [learn.math.algebra-basics.exercises.pot - 43%](https://crowdin.com/translate/khanacademy/30006/enus-de) -- Search for `What is <code>X</code>?` text as per example above.

PR #3309, opened 2015-03-16
Cherry pick changes from develop

MCGallaspy:
See #3265 for context.


MCGallaspy:
If these pass and are merged, then we can update the submodule in [239](https://github.com/fle-internal/ka-lite-central/pull/239), close [238](https://github.com/fle-internal/ka-lite-central/issues/238), and finally revisit #3215.

aronasorman:
Nondeterministic test errors. Merging.

PR #3313, opened 2015-03-17
Assessment zip file handling

### LINKED ISSUE #3275: Ask user if they have a local assessment zipfile
To be able to handle completely offline installations, users installing through Git should be made aware through a message that:
1. There is now this big zip file they have to download
2. What the URL is to that zip file (so they can download it)
3. Whether they have a local copy of that zip file.

MCGallaspy:
Fixes #3275. Is the language in the docs sufficiently clear? Should we take the opportunity in the release notes to explain what exactly the assessment items are and the various options users have to download them or unpack them later with the `contentload` management commands?

### DIFF
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -169,7 +169,7 @@ class Command(BaseCommand):
             action='store_true',
             dest='force-assessment-item-dl',
             default=False,
-            help='Downloads assessment items from the url specified by settings.ASSESSMENT_ITEM_ZIP_URL, without interaction'),
+            help='Downloads assessment items from the url specified by settings.ASSESSMENT_ITEMS_ZIP_URL, without interaction'),
     )
 
     def handle(self, *args, **options):
@@ -284,13 +284,13 @@ class Command(BaseCommand):
         # download assessment items
         # This can take a long time and lead to Travis stalling. None of this is required for tests.
         if options['force-assessment-item-dl']:
-            call_command("unpack_assessment_zip", settings.ASSESSMENT_ITEM_ZIP_URL)
+            call_command("unpack_assessment_zip", settings.ASSESSMENT_ITEMS_ZIP_URL)
         elif not settings.RUNNING_IN_TRAVIS and options['interactive']:
             print("\nStarting in version 0.13, you will need an assessment items package in order to access many of the available exercises.")
             print("If you have an internet connection, you can download the needed package. Warning: this may take a long time!")
             print("If you have already downloaded the assessment items package, you can specify the file in the next step.")
             print("Otherwise, we will download it from {url}.".format(settings.ASSESSMENT_ITEMS_ZIP_URL))
-           
+
             if raw_input_yn("Do you wish to download the assessment items package now?"):
                 ass_item_filename = settings.ASSESSMENT_ITEMS_ZIP_URL
             elif raw_input_yn("Have you already downloaded the assessment items package?"):
@@ -302,9 +302,9 @@ class Command(BaseCommand):
                 logging.error("No assessment items package file given. You will need to download and unpack it later.")
             else:
                 call_command("unpack_assessment_zip", ass_item_filename)
-        else:        
+        else:
             logging.error("No assessment items package file given. You will need to download and unpack it later.")
-                
+
 
         # Individually generate any prerequisite models/state that is missing
         if not Settings.get("private_key"):
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -152,6 +152,11 @@ class Command(BaseCommand):
             dest='interactive',
             default=True,
             help='Run in non-interactive mode'),
+        make_option('-a', '--dl-assessment-items',
+            action='store_true',
+            dest='force-assessment-item-dl',
+            default=False,
+            help='Downloads assessment items from the url specified by settings.ASSESSMENT_ITEM_ZIP_URL, without interaction'),
     )
 
     def handle(self, *args, **options):
@@ -265,7 +270,9 @@ class Command(BaseCommand):
 
         # download assessment items
         # This can take a long time and lead to Travis stalling. None of this is required for tests.
-        if not settings.RUNNING_IN_TRAVIS and options['interactive']:
+        if options['force-assessment-item-dl']:
+            call_command("unpack_assessment_zip", settings.ASSESSMENT_ITEM_ZIP_URL)
+        elif not settings.RUNNING_IN_TRAVIS and options['interactive']:
             print("\nStarting in version 0.13, you will need an assessment items package in order to access many of the available exercises.")
             print("If you have an internet connection, you can download the needed package. Warning: this may take a long time!")
             print("If you have already downloaded the assessment items package, you can specify the file in the next step.")
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -44,7 +44,7 @@ def raw_input_yn(prompt):
         ans = raw_input("%s (yes or no) " % prompt.strip()).lower()
         if ans in ["yes", "no"]:
             break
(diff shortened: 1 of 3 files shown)


kalite/distributed/management/commands/setup.py
```diff
@@ -96,6 +96,25 @@ def get_hostname_and_description(hostname=None, description=None):
     return (hostname, description)
 
 
+def get_assessment_items_filename():
+    def validate_fn(fn):
+        try:
+            open(fn, "r").close()
+            return True
+        except IOError:
+            sys.stderr.write("Error: couldn't open the specified file: \"%s\"\n" % fn)
```
aronasorman: Why not just `logging.Error`?

kalite/distributed/management/commands/setup.py
```diff
@@ -250,11 +269,25 @@ def handle(self, *args, **options):
...
         call_command("migrate", merge=True, verbosity=options.get("verbosity"))
 
+        # download assessment items
         # This can take a long time and lead to Travis stalling. None of this is required for tests.
-        if not settings.RUNNING_IN_TRAVIS:
-            # download assessment items
-            # TODO-BLOCKER (aron): do not hardcode this, and put in the proper URL once we have a redirect set up
-            call_command("unpack_assessment_zip", "http://eslgenie.com/media/assessment_item_resources.zip")
+        if not settings.RUNNING_IN_TRAVIS and options['interactive']:
+            sys.stdout.write("\nStarting in version 0.13, you will need an assessment items package in order to access many of the available exercises.\n")
```
aronasorman: I see the value of explicitly saying `sys.(pipe).write`, but I stylistically prefer a `print` function in here.

MCGallaspy:
Regarding your comments @aronasorman I adopted the style of the file. Think I should change it?

aronasorman:
Yeah, that would be great. Thanks!

kalite/distributed/management/commands/setup.py
```diff
@@ -250,11 +269,25 @@ def handle(self, *args, **options):
...
+            sys.stdout.write("If you have an internet connection, you can download the needed package. Warning: this may take a long time!\n")
+            sys.stdout.write("If you have already downloaded the assessment items package, you can specify the file in the next step.\n")
+            
+            if raw_input_yn("Do you wish to download the assessment items package now?"):
+                ass_item_fn = settings.ASSESSMENT_ITEMS_ZIP_URL
```
jamalex: This variable name sounds like it's a function, when it's a URL string?

kalite/distributed/management/commands/setup.py
jamalex: 0.13.0

kalite/distributed/management/commands/setup.py
```diff
@@ -96,6 +96,25 @@ def get_hostname_and_description(hostname=None, description=None):
...
+def get_assessment_items_filename():
+    def validate_fn(fn):
+        try:
+            open(fn, "r").close()
+            return True
+        except IOError:
+            sys.stderr.write("Error: couldn't open the specified file: \"%s\"\n" % fn)
+            return False
+
+    prompt = "Please enter the filename of the assessment items package you have downloaded (leave blank to skip this step): "
```
jamalex: Would be good to allow this to be a full path (and ask them for it), in case they downloaded it to elsewhere. But still allow relative/filename-only entries.

kalite/distributed/management/commands/setup.py
```diff
@@ -250,11 +269,25 @@ def handle(self, *args, **options):
         call_command("syncdb", interactive=False, verbosity=options.get("verbosity"))
         call_command("migrate", merge=True, verbosity=options.get("verbosity"))
 
+        # download assessment items
         # This can take a long time and lead to Travis stalling. None of this is required for tests.
-        if not settings.RUNNING_IN_TRAVIS:
-            # download assessment items
-            # TODO-BLOCKER (aron): do not hardcode this, and put in the proper URL once we have a redirect set up
-            call_command("unpack_assessment_zip", "http://eslgenie.com/media/assessment_item_resources.zip")
+        if not settings.RUNNING_IN_TRAVIS and options['interactive']:
```
jamalex: Might be good to either default to downloading the assessment items in non-interactive mode, or at least have an additional command line argument that can force it to? Although I guess any automated deployment systems can also manually run `unpack_assessment_zip` afterwards as well...

kalite/distributed/management/commands/setup.py
jamalex: Would also be good to do a very basic search (current directory, and parents of current directory) for a file named `assessment_item_resources.zip`, and offer that as a default if found, allowing them to just press enter.

sphinx-docs/installguide/release_notes.rst
```diff
@@ -16,7 +16,12 @@ On Windows, use the ``bin\windows\kalite.bat`` in the cmd.exe prompt::
 
     bin\windows\kalite.bat manage setup
 
-When you are asked whether or not to delete your database, you should choose to keep your database!
+When you are asked whether or not to delete your database, you should choose to keep your database! You will also be prompted to download an assessment items package, or to specify the location if you have already downloaded it. If you wish to download the package and specify the location during the setup process: 
+
+* Download the assessment items package `here <http://eslgenie.com/media/assessment_item_resources.zip>`_. Save it in the same folder as the setup script. Take note of the filename, as you will need to enter it exactly!
```
jamalex: > you will need to enter it exactly

This would be nice to avoid, via the recommendation in https://github.com/learningequality/ka-lite/pull/3313/files#r26718375

kalite/distributed/management/commands/setup.py
MCGallaspy: I didn't want to assume that the user was networked in non-interactive mode, but you're right that adding a flag to force a download would be a good idea.

sphinx-docs/installguide/release_notes.rst
MCGallaspy: I'm not sure what you mean. Should the script suggest a filename?

MCGallaspy: I see, the comment was on an outdated diff.

kalite/distributed/management/commands/setup.py
MCGallaspy: Full path should work (it's just undocumented).

MCGallaspy: By "should work" I mean the filename is just passed directly to `open` via `unpack_assessment_zip` management command.

kalite/distributed/management/commands/setup.py
```diff
@@ -43,7 +44,7 @@ def raw_input_yn(prompt):
         ans = raw_input("%s (yes or no) " % prompt.strip()).lower()
         if ans in ["yes", "no"]:
             break
-        sys.stderr.write("Please answer yes or no.\n")
+        logging.error("Please answer yes or no.\n")
```
aronasorman: `error` sounds too severe. `warning` feels better.

kalite/distributed/management/commands/setup.py
```diff
@@ -139,46 +177,41 @@ def handle(self, *args, **options):
...
+        print(" | | / / / _ \  | |   (_) |       ")
+        print(" | |/ / / /_\ \ | |    _| |_ ___  ")
+        print(" |    \ |  _  | | |   | | __/ _ \ ")
+        print(" | |\  \| | | | | |___| | ||  __/ ")
+        print(" \_| \_/\_| |_/ \_____/_|\__\___| ")
+        print("                                  ")
+        print("http://kalite.learningequality.org")
+        print("                                  ")
+        print("         version %s" % VERSION)
+        print("                                  ")
```
aronasorman: Ha! Nice.

aronasorman:
Tested with zip in the same directory, works. Tested with `--noinput` we just don't download it. That's the behaviour i was expeting. I think this is good to go.

PR #3322, opened 2015-03-17
Partially Fixes #3310 

### LINKED ISSUE #3310: Details on download error?
When there is an error while downloading videos, we have a gentle message that warn us about this fact, which is nice:

screenshot from 2015-03-16 22 18 18

but having a bit more details (file unavailable? network issue? missing space on device?…) would be very useful :)

sharprakeshverma:
Added IOError exception in case URL isn't found.

### DIFF
--- a/kalite/updates/management/commands/videodownload.py
+++ b/kalite/updates/management/commands/videodownload.py
@@ -216,6 +216,14 @@ class Command(UpdatesDynamicCommand, CronCommand):
                 "num_total_videos": len(handled_youtube_ids) + len(failed_youtube_ids),
             })
             caching.initialize_content_caches()
+            
+        except TypeError:
+            failed_youtube_ids.append(video.youtube_id)
+            logging.info("TypeError due to internet disconnect")
+            
+        except IOError:
+            failed_youtube_ids.append(video.youtube_id)
+            logging.info("Either file not found or Disk is Full")
 
         except Exception as e:
             self.cancel(stage_status="error", notes=_("Error: %(error_msg)s") % {"error_msg": e})


aronasorman:
@sharprakeshverma please follow the guidelines for PRs in https://github.com/learningequality/ka-lite/blob/develop/CONTRIBUTING.md.

sharprakeshverma:
@aronasorman , thanks for the guidelines. Changed the title and added relevant information as well.

aronasorman:
Any unit tests for this @sharprakeshverma?

kalite/updates/management/commands/videodownload.py
```diff
@@ -216,6 +216,14 @@ def youtube_dl_cb(stats, progress_callback, *args, **kwargs):
                 "num_total_videos": len(handled_youtube_ids) + len(failed_youtube_ids),
             })
             caching.initialize_content_caches()
+            
+        except TypeError:
+            failed_youtube_ids.append(video.youtube_id)
+            logging.info("TypeError due to internet disconnect")
```
aronasorman: I think it's better to just use the string-converted error message, as that might give more detail than assuming what the error could be. See how to use `str` with `BaseException`: https://docs.python.org/2/library/exceptions.html#exceptions.BaseException.

sharprakeshverma:
@aronasorman, i am a newbie to python as well as ka-lite, i don't know how to write a unit test for the PR. I am just trying to contribute as much as i can. Can you guide me with the unit test thing once?

aronasorman:
Sure thing. It might be good to read up on http://docs.python-guide.org/en/latest/writing/tests/ and any unit testing guide out there to get an idea how to do it. After that, take a look at our own guide: https://github.com/learningequality/ka-lite/wiki/Guide:-Writing-unit-tests---Part-1

Since this adds new strings, this will get postponed to 0.14, which gives us 3 months for you to be able to learn unit testing and mocking, and get this merged in :)

sharprakeshverma:
@aronasorman sure and Thanks a ton.

aronasorman:
This PR has become stale, I believe. @sharprakeshverma please open another PR once the comments have been addressed.

PR #3323, opened 2015-03-17
dummy language pack

aronasorman:
Solves https://github.com/fle-internal/ka-lite-central/issues/229.

Does the following:
1. Pulls in the `BASE_LANGUAGE_PACK` from the central server (currently english.)
2. Creates the `TARGET_LANGUAGE_PACK` directory in the central server (currently esperanto).
3. Copies over `BASE_LANGUAGE_PACK` mo files to `TARGET_LANGUAGE_PACK` and adds accenting to them through the `accenting` package.

Remaining issues:
- ~~I need to flesh out the `eo_metadata.json` before the `update_language` screen would work again.~~


kalite/i18n/management/commands/create_dummy_language_pack.py
```diff
@@ -0,0 +1,94 @@
...
+        json.dump(barebones_metadata, f)
+
+    # Now create the actual MO files
+    mo_file_path = os.path.join(molocation, filename)
+    # Alright, so polib can only read (p|m)ofiles already written to
+    # disk, so let's go write it to a temporary file first
+    # with tempfile.NamedTemporaryFile(delete=True) as tmp:
+    #     tmp.write(filecontents)
+    with open(mo_file_path, "w") as f:
+        f.write(filecontents)
```
MCGallaspy: Why no tempfile? (Is it deleted when exiting the `with` statement?)

kalite/i18n/tests/create_dummy_language_pack_tests.py
```diff
@@ -0,0 +1,62 @@
...
+        get_method.assert_called_once_with(get_language_pack_url(lang))
+        self.assertIsInstance(result, zipfile.ZipFile)
+
+
+    def test_retrieve_mo_files(self):
+        dummy_lang_pack = MagicMock(autospec=zipfile.ZipFile)
+
+        result = mod.retrieve_mo_files(dummy_lang_pack)
+
+        dummy_lang_pack.read.call_args_list == [call("LC_MESSAGES/django.mo"),
```
MCGallaspy: Is this supposed to be a comparison?

MCGallaspy: Did you mean to make this an assertion?

kalite/i18n/tests/create_dummy_language_pack_tests.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        with patch('%s.open' % mod.__name__, mock_open(), create=True) as mopen:
+            dummycontent = "writethis"
+            dummylocation = "/this/doesnt/exist"
+            mofile_class.return_value.__iter__ = MagicMock(return_value=iter([MagicMock(), MagicMock(), MagicMock()]))  # 3 "MOEntries"
+            mofile_class.save = MagicMock()  # so we can simulate a save call
+
+            mod.create_mofile_with_dummy_strings(dummycontent, dummylocation)
+
+            mopen.call_args_list == [call(dummylocation, 'w'),
```
MCGallaspy: Again, is this meant to be an assertion?

aronasorman:
I'm having a hard time replicating this test failure. Will hack on it tomorrow.

aronasorman:
@MCGallaspy addressed your comments.

python-packages/accenting.py
```diff
@@ -0,0 +1,160 @@
...
+"""
+import re
+import itertools
+
+
+# Acquired from
+# https://github.com/edx/i18n-tools/blob/master/i18n/dummy.py and
+# https://github.com/edx/i18n-tools/blob/master/i18n/converter.py
+class Converter(object):
+    r"""Converter is an class that transforms strings.  It hides embedded
```
MCGallaspy: A regex literal?

kalite/i18n/tests/create_dummy_language_pack_tests.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @patch.object(requests, "get", autospec=True)
+    def test_download_language_pack(self, get_method):
+
+        # so the next n lines before a newline separator are all about
+        # creating a dummy zipfile that is readable by zipfile.ZipFile
+        dummy_file = StringIO()
+        zf = zipfile.ZipFile(dummy_file, mode="w")
+        zf.close()
+        get_method.return_value.content = dummy_file.getvalue()  # should still be convertible to a zipfile
```
MCGallaspy: I suspect the 2.6 error is because `get` returns a response object, so it should be something like this
```python
get_method.return_value = resp_mock = MagicMock()
resp_mock.content = dummy_file.getvalue()
```

aronasorman:
Now featuring a fix for to change `adhoc.learningequality.org` to `staging.learningequality.org`! Also, we now have a `CENTRAL_SERVER_URL` settings var.

kalite/i18n/management/commands/create_dummy_language_pack.py
```diff
@@ -0,0 +1,100 @@
...
+    url = get_language_pack_url(lang)
+
+    try:
+        resp = requests.get(url)
+        resp.raise_for_status()
+    except requests.ConnectionError as e:
+        logging.error("Error downloading %s language pack: %s" % (lang, e))
+
+    logging.debug("Successfully downloaded base language pack %s" % lang)
+    return zipfile.ZipFile(StringIO(resp.content))
```
MCGallaspy: So if an exception is caught you'll execute both these lines (58 & 59) anyway? What is the value of `resp.content` if a ConnectionError is raised? (Namely, will it cause an error in line 59?)

Perhaps you could move lines 58 & 59 into your try block and return something else in the case of an exception. (Maybe `None` or an empty ZipFile?)

aronasorman: Nope, if it can't download it, we should just error out. I'll add a `sys.exit` call so we can shut down immediately.

aronasorman:
Signed off by mike.

PR #3328, opened 2015-03-17
Look for khan-exercises only when the exercise isn't a perseus exercise

### LINKED ISSUE #3311: get_installed_language_packs gives very different output from one call to the next
Branch: develop

Expected behaviour: Unless a new language pack has been installed, the output of get_installed_language_packs should be identical from call to call

Current behaviour: Repeated calls to the function result in seemingly radically different outputs.

Steps to reproduce:
Inject a logging.debug statement inside 'get_best_available_language' and log the output of the call to get_installed_language_packs.

aronasorman:
Fixes #3311. Does what it says in the title. We now also log what languages we choose from when selecting the best language.

### DIFF
--- a/kalite/i18n/__init__.py
+++ b/kalite/i18n/__init__.py
@@ -394,6 +394,7 @@ def select_best_available_language(target_code, available_codes=None):
     target_code = lcode_to_django_lang(target_code)
     if available_codes is None:
         available_codes = get_installed_language_packs().keys()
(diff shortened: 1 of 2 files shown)


aronasorman:
We shouldn't get as many `requested lang, got lang lang` logs anymore.

rtibbles:
These look like the right fixes.

Issue #3315, opened 2015-03-17
sidebar handle overlaps the text field when resize the window to its minimum width


shubhendras1992:
Branch - develop 

1. There are possible ways to fix this. Either repositioning of sidebar handle to the top or bottom, so its     wont overlap the text field.

2.By changing its padding-left attribute. 
![screen shot 2015-03-17 at 12 48 36 pm](https://cloud.githubusercontent.com/assets/9433264/6682798/80d0eedc-cca3-11e4-87ca-26c9756e90aa.png)

mjptak:
Awesome to see someone looking at how the exercises can become more small
screen friendly!  Moving the submit button to the bottom would in my
experience greatly increase the user experience for a large number of the
exercises...especially the "rote" ones.

66eli77:
my two cents
![screen shot 2015-03-17 at 9 53 43 am](https://cloud.githubusercontent.com/assets/5577205/6692717/99173f18-cc8b-11e4-9a2b-76bc9f49caee.png)
![screen shot 2015-03-17 at 9 54 01 am](https://cloud.githubusercontent.com/assets/5577205/6692720/9c1d98ba-cc8b-11e4-9d3d-0f1b505c0514.png)

rtibbles:
I agree with @mjptak that we should look at the responsiveness of the exercise template as well. @66eli77 or @jtamiace could we think about this for 0.14?

shubhendras1992:
Do I fix it ?

rtibbles:
@Antrikshy could you take a look at this?

Antrikshy:
@rtibbles Sure.

Antrikshy:
Should I make it move out of the way only on smaller screens, @rtibbles?

rtibbles:
Yeah, that should work, I think. What happens to it when you click on the burger to expand the collapsed top nav?

Antrikshy:
It overlaps. :(

I'll fiddle with it.

Antrikshy:
@rtibbles Check out #3537!

Antrikshy:
@MCGallaspy You can close this because #3537 was merged.

PR #3333, opened 2015-03-18
"fixes issue #3315" Fixed sidebar when resizing

shubhendras1992:
issue #3315 
branch - develop 

Current behaviour - Shifted sidebar-tab to top of the page.

screen shot 2015-03-18 at 1 15 49 am

### DIFF
--- a/kalite/distributed/static/css/distributed/sidebar.css
+++ b/kalite/distributed/static/css/distributed/sidebar.css
@@ -89,25 +89,25 @@ div.fade {
     z-index: 10000;
     position: fixed;
     left: 0;
-    top: 40%;
+    top: 0%;
     background: #5AA685;
     color: #fff;
     font-weight: bold;
     font-size: 1.9em;
-    padding: 30px 10px;
-    padding-left: 14px;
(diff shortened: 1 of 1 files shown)


rtibbles:
This should only happen on small media sized screens. Please use media selectors to make this behaviour conditional.

Antrikshy:
@shubhendras1992 I agree with @rtibbles. I would also suggest moving it down so it doesn't cover the logo like that. I also like the longer tab in the mockup on the original issue.

shubhendras1992:
@Antrikshy @rtibbles how about if we replace that arrow with the logo KA LITE?

Antrikshy:
@shubhendras1992 If we're looking into changing the icon, I would like just the leaf, like in the mockups. But I really like the current arrow as well.

rtibbles:
I concur, @Antrikshy.

Also, probably good to do a rebase to latest develop, @shubhendras1992 to resolve the merge conflict.

rtibbles:
This is stale, and comments have not been addressed. Closing.

PR #3334, opened 2015-03-18
Don't split args on whitespace

### LINKED ISSUE #3330: In Windows XP, `bin/kalite start` fails
In `develop` branch, `bin/kalite start` command fails in XP if there's whitespace in the path name of e.g. the `--startup-lock-file` argument, which will almost certainly be the case. Simply escaping the arguments in `kalitectl.py` sadly fails to work. This is apparently only an issue in XP (can someone please verify? @EdDixon?). Perhaps it's an issue with the `kaserve` management command?

MCGallaspy:
Because valid paths can have whitespace in them.
Don't you like paths with whitespace in them? :(

Fixes #3330. Before we merge this, can we get @EdDixon to verify that this works on Windows 7 & XP? (And Vista now too @jamalex?) Ed, just make sure your home directory has white space in it (or if it doesn't ping me and we can just add some whitespace to one of the directories in `kalitectl.py`), pull my branch and try to run the `bin/windows/kalite.bat start` command. You should no longer get a `CommandError`, but instead the server should start normally.

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -349,16 +349,17 @@ def start(debug=False, args=[], skip_job_scheduler=False):
         manage(
             'cronserver',
             in_background=True,
-            args=[
-                '--daemon', '--pid-file={0000:s}'.format(PID_FILE_JOB_SCHEDULER)]
+            args=['--daemon',
+                  '--pid-file={0000:s}'.format(PID_FILE_JOB_SCHEDULER)
+                  ]
         )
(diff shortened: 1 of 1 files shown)


EdDixon:
Vista! You slave drivers! LOL WIll check XP first it was definitely a show stopper there before. Looking for Vista install iso's here now ROFL....

MCGallaspy:
@EdDixon hold off on Vista, was just a tongue-in-cheek comment from a conversation in the office earlier.

aronasorman:
Tests passing. Merging.

Issue #3258, opened 2015-03-11
Check for and resolve 0.13 TODO-BLOCKERS


jamalex:
Running `./manage.py todolist` (or grepping the codebase) identifies a number of things we marked as `TODO-BLOCKER`s, which we should resolve (either through fixing or demoting to simple `TODO`s) before releasing 0.13.

Current list:
```
coachreports/api_resources.py:        #TODO-BLOCKER(66eli77): need to find a way to include exercises that are not completed yet.
coachreports/api_resources.py:        #TODO-BLOCKER(66eli77): need to find a way to include exercises that are not completed yet.
distributed/api_views.py:        # TODO-BLOCKER(jamalex): re-enable this conditional once tastypie endpoints invalidate cached session value
distributed/management/commands/setup.py:            # TODO-BLOCKER (aron): do not hardcode this, and put in the proper URL once we have a redirect set up
distributed/management/commands/todolist.py:        ret = subprocess.call(["git", "--no-pager", "grep", "\#.*TODO-BLOCKER"])
main/api_resources.py:        # TODO-BLOCKER (MCGallaspy) this is inappropriate if multiple channels are active at once
store/models.py:    # TODO-BLOCKER(rtibbles): Update this to "0.13.0" (or whatever the release version number is at the time this goes upstream)
student_testing/middleware.py:            "^.*/_generated/*"    # TODO-BLOCKER(aron): change this to make use of ab_testing app once that's implemented
```

PR #3338, opened 2015-03-18
Downgrade TODO-BLOCKERS to TODOS

MCGallaspy:
I removed @aronasorman's todo because he told me to. The rest I just downgraded. There's still one more, not sure how to handle it yet, so do not merge. Ultimately will address #3258.

### DIFF
--- a/kalite/store/models.py
+++ b/kalite/store/models.py
@@ -51,7 +51,7 @@ class StoreTransactionLog(DeferredCountSyncedModel):
     Model to track student 'expenditure' of points.
     """
 
-    # TODO-BLOCKER(rtibbles): Update this to "0.13.0" (or whatever the release version number is at the time this goes upstream)
+    # TODO(rtibbles): Update this to "0.13.0" (or whatever the release version number is at the time this goes upstream)
 
     minversion = "0.13.0"
 
--- a/kalite/coachreports/api_resources.py
+++ b/kalite/coachreports/api_resources.py
@@ -160,7 +160,8 @@ class ScatterReportExerciseResource(ExerciseSummaryResource):
         return bundle
 
     def get_object_list(self, request):
-        #TODO-BLOCKER(66eli77): need to find a way to include exercises that are not completed yet.
+        #TODO(66eli77): need to find a way to include exercises that are not completed yet.
+        #TODO(MCGallaspy): Probably by adding an "attempted_timestamp" field to the ExerciseLog model
         if not request.GET.get('facility_id'):
             if not request.GET.get('group_id'):
                 return super(ExerciseSummaryResource, self).get_object_list(request).filter(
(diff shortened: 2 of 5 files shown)


kalite/store/models.py
```diff
@@ -51,7 +51,7 @@ class StoreTransactionLog(DeferredCountSyncedModel):
     Model to track student 'expenditure' of points.
     """
 
-    # TODO-BLOCKER(rtibbles): Update this to "0.13.0" (or whatever the release version number is at the time this goes upstream)
+    # TODO(rtibbles): Update this to "0.13.0" (or whatever the release version number is at the time this goes upstream)
```
rtibbles: Looking at the minversion below, I think this can actually be deleted, no?

MCGallaspy: Sure. But the store app isn't activated in 0.13.0, right? Not even clear to me where the minversion is used, is this just a duplicate of information that could be found in `version.py`?

MCGallaspy:
@jamalex the unresolved TODO-BLOCKER is from you, so can you consult me on it when you get a moment?

MCGallaspy:
@jamalex context: (in `distributed/api_views.py`)
```python
def status(request):
```
some lines removed
```python
    # Override properties using facility data
    if "facility_user" in request.session:  # Facility user
        user = request.session["facility_user"]
        data["is_logged_in"] = True
        data["username"] = user.get_name()
        # TODO-BLOCKER(jamalex): re-enable this conditional once tastypie endpoints invalidate cached session value
        # if "points" not in request.session:
        request.session["points"] = compute_total_points(user)
        data["points"] = request.session["points"] if request.session["points"] else 0
        data["user_id"] = user.id
        data["user_uri"] = reverse("api_dispatch_detail", kwargs={"resource_name": "user", "pk": user.id})
        data["facility_id"] = user.facility.id

    # Override data using django data
    if request.user.is_authenticated():  # Django user
        data["is_logged_in"] = True
        data["username"] = request.user.username

    return JsonResponse(data)
```

kalite/store/models.py
rtibbles: Correct on the first. @jamalex should we just be deleting most of the store or nay?

Min-version is used on specific models for syncing purposes. We won't try to sync models that are 0.13 min-version to a 0.12 system, for example.

jamalex: The store app isn't activated on the distributed server in 0.13.0, but it's activated on the central server still, since otherwise nalanda would lose all the data they have (and are actively using) in there. So, we need to keep the models in there for now.

jamalex:
> `if "points" not in request.session:`

It would be great to be able to re-enable this caching, since `compute_total_points` is expensive, but it could be downgraded to a `TODO` and an issue created for 0.13-bugfixes.

MCGallaspy:
@jamalex how do you suggest the endpoint should invalidate a cached "points" value? I'm not sure where to start, but if it's quick enough I can make these changes. Otherwise I'll just downgrade it.

rtibbles:
I would guess that we would delete request["points"] from the request object any time an exerciselog or videolog was saved and it resulted in a points change on that log model?

MCGallaspy:
It looks like there are genuinely few requests to the status model endpoint anyway, so I'm just going to downgrade this for now.

MCGallaspy:
Ready to be merged. One extant TODO-BLOCKER that's resolve in #3313.

jamalex:
(Note: The status endpoint is loaded on every pageload. At least now, with a bit more stuff pushed to a frontend app, pageloads don't happen as often)

MCGallaspy:
>At least now, with a bit more stuff pushed to a frontend app, pageloads don't happen as often

That's what I was referring to.

MCGallaspy:
And compute_total_points should be pretty quick for coaches (with few or no associated *Log models) who will be most likely to stray off the learn page.

PR #3339, opened 2015-03-18
Auth redirect

### LINKED ISSUE #3256: Cheery Pick the Redirect when unauthorized issue fix
Its fixed on nalanda branch for 403 unauthorized error.
But, its still on the develop and it shows Contact FLE Error ( Object Object) and doesn't redirect to login page.

Cheery pick commits from here.
https://github.com/learningequality/ka-lite/pull/3092

rtibbles:
Fixes #3256

Summary of changes:
* Redirects to login page when API returns a 401/403.
* Fixes session timeout middleware to return 401 on Ajax calls.

Note: Any other reason for a session ending will not get a message.

### DIFF
--- a/kalite/distributed/tests/browser_tests/quiz.py
+++ b/kalite/distributed/tests/browser_tests/quiz.py
@@ -9,8 +9,10 @@ from django.utils import unittest
 
 from kalite.testing.base import KALiteBrowserTestCase
 from kalite.testing.mixins import BrowserActionMixins, FacilityMixins
+from selenium.webdriver.common.keys import Keys
 
 PLAYLIST_ID = "g3_p1"
+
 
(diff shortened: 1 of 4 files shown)


kalite/distributed/tests/browser_tests/quiz.py
```diff
@@ -9,9 +9,11 @@
 
 from kalite.testing.base import KALiteBrowserTestCase
 from kalite.testing.mixins import BrowserActionMixins, FacilityMixins
+from selenium.webdriver.common.keys import Keys
```
MCGallaspy: Unused import?

rtibbles: Ah yes, cherry-pick residue.

PR #3348, opened 2015-03-19
Use valid mode

### LINKED ISSUE #3347: unpack_assessment_zip managment command fails on windows
Using the installer built from `ka-lite-installer-windows` on the branch `master`, I tried both starting the server for the first time and running the command `bin/windows/kalite.bat start`. The first fails quietly, while the latter has the following error:
```cmd
  File "python-packages\django\core\management\base.py", line 224, in run_from_argv
  File "python-packages\django\core\management\base.py", line 263, in execute
  File "C:\Program Files\KA Lite4\ka-lite\kalite\contentload\management\commands\unpack_assessment_zip.py", line 58, in handle
    f = tempfile.TemporaryFile("r+w")
  File "C:\Python27\lib\tempfile.py", line 455, in NamedTemporaryFile
    file = _os.fdopen(fd, mode, bufsize)
ValueError: Invalid mode ('r+w')
```

@aronasorman seems like "r+w" [isn't truly a valid mode](https://docs.python.org/2/library/functions.html#open), but works ok enough in linux. Maybe the desired mode is "r+"?

MCGallaspy:
r+w as a file mode works ok on linux but fails
on windows, but isn't a valid mode at any rate.
Probably meant to use r+.

Fixes #3347. Important so that a windows installer built with included submodule pointing to `develop` will work properly.

### DIFF
--- a/kalite/contentload/management/commands/unpack_assessment_zip.py
+++ b/kalite/contentload/management/commands/unpack_assessment_zip.py
@@ -55,7 +55,7 @@ class Command(BaseCommand):
             logging.warn("Downloading assessment item data from a remote server. Please be patient; this file is big, so this may take some time...")
             # this way we can download stuff larger than the device's RAM
             r = requests.get(ziplocation, prefetch=False)
-            f = tempfile.TemporaryFile("r+w")
+            f = tempfile.TemporaryFile("r+")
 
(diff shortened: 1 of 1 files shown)


aronasorman:
Thanks!

PR #3342, opened 2015-03-19
Fixes #3082: Update topic tree (removing towers of hanoi) and replace all image and content links properly

jamalex:
As per #3082, this:

- removes the towers-of-hanoi exercise, since it's an iframe onto the KA site
- downloads a fuller set of images (includes *.jpeg and *.JPG, etc)
- replaces all links to other KA resources to KA Lite's local copy (or removes links+text if KA Lite doesn't have them)

PR #3354, opened 2015-03-19
Translate the text inside the question area

### LINKED ISSUE #2729: Fetch translated exercises
Right now we only support officially internationalized KA sites like fr, pt-br and es. We wanna expand that to any user-translated exercises like de. We should also fetch translations for the Perseus exercises.

aronasorman:
Fixes #2729. 

Before, the actual questions aren't getting translated.
screen shot 2015-03-19 at 9 18 17 am

This was due to an encoding issue. PO files are read as ascii, but we are passing unicode strings to `gettext`, and hence they are not getting matched. With this fix (+ a couple of others) we now have this:
screen shot 2015-03-19 at 12 38 30 pm

To test:
1. Run `bin/kalite manage create_dummy_language_pack --traceback`
2. Run the server, and change your language to `Esperanto`.
3. Go to a perseus exercise, preferably some of the early math ones (as the later ones don't even have entries in the PO file).

Remaining issues:
- Need to get `backbone.js` strings translated.

### DIFF
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -37,6 +37,17 @@ CONTENT_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "contents.json")
 
 CACHE_VARS = []
 
+ASSESSMENT_DATA_TRANSLATABLE_FIELDS = [
+    'widgets',
+    'hints',
+    'question',
+    'answerArea',
+    'radio 1',
+    'choices',
+    'options',
+]
+
+
 if not os.path.exists(settings.CHANNEL_DATA_PATH):
     logging.warning("Channel {channel} does not exist.".format(channel=settings.CHANNEL))
 
@@ -496,31 +507,38 @@ def get_assessment_item_data(request, assessment_item_id=None):
 
     # Enable internationalization for the assessment_items.
     try:
+
         item_data = json.loads(assessment_item['item_data'], object_hook=json_ascii_decoder)
-        question_content = _(item_data['question']['content'])
-        answerarea_content = item_data['answerArea']['options']['content']
-        if answerarea_content:  # Not an empty string
-            answerarea_content = _(answerarea_content)
-
-        # Loop over contents for multiple hints.
-        # Pickup the wrapped strings and append to hints dict
-        new_hints = []
-        for contents_dict in item_data['hints']:
-            content_value = contents_dict.get('content')
-            converted_content = _(content_value)
-            contents_dict['content'] = converted_content
-            new_hints.append(contents_dict)
-
-        item_data['hints'] = new_hints
-        item_data['question']['content'] = question_content
-        item_data['answerArea']['options']['content'] = answerarea_content
-        # dump the data for a proper json format.
+        item_data = smart_translate_item_data(item_data)
         assessment_item['item_data'] = json.dumps(item_data)
 
     except KeyError as e:
         logging.error("Assessment item did not have the expected key %s. Assessment item: \n %s" % (e, assessment_item))
 
     return assessment_item
+
+
+def smart_translate_item_data(item_data):
+    """Auto translate the content fields of a given assessment item data.
+
+    An assessment item doesn't have the same fields; they change
+    depending on the question. Instead of manually specifying the
+    fields to translate, this function loops over all the relevant
+    fields of item_data and translates only the content field.
+    """
+    if 'content' in item_data:
+        item_data['content'] = _(item_data['content'])
+
+    for field in ASSESSMENT_DATA_TRANSLATABLE_FIELDS:
+        if field in item_data:
+            field_data = item_data[field]
+            if isinstance(field_data, dict):
+                item_data[field] = smart_translate_item_data(field_data)
+            elif isinstance(field_data, list):
+                item_data[field] = map(smart_translate_item_data, field_data)
+
+    return item_data
+
 
 
 def get_content_data(request, content_id=None):
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -37,16 +37,6 @@ CONTENT_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "contents.json")
 
 CACHE_VARS = []
 
-ASSESSMENT_DATA_TRANSLATABLE_FIELDS = [
-    'widgets',
-    'hints',
-    'question',
-    'answerArea',
-    'radio 1',
-    'choices',
-    'options',
-]
-
 
 if not os.path.exists(settings.CHANNEL_DATA_PATH):
     logging.warning("Channel {channel} does not exist.".format(channel=settings.CHANNEL))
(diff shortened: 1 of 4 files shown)


kalite/topic_tools/__init__.py
```diff
@@ -496,25 +507,9 @@ def get_assessment_item_data(request, assessment_item_id=None):
...
-        new_hints = []
-        for contents_dict in item_data['hints']:
-            content_value = contents_dict.get('content')
-            converted_content = _(content_value)
-            contents_dict['content'] = converted_content
-            new_hints.append(contents_dict)
-
-        item_data['hints'] = new_hints
-        item_data['question']['content'] = question_content
-        item_data['answerArea']['options']['content'] = answerarea_content
```
aronasorman: All deletions have been compressed to `smart_translate_item_data`.

aronasorman:
also @cpauya please review.

aronasorman:
Been scanning some of the exercises where translations aren't working. All of them either:
- have an unlocalized url:
screen shot 2015-03-19 at 4 43 27 pm
screen shot 2015-03-19 at 4 44 27 pm

- or don't exist in the po file at all:
screen shot 2015-03-19 at 4 46 25 pm
screen shot 2015-03-19 at 4 26 45 pm

MCGallaspy:
I get the following error when I run `bin/kalite manage create_dummy_language_pack --traceback`
```sh
  File "/home/gallaspy/Desktop/ka-lite/python-packages/django/core/management/base.py", line 224, in run_from_argv
  File "/home/gallaspy/Desktop/ka-lite/python-packages/django/core/management/base.py", line 263, in execute
  File "/home/gallaspy/Desktop/ka-lite/python-packages/django/core/management/base.py", line 393, in handle
    return self.handle_noargs(**options)
  File "/home/gallaspy/Desktop/ka-lite/kalite/i18n/management/commands/create_dummy_language_pack.py", line 43, in handle_noargs
    django_mo_contents, djangojs_mo_contents = retrieve_mo_files(langpack_zip)
  File "/home/gallaspy/Desktop/ka-lite/kalite/i18n/management/commands/create_dummy_language_pack.py", line 68, in retrieve_mo_files
    langpack_zip.read("LC_MESSAGES/djangojs.mo"))
  File "/usr/lib/python2.7/zipfile.py", line 935, in read
    return self.open(name, "r", pwd).read()
  File "/usr/lib/python2.7/zipfile.py", line 961, in open
    zinfo = self.getinfo(name)
  File "/usr/lib/python2.7/zipfile.py", line 909, in getinfo
    'There is no item named %r in the archive' % name)
KeyError: "There is no item named 'LC_MESSAGES/djangojs.mo' in the archive"
```

MCGallaspy:
Because I'm pointing to a local instance of the central server. Nevermind.

MCGallaspy:
At what point are urls supposed to be converted? That's not a problem with `create_dummy_language_pack`, but rather something has gone wrong when the central server built the base language pack (`de` in this case)?

MCGallaspy:
@jamalex @aronasorman

python-packages/fle_utils/general.py
```diff
@@ -215,12 +215,12 @@ def _decode_list(data):
...
             item = _decode_list(item)
         elif isinstance(item, dict):
-            item = _decode_dict(item)
+            item = json_ascii_decoder(item)
         rv.append(item)
     return rv
 
 
-def _decode_dict(data):
+def json_ascii_decoder(data):
```
MCGallaspy: Might be good to add a docstring indicating briefly what this does, i.e.:
```python
""" Takes a deserialized JSON object and returns a copy of that JSON object with all unicode strings encoded as utf-8 """
```

kalite/topic_tools/__init__.py
```diff
@@ -523,6 +518,29 @@ def get_assessment_item_data(request, assessment_item_id=None):
...
+
+    for field in ASSESSMENT_DATA_TRANSLATABLE_FIELDS:
+        if field in item_data:
+            field_data = item_data[field]
+            if isinstance(field_data, dict):
+                item_data[field] = smart_translate_item_data(field_data)
+            elif isinstance(field_data, list):
+                item_data[field] = map(smart_translate_item_data, field_data)
+
+    return item_data
```
MCGallaspy: This function is doubleplus good on readability. :+1:

aronasorman:
@MCGallaspy yeah, they're supposed to be built when the language packs are generated on the central server. I know that the current `convert_urls` function in there isn't comprehensive, and hence we need to get Jamie's #3342 PR in there soon.

kalite/topic_tools/__init__.py
```diff
@@ -523,6 +518,29 @@ def get_assessment_item_data(request, assessment_item_id=None):
...
+
+    An assessment item doesn't have the same fields; they change
+    depending on the question. Instead of manually specifying the
+    fields to translate, this function loops over all the relevant
+    fields of item_data and translates only the content field.
+    """
+    if 'content' in item_data:
+        item_data['content'] = _(item_data['content'])
+
+    for field in ASSESSMENT_DATA_TRANSLATABLE_FIELDS:
```
MCGallaspy: How did you determine the members of `ASSESSMENT_DATA_TRANSLATABLE_FIELDS`? Can we avoid using a magic constant like this if we just iterate over all the keys in item_data:
```python
        for field, field_data in item_data.iteritems()
            if isinstance(field_data, dict):
                item_data[field] = smart_translate_item_data(field_data)
            elif isinstance(field_data, list):
                item_data[field] = map(smart_translate_item_data, field_data)
```
Or is there something I'm missing? Is it much more expensive to consider all of the keys rather than a limited subset?

kalite/topic_tools/tests/utils_tests.py
```diff
@@ -0,0 +1,47 @@
...
+
+DUMMY_STRING = "maize"
+TRANS_STRING = "trans"
+
+
+ugettext_dummy = MagicMock(return_value=DUMMY_STRING)
+
+
+@patch.object(mod, '_', new=ugettext_dummy)
+class SmartTranslateItemDataTests(KALiteTestCase):
```
MCGallaspy: Perhaps good to include a test where we don't expect internationalization, like if `a in ASSESSMENT_DATA_TRANSLATABLE_FIELDS` is `False`:
```python
def test_not_translated(self):
    test_data = {a: {'content': 'this should not be translated'}}
    # assert that ugettext_dummy called zero times
```

kalite/topic_tools/__init__.py
MCGallaspy: >> Is it much more expensive to consider all of the keys rather than a limited subset?

Obviously if there are a lot more than 7 keys the answer is yes, but is there something else I'm not considering?

kalite/topic_tools/tests/utils_tests.py
MCGallaspy: Of course if you made the changes I suggested so that `'content'` is more thoroughly searched for then this test would fail.

kalite/topic_tools/__init__.py
aronasorman: I was hesitant in simply scanning all the fields initially, but upon inspection of a lot of assessment items it seems that most of them are 3-7 levels deep, so I think we can handle just scanning through all fields.

aronasorman:
@MCGallaspy removed that magic variable, and now the function is even simpler \o/

kalite/topic_tools/tests/utils_tests.py
aronasorman: I didn't write this test, as we just now recurse on all dict/list fields.

MCGallaspy:
:+1: @aronasorman do you want anyone else to review this?

aronasorman:
Think that's it, I'll put the backbone.js translation fixes in another PR.

Issue #3365, opened 2015-03-20
Accessbility Issue: Navigation sidebar put above main content in HTML flow


sopu:
Branch: develop
Current:  
One of the problems with using a screen reader to read a web page is that it sees the web page as a stream of words, appearing one at a time. As the HTML for ka-lite begins with topbar and then sidebar with all the navigation buttons, a typical screen reader will proceed to read the sidebar first, through the whole list, before it reaches the main content. 
Scrrenshot:
![screenshot from 2015-03-20 11 30 04](https://cloud.githubusercontent.com/assets/6399612/6746804/7e6ca3b8-cef4-11e4-905b-5c5acb428631.png)

Improvements suggested:
1. Put a "Skip to main content" link at the top of the page.(Used by many accessibility-aware sites.) OR
2. Put the main content before sidebar in the HTML code.

sopu:
Please suggest the kind of improvement would be preferred.
I would like to work on it.

MCGallaspy:
Apparently resolved.

PR #3366, opened 2015-03-20
Puts content-area above sidebar-container in html flow for accessibility improvements

sopu:
Fixes issue #3365 .
Puts content-area above sidebar-container in html flow for better accessibility using screen readers.
The navigation links are put below the main content.
Screenshot using fangs Screen Reader:
screenshot from 2015-03-20 17 48 25


aronasorman:
I don't feel qualified enough to review this, but the change looks small enough. Tests pass too, so let's merge!

PR #3369, opened 2015-03-20
Further i18n fixes

aronasorman:
These commits are self-explanatory. I'll put in justifications in the comments.


kalite/distributed/templates/distributed/base.html
```diff
@@ -54,7 +54,7 @@
         {% endif %}
 
         <!-- JS translation files dynamically generated by django, then saved to the static file below -->
-        <script type="text/javascript" src="{% static 'js/i18n/'|add:request.language|add:'.js' %}"></script>
+        <script type="text/javascript" src="{% static 'js/i18n/'|add:request.language|add:'.js' True %}"></script>
```
aronasorman: We now let the browser cache the JS translation catalog. This is because this file is really big now: 18.7 MB here in my system.

kalite/i18n/__init__.py
```diff
@@ -371,7 +371,7 @@ def update_jsi18n_file(code="en"):
     request.path = output_file
     request.session = {settings.LANGUAGE_COOKIE_NAME: code}
 
-    response = javascript_catalog(request, packages=('ka-lite.locale',))
+    response = javascript_catalog(request, packages=('ka-lite.locale',), domain="django")
```
aronasorman: Why did the JS catalog become so big? Because we now have to include the entire `django` domain in order to translate as much exercises as possible, the `djangojs` domain is almost a subset of `django`. This means the JS catalog file would be really big, but see my workaround to make sure it's only downloaded once per user.

kalite/i18n/management/commands/create_dummy_language_pack.py
```diff
@@ -53,6 +53,7 @@ def download_language_pack(lang):
     url = get_language_pack_url(lang)
 
     try:
+        logging.debug("Downloading from {url}.".format(url=url))
```
aronasorman: Just another logging statement (tm).

kalite/updates/views.py
```diff
@@ -49,7 +49,7 @@ def update_context(request):
 @render_to("updates/update_videos.html")
 def update_videos(request, max_to_show=4):
     context = update_context(request)
-    messages.warning(request, 'For low-powered devices like the Raspberry Pi, please download videos one at a time.')
+    messages.warning(request, _('For low-powered devices like the Raspberry Pi, please download videos one at a time.'))
```
aronasorman: As I was scanning through strings I noticed that this wasn't i18nized yet. Here you go!

kalite/i18n/__init__.py
MCGallaspy: So if I'm understanding this correctly, some of the internationalized strings in JS files are not translated, but since the default domain `djangojs` is almost a superset of the `django` domain, your suggestion is that we should just use the `django` domain so that as much as possible is translated?

MCGallaspy:
So finally your suggested workaround for untranslated JS strings is to manually append translated strings in the `djangojs` domain to the `django` domain when they do not already appear there, yes?

aronasorman:
Actually, looks like we already combine `django` and `djangojs.po` files in the central server. Not sure why we're still missing some strings.

aronasorman:
Ok, I think this is ready.

PR #3374, opened 2015-03-21
Disabled reports tab for teachers

jtamiace:
No need for it since it's the only page that teachers need to access on the screen

no more reports


jamalex:
Please target 0.13.x.

PR #3377, opened 2015-03-21
Creates minimum width for the dropdown username tab

jtamiace:
To address this issue:

Tested in German as well, and it works out.


rtibbles:
Screenshot after?

jtamiace:
fix

PR #3400, opened 2015-03-24
Don't show `manage.py update` message.

aronasorman:
Remove this. We should not let them run this command, as we don't support it.


aronasorman:
nondeterministic test failure.

benjaoming:
Agreed, also agreed that the test failure was one of these silly things that I think won't happen much in the future.

PR #3401, opened 2015-03-24
Javascript testing framework proof of concept

rtibbles:
Adds a Javascript Testing Framework

Summary of changes:
* Uses Karma as a test runner for Javascript tests
* Uses QUnit for Javascript Unit Testing
* Uses Django-Compressor to compile static assets into concatenated Javascript and CSS files (currently happening within the request/response cycle if Debug=False).
* Adds a custom management command (buildassets) to compile assets offline, and create a map so that the hex keys are mapped to more meaningful identifiers.
* Imports these concatenated files into Karma for testing.
* Removes the BUILD_HASH addending to static files, as this will be superseded by the Cache invalidation of Django-Compressor (either by the client regenerating the files, or the build process creating a new set).
* Removes the repeatblock custom template tag that broke templates.
* Adds a sessionModel for storing important variables that were otherwise just accessed from the global scope.
* Adds an example test for the default values of the newly added sessionModel to demonstrate writing unit tests in QUnit.

Travis is not currently set up to deal with the requirements.txt this uses, nor to handle the karma testing properly.

To test this out manually first:

```pip install```
then add the following to local settings:
```
COMPRESS_ENABLED = True
COMPRESS_OFFLINE = True
```
You can then use ```./bin/kalite manage bundleassets``` to build the assets.

To run the Javascript tests:
 ```(sudo) npm install``` to install the required npm packages.
```karma start```.

### CODE REVIEW

**Comment by aronasorman** [comment]
Looks like tests are failing due to an invalid config @rtibbles

**Comment by rtibbles** [comment]
Yes, this requires our pip install pipeline to be set up to work on Travis. See the PR text.

File: kalite/distributed/templates/distributed/base.html [hunk 1]
```diff
@@ -38,12 +41,17 @@
...
         <script type="text/javascript" src="{% url 'dynamic_js' %}"></script>
 
-        {# Older versions of IE didn't have a JSON object, so this library adds it back in if needed #}
-        <script type="text/javascript" src="{% static 'js/json2.js' %}"></script>
-        <!-- JS translation file statically included as a back-up, so that if the saved file below isn't available, we don't choke. -->
-        {% if current_language != "en" %}
-        <script type="text/javascript" src="{% static 'js/i18n/en.js' %}"></script>
-        {% endif %}
+        {# Use JS to help IE5.5-IE8 support modern browser features, especially for CSS and PNGs #}
+        <!--[if lt IE 8]>
```
aronasorman: Delete, now that we don't support IE 8.

kalite/distributed/templates/distributed/base.html
```diff
@@ -53,26 +61,25 @@
...
         {% endif %}
 
+        <!-- JS translation file statically included as a back-up, so that if the saved file below isn't available, we don't choke. -->
+        {% if current_language != "en" %}
+        <script type="text/javascript" src="{% static 'js/i18n/en.js' %}"></script>
+        {% endif %}
+
         <!-- JS translation files dynamically generated by django, then saved to the static file below -->
-        <script type="text/javascript" src="{% static 'js/i18n/'|add:request.language|add:'.js' True %}"></script>
+        <script type="text/javascript" src="{% static 'js/i18n/'|add:request.language|add:'.js' %}"></script>
```
aronasorman: Shouldn't these JS catalog files be part of the compress blocks too?

kalite/distributed/templates/distributed/base.html
```diff
@@ -140,6 +119,37 @@
...
+          DELETE_GROUPS_URL           : "{% url 'group_delete' %}",
+          MOVE_TO_GROUP_URL           : "{% url 'move_to_group' %}",
+          STATIC_URL                  : "{% static "" %}",
+
+          ALL_ASSESSMENT_ITEMS_URL    : "{% url 'api_dispatch_list' resource_name='assessment_item' %}",
+          GET_VIDEO_LOGS_URL          : "{% url 'api_dispatch_list' resource_name='videolog' %}",
+          GET_EXERCISE_LOGS_URL       : "{% url 'api_dispatch_list' resource_name='exerciselog' %}",
+          GET_CONTENT_LOGS_URL        : "{% url 'api_dispatch_list' resource_name='contentlog' %}",
+
+          KHAN_EXERCISES_SCRIPT_URL : "{% static 'perseus/ke/khan-exercise.js' %}",
```
aronasorman: Haha, I like that even this one was kept unaligned from the others (like how it was before).

kalite/distributed/templates/distributed/base.html
MCGallaspy: Disagree... if we have vague plans to add support later, let's leave it in, since it's simply a no-op on other browsers.

aronasorman: This small snippet should be easy to add in though, if we ever want to revisit IE 8. I don't see why we want to clutter our code with something we explicitly don't support right now.

aronasorman:
Apart from the code, one thing I'd ask once this is merged is perhaps a wiki article instructing us how to make frontend tests. The most basic one would include links to read, dependencies to download and what commands to run so we can get tests running on our local machine.

MCGallaspy:
@aronasorman Perhaps you're suggesting we add a page to our burgeoning developer docs?

aronasorman:
Ah, developer docs it is. Is this the right directory @MCGallaspy? https://github.com/learningequality/ka-lite/tree/develop/sphinx-docs/developer_docs

MCGallaspy:
Right, you can add an .rst file there and it will be built automatically.

kalite/distributed/templates/distributed/base.html
benjaoming: I have a pretty sad feeling that IE8 has to be supported for a project in Denmark :/

kalite/distributed/templates/distributed/base.html
rtibbles: Having Django template variables inside the compression blocks makes them not work in a reliable enough way for offline compression, so I have left these dynamically loaded files out for now.

kalite/distributed/templates/distributed/base.html
rtibbles: We should look at how we can support IE8 then. Is it just the sidebar that is the issue? We should provide some graceful fallback that might be helpful for accessibility too.

rtibbles:
Now with extra docs! https://github.com/rtibbles/ka-lite/blob/javascript_testing/sphinx-docs/developer_docs/javascript_testing.rst Yummy!

PR #3404, opened 2015-03-24
Coverage and coveralls.io support

benjaoming:
Now tested:

https://coveralls.io/r/benjaoming/ka-lite

Sorry about the many commits, had to debug Coveralls a bit and the non-deterministic test weirdness.


benjaoming:
Oh btw... just wait for the last Travis build to finish to know if the --omit option worked... if coverage increases, it worked...

benjaoming:
Okay so I have no idea why it says coverage went down, when at the same time it reports no changes to any individual files....

https://coveralls.io/builds/2186276

Anyways, I think it's good to merge :) When it's merged, I'll have a look to configure Coveralls to report back in PRs about their coverage impact.

rtibbles:
Could just be our magical tests at work!

benjaoming:
Ah yes, actually that's true... but still I don't understand how it can report a decrease in coverage without saying which files have decreased.

rtibbles:
Odd - weird test failure on Python 2.7...

MCGallaspy:
@rtibbles @benjaoming take a look at `UnicodeModelsTest` from `fle_utils`:

```python
class UnicodeModelsTest(unittest.TestCase):
    korean_string = unichr(54392)

    def check_unicode_class_coverage(self, models_module, known_classes):

        # Make sure we're testing all classes
        #   NOTE: we're not testing UserProfile
        found_classes = filter(lambda class_obj: hasattr(class_obj, "__unicode__"), all_classes_in_module(models_module))
        self.assertTrue(not set(found_classes) - set(known_classes), "The following models have __unicode__ methods that need to have tests written: %s" % ", ".join([str(cls) for cls in set(found_classes) - set(known_classes)]))
```

This method (which is the cause of the error) tries to assert that the classes in a module have a `__unicode__` method, but fail because of tricky importing magic (the success of the test depends on which `manage.py` you use to start the test runner). Can someone explain why this test is here? And the necessity of `MainUnicodeModelsTest`? When are we converting models into a unicode representation?

rtibbles:
I cannot explain it, no. Happy to delete.

As part of @benjaoming's clean up of python-packages, we should also sort out the proper place for what are ostensibly our own packages in here - and refactor as necessary.

benjaoming:
@rtibbles thanks, *ostensibly* shall be my word of the day :)

Since ostensibly there are no objections to using coverage, not even from the latest Travis, I shall merge this :) Hope this Coverage stuff does us well.

I case it gets annoying, we can remove the big, long text report (`coverage report`)  at the end of the .travis.yml config.

PR #3405, opened 2015-03-25
Behave proof of concept - BDD integration tests

MCGallaspy:
What's in here:
* A management command (`run_behave_tests`) which searches out `behave` tests in the project directory and runs them.
* A stubbed-out test for the learn tab 
* My thoughts on where it's appropriate to use integration tests vs. unit tests

What's coming:
* A better example test

My thoughts:

This is still a work in progress. I wanted to write a nontrivial (as in not just `pass` statements) integration test for the `/learn/` page, but I encountered a few stumbling blocks.

First, because it takes so dang long to load the content cache *et al* into memory on my machine, tests which try to load pages that use cached variables (like through ajax requests) time out. Secondly, the learn tab is not easily testable as a single-page JS app because it's hard to avoid race conditions -- this problem is carried over from our existing test suite and is a "feature" of the app itself.

For the learn page (and for other single-page JS features we develop in the future), we need to carefully distinguish between what should be tested using a front-end testing framework with unit tests, and what should be tested using integration tests. For example, some existing tests for the learn page are:

* Test that watching videos increases points
* Test that the displayed points are updated correctly
* Test various interactions with exercises                
                  
None of which are appropriate as integration tests, because they can be tested much more easily as unit tests on our Backbone views.

For the learn page some tests which *are* appropriate as integration tests since they test client-server interaction through API requests and responses (and might've been helpful during development using BDD but now would only serve as regression tests) are:

* Test that navigating the sidebar prior to a terminal choice gives you a blank content area.
* Test that navigating to an available content item loads that content item.
* Test that navigating to an unavailable content item displays an appropriate error.

The lesson here for single-page JS apps is that if we're purely testing *client-side* behavior let's make it a unit test, but if we're testing the server's response to user actions let's make it an integration test, and spec it out beforehand so we can use BDD. (This may be obvious to you guys, but I've only finally managed to fully grok it. :/ )

**Finally** I'm putting this PR and my novel-length comment up for review, but I'd also like to write a non-trivial test for another feature. (Which I am presently doing.)


rtibbles:
At the risk of starting a dialectic, I would argue that if behaviour is only on the client side does not mark it out as integration or unit test. Unit tests should test units, whereas integration tests should test interaction between components. If those components are entirely client side, then it could be a client side integration test - the question then becomes whether we want to have integration tests in both the Python testing framework and the Javascript testing framework (we could use a Gherkin specification for both, so the separation would not be as terrible as it might otherwise seem).

kalite/distributed/tests/features/exercise.feature
```diff
@@ -0,0 +1,15 @@
...
+
+    Scenario: Exercise is not available
+        Given I open an exercise
+        And the exercise is not available
+        Then I should see an alert
+
+    Scenario: Exercise is available
+        Given I open an exercise
+        And the exercise is available
+        Then I will be happy
```
rtibbles: Testing for happiness - could get yourself in a pickle!

MCGallaspy:
Started to write something a bit more meaty.

What you'll find inside:
* Starting to get proof-of-concept tests for the control_panel app
* Some functions I anticipate will be used widely in `testing/behave_helpers.py`
* In `control_panel/tests/features/control_panel.feature`, some examples of what I consider appropriate integration tests (Are they? What do you think?)
* In `control_panel/templates/control_panel/partials/_facility_table.html` I added some IDs to a template to make testing less brittle (in anticipation of wanting to use those elements in tests)
in `control_panel/tests/features/steps/control_panel.py` some steps definitions (with quite a few missing)

kalite/distributed/tests/features/steps/exercises.py
```diff
@@ -0,0 +1,45 @@
+from behave import *
+from django.conf import settings
+from selenium.webdriver.common.by import By
+from selenium.common.exceptions import TimeoutException
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+
+TIMEOUT = 3
+
+@given("I open an exercise")
```
rtibbles: There is something so appealing about this way of labeling steps. Really hope we can build up a good library of steps and encourage non-devs to create Gherkin test specifications with them.

jamalex: `@given("I curse in frustration")`

`@given("I really wish I had paid attention in Ms Wilbur's Grade 3 class....")`

MCGallaspy:
> Unit tests should test units, whereas integration tests should test interaction between components. If those components are entirely client side, then it could be a client side integration test

Fair enough. I'm abusing the term "unit test" here.

> the question then becomes whether we want to have integration tests in both the Python testing framework and the Javascript testing framework (we could use a Gherkin specification for both, so the separation would not be as terrible as it might otherwise seem).

My point is that it we'll have to jump through some hoops to do integration testing of single-page JS apps using Selenium. Maybe we can come up with a scheme where we write all our tests using Gherkin, and then our test running commands intelligently figure out which framework to run them under? I guess we're kinda backed into a corner given that our app is a chimera of JS and Python -- it's hard to test one using the other, but it also adds complexity to have multiple test frameworks.

rtibbles:
Well, we will soon already have two, so having something commensurate between the two of them (Gherkin) is a bit of a salve on that.

kalite/control_panel/tests/features/environment.py
```diff
@@ -0,0 +1,63 @@
...
+    hijacked_test_case = context.hijacked_test_case = HijackTestCase()
+    hijacked_test_case.setUpClass()
+     
+    def browser_url(url):
+        return urljoin(hijacked_test_case.live_server_url, url)
+    
+    context.browser_url = browser_url
+
+
+def after_all(context):
```
rtibbles: Seems like some of these functions could be inherited from a base environment - or were you trying to avoid that?

rtibbles:
Seems good - my only question about implementation is whether we want to be explicit about everything in environment.py - I can see the argument for keeping everything explicit in there, but also think it is a little unDRY to have the same functions in multiple environment.pys

MCGallaspy:
> but also think it is a little unDRY to have the same functions in multiple environment.pys

Agreed. Maybe a better idiom is to import the names from a base_environment module, like you suggested. But we also want the test writer to be able to extend it. Maybe:
```python
from base_environment import base_before_all as before_all
```
and if the user wants to extend it then
```python
from base_environment import base_before_all
def before_all(context):
    # Do something
    base_before_all(context)
```
What say ye @rtibbles?

kalite/distributed/tests/features/steps/exercises.py
```diff
@@ -0,0 +1,45 @@
...
+    # It's hard for me and makes me sad. :(
+    #context.browser.get(context.browser_url("learn/some/random/exercise"))
+    pass
+
+@given("the exercise is not available")
+def step_impl(context):
+    # See above. This is hard for me to implement. :(
+    pass
+
+@then("I should see an alert")
```
aronasorman: We should have a test expectation with:

```python
@then("I should fail non-deterministically")
def  step_impl(context):
    import random; assert random.choice([True, False])
```

Just so we can replicate our current test suite test-by-test.

rtibbles:
Yes, I agree with the DRY with options approach you outline above.

kalite/distributed/tests/features/steps/exercises.py
MCGallaspy: Veto.

rtibbles:
I would say DAMPen this up, and add Docs on writing code, and this is good to merge.

MCGallaspy:
Okay, this is much closer to completion (but not yet there, just pushing my work so far). The biggest change is the removal of the management command and cherry-picking some test runner stuff from [django-behave](https://github.com/django-behave/django-behave) to get more better-er integration with Django re: setting up test database, testing server, and running behave tests. Now, when you run the `test` management command, in addition to automatically discovering `TestCase`s in the `test` directory of an app, it will also seek behave tests in the `feature` directory and run those with behave.

MCGallaspy:
Running the whole test suite generally fails for me due to performance reasons. Let's see how Travis likes it.

To-do: Add a docs page.

kalite/control_panel/features/environment.py
```diff
@@ -0,0 +1,28 @@
...
+could be used to bootstrap other integration tests in our project.
+It sets up a test server and test database by using the LiveServerTestCase
+machinery.
+"""
+from behave import *
+from httplib import CannotSendRequest
+from selenium import webdriver
+from urlparse import urljoin
+
+from django.test.testcases import LiveServerTestCase
```
MCGallaspy: Guess I can remove some of these.

.travis.yml
```diff
@@ -10,7 +10,7 @@ before_install:
 install:
 -  npm install
 -  npm install -g grunt-cli
--  pip install -r requirements.txt
+-  pip install -r dev_requirements.txt
```
rtibbles: The way django-compressor is currently implemented, we need it for production as well, so I would rather keep both requirements.txt and dev-requirements.txt side by side.

dev_requirements.txt
```diff
@@ -0,0 +1,2 @@
+behave==1.2.4
+django-compressor==1.4
```
rtibbles: See above.

kalite/control_panel/features/environment.py
```diff
@@ -0,0 +1,21 @@
+"""
+environment.py defines setup and teardown behaviors for behave tests.
+The behavior in this file is appropriate for integration tests, and
+could be used to bootstrap other integration tests in our project.
+It sets up a test server and test database by using the LiveServerTestCase
+machinery.
+"""
+from kalite.testing.behave_helpers import login_as_admin, logout
+from kalite.testing.base_environment import before_all, after_all
```
rtibbles: DRY WIT! (Don't Repeat Yourself Writing In Tests)

rtibbles: :+1:

kalite/testing/behave_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+from kalite.testing.mixins.browser_mixins import BrowserActionMixins
+from kalite.testing.mixins.django_mixins import CreateAdminMixin
+
+
+# Maximum time to wait when trying to find elements
+MAX_WAIT_TIME = 1
+
+
+def find_css_class_with_wait(context, css_class, **kwargs):
+    """ Tries to find an element with given css class with an explicit timeout.
```
rtibbles: :+1: for doc strings.

kalite/testing/behave_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+      # TODO(MCGallaspy): Get rid of old integration tests and refactor the mixin methods
+      # as functions here.
+      class ContextWithMixin(CreateAdminMixin):
+          def __init__(self):
+              self.browser = context.browser
+      context_wm = ContextWithMixin()
+      context_wm.create_admin(username=admin_name, password=admin_pass)
+    data = json.dumps({"username": admin_name, "password": admin_pass})
+    url = reverse("api_dispatch_list", kwargs={"resource_name": "user"}) + "login/"
+    post(context, url, data)
```
rtibbles: Handy.

rtibbles:
Looks good to merge, just the one adjustment to the requirements.txt if you would. Also, when this is merged, please update the 'getting started' wiki page to add the `pip install -r dev-requirements.txt` step

PR #3406, opened 2015-03-25
Don't access settings from __init__ this gives circularity

benjaoming:
I'm currently refactoring how settings are loaded so they are not loaded using `execfile()` but just by resolving the module name.

This is all just a dumb quick fix, but it does show some other problems along the way....

### DIFF
--- a/kalite/testing/__init__.py
+++ b/kalite/testing/__init__.py
@@ -1,4 +1,0 @@
-from django.conf import settings
-
-if settings.USE_DEBUG_TOOLBAR:
-    assert settings.CACHE_TIME == 0, "Debug toolbar must be set in conjunction with CACHE_TIME=0"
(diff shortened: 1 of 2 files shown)


MCGallaspy:
For my edification, can you describe how this causes circularity?

benjaoming:
If you try to directly import `testing.settings` in `kalite.settings`, you get the problem, because `django.conf.settings` is a pre-requisite for loading `django.conf.settings`.

I mean, just put this line in kalite.settings and you'll see the problem...

    from kalite.testing.settings import *

benjaoming:
** So it's not a problem now, but it will be soon, when I change the execfile() based settings load to an __import__ based one...

benjaoming:
Okay, I will just wait with this for the bigger setuptools PR.

PR #3409, opened 2015-03-25
Simplified login for develop

rtibbles:
Let's get this into develop at least.

### DIFF
--- a/kalite/testing/mixins/browser_mixins.py
+++ b/kalite/testing/mixins/browser_mixins.py
@@ -1,3 +1,4 @@
+import json
 import time
 import re
 
@@ -17,6 +18,8 @@ from kalite.topic_tools import get_content_cache
 from django.contrib.auth.models import User
 
 from random import choice
+
+FIND_ELEMENT_TIMEOUT = 3
 
 class BrowserActionMixins(object):
 
@@ -76,13 +79,13 @@ class BrowserActionMixins(object):
         browser = browser or self.browser
         if not elem:
             if id:
-                elem = browser.find_element_by_id(id)
+                elem = WebDriverWait(browser, FIND_ELEMENT_TIMEOUT).until(EC.presence_of_element_located((By.ID, id)))
             elif name:
-                elem = browser.find_element_by_name(name)
+                elem = WebDriverWait(browser, FIND_ELEMENT_TIMEOUT).until(EC.presence_of_element_located((By.NAME, name)))
             elif tag_name:
-                elem = browser.find_element_by_tag_name(tag_name)
+                elem = WebDriverWait(browser, FIND_ELEMENT_TIMEOUT).until(EC.presence_of_element_located((By.TAG_NAME, tag_name)))
             elif css_class:
-                elem = browser.find_element_by_class_name(css_class)
+                elem = WebDriverWait(browser, FIND_ELEMENT_TIMEOUT).until(EC.presence_of_element_located((By.CLASS_NAME, css_class)))
         elem.click()
 
     def browser_send_keys(self, keys, browser=None):
@@ -319,36 +322,19 @@ class BrowserActionMixins(object):
         self.browser.find_element_by_id("id_username").submit()
 
     def browser_login_user(self, username, password, facility_name=None, browser=None):
-        """
-        Tests that an existing admin user can log in.
-        """
-        browser = browser or self.browser
-
-        login_url = self.reverse("homepage")
-        self.browse_to(login_url, browser=browser)  # Load page
-        WebDriverWait(self.browser, 10).until(EC.visibility_of_element_located((By.ID, "nav_login")))
-        self.browser.find_element_by_id("nav_login").click()
-        WebDriverWait(browser, 10).until(EC.visibility_of_element_located((By.ID, "id_username")))
-
-        # Focus should be on username, password and submit
-        #   should be accessible through keyboard only.
         try:
-            if facility_name and browser.find_element_by_id("id_facility"):
-                self.browser_activate_element(id="id_facility")
-                self.browser_send_keys(facility_name)
-        except NoSuchElementException:
-            pass
+            facility = Facility.objects.get(name=facility_name)
+        except Facility.DoesNotExist:
+            facility = Facility.objects.all()[0]
+        data = json.dumps({
+            "username": username,
+            "password": password,
+            "facility": facility.id,
+        })
+        url = self.reverse("api_dispatch_list", kwargs={"resource_name": "user"}) + "login/"
+        self.browser.execute_script('window.FLAG=false;$.ajax({type: "POST", url: "%s", data: \'%s\', contentType: "application/json", success: function(){window.FLAG=true}})' % (url, data))
+        self.browser_wait_for_js_condition("window.FLAG")
 
-        username_field = browser.find_element_by_id("id_username")
-        username_field.clear()  # clear any data
-        username_field.click()  # explicitly set the focus, to start
-        self.browser_form_fill(username, browser=browser)
-        self.browser_form_fill(password, browser=browser)
-        browser.find_element_by_class_name("login-btn").click()
-        # self.browser_send_keys(Keys.RETURN)
-
-        # wait for 5 seconds for the modal to disappear
-        WebDriverWait(browser, 5).until(EC.invisibility_of_element_located((By.ID,"id_username")))
 
     def browser_login_admin(self, username=None, password=None, browser=None):
         self.browser_login_user(username=username, password=password, browser=browser)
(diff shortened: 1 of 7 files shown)


kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -77,6 +80,72 @@ var StatusModel = Backbone.Model.extend({
...
+            contentType: 'application/json',
+            dataType: 'json',
+            type: 'POST',
+            data: JSON.stringify(data),
+            success: function(data, status, response) {
+                if (data.redirect) {
+                    window.location = data.redirect;
+                } else {
+                    // TODO (rtibbles) Reinstate the code below once
+                    // the front end app responds better to statusModel changes
```
MCGallaspy: Can you elaborate?

MCGallaspy: Also, why comment this out but not on [line 133](https://github.com/learningequality/ka-lite/pull/3409/files#diff-0494fde1b1ac12ae91bc3d53e098182eR133)?

rtibbles: At the moment, logging in requires a page refresh still, as our front end code is not set up to handle changes of user state without a page refresh.

rtibbles: It will require some refactoring of the front end code (particularly exercise, video, and content views) in order to allow them to better handle user data when the statusModel changes.

rtibbles: The app handles logouts fine - mostly because behaviour is restricted when 'is_logged_in' is false.

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -0,0 +1,285 @@
...
+    },
+
+    facility_change: function(event) {
+        this.facility = this.$("#id_facility").val();
+        if (this.model.get("simplified_login")) {
+            this.set_autocomplete();
+        }
+    },
+
+    set_autocomplete: function() {
```
MCGallaspy: Just make a comment that this initializes `self.student_usernames` for `start_autocomplete`

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -0,0 +1,285 @@
...
+            }
+        });
+        // Cannot disable autocomplete before it has been enabled
+        // So start this button off disabled and enable it when autocomplete
+        // has finished loading.
+        this.$(".password-btn").removeAttr("disabled");
+    },
+
+    key_user: function(event) {
+        if (event.which == 13) {
```
MCGallaspy: `13` corresponds to "Enter"?

rtibbles: Yup

kalite/distributed/static/js/distributed/user/views.js
rtibbles: :+1:

kalite/distributed/tests/browser_tests/distributed.py
```diff
@@ -557,7 +541,8 @@ def test_logout_link_visible(self):
     def test_logout_link_visible_small_browser_size(self):
         # Check if browser size is too small and menu is collapsed, for mobile.
         self.browser.set_window_size(640, 480)
-        expand_menus_button = self.browser.find_element_by_xpath("//*[@id=\"wrapper\"]/div[1]/div/div/div[1]/button")
+        # import pdb; pdb.set_trace()
```
MCGallaspy: DESTROY

MCGallaspy:
It's pretty, but I see this in chrome:
ss

MCGallaspy:
Tried logging in as a Django `User` without enabling the password prompt. Pressing enter causes the modal to disappear, but the dark overlay remains and the rest of the page is not interactable. Same thing happens with a username that doesn't exist.

MCGallaspy:
The password error highlighting is maybe a bit too subtle:
ss

MCGallaspy:
Now when I go to the learn tab when I'm not logged in, I don't see the sidebar.

kalite/distributed/tests/browser_tests/distributed.py
rtibbles: Woops.

rtibbles:
Ooh, fun and games!

MCGallaspy:
I think the prior behavior is that `/learn/` is redirected to `/learn/khan/` (which works as expected). Also, I have an empty `UserView` when I'm not logged in on `/learn/khan/`

MCGallaspy:
I mainly want to get this merged so I can create sessions with an api endpoint instead of fiddling around with the old login view stuff. Noted a few issues above. Perhaps someone else should review prior to merging, given there are pretty significant changes?

rtibbles:
Addressed all of @MCGallaspy's comments in the latest changes. If someone else can give it the once over, I can resolve the merge conflicts.

kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -77,6 +80,75 @@ var StatusModel = Backbone.Model.extend({
         return (new Date(new Date() - this.get("client_server_time_diff"))).toISOString().slice(0, -1);
     },
 
+    login: function(username, password, facility, callback) {
+        // Add a callback to allow functions calling this method to do things depending on the result of
```
aronasorman: I think it will be better if we expand this into a full description of this function, specifically what types are expected for each parameter, and how is it expected to be called.

kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -77,6 +80,75 @@ var StatusModel = Backbone.Model.extend({
...
+        // the login - failure, success, and particular errors that can be noted on the UI (such as incorrect username)
+        var self = this;
+
+        data = {
+            username: username || "",
+            password: password || "",
+            facility: facility || ""
+        };
+
+        $.ajax({
```
aronasorman: I'm confused on whether we should use `$.ajax`, or use our own `doRequest` function.

kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -77,6 +80,75 @@ var StatusModel = Backbone.Model.extend({
...
+            url: USER_URL + "login/",
+            contentType: 'application/json',
+            dataType: 'json',
+            type: 'POST',
+            data: JSON.stringify(data),
+            success: function(data, status, response) {
```
aronasorman: Can you move this success callback into its own function? I think that was the style in the other backbone files we have.

kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -77,6 +80,75 @@ var StatusModel = Backbone.Model.extend({
...
+                    // TODO (rtibbles) Reinstate the code below once
+                    // the front end app responds better to statusModel changes
+                    window.location.reload();
+                    // self.load_status();
+                    // if (callback) {
+                    //     callback(response);
+                    // }
+                }
+            },
+            error: function(response) {
```
aronasorman: Refactor this function out too.

kalite/distributed/static/js/distributed/distributed-server.js
rtibbles: Should be using doRequest to ensure CSRF.

kalite/distributed/templates/distributed/base.html
```diff
@@ -93,11 +93,10 @@
          // TODO(jamalex): include these in some Backbone Model state instead, synced via AJAX
          var INSTALLED_LANGUAGES_URL     = "{% url 'installed_language_packs' %}";
          var SEARCH_TOPICS_URL           = "{% url 'flat_topic_tree' lang_code=current_language %}";
-         var STATUS_URL                  = "{% url 'status' %}";
+         var USER_URL                  = "{% url 'api_dispatch_list' resource_name='user' %}";
```
aronasorman: nitpick: `=` not aligned with the rest :(

kalite/distributed/tests/browser_tests/distributed.py
```diff
@@ -547,7 +531,7 @@ def test_logout_link_visible(self):
         )
         self.assertFalse(nav_logout.is_displayed(), "The dropdown menu logout item must not be displayed yet!")
         # Activate the dropdown menu and see if the logout link is visible.
-        dropdown_menu = self.browser.find_element_by_xpath("//*[@id=\"wrapper\"]/div[1]/div/div/div[2]/ul/li[7]/a")
+        dropdown_menu = self.browser.find_element_by_id("username")
```
aronasorman: :+1:

aronasorman:
Tested a bit. Working. We'll want to update the documentation too to reflect that logins now happen through a modal. @MCGallaspy would this require changes to your screenshot scripts in order to trigger the modal?

kalite/distributed/static/js/distributed/distributed-server.js
MCGallaspy: On a slight tangent, is there a way we can namespace our own functions?

rtibbles: ohhh.. Yeah, I remember why I did this now. I wanted to circumvent the default callbacks for handling failed API responses.

MCGallaspy:
@aronasorman yeah potentially. I'll build the docs and see if there are any hiccups.

MCGallaspy:
@aronasorman actually, we only see this behavior if `SIMPLIFIED_LOGIN=True` right? Do we want the screenshots to have this setting enabled? At any rate it would be good to document this feature, too.

aronasorman:
@MCGallaspy I didn't have to enable anything to have the login as a modal. So even the default behaviour is changed.

rtibbles:
Default behaviour is now modal also.

kalite/distributed/static/js/distributed/distributed-server.js
rtibbles: Have addressed this in this issue: https://github.com/learningequality/ka-lite/issues/3411

MCGallaspy:
Note that in `kalite/settings.py` and 'python-packages/securesync/settings.py` the following lines appear:
```python
# TODO(bcipolli): change these to "login" and "logout", respectively, if/when
#  we migrate to a newer version of Django.  Older versions require these
#  to be set if using the login_required decorator.
LOGIN_URL = "/securesync/login/"
LOGOUT_URL = "/securesync/logout/"
```
Though these are no longer valid urls.

rtibbles:
Interesting. Do they actually get used anywhere?

MCGallaspy:
The decorator mentioned is used in the securesync package, but otherwise no.

kalite/testing/mixins/browser_mixins.py
```diff
@@ -316,25 +324,31 @@ def browser_login_user(self, username, password, facility_name=None, browser=Non
         """
         browser = browser or self.browser
 
-        login_url = self.reverse("login")
+        login_url = self.reverse("homepage")
```
MCGallaspy: Why go through all this rigamarole when we can just use the new API endpoint you made? (This is pretty brittle, unfortunately.) Same goes for logout. Working on something in a PR, along with some screenshot fixes.

rtibbles: I did the minimal change to get it working. Very happy to make it use the endpoint instead.

rtibbles:
@jamalex Could you weigh in how to handle the URLs that @MCGallaspy has found up above? I can't see the right way to resolve this. Not sure why we are using Django's login_required decorator for the securesync stuff and nowhere else (why are we not using our own require_admin decorator, for example?).

kalite/testing/mixins/browser_mixins.py
MCGallaspy: > Very happy to make it use the endpoint instead.

Great, I'll open a PR targeting your branch soon.

rtibbles: I am glad you understood me when I misspoke.

> Glad to have you make it use the endpoint instead.

Is of course what I meant.

aronasorman:
Changes look good @MCGallaspy.

MCGallaspy:
[implicitly_wait](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.remote.webdriver.WebDriver.implicitly_wait) sets a timeout for certain methods when there's no explicit timeout. I'm hedging my bets here against brittle `browser_mixin` methods.

>What does implicitly_wait wait for?

aronasorman:
Alright. Just give the heads @MCGallaspy if the docs are now working with this PR.

MCGallaspy:
Docs work fine and I built them to test it out. Only thing now will be to write integration tests for this new feature (but obviously we'll need to merge the integration testing framework first).

PR #3412, opened 2015-03-26
enable Gzip

66eli77:
if we have lots of binary files to serve, might need to watch out for Gzip backfire

### DIFF
--- a/kalite/settings.py
+++ b/kalite/settings.py
@@ -117,6 +117,7 @@ else:
     INSTALLED_APPS += getattr(local_settings, 'INSTALLED_APPS', tuple())
 
 MIDDLEWARE_CLASSES = (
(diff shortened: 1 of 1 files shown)


aronasorman:
I believe there was some contention regarding enabling GZip.

What I think is: we don't enable it right now, but get it in for develop. From there, A backend benchmark suite (ehem ehem) can test out whether this significantly increases our CPU usage. If not, we keep it. If yes, we make another PR that removes this line :)

aronasorman:
So given this resolution, I'm closing this PR. Can you open another targeting develop?

aronasorman:
Also, please follow the format given by https://github.com/learningequality/ka-lite/blob/develop/CONTRIBUTING.md on PR descriptions.

66eli77:
@aronasorman  opened a new PR #3417 pointing to develop

PR #3414, opened 2015-03-26
Learn navigation clean up

### LINKED ISSUE #3391: Title doesn't change during sidebar navigation
Branch: 0.13.x

Expected Behaviour: Page title should reflect current URL (i.e. topic or content item), so that when using the back and forward buttons, the information shown in the history is meaningful.

Current behaviour: Title remians the same throughout.

image

### LINKED ISSUE #3410: Pressing the Learn tab when already under /learn/ should just expand sidebar to top-level topic
Branch: 0.13.x

Observed:
Pressing the Learn tab when already on a sub-page of /learn/ does a full page reload. Since the Learn tab takes a long time to load, this is a bit painful. Observed a (new) user tap it repeatedly, trying to figure out where to go next (after collapsing sidebar).

Desired:
Just like we do for autocomplete results etc, we should intercept that click and route it via the Backbone router to open the sidebar and have it collapse back to the top-level of the topic tree.

Thoughts, @rtibbles / @Antrikshy?

rtibbles:
Fixes #3410 and fixes #3391, plus two issues I found while fixing #3410.

Summary of changes:
* Intercept click on 'nav_learn' button while on the learn tab and use the Backbone Router to resolve it.
* Do not pass the empty string that we get from splitting our URLs by "/" due to the trailing slash we have.
* The above then prevents too many inner views on the sidebar being closed when go to a higher level in the hierarchy.
* Clearing any existing callbacks from load of assessment item data, or loading Perseus files - this prevents Perseus trying to render into non-existent DOM elements if someone navigates away before it has finished loading.
* Change the title of the page as we navigate through the topic tree, now instead of always being "Learn | KA Lite", we will instead get the title of the bottom most selected node, e.g. "Learn: Basic Addition 1 | KA Lite".

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -337,22 +337,34 @@ window.ExerciseView = Backbone.View.extend({
 
         clear_messages();
 
-        item.fetch({
-            success: self.render_perseus_exercise,
-            error: function() {
-                self.get_assessment_item(attempts+1);
-            }});
+
+        // Do this in this way, so that if the view is closed prior to the fetch completing
+        // successfully, the listens will have been unbound, and the callbacks will not get
+        // called.
+        this.listenToOnce(item, "sync", self.render_perseus_exercise);
+        this.listenToOnce(item, "error", function() {
+            self.get_assessment_item(attempts+1);
+        });
+
+        item.fetch();
     },
 
     render_perseus_exercise: function(item) {
+        var self = this;
         require([KHAN_EXERCISES_SCRIPT_URL], function() {
-            Exercises.PerseusBridge.load().then(function() {
-                Exercises.PerseusBridge.render_item(item.get_item_data());
-                $(Exercises).trigger("newProblem", {
-                    userExercise: null,
-                    numHints: Exercises.PerseusBridge.itemRenderer.getNumHints()
+            // The exercise view can get closed before these loads finish happening, so check
+            // that it is still open before proceeding - otherwise, errors can ensue.
+            if (!self.closed) {
+                Exercises.PerseusBridge.load().then(function() {
+                    if (!self.closed) {
+                        Exercises.PerseusBridge.render_item(item.get_item_data());
+                        $(Exercises).trigger("newProblem", {
+                            userExercise: null,
+                            numHints: Exercises.PerseusBridge.itemRenderer.getNumHints()
+                        });
+                    }
                 });
-            });
+            }
         });
     },
 
@@ -431,6 +443,7 @@ window.ExerciseView = Backbone.View.extend({
             this.related_video_view.remove();
         }
         this.$("input").qtip("destroy", /* immediate */ true);
+        this.closed = true;
         this.remove();
     }
 
--- a/kalite/distributed/static/js/distributed/topics/router.js
+++ b/kalite/distributed/static/js/distributed/topics/router.js
@@ -1,4 +1,4 @@
-ChannelRouter = Backbone.Router.extend({
+TopicChannelRouter = Backbone.Router.extend({
     initialize: function(options) {
         _.bindAll(this);
         this.default_channel = options.default_channel;
@@ -33,12 +33,7 @@ ChannelRouter = Backbone.Router.extend({
             });
             this.channel = channel;
         }
-        splat = splat || "/";
-        if (splat.indexOf("/", splat.length - 1)==-1) {
-            splat += "/";
-            this.navigate(Backbone.history.getFragment() + "/", {replace: true});
-        }
-        this.control_view.navigate_paths(splat.split("/"));
+        this.navigate_splat(splat);
     },
 
     add_slug: function(slug) {
@@ -67,16 +62,3 @@ ChannelRouter = Backbone.Router.extend({
     }
 });
 
-TopicChannelRouter = ChannelRouter.extend({
-    navigate_channel: function(channel, splat) {
-        if (this.channel!==channel) {
-            this.control_view = new SidebarView({
-                channel: channel,
-                entity_key: "children",
-                entity_collection: TopicCollection
-            });
-            this.channel = channel;
-        }
-        this.navigate_splat(splat);
-    }
-});
--- a/kalite/distributed/static/js/distributed/topics/router.js
+++ b/kalite/distributed/static/js/distributed/topics/router.js
@@ -54,7 +54,7 @@ TopicChannelRouter = Backbone.Router.extend({
             splat += "/";
             this.navigate(Backbone.history.getFragment() + "/");
         }
-        this.control_view.navigate_paths(splat.split("/"));
+        this.control_view.navigate_paths(splat.split("/").slice(0,-1));
     },
 
     trigger_navigation_callback: function() {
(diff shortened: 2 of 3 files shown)


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -337,22 +337,34 @@ window.ExerciseView = Backbone.View.extend({
...
+
+        // Do this in this way, so that if the view is closed prior to the fetch completing
+        // successfully, the listens will have been unbound, and the callbacks will not get
+        // called.
+        this.listenToOnce(item, "sync", self.render_perseus_exercise);
+        this.listenToOnce(item, "error", function() {
+            self.get_assessment_item(attempts+1);
+        });
+
+        item.fetch();
```
jamalex: Yes, great call. This is probably the way we should always be doing things, rather than using the XMLHttpRequest callbacks.

kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -337,22 +337,34 @@ window.ExerciseView = Backbone.View.extend({
...
+        var self = this;
         require([KHAN_EXERCISES_SCRIPT_URL], function() {
-            Exercises.PerseusBridge.load().then(function() {
-                Exercises.PerseusBridge.render_item(item.get_item_data());
-                $(Exercises).trigger("newProblem", {
-                    userExercise: null,
-                    numHints: Exercises.PerseusBridge.itemRenderer.getNumHints()
+            // The exercise view can get closed before these loads finish happening, so check
+            // that it is still open before proceeding - otherwise, errors can ensue.
+            if (!self.closed) {
```
jamalex: This outer one is just for efficiency, I guess? Or does the view getting closed cause problems for the loading of Perseus itself?

kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -431,6 +443,7 @@ window.ExerciseView = Backbone.View.extend({
             this.related_video_view.remove();
         }
         this.$("input").qtip("destroy", /* immediate */ true);
+        this.closed = true;
```
jamalex: This seems like something general purpose, that should either already be part of Backbone or we should add as part of our ancestor class, in our `close` method?

jamalex: Although I guess we override that often... hmmm. But it could be done inside the `remove` method, from which we call the super, but never override in leaf classes.

kalite/distributed/static/js/distributed/exercises/views.js
rtibbles: Yeah, efficiency more than anything, and a precautionary approach, as I could never catch it at this point in my manual testing.

kalite/distributed/static/js/distributed/exercises/views.js
rtibbles: Yep, definitely worth doing when we refactor all of the Backbone View stuff to use our BaseView.

kalite/distributed/static/js/distributed/topics/router.js
```diff
@@ -1,6 +1,8 @@
-ChannelRouter = Backbone.Router.extend({
+TopicChannelRouter = Backbone.Router.extend({
     initialize: function(options) {
+        _.bindAll(this);
         this.default_channel = options.default_channel;
+        $("#nav_learn").click(this.intercept_learn_nav);
```
jamalex: I wonder whether we could handle this in a more general way (perhaps not now, to keep risk low, but as a TODO in there). It could work like this:

Intercept any click event that bubbles up to `$('body')`, and check whether the `href` begins with the root on the current base Backbone router. If so, route through there and return `false`. Otherwise, let it carry on with a page nav.

jamalex: That seems like it should be a problem that's been solved by Backbone or its community? But maybe not.

kalite/distributed/static/js/distributed/topics/router.js
```diff
@@ -16,6 +18,12 @@ ChannelRouter = Backbone.Router.extend({
         this.navigate(this.default_channel + "/", {trigger: true, replace: true});
     },
 
+    intercept_learn_nav: function(event){
+        event.preventDefault();
```
jamalex: [According to the source](http://stackoverflow.com/a/1357151), `return`ing `false` will also already do `preventDefault`.

kalite/distributed/static/js/distributed/topics/router.js
```diff
@@ -25,12 +33,7 @@ ChannelRouter = Backbone.Router.extend({
             });
             this.channel = channel;
         }
-        splat = splat || "/";
-        if (splat.indexOf("/", splat.length - 1)==-1) {
-            splat += "/";
-            this.navigate(Backbone.history.getFragment() + "/", {replace: true});
-        }
-        this.control_view.navigate_paths(splat.split("/"));
+        this.navigate_splat(splat);
```
jamalex: Can you explain this miracle? Is it moving some legacy routing stuff into the new standard approach? Do we still auto-add trailing slashes?

kalite/distributed/static/js/distributed/topics/router.js
rtibbles: Hrm, now you mention it, we do already have a catch all splat that checks for navigation events. I should probably just intercept there instead.

kalite/distributed/static/js/distributed/topics/router.js
rtibbles: This method was previously being overwritten by the navigate_channel method in TopicChannelRouter (ChannelRouter was not being used anywhere, by anything, so it seemed pointless to leave it lying around).

rtibbles: navigate_splat has all the code that used to be here.

kalite/distributed/static/js/distributed/topics/router.js
rtibbles: Fine!

kalite/distributed/static/js/distributed/topics/router.js
jamalex: Cool, thx

kalite/distributed/static/js/distributed/topics/router.js
```diff
@@ -51,7 +54,11 @@ ChannelRouter = Backbone.Router.extend({
             splat += "/";
             this.navigate(Backbone.history.getFragment() + "/");
         }
-        this.control_view.navigate_paths(splat.split("/"));
+        this.control_view.navigate_paths(splat.split("/").slice(0,-1), this.set_page_title);
+    },
+
+    set_page_title: function(title) {
+        document.title = document.title.replace(/(Learn)( |:+[^|]*)/, sprintf("$1: %s ", title));
```
jamalex: i18n? can it perhaps just replace whatever's after the | or something instead?

kalite/distributed/static/js/distributed/topics/router.js
rtibbles: Doesn't work. Will leave this as it is until we have a more general framework for routing across the app.

jamalex:
This looks great! What's your thought on risk vs benefit of targeting each of these pieces to 0.13.0 vs later? (in terms of testing, test coverage, etc)

kalite/distributed/static/js/distributed/topics/router.js
rtibbles: Fixed this.

rtibbles:
I don't think this is particularly high risk, considering most of it is just clean up of pre-existing code. I would rather have this in for 0.13.0 for the reasons you cited in your original issue.

rtibbles:
Got the ok from the big man to push the button. Merging.

PR #3426, opened 2015-03-27
Develop fix

MCGallaspy:
My machine runs out of memory running the test suite... :(

This ought to do it, though.


aronasorman:
Still failing.

rtibbles:
Passes. Merging.

PR #3429, opened 2015-03-28
Convert web+graphie: URLs into local paths, and download into assessment item packs

jamalex:


rtibbles:
Looks good. Tests pass on new content functions, so they seem to do what they are meant to.

PR #3431, opened 2015-03-29
fixes bug 3430

tarunsharma1:
Added placeholder for password in login.handlebars
before : 
screenshot 142
after : 
screenshot 143


rtibbles:
I had originally left this out deliberately, but can see how this might be jarring to a user.

PR #3435, opened 2015-03-29
Add appropriate HTTP status codes to JSON response errors.

### LINKED ISSUE #3177: Non-semantic HTTP Response codes
At the moment, many of our JSON response errors are 500s - indicating that something has gone terribly wrong inside the server. This is simply not the case, as the response has been returned deliberately, and hence should almost always return an appropriate 4xx code.

georgyberdyshev:
Adds appropriate HTTP status codes to JSON response errors.
Fixes #3177 Non-semantic HTTP Response codes

Branch: develop
Expected behavior: Provide meaningful error status codes
Steps to reproduce: User registration or login are an example. When a client device tries to register and fails to validate the chain of trust, HTTP 401 unauthorized should be returned, which is not the case.

At the moment JSON response codes are mostly 500 errors or do not contain any status message.
The changes distinguish between client and server side errors and can lead the user / developer in the right direction during the verification of returned JSON error responses.

Summary of changes:
-added missing status kwarg to JsonResponseMessageError
-added missing import of JsonResponseMessageError (kalite/facility/api_views.py)

Thanks, Georgy

### DIFF
--- a/python-packages/securesync/devices/api_views.py
+++ b/python-packages/securesync/devices/api_views.py
@@ -71,7 +71,7 @@ def register_device(request):
             # But still, good to keep track of!
             UnregisteredDevicePing.record_ping(id=client_device.id, ip=get_request_ip(request))
 
-            return JsonResponseMessageError("Failed to validate the chain of trust (%s)." % e, code=EC.CHAIN_OF_TRUST_INVALID, status=401)
+            return JsonResponseMessageError("Failed to validate the chain of trust (%s)." % e, code=EC.CHAIN_OF_TRUST_INVALID, status=500)
 
     if not zone: # old code-path
         try:
@@ -87,9 +87,9 @@ def register_device(request):
                 # A redirect loop here is also possible, if a Device exists in the central server database 
                 # corresponding to the client_device, but no corresponding RegisteredDevicePublicKey exists
                 device = Device.objects.get(public_key=client_device.public_key)
-                return JsonResponseMessageError("This device has already been registered", code=EC.DEVICE_ALREADY_REGISTERED, status=400)
+                return JsonResponseMessageError("This device has already been registered", code=EC.DEVICE_ALREADY_REGISTERED, status=409)
             except Device.DoesNotExist:
-                return JsonResponseMessageError("Device registration with public key not found; login and register first?", code=EC.PUBLIC_KEY_UNREGISTERED, status=400)
+                return JsonResponseMessageError("Device registration with public key not found; login and register first?", code=EC.PUBLIC_KEY_UNREGISTERED, status=404)
 
     client_device.save(imported=True)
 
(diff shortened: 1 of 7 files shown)


python-packages/securesync/devices/api_views.py
```diff
@@ -71,7 +71,7 @@ def register_device(request):
             # But still, good to keep track of!
             UnregisteredDevicePing.record_ping(id=client_device.id, ip=get_request_ip(request))
 
-            return JsonResponseMessageError("Failed to validate the chain of trust (%s)." % e, code=EC.CHAIN_OF_TRUST_INVALID)
+            return JsonResponseMessageError("Failed to validate the chain of trust (%s)." % e, code=EC.CHAIN_OF_TRUST_INVALID, status=401)
```
rtibbles: This should be a 500, I think, as it is entirely internal to the server.

python-packages/securesync/devices/api_views.py
```diff
@@ -87,9 +87,9 @@ def register_device(request):
                 # A redirect loop here is also possible, if a Device exists in the central server database 
                 # corresponding to the client_device, but no corresponding RegisteredDevicePublicKey exists
                 device = Device.objects.get(public_key=client_device.public_key)
-                return JsonResponseMessageError("This device has already been registered", code=EC.DEVICE_ALREADY_REGISTERED)
+                return JsonResponseMessageError("This device has already been registered", code=EC.DEVICE_ALREADY_REGISTERED, status=400)
```
rtibbles: This might be a case for a 409 - Conflict.

python-packages/securesync/devices/api_views.py
```diff
@@ -87,9 +87,9 @@ def register_device(request):
...
             except Device.DoesNotExist:
-                return JsonResponseMessageError("Device registration with public key not found; login and register first?", code=EC.PUBLIC_KEY_UNREGISTERED)
+                return JsonResponseMessageError("Device registration with public key not found; login and register first?", code=EC.PUBLIC_KEY_UNREGISTERED, status=400)
```
rtibbles: 404?

rtibbles:
Generally good - a couple of things I would adjust.

python-packages/securesync/devices/api_views.py
georgyberdyshev: I was not sure here initially, but as it is server side 500 makes sense.

python-packages/securesync/devices/api_views.py
georgyberdyshev: I thought initially of 401 as it related to device registration.
Agree here with you that 404 makes sense, due to the not found.

georgyberdyshev:
Hello @rtibbles ,
I have modified my pull request.

Thanks, Georgy

benjaoming:
@georgyberdyshev -- this looks great and in line with @rtibbles recommendations, merging.

Thanks a bunch!

georgyberdyshev:
Also big thanks @rtibbles and @benjaoming.

PR #3446, opened 2015-03-30
Change version to 0.14dev

benjaoming:


aronasorman:
This would mean language packs will be borken for develop for a while.

benjaoming:
@aronasorman how so? :)

PR #3451, opened 2015-03-30
Commented out "dev-only" alert in Perseus that gets shown to users.

jamalex:


jamalex:
@rtibbles Not sure of the best way to make these mods now.

Steps to replicate issue:

- http://127.0.0.1:8008/learn/khan/math/geometry/transformations/hs-geo-reflections/defining-reflections/
- Click on the "Reflection" button.

rtibbles:
I guess we'd better off just using a built version of react-with-addons that should exclude these developer messages?

rtibbles:
"alert" not found in:

http://fb.me/react-with-addons-0.13.1.min.js so we should just use that (or whatever version is appropriate for the version of Perseus we have).

Looks like we should be using 0.12:

http://fb.me/react-with-addons-0.12.0.min.js

Can you put this in place of the react we have?

jamalex:
Seems like a good plan for 0.14, but too risky for the imminent 0.13 release without further testing? I was trying to go as low-friction as possible.

rtibbles:
Sure - we should probably update Perseus on develop as well.

rtibbles:
Merging as a quick fix for 0.13.

Issue #3413, opened 2015-03-26
Some assessment item API endpoints returning Status 500 errors


EdDixon:
In 0.13 alpha (pre-release) discovered several assessment items that were missing the graphical elements they seemed to rely. Jamie recently became aware of an issue with the way SVG data is being handled within the offical Khan API which may be related and lead to solutions to this particular assessment item and other other assessment items similarly affected . 

All that said the particular issue can be reproduced by attempting to run the Light Puzzle 1 assessment item. Screen shots attached. There seems to be a missing but continually changing url needed which can be seen referenced via the chrome inspection tool in the second screen shot. This same issue appears to be reproduced in all the light puzzle assessment items, transformation puzzles assessment items, and recreational maps, under the math assessments. 
![capture119](https://cloud.githubusercontent.com/assets/4349806/6855542/e8274b36-d3be-11e4-9311-e3fea66d3b72.PNG)
![capture117](https://cloud.githubusercontent.com/assets/4349806/6855546/efdbafe8-d3be-11e4-8fcf-0076e1fd03fb.PNG)

jamalex:
Hmm, that looks like a different issue -- it's returning 500s on the API endpoints. Might not have anything to do with images.

jamalex:
Can you set `DEBUG = True` and inspect the response on that API endpoint?

jamalex:
Also, that X in the top right is to clear the error message. But what happened to the color on the error message bubble?

EdDixon:
Was trying to highlight that not sure why the x is there?

rtibbles:
It's from the error message alert box.

EdDixon:
I should also add this was while testing without internet availability.

jamalex:
I went ahead and tested this out, with `DEBUG = True`. Here's the traceback:

```
Traceback (most recent call last):
File "ka-lite/python-packages/tastypie/resources.py", line 195, in wrapper response = callback(request, *args, **kwargs)
File "ka-lite/python-packages/tastypie/resources.py", line 435, in dispatch_detail return self.dispatch('detail', request, **kwargs)
File "ka-lite/python-packages/tastypie/resources.py", line 458, in dispatch response = method(request, **kwargs)
File "ka-lite/python-packages/tastypie/resources.py", line 1295, in get_detail obj = self.cached_obj_get(bundle=basic_bundle, **self.remove_api_resource_names(kwargs))
[...]

PR #3453, opened 2015-03-30
smart_translate_item_data: translate strings automatically

aronasorman:
Fixes https://github.com/learningequality/ka-lite/issues/3413.

Upon reading [KA's code](https://gist.github.com/aronasorman/e8f77ad7444d55c5a5e1#file-models-py-L127) we apparently don't translate a list of strings inside an assessment item properly. The symptom of this is the 500 error code which @EdDixon caught. So now, when we see a string (perhaps passed as an argument in `item_data[field] = map(smart_translate_item_data, field_data)`, we now just translate that string directly.

Now also features tests! (tm)

### DIFF
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -517,15 +517,18 @@ def smart_translate_item_data(item_data):
     item_data and translates only the content field.
 
     """
-    if isinstance(item_data, dict):
-        if 'content' in item_data:
-            item_data['content'] = _(item_data['content']) if item_data['content'] else ""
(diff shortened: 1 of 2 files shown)


jamalex:
Looks good to me (pending testing). Will merge now, but @rtibbles please let us know if you see any issues (as this reverts and replaces https://github.com/learningequality/ka-lite/commit/82fe84145e21085648b96ca3e6e087ecc1b3d0f9)

rtibbles:
Should be fine, as long as all the data that gets passed to the function is
only ever a string or a dictionary.

PR #3454, opened 2015-03-30
0.13.x to central-develop

aronasorman:
Needed for the i18n fixes.


jamalex:
Going to have a few more things coming in here, related to reg, shortly. Then we can merge.

PR #3457, opened 2015-03-31
fixes#3336

tarunsharma1:
Added +1 to the end date and hence -32 to the begining date instead of -31
screenshot 150
as you can see, the actual date is 31st march but end date is now 1st april. Now all student progress is shown

### DIFF
--- a/kalite/coachreports/templates/coachreports/timeline_view.html
+++ b/kalite/coachreports/templates/coachreports/timeline_view.html
@@ -40,8 +40,9 @@
             });
             
             var d = new Date();
+            d.setDate(d.getDate() + 1);
             var current_date = $.datepicker.formatDate('yy-mm-dd', d);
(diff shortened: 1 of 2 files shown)


kalite/coachreports/templates/coachreports/timeline_view.html
```diff
@@ -50,4 +51,4 @@
             TimelineplotTopics(null);
         });
     </script>
-{% endblock headjs %}
\ No newline at end of file
+{% endblock headjs %}
```
tarunsharma1: I did not change anything here....i dont why why this is showing i didnt even press enter.. should i recommit or is it fine?

rtibbles:
I don't think this is the right fix. The problem is that our date filtering is not inclusive of the final date shown. This should be fixed in the way that we either pass a value to the filter, or the way the filter works. Not by changing the UI.

PR #3463, opened 2015-04-01
Global variable exposed now in sessionModel, so use it

### LINKED ISSUE #3460: "Unexpected error" and improperly displayed alert on language tab
On `develop`, logged in as admin, with `DEBUG=True`, I navigated to the language tab (`/update/languages`), and I get a big alert behind some of the page elements (see screenshot). I'm also apparently unable to download any language packs. (That could be due to tinkering on the staging server at the moment?) The alert display is evidently an error, at least.

Expected behavior: No alert at all and the ability to download language packs, or alternatively a more meaningful message than "Unexpected Error" and an alert that is displayed properly.

ss

MCGallaspy:
Fixes #3460, in the sense that there's no error. But in the case that there is an error on the page, it's still a really ugly alert. The problem was that a `$.ajax` request was made with `url: null` (because of the syntax error fixed in the PR), which resulted in our error handler in `khan-lite.js` trying to deal with a failed response that had status code 200. There probably isn't a good way in general to deal with failed responses that have status 200. Suggested regression test approach?


rtibbles:
Ugh, my bad. I thought I had caught all these globals.

aronasorman:
re: regression tests: is there a way to see if we use a variable?

aronasorman:
Maybe make `CENTRAL_SERVER` a test string, then make sure `LANGUAGE_PACK_URL` doesn't contain that string?

rtibbles:
The other alternative is to start migrating this inline javascript into separate Javascript files, import inside a compress tag, and then do unit tests to check that particular variables are actually defined.

MCGallaspy:
> The other alternative is to start migrating this inline javascript into separate Javascript files, import inside a compress tag, and then do unit tests to check that particular variables are actually defined

Will do.

MCGallaspy:
@rtibbles I've followed your pattern of extrapolating variables into `sessionModel`, and wrote a test to ensure that they have an empty default value. But this won't necessarily protect us from the error I encountered in #3460. Is this a case where we would like to retroactively write integration tests? Or wait until the code is refactored to make it more testable?

rtibbles:
The failure in #3640 was from a variable failing to be defined, so at
least, we are protecting against that. Exactly how we should test that
javascript variables set to equal django context variables take on the
correct value, I am unsure.

MCGallaspy:
> The failure in #3640 was from a variable failing to be defined

More specifically, it was from the variable failing to be a valid url. So the default value (the empty string) could still cause this error. But I agree I don't see an easy way to test this, given that the URL will be generated by the server. Hence an integration test?

MCGallaspy:
karma error? Works fine for me.

MCGallaspy:
@rtibbles I have no idea what could be causing this test error. The failing command runs fine on my machine. Any suggestions?

rtibbles:
I think the reason the test is failing on Travis is because of line 65 in base.html
My guess would be that the current_language variable is somehow set to "en" on Travis during bundleassets, and not on your machine.
So, that i18n js is getting bundled in locally, but not when it runs on travis.
As none of the other JS that has been bundled in for Karma so far relied on gettext, this error was not apparent.

kalite/distributed/templates/distributed/base.html
```diff
@@ -61,15 +61,13 @@
             <script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
         {% endif %}
 
+        {% compress js file basejs %}
         <!-- JS translation file statically included as a back-up, so that if the saved file below isn't available, we don't choke. -->
-        {% if current_language != "en" %}
```
rtibbles: Yeah, this seems like the best solution, with no drawback really. Could, perhaps, put these around the script tag below?

My main interest is how these compiled tags would work at run time - it is something I am not clear on, but I assume it will pull in whichever i18n/<lang>.js file was active when we ran the compressor.

MCGallaspy: Gotcha

rtibbles:
This will work for now unless we decide to do offline compression for the build process. Merging.

PR #3467, opened 2015-04-02
More i18n fixes

aronasorman:
Summary of changes:
1. Match the KA channel description with that's in the po file (already fixed before, but someone edited it.)
screen shot 2015-04-02 at 9 35 01 am

2. i18nize the exercise title (don't mind the exercise text, this is a khan exercise).
screen shot 2015-04-02 at 9 34 40 am

3. i18nize the hint button.
screen shot 2015-04-02 at 9 36 17 am

Remaining issues:

Couldn't i18nize the "Remaining hints" text in the hint button:
screen shot 2015-04-02 at 9 37 13 am

Fixing this properly would require more extensive changes to the perseus codebase, related to integrating Django's and `jed.js`'s catalog file. I'm getting there, but I need more time, and I believe this shouldn't be a blocker for 0.13.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -397,7 +397,7 @@ window.ExerciseView = Backbone.View.extend({
     },
 
     update_title: function() {
-        this.$(".exercise-title").text(this.data_model.get("title"));
+        this.$(".exercise-title").text(gettext(this.data_model.get("title")));
     },
 
     hint_used: function() {
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -397,7 +397,7 @@ window.ExerciseView = Backbone.View.extend({
     },
 
     update_title: function() {
-        this.$(".exercise-title").text(gettext(this.data_model.get("title")));
+        this.$(".exercise-title").text(this.data_model.get("title"));
     },
 
     hint_used: function() {
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -165,7 +165,7 @@ def get_exercise_cache(force=False, language=settings.LANGUAGE_CODE):
                 exercise_template = os.path.join(exercise_lang, exercise_file)
 
 
-            with i18n.translate_block(exercise_lang):
+            with i18n.translate_block(language):
                 exercise["available"] = available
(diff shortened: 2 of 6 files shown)


kalite/topic_tools/__init__.py
```diff
@@ -534,8 +534,7 @@ def smart_translate_item_data(item_data):
             elif isinstance(field_data, list):
                 item_data[field] = map(smart_translate_item_data, field_data)
 
-
-            return item_data
+        return item_data
```
aronasorman: Heh, the dangers of wrong indentation. Previously, this would only iterate on a field once.

kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -397,7 +397,7 @@ window.ExerciseView = Backbone.View.extend({
     },
 
     update_title: function() {
-        this.$(".exercise-title").text(this.data_model.get("title"));
+        this.$(".exercise-title").text(gettext(this.data_model.get("title")));
```
jamalex: Fix translation in API, rather than client-side.

jamalex:
Everything else looks good. After addressing that gettext thing as discussed, feel free to merge.

kalite/topic_tools/__init__.py
```diff
@@ -165,7 +165,7 @@ def get_exercise_cache(force=False, language=settings.LANGUAGE_CODE):
                 exercise_template = os.path.join(exercise_lang, exercise_file)
 
 
-            with i18n.translate_block(exercise_lang):
+            with i18n.translate_block(language):
```
aronasorman: @rtibbles can you see if this change breaks anything? The only one I can think of is since we don't use `select_best_available_language`, we won't be able to fall back to the  toplevel languages when the dialect versions aren't present (i.e. `pt-BR` won't fall back to `pt` when the former is absent, but the latter is present, It always reverts to `en`).

rtibbles: Yeah, that seems to be the right intuition for what will happen.

kalite/topic_tools/__init__.py
rtibbles: Oh, yikes!

kalite/topic_tools/__init__.py
aronasorman: Looking at our central server's `supported_languages.json` we don't have these cases yet, so this regression is not a blocker for me.

aronasorman:
Tests pass. Merge!

PR #3468, opened 2015-04-02
Registration tweaks (destined for central server submodule)

jamalex:
Two small fixes in here:
- Send us a notification with diagnostic info if the "Client device did not verify" occurs again in production.
- Ensure public key submitted for registration uses "+" rather than " ", to avoid reg issue caused by parsing oddities in some sort of weird browser.

### DIFF
--- a/python-packages/securesync/devices/api_views.py
+++ b/python-packages/securesync/devices/api_views.py
@@ -54,7 +54,10 @@ def register_device(request):
     if not isinstance(client_device, Device):
         return JsonResponseMessageError("Client device must be an instance of the 'Device' model.", code=EC.CLIENT_DEVICE_NOT_DEVICE)
     if not client_device.verify():
-        return JsonResponseMessageError("Client device must be self-signed with a signature matching its own public key.", code=EC.CLIENT_DEVICE_INVALID_SIGNATURE)
+        # We've been getting this verification error a lot, even when we shouldn't. Send more details to us by email so we can diagnose.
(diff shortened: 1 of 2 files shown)


jamalex:
Note that these have no impact on the distributed server, but are coming through here to minimize divergence between the release branches and the central-x branches.

python-packages/securesync/devices/api_views.py
```diff
@@ -54,7 +54,10 @@ def register_device(request):
     if not isinstance(client_device, Device):
         return JsonResponseMessageError("Client device must be an instance of the 'Device' model.", code=EC.CLIENT_DEVICE_NOT_DEVICE)
     if not client_device.verify():
-        return JsonResponseMessageError("Client device must be self-signed with a signature matching its own public key.", code=EC.CLIENT_DEVICE_INVALID_SIGNATURE)
+        # We've been getting this verification error a lot, even when we shouldn't. Send more details to us by email so we can diagnose.
+        msg = "\n\n".join([request.body, client_device._hashable_representation(), str(client_device.validate()), client_device.signed_by_id, client_device.id, str(request)])
+        send_mail("Client device did not verify", msg, "kalite@learningequality.org", ["errors@learningequality.org"])
```
aronasorman: Does this get triggered on the distributed server?

jamalex: Nope, these api_views are only mounted at all on central.

jamalex: Hmm, at least, only `used` on central. Looking into the mounting now.

aronasorman:
Just waiting for tests to pass.

PR #3469, opened 2015-04-02
Tweaks to help with debugging reg errors on the central server

jamalex:


kalite/__init__.py
```diff
@@ -27,3 +28,8 @@ def is_installed():
         # except:
         #     return False
         return True
+
+
+# Modify Request's default user agent to include the KA Lite version number
+base_headers = requests.defaults.defaults["base_headers"]
+base_headers["User-Agent"] = ("ka-lite/%s " % VERSION) + base_headers["User-Agent"]
```
aronasorman: Looks like you found a better solution than `patch`ing requests!

jamalex: Turns out the requests code I had been inspecting was for a different version than we use in KA Lite. The version we use allows the above approach.

aronasorman:
Passing now.

PR #3470, opened 2015-04-03
Inline help features

MCGallaspy:
Not yet done, but opening a PR to keep track of the changes. Reviewers, please take note of the feature spec in particular: `kalite/distributed/features/inline-help.feature`.

**What's inside so far:**
1. Behave test specs for the inline help feature
2. Adding some test helper methods
3. Re-factoring the behave environment a bit (for more general use)
3. Also fixes an issue I noticed: running the test command without any test labels wasn't running *any* behave tests, when in fact we'd like to run them *all* in that case. Fixes that in `kalite/testing/testrunner.py`

**What's next:**
1. Implement test steps.

### DIFF
--- a/kalite/testing/behave_helpers.py
+++ b/kalite/testing/behave_helpers.py
@@ -1,13 +1,31 @@
 """
 These methods will probably be used again and again in behave tests.
-We'll make a few assumptions: every function here takes a behave context
-as the first positional argument, and should assert what it expects from 
-that context in order to function (instead of just silently failing).
+We'll make an assumptions: every function here takes a behave context
+as the first positional argument. 
+
+Useful functions you should know about and use include:
+
+For finding and interacting with elements safely:
+* elem_is_invisible_with_wait
+* find_css_class_with_wait
+* find_id_with_wait
+
+For navigating the site:
+* build_url
+
+For logging in and out:
+* login_as_coach
+* login_as_admin
+* logout
+
+For interacting with the API:
+* post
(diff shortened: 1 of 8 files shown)


MCGallaspy:
Also, we'll see how Travis likes behave tests here.

MCGallaspy:
Behave tests run beautifully, but I guess it would help if I implemented the steps.

MCGallaspy:
I expect this to fail until we finish the inline help feature. But until then it can stand as a brave *test*ament to testing. Perhaps those who wish to write their own tests can use it as a springboard and/or sniff out smelly code within.

kalite/testing/behave_helpers.py
```diff
@@ -1,13 +1,31 @@
 """
 These methods will probably be used again and again in behave tests.
-We'll make a few assumptions: every function here takes a behave context
-as the first positional argument, and should assert what it expects from 
-that context in order to function (instead of just silently failing).
+We'll make an assumptions: every function here takes a behave context
```
aronasorman: *assumption

MCGallaspy:
Oh, why'd you merge this? Test will be failing now for quite some time.

aronasorman:
Oh, I thought this was the central develop PR.

MCGallaspy:
Actually, I think github closed this automatically because you merged something that supersedes it (the PR I made where I cherry picked commits from here, but also a couple extra commits).

Issue #3459, opened 2015-03-31
Redundant label/placeholder on login modal


Antrikshy:
![screen shot 2015-03-31 at 12 47 16 pm](https://cloud.githubusercontent.com/assets/6601788/6927741/26edbf1a-d7a4-11e4-8bea-71ed814d5d3a.png)

>Username Username
>Password Password

ಠ_ಠ

Shouldn't we have one or the other? Maybe remove the label and make it `sr-only`? Or just get rid of the placeholder? I'm a fan of this login form: http://getbootstrap.com/examples/signin/

Assign me to this and send over your recommendations.

rtibbles:
I couldn't give a monkeys' either way, to be honest. The problem with *just* the prefill though, is that 'Facility' is a selection drop down, so it won't be consistent.

radinamatic:
Btw, sorry for "stealing" this one under your nose @Antrikshy , I literally just saw your last sentence about wanting to be assigned this... :(
   
The problem with Facility drop down that @rtibbles mentioned is avoided by including a "dummy" "selected disabled" option. I thought it would create problems with the screen readers, but NVDA read out everything just fine.

Antrikshy:
Oh, that's no problem. I'll find other stuff to work on. Right now, I'm getting used to a very busy quarter schedule anyway.

I was thinking of a blank option as well. Yours looks very nice.

MCGallaspy:
@radinamatic looks like you decided to tackle this one. Be sure to tag this issue in a PR when you're ready, so we know we can close it.

radinamatic:
You can close it already, solved in #3647.

radinamatic:
@MCGallaspy, this can be closed...

PR #3472, opened 2015-04-03
Reworked the login dialog to avoid redundant label/placeholder #3459

radinamatic:
Not exactly as the example @Antrikshy posted, but close.
A11y tested with NVDA on FF.


Antrikshy:
@radinamatic Screenshots?

radinamatic:
You're reading my mind...? Just was making one... :) 

login-dialog

kalite/distributed/hbtemplates/user/login.handlebars
```diff
@@ -4,9 +4,10 @@
             <form class="form-horizontal">
             {{#if facilities}}
             <div class="form-group form-group-lg" id="id_facility-container">
-              <label class="col-sm-3 control-label h3" for="id_facility" id="id_facility-addon">{{_ "Facility" }}</label>
-              <div class="col-sm-9">
-                <select class="form-control input-lg" id="id_facility" aria-describedby="id_facility-addon">
+              <label class="sr-only control-label h3" for="id_facility">{{_ "Facility" }}</label>
```
rtibbles: Any reason for removing the id tag here?

kalite/distributed/static/css/distributed/bootstrap-overrides.css
```diff
@@ -290,6 +290,11 @@ form .input-group {
     background-color: #40775F !important;
 }
 
+.login-btn, btn-block {
+    display: block;
+    width: 100%;
```
rtibbles: Can you set SIMPLIFIED_LOGIN=True in your local_settings.py and see how this width setting interacts with the additional button that appears with this setting?

kalite/distributed/static/css/distributed/bootstrap-overrides.css
```diff
@@ -455,6 +460,18 @@ footer {
     width: 82px;
 }
 
+/* ===================================
+    Modal Dialog Login
+=================================== */
+
+.modal-dialog {
```
rtibbles: Why do we want to override all bootstrap modal dialog boxes? Is there an issue with the way that bootstrap handles the sizing?

kalite/distributed/static/css/distributed/bootstrap-overrides.css
```diff
@@ -455,6 +460,18 @@ footer {
...
+    width: 400px;
+}
+
+div.panel-body form.form-horizontal .form-group {
```
rtibbles: Again, I would rather not do such non-specific selectors there. Can you make this more specific to the login box where we want this behaviour to occur (and hence move it out of bootstrap-overrides also).

rtibbles:
A few concerns with over generalized CSS selectors (using a hammer to crack a nut), also the login button width might be a problem in the situation mentioned above.

Lastly, error handling will now be somewhat subtle. There is a popover that will show the error message, but the highlighting of the input box is very subtle on bootstrap defaults. Please have a look at the behaviour there with the addition of the 'has-error' class to the containing div of the form elements.

kalite/distributed/static/css/distributed/bootstrap-overrides.css
radinamatic: Ok, will work on more-specific-selector solution.
Question: if the changes are to be moved **from** bootstrap-overrides, where do we move them to?

rtibbles: Probably into khan-lite.css would be best.

kalite/distributed/static/css/distributed/bootstrap-overrides.css
radinamatic: No reason I can think of right now, I will eliminate this.

kalite/distributed/static/css/distributed/bootstrap-overrides.css
radinamatic: This is how it looks with simplified login:

simplified-login-dialog

kalite/distributed/hbtemplates/user/login.handlebars
radinamatic: Must be the Git gremlin, I'll put it back... :)

kalite/distributed/static/css/distributed/bootstrap-overrides.css
rtibbles: Odd, the password box should be hidden, and the login button should be disabled until a username is typed in.

rtibbles: Also, could you set the lower button to be centrally aligned? I think it is set to float: right currently.

rtibbles: Or is this after pressing the 'Admin/Coach' button?

radinamatic: Check the latest commit:

a) Login dialog loaded
simplified-login-dialog2

b) Wrong username no password
simplified-login-dialog3

c) Wrong username & password
simplified-login-dialog4

One thing I noticed is that the error message tooltip stays on username even when it is correctly inputed for the given facility, and the password is the incorrect one. I'm guessing lines 219-220 in /facility/forms.py?

radinamatic:
I possibly overdid a bit the "un-subtling" of the input fields with 'has-error', but that's easily adjustable.

Will have to work on accessibility of those popover tooltips though,  they only get read if I put aria-live="assertive" which is another hammer... :)

kalite/distributed/static/css/distributed/bootstrap-overrides.css
rtibbles: No - this validation code is either happening in facility/api_resources.py or in distributed/static/js/user/views.js

Incorrect popover behaviour is probably happening in the latter.

rtibbles:
Sorry for letting this get stale, @radinamatic - could you rebase, and maybe tone the has-error back down a little and then I'll merge this.

radinamatic:
Just seeing your last comment here @rtibbles, I'm not totally satisfied with the manner SRs deal with the tooltips here... As soon as I find the acceptable solution, I'll rebase, adjust the CSS and commit again.

PR #3473, opened 2015-04-03
Replace urllib.urlopen with more robust requests.get method.

jamalex:
This avoids a "peer did not return certificate" error being thrown; urllib.urlopen is a bit flaky.


aronasorman:
Nondeterministic test failure. Merging.

PR #3474, opened 2015-04-03
Replaces react-with-addons with minned production version.

rtibbles:
This is a parallel fix to #3451, but we didn't make this drastic a change there for stability.


aronasorman:
Do we need to this when we already have `django-compressor`? I find that including the minified versions makes debugging harder.

rtibbles:
Universally, no. But, the react development code has alerts in it. Yes, alerts. See @jamalex's PR referenced above.

PR #3476, opened 2015-04-03
Fixes #3168: New responsive menu for mobile devices and small desktop screen views

### LINKED ISSUE #3168: Teacher (and presumably other navs) aren't properly mobile responsive
While use on cell phones is not our primary use case by any means, I think the second layer nav system for teachers and admins would be better if it could responsively collapse like the top level nav (maybe using a hamburger, maybe using some other format).

See here a screenshot on the Nexus 5:

image

@jtamiace Can you take a look at making this a more responsive design over all?

jtamiace:
This pull request fixes the closed issue https://github.com/learningequality/ka-lite/issues/3168

The dropdown menu for smaller screens is now much more aesthetically pleasing. 
Also, the second search bar has been removed from distributed/base.html. @rtibbles please check to see that this hasn't adversely affected the autofill; seems to work on my end though. 

Following are screenshots of the new mobile navigation. Critiques and suggestions are very welcome!

Not logged in screen:
mobile not logged in

Unopened menu for admins/teachers:
mobile menu closed

Opened menu for admins/teachers:
mobile menu open

Second level navigation for teachers:
teacher manage mobile

Dropdown for students:
mobile student menu

### DIFF
--- a/kalite/distributed/templates/distributed/base.html
+++ b/kalite/distributed/templates/distributed/base.html
@@ -89,18 +89,32 @@
             FastClick.attach(document.body);
         });
 
-         /* Jessica TODO: Backbonify this function and this page... when the time comes.
+         /* TODO(jtamiace): Backbonify this function and this page... when the time comes.
             This function addresses Bootstrap's limitation of having a dropdown menu in an already collapsed menu*/
-          window.addEventListener("resize", function (e) {
-              if ( $('body').innerWidth() <= 750 ) {
+          var collapsedNav = $(function () {
+            if ( $('body').innerWidth() <= 750 ) {
                 document.getElementById("user-name-a").removeAttribute("data-toggle");
                 document.getElementById("user-name").classList.add("open");
-              }
-              if ( $('body').innerWidth() >= 751 ) {
+              } 
+            if ( $('body').innerWidth() >= 751 ) {
                 document.getElementById("user-name-a").setAttribute("data-toggle", "dropdown");
                 document.getElementById("user-name").classList.remove("open");
               }
             });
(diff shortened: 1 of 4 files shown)


rtibbles:
@jtamiace I don't think this will affect the auto-fill, as you deleted the one without the ID. If anything, I think the autofill may have been missing on mobile previously.

rtibbles:
Yep... autocomplete did not previously work on the collapsed version.

jtamiace:
cool lucky guess!

Antrikshy:
Ooooh pretty!

kalite/distributed/templates/distributed/base.html
```diff
@@ -82,9 +84,26 @@
 
         <script type="text/javascript" src="{% static 'js/fastclick.js' %}"></script>
         <script>
+
         $(function() {
             FastClick.attach(document.body);
         });
+
+         /* TODO(jtamiace): Backbonify this function and this page... when the time comes.
```
rtibbles: @jtamiace Do you think you'll have time during the 0.14 cycle to tackle doing this? Wrapping the whole navbar into a Backbone view that is then toggled properly by changes in the statusModel would be great. This is also going to be helpful as part of the clean up required for the API driven login as well (at the moment we force a page refresh because our Javascript does not respond well to dynamic changes in the statusModel).

jtamiace: @rtibbles Definitely will tackle this when I'm not working on content curation; great chance to practice backbone. There's a couple of months left in this cycle right?

rtibbles: Yep, got until the end of the quarter!

PR #3480, opened 2015-04-06
Docs translation changes for central server

MCGallaspy:
**What's inside:**
* Add a command line option to the `screenshots` management command to set the language for the session when taking screenshots. Note: the language pack must be downloaded already for this to work.
* Update the sphinx-docs screenshot directive to use recommended `kalitectl.py` now that `manage.py` has been deprecated and removed.
* Fix a bug in `kalitectl.py` when `KALITE_DIR` is not set to include appropriate directories in the python path.
* Update the README in `sphinx-docs` to add notes about building translated docs.

These changes are needed for the `update_pot` management command to run successfully on the central server. [See here.](https://github.com/fle-internal/ka-lite-central/pull/252)


aronasorman:
I get this error when trying to run `sphinx-intl update -p _build/locale -l xx` on my virtualenv:

a

I guess this doesn't work for python 2.6? Which is okay, but it might be good to add that to your instructions.

aronasorman:
I did a `pip install ordereddict` and it's working now. Please add that to your `requirements.txt`.

aronasorman:
Oh, if this needs to be in `central-develop`, then we can cherry-pick the changes in here after we merge this.

PR #3488, opened 2015-04-07
Cherry pick behave

MCGallaspy:
Cherry pick changes from #3470 that might be useful for other people wishing to write tests. The main changes are fixing the test runner to run behave tests even when no app is specified from the command line, and adding several helper methods which can be used in steps.


kalite/distributed/features/steps/exercises.py
```diff
@@ -9,7 +8,7 @@
 
 @given("I open an exercise")
 def step_impl(context):
-    context.browser.get(context.browser_url("/"))
+    pass
```
aronasorman: Who'll implement this?

kalite/testing/behave_helpers.py
```diff
@@ -1,13 +1,31 @@
 """
 These methods will probably be used again and again in behave tests.
-We'll make a few assumptions: every function here takes a behave context
-as the first positional argument, and should assert what it expects from 
-that context in order to function (instead of just silently failing).
+We'll make an assumptions: every function here takes a behave context
```
aronasorman: *assumption

kalite/distributed/features/steps/exercises.py
MCGallaspy: No plan to implement this at this point. More of an example than anything. The backbone stuff is hard to interact with using Selenium presumably because it wasn't designed with that in mind.

aronasorman:
Just one comment about the typo.

PR #3489, opened 2015-04-07
3-column enhancement for search page

Antrikshy:
I always thought the search page could do with a refresh, but I couldn't quite put a finger on what was wrong with it. I don't think anything was wrong with it, but I decided to make a few enhancements anyway.

1. It does not look as bland
2. Requires less scrolling (mobile optimization FTW!)
3. No more wasted whitespace

Here's what it looks like:

screen shot 2015-04-07 at 3 22 21 pm

screen shot 2015-04-07 at 3 22 50 pm

I'm not sure if there are any more categories than the ones pictured above, but this new layout should scale gracefully, stacking them underneath, since it uses a simple Bootstrap grid.

Suggestions for enhancements like splashes of color etc. are welcome.

### DIFF
--- a/kalite/distributed/static/css/distributed/bootstrap-overrides.css
+++ b/kalite/distributed/static/css/distributed/bootstrap-overrides.css
@@ -22,6 +22,7 @@ footer {
     width: 100%;
     bottom: 0;
     left: 0;
+    padding: 10px 0;
 }
 
 a:hover, a:focus {
@@ -445,8 +446,7 @@ footer {
     margin: 0;
 }
 #fle-logo {
-    margin-top: 15px;
-    margin-bottom: 5px;
+    margin: 5px 0;
     height: 18px;
     width: 268px;
 }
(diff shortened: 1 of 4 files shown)


Antrikshy:
I just decided to add a tiny fix for the sidebar back button into this PR. It just gets rid of the awkward 2-4 pixel overlapping of the vertical bar.

Antrikshy:
Also just added a very minor fix to the main site footer. There is now some 10px padding, so the Creative Commons badge does not awkwardly touch the very bottom of the page.

jtamiace:
Looks really good Ant :) Giving it extra space makes it a lot nicer to look at. Good job taking out the tables too!

Antrikshy:
Yeah, it's a pretty neat enhancement without messing up the structure of the page. Surprisingly clean to implement. I heard @alexkchan wanted to work on search pagination to show more than 25 results. That should still work in the exact same way as before.

PR #3492, opened 2015-04-08
Tweak release notes

MCGallaspy:
Now reflects actual use.


MCGallaspy:
Test failure due to a previously undetected race condition in my brand new tests. I'm cooking up a fresh new solution right now.

PR #3496, opened 2015-04-09
Fixes #2470: "Cant register because Client device must be self-signed with a signature matching its own public key"

### LINKED ISSUE #2470: Cant register because Client device must be self-signed with a signature matching its own public key
Fresh KA Lite install on Arch Linux for personal use. The server is being manually launched with ./start.sh each time in /home/matt/Apps/ka-lite. Google brought me to someone who had the same thing happen to them, and it was closed as manually launching the server prevented this problem.

https://github.com/learningequality/ka-lite/issues/2430

Upon signing in as an administrator and clicking on 'videos' to download some, I get the following after attempting registration:

Sorry, there was an error during registration. Please make sure your device is fully up to date (run python manage.py update and then stop/start the server).

The error message was: Client device must be self-signed with a signature matching its own public key.

This is what I get when I try doing python manage.py update.
[matt@arch ka-lite]$ cd kalite
[matt@arch kalite]$ python manage.py update
  File "manage.py", line 58, in <module>
    from django.core.management import execute_from_command_line
  File "/home/matt/Apps/ka-lite/python-packages/django/core/management/__init__.py", line 5, in <module>
    import imp
  File "/usr/lib/python3.4/imp.py", line 19, in <module>
    from importlib._bootstrap import SourcelessFileLoader, _ERR_MSG, _SpecMethods
ImportError: No module named 'importlib._bootstrap'

Anyone have any idea how to fix it?

jamalex:
For now, this is just the regression test, as I want to see whether it fails on Travis.

### DIFF
--- a/kalite/facility/models.py
+++ b/kalite/facility/models.py
@@ -90,7 +90,7 @@ class FacilityGroup(DeferredCountSyncedModel):
 
     def __init__(self, *args, **kwargs):
         super(FacilityGroup, self).__init__(*args, **kwargs)
-        self._unhashable_fields.append("description") # since it's being stripped out by minversion, we can't include it in the signature
+        self._unhashable_fields += ("description",) # since it's being stripped out by minversion, we can't include it in the signature
 
     class Meta:
         app_label = "securesync"  # for back-compat reasons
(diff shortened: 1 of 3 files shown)


jamalex:
@aronasorman Looks like there's no need to cherry-pick this against master (0.12), since the code modifying `_unhashable_fields` didn't yet exist on that branch. The reason 0.12.11 fails is that even though it's doing everything right, the central server is stripping out the description before checking the signature. Merging this to the central server will fix for all clients.

jamalex:
Ok, got it to fail in the right places, and pass in the right places, I believe. Merging and moving onwards!

(The current failure on the HEAD is a stochastic, unrelated one)

PR #3497, opened 2015-04-09
Add method to wait for page load after a click

MCGallaspy:
After a behave test failed on python 2.6, I realized I was flirting with race conditions. The upshot is that if one clicks an element in selenium and causes a page load, selenium will not wait for the page to load. In general, the `WebElement.click` method will not wait for anything and so should be used with great caution. See [this blog post](http://www.obeythetestinggoat.com/how-to-get-selenium-to-wait-for-page-load-after-a-click.html) for some context.

**What's inside:**
* A helper method for the common case that you want to click an element and wait for a new page to load as a result.
* Fixed an existing test to explicitly wait for a page load.

### DIFF
--- a/kalite/testing/behave_helpers.py
+++ b/kalite/testing/behave_helpers.py
@@ -5,6 +5,9 @@ as the first positional argument.
 
 Useful functions you should know about and use include:
 
+For clicking elements that cause a page load (`click` is not safe!):
+* click_and_wait_for_page_load
+
 For finding and interacting with elements safely:
 * elem_is_invisible_with_wait
 * find_css_class_with_wait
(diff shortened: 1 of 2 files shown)


kalite/testing/behave_helpers.py
```diff
@@ -45,6 +48,26 @@
...
+    a TimeoutException. (So if you pass an element to click that doesn't cause a 
+    page load, then you'll probably get a TimeoutException.)
+    context: a behave context
+    elem: a WebElement to click.
+    wait_time: Optional. Max wait time for the page to load. Has a default value.
+    """
+    # The body element should always be on the page.
+    wait_elem = context.browser.find_element_by_tag_name("body")
+    elem.click()
+    WebDriverWait(context.browser, wait_time).until(
```
rtibbles: Yep, this seems like the right way to do it. (reference for others interested: http://www.obeythetestinggoat.com/how-to-get-selenium-to-wait-for-page-load-after-a-click.html )

rtibbles:
This seems right. Will merge when tests pass.

PR #3503, opened 2015-04-10
Handle case where num_messages is 0

### LINKED ISSUE #3307: browser_check_django_message no longer works correctly with num_messages=0
There are several tests on ka-lite-central that call `browser_check_django_message` with `num_messages=0`, in order to assert that no alerts were displayed.

The added line:
`WebDriverWait(self.browser, 30).until(EC.presence_of_element_located((By.CLASS_NAME, "alert")))`
is not set up to handle this case, and hence pauses 30 seconds before failing out.

It's uncertain how long should be waited in order to check for lack of alert messages, but perhaps we could get away with skipping that `WebDriverWait` line if `num_messages=0`?

MCGallaspy:
Fixes #3307. Should I target this to `0.13.x` instead?

### DIFF
--- a/kalite/testing/mixins/browser_mixins.py
+++ b/kalite/testing/mixins/browser_mixins.py
@@ -94,11 +94,11 @@ class BrowserActionMixins(object):
         """Both central and distributed servers use the Django messaging system.
         This code will verify that a message with the given type contains the specified text."""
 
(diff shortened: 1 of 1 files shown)


MCGallaspy:
What's our strategy here? Merging `0.13.x` back into `develop` at some point?

aronasorman:
re: 0.13 to develop --  yes. So this PR is redundant. Duplicate of https://github.com/learningequality/ka-lite/pull/3504

Issue #3501, opened 2015-04-10
Deprecated playlist on homepage preventing installer upgrades


EdDixon:
In upgrading from 0.12 to 0.13 using installer the playlists.js references appear to break proper formatting of the landing page.![screenshot from 2015-04-10 10-20-46](https://cloud.githubusercontent.com/assets/4349806/7092285/9de4f388-df6d-11e4-8198-865634058f08.png)

PR #3506, opened 2015-04-10
Resolves depricated playlist in 0.13 release with more clean up needed

EdDixon:
Issue discovered in using the installers to update 0.12 to 0.13 this should resolve those issues as a temp fix however since the entire playlist is deprecated we should go back through this and remove other parts such as divs and tables for 0.14 that could right now affect release of 0.13.


MCGallaspy:
Fixes #3501.

MCGallaspy:
Seems ok. Will merge after tests pass.

PR #3507, opened 2015-04-10
Exercises refactor

### LINKED ISSUE #2228: Refactor Backbone Views for Exercises, Quizzes, and Tests
Current state:
Code is currently repeated redundantly across three different instantiations of the exercises.

Desired state:
Same behaviour - but with code modularized so the same wrapper functions are called with different injections for API endpoints and feedback and response behaviour.

rtibbles:
Fixes #2228

Summary of changes:
* Makes all views in exercise views.js use BaseView as the BaseClass.
* Abstracts redundant code from Practice, Test, and Quiz views into a Base class for all.
* Cleans up some old Nalanda code

I have manually tested this on exercises and tests. Both seem to function as intended.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -744,30 +744,23 @@ window.ExercisePracticeView = ExerciseWrapperBaseView.extend({
 });
 
 
-window.ExerciseTestView = BaseView.extend({
+window.ExerciseTestView = ExerciseWrapperBaseView.extend({
 
     start_template: HB.template("exercise/test-start"),
 
     stop_template: HB.template("exercise/test-stop"),
 
-    initialize: function() {
+    load_user_data: function() {
 
-        _.bindAll(this);
+        // load the data about this particular test
+        this.test_model = new TestDataModel({test_id: this.options.test_id});
+        var test_model_deferred = this.test_model.fetch();
 
-        if (window.statusModel.get("is_logged_in")) {
+        // load the data about the user's overall progress on the test
+        this.log_collection = new TestLogCollection([], {test_id: this.options.test_id});
+        var log_collection_deferred = this.log_collection.fetch();
 
-            // load the data about this particular test
-            this.test_model = new TestDataModel({test_id: this.options.test_id});
-            var test_model_deferred = this.test_model.fetch();
-
-            // load the data about the user's overall progress on the test
-            this.log_collection = new TestLogCollection([], {test_id: this.options.test_id});
-            var log_collection_deferred = this.log_collection.fetch();
-
-            this.user_data_loaded_deferred = $.when(log_collection_deferred, test_model_deferred).then(this.user_data_loaded);
-
-        }
-
+        this.user_data_loaded_deferred = $.when(log_collection_deferred, test_model_deferred).then(this.user_data_loaded);
     },
 
     finish_test: function() {
@@ -796,7 +789,9 @@ window.ExerciseTestView = BaseView.extend({
         } else {
 
             if(this.log_model.get("complete")){
+
                 this.finish_test();
+
             } else {
 
                 this.listenTo(this.log_model, "complete", this.finish_test);
@@ -835,17 +830,7 @@ window.ExerciseTestView = BaseView.extend({
 
         this.exercise_view.suppress_button_feedback();
 
-        // increment the response count
-        this.current_attempt_log.set("response_count", this.current_attempt_log.get("response_count") + 1);
-
-        this.current_attempt_log.add_response_log_event({
-            type: "answer",
-            answer: data.guess,
-            correct: data.correct
-        });
-
-        // update and save the exercise and attempt logs
-        this.update_and_save_log_models("answer_given", data);
+        return ExerciseWrapperBaseView.prototype.check_answer.call(this, data);
     },
 
     update_and_save_log_models: function(event_type, data) {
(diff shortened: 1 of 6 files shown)


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -436,8 +436,7 @@ window.ExerciseView = Backbone.View.extend({
 
 });
 
-
-window.ExercisePracticeView = Backbone.View.extend({
+window.ExerciseWrapperBaseView = BaseView.extend({
```
MCGallaspy: Can you please add comments explaining the intended use? This appears to be an abstract base class.

MCGallaspy: Comments explaining which methods are intended to be overwritten or what functionality needs to be implemented by views extending this would be good.

kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -673,103 +580,201 @@ window.ExercisePracticeView = Backbone.View.extend({
...
 
-        var self = this;
+    get_points_per_question: function() {
+        return this.attempt_collection.calculate_points_per_question(this.exercise_view.data_model.get("basepoints"));
+    },
 
-        if (this.user_data_loaded_deferred) {
 
-            this.user_data_loaded_deferred.then(function() {
+    display_message: function() {
```
MCGallaspy: I would like to see comments giving some context to these functions. For example, what kind of message does this display? How is it meant to be used by the developer? It's not immediately obvious how to follow the control flow to this function, which is reached by passing control from `ExerciseWrapperBaseView.initialize` -> `ExercisePracticeView.load_user_data` -> `ExercisePracticeView.user_data_loaded` -> `ExerciseWrapperBaseView.display_message`. It's also not used by any of the other descendants of `ExerciseWrapperBaseView`, which suggests it should be defined in `ExercisePracticeView`.

rtibbles: Too true!

kalite/distributed/static/js/distributed/exercises/views.js
rtibbles: Comments added.

MCGallaspy:
Looking better. How should I test this? Just try out some exercises?

rtibbles:
Yeah, probably the best way. I would say 'tests pass, of course it works' but I don't think that's true.

rtibbles:
Mostly just make sure that stuff is updating properly in terms of the attempt logs, points, etc.

MCGallaspy:
I've let this age long enough, let me give it a test ride real quick, then I'll merge it.

rtibbles:
Like a fine wine, it has only gotten better.

MCGallaspy:
I keep getting "Content was not found" errors on Khan exercises. Could be an issue with my own setup? I'm slowly downloading the assessment items as well.

rtibbles:
After you download assessment items, you'll need to restart the server is the main thing. Otherwise content won't be there.

MCGallaspy:
Seems like attempt logs are not updating correctly. I created a new account, and already it's suggesting that I have an incorrect answer:
ss

Subsequent correct or incorrect answers don't seem to be accurately reflected here. Not sure if this is an issue in `develop` or not, let me check on that branch.

rtibbles:
This is probably a result of the refactor. Will take a look tomorrow.

MCGallaspy:
On develop, I get the following text at the top of an exercise (`/learn/khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/`):

> no such column: main_attemptlog.response_count
×Traceback (most recent call last): File "python-packages/tastypie/resources.py", line 195, in wrapper response = callback(request, *args, **kwargs) File "python-packages/tastypie/resources.py", line 426, in dispatch_list return self.dispatch('list', request, **kwargs) File "python-packages/tastypie/resources.py", line 458, in dispatch response = method(request, **kwargs) File "python-packages/tastypie/resources.py", line 1275, in get_list for obj in to_be_serialized[self._meta.collection_name]: File "python-packages/django/db/models/query.py", line 123, in _result_iter self._fill_cache() File "python-packages/django/db/models/query.py", line 927, in _fill_cache self._result_cache.append(next(self._iter)) File "python-packages/django/db/models/query.py", line 301, in iterator for row in compiler.results_iter(): File "python-packages/django/db/models/sql/compiler.py", line 775, in results_iter for rows in self.execute_sql(MULTI): File "python-packages/django/db/models/sql/compiler.py", line 840, in execute_sql cursor.execute(sql, params) File "python-packages/django/db/backends/util.py", line 41, in execute return self.cursor.execute(sql, params) File "python-packages/django/db/backends/sqlite3/base.py", line 366, in execute six.reraise(utils.DatabaseError, utils.DatabaseError(*tuple(e.args)), sys.exc_info()[2]) File "python-packages/django/db/backends/sqlite3/base.py", line 362, in execute return Database.Cursor.execute(self, query, params) DatabaseError: no such column: main_attemptlog.response_count

rtibbles:
Radina has reported this issue also. I believe I have had it appear for me
as well, but I can't recall how I resolved it.

radinamatic:
Yep, tried syncdb --migrate, no luck.

MCGallaspy:
Verified that this is not an issue on `develop`. Resolved the apparently unrelated error by creating a new database.

rtibbles:
Issue with the additional cross appearing straight away should now be resolved.

MCGallaspy:
Hmmm, I've been trying to determine why there's a `TransactionManagementError: This code isn't under transaction management`. Otherwise this seems to run fine, but if it's not critical I think it's best to hold off merging for now.

rtibbles:
Not critical. I have noticed this error in other travis output too. I think
it happens when behave quits while an API request is still processing,
possibly?

MCGallaspy:
Hmm, seems like your intuition was right: https://code.djangoproject.com/ticket/20382

MCGallaspy:
I'm guessing some very long-running API response is being generated. Merging this, opening a separate issue.

MCGallaspy:
I mean I'll merge it as soon as I check it out manually.

MCGallaspy:
Looks ok.

PR #3509, opened 2015-04-10
delete power-mode images.

aronasorman:
They seem to be used for the old knowledge map, which doesn't exist for 0.13.


rtibbles:
Fair enough.

66eli77:
Merge with caution!

PR #3510, opened 2015-04-10
Refactors Search Autocomplete into Backbone

### LINKED ISSUE #2640: Refactor search autocomplete to use Backbone
At the moment we are overriding the majority of the functionality for autocomplete-ui, and having to do extra things to stop it from doing its default actions. We may as well remake it in Backbone to give us more flexibility in styling and integration with the rest of the site.

rtibbles:
Fixes #2640

Summary of changes:
* Deletes old search_autocomplete.js
* Instantiates a backbone view that handles the search bar
* Creates handlebars templates for the search bar and auto complete items.


aronasorman:
Travis errors look significant.

aronasorman:
Also, tests please.

rtibbles:
Woops, accidentally put a handlebars template tag inside a django-compressor block. I should probably put it in the docs that this breaks stuff too..

rtibbles:
Tests errors should be rectified. Added BDD tests for the autocomplete behaviour. Most recent test run seemed to show a non-deterministic failure, unfortunately :/

kalite/testing/behave_helpers.py
```diff
@@ -47,7 +47,7 @@
 
 
 # Maximum time to wait when trying to find elements
-MAX_WAIT_TIME = 1
+MAX_WAIT_TIME = 5
```
MCGallaspy: I'm hesitant to increase the wait time across the board. Let's try to determine which step is failing, and address that instead.

rtibbles: It's only a problem for Python 2.6, waiting for a new page to load just takes longer on Travis. The wait time is only a problem if it *needs* to wait that long, if the condition is satisfied sooner it won't wait that long.

MCGallaspy:
```shell
Assertion Failed: Assertion failed. context.browser.current_url: http://localhost:9000/learn/khan/math/early-math/cc-early-math-add-sub-topic/basic-addition-subtraction/
      build_url: http://localhost:9000/learn/khan/math/arithmetic/addition-subtraction/basic_addition/basic-addition/
```
Seems like the expected URL is simply wrong. I wonder if this isn't an unreasonable expectation, given that it depends on the topic tree, but what we're trying to test is the search autocomplete behavior. If the search space or search results change for any reason, the test will fail. Perhaps the steps
```gherkin
When I search for Basic Addition
Then I should see a list of options
When I click on the first option
Then I should navigate to Basic Addition
```
Can be rephrased as follows:
```gherkin
When I search for [[something nonspecific]]
Then I should see a list of options
When I click on the first option
Then I should navigate to a content item page
```

rtibbles:
Sure, we can do that.

MCGallaspy:
Error seems unrelated to the (otherwise passing) tests.

rtibbles:
Will rebase to latest develop.

aronasorman:
Tests pass. Merging.

PR #3514, opened 2015-04-12
Test and fix for regression error on SyncedModel.full_clean(imported=True)

jamalex:
This started showing up on the central server tests, even though it was passing here (since we never do `full_clean(imported=True)` in the distributed server tests, only in the ecosystem tests on the central server). Adding the fix here, along with a test so that any regressions will be caught here directly.

### DIFF
--- a/python-packages/securesync/engine/models.py
+++ b/python-packages/securesync/engine/models.py
@@ -230,7 +230,7 @@ class SyncedModel(ExtendedModel):
         """
         exclude = exclude or []
         if imported:
-            exclude = list(set(exclude + self._import_excluded_validation_fields))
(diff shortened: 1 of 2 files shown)


aronasorman:
Heh. Travis is trying to pull the wrong PR (#3515). WHY

aronasorman:
Is this the one causing the registration errors?

MCGallaspy:
Looks groovy, let me just try to figure out what's going on with Travis.

MCGallaspy:
Test runs fine locally.

jamalex:
> Is this the one causing the registration errors?

This was the one causing the syncing errors (and was caused by the fixes *for* the reg errors).

PR #3521, opened 2015-04-14
Refactor CoachHasLogoutLinkTest into behave testing framework

### LINKED ISSUE #3502: Fix a broken test
Namely, the coach logout link test that keeps failing on `develop`.

MCGallaspy:
Fixes #3502. Starts moving old "browser tests" into the behave framework. We need to make a decision about how to phase out old browser tests. It's not faster to run these tests under behave than it is to run them under the old unittest framework, but hopefully it's nicer and more maintainable. The question then is which of the old browser tests do we port over, and which do we nix? Perhaps some of the code being tested could be refactored and tested as units instead, which would be considerably faster.

Perhaps calling this a regression test is also a misnomer. Do we want to write a suite of tests to ensure there's no regression for basic UI features? Kind of like retroactive BDD. The drawback is that it would be slower since it's adding more integration tests, but perhaps if it's coupled with a judicious removal/refactoring of existing browser tests the net effect could at least be neutral.


benjaoming:
For me, I don't care if it's BDD :) Where I see the big benefit is getting rid of Selenium for as many things as possible... if tests are about calling a view and seeing if some element is present in the DOM, I don't think that Selenium is the answer.

A counter-argument would be: "But we are using handlebars and JS"... but those tools will not only slow down our tests and make things less test-able, they will also slow down the user experience.

In this case, though, we are talking about detecting an element in the DOM that will be present regardless of any JS stuff.

jamalex:
> "But we are using handlebars and JS"... but those tools will not only slow down our tests and make things less test-able, **they will also slow down the user experience**

Wait, what? We sped up the user experience like 1000x by switching from pageloads for topic tree nav to the sidebar JS / Handlebars stuff... Or did you mean something else? :confused:

MCGallaspy:
> In this case, though, we are talking about detecting an element in the DOM that will be present regardless of any JS stuff.

So what's the best way to test this? I agree that Selenium is overkill for something like this.

benjaoming:
(rant warning)

> Wait, what? We sped up the user experience like 1000x by switching from pageloads for topic tree nav to the sidebar JS / Handlebars stuff... Or did you mean something else? :confused:

Yeah, somewhat and especially for the huge sidebar.. but that's still very dependent on user behaviour...

This was just a reflection on the menu items in the test case and why it failed and seems like a non-trivial test case.

In this case with the logout menu item (and others with the `.logged-in-only` CSS class), it's a little odd... a mix of both worlds..... most items in the menu are shown or hidden using template tag conditions `{% if request.user.is_authenticated %}` like conditions... but the drop down menu that's shown is always present in the DOM, it just happens to only show when its parent menu item is clicked... which is only present in the DOM because of a javascript injection... and then the text nodes are populated with JS, such as the username. This means that every time a page is loaded, we need JS to populate something that would otherwise have been a pretty static element right from the beginning of Django's request object generation....

More generally, I mean is that when you switch dynamic content generation from a database+python driven mode to something JS / Handlebars, then it will be slower because essentially, you move data into the browser and then populate the DOM there.

"Let's have everything as JSON data and populate all DOM nodes with JS" is not really a good idea performance-wise. "Let's generate all content (i.e. populate all the DOM) by default on the server-side and use JSON data to update elements that can change from Ajax callbacks" would make more sense IMO.

The extreme end of this spectre would be leaving a whole database to the browser and just building the app from there. In many cases there's a reason to move content generation to the browser layer because of the user experience creating a case where it's desirable to update specific elements rather than reloading the whole page.. what I'm arguing is that it's just really not good for testability, nor overall performance...

So here we are with the BDD that should say something like: "Check if there is a logout button for a logged in user"... this just became somewhat harder because we have to go "Check if there is a visible logout element for a logged in user and that it's in a hidden dropdown menu that has an element that makes it possible to toggle the visiblity of the dropdown menu by clicking it".

Also, the reverse test becomes very unintuitive... if you want to test if non-logged in users aren't displayed a logout button, you'd have to check that it's there but is never shown.

Am I making any sense or should I go home? :)

benjaoming:
> So what's the best way to test this? I agree that Selenium is overkill for something like this.

My suggestion: Make the top menu bar completely server side generated, i.e. all the elements of the menu are generated in the django template.

A test that avoids using Selenium can then use Django's much faster test client to create a logged in session and call the main page view to see that it has an element that points to the correct logout URL... i.e. reverse('logout')

That would make a pretty fast and robust test.

The change could happen in the `behave_helpers` rather than the test case..

rtibbles:
Just to throw another spanner in the works here - an additional issue is that the login is now API driven. And precisely because of the hybrid nature of the parameterization of the navbar that @benjaoming mentioned, this doesn't actually work properly, because the client side JS cannot properly update the navbar based on user login and logout events due to the fact that some DOM elements just aren't there.

Further, this is not an issue of Django rendering vs Javascript/Handlebars rendering. It is more a matter of our previous over use of integration tests when we should have been using more unit tests. Although the Javascript unit tests are currently strangled due to the way we have implemented Handlebars, exactly the kind of test that @benjaoming suggests we could do in Django without firing up the live server and Selenium, we can also do very quickly with proper Javascript unit testing by checking the output of Backbone View render functions.

jamalex:
(Not addressing the testing stuff here -- just engaging with the performance rant :grin:)

I think that intelligently shifting many things (but not everything) to the frontend can make things dramatically faster and smoother for the user. Most of the current codebase does not yet reflect this, as much of the code is legacy (read: crappy) and needs to be brought in line with our new approaches (Backbone, Handlebars, solid API, etc).

> Yeah, somewhat and especially for the huge sidebar.. but that's still very dependent on user behaviour...

Curious about what you mean by "dependent on user behaviour". Particularly on a slow server -- like the Raspberry Pi, Sandisk, or python-for-android -- Django pageloads take a few seconds, whereas navigation in the frontend-rendered sidebar takes 10's of milliseconds.

> when you switch dynamic content generation from a database+python driven mode to something JS / Handlebars, then it will be slower because essentially, you move data into the browser and then populate the DOM there

That definitely doesn't necessarily follow (if things are done right). The data is being sent in both cases (either in the HTML or in the JSON). The DOM is being constructed and rendered in either case. JS is [faster than CPython](http://www.cdotson.com/2014/08/nodejs-vs-python-vs-pypy-a-simple-performance-comparison/). And it's especially true when any of these conditions are met:
- The amount of data is not huge (the API should be providing the appropriate pagination/processing/filtering to ensure this). Anything like aggregate stats should be calculated server-side.
- There is one server (often a very slow one) with many clients. Thus, the number of CPUs available to render templates scales in line with the number of users, putting less load on the server.
- A lot of interaction happens on the client-side without needing to fetch additional data from the server. Quickly toggling visibility or state of elements in the DOM is much (++) more efficient than reloading the page, and provides more of an interactive "app" feel than a static browsing experience.

Also, pre-compiled Handlebars templates are _blazing_ fast. Really. We're not yet pre-compiling them (as this is waiting on the node.js build process @aronasorman and @rtibbles are working on). The templates can be compiled down to raw, optimized JS functions that take the context as input and return the HTML. I challenge you to a Django template vs precompiled Handlebars template speed race! :racehorse: 

> "Let's generate all content (i.e. populate all the DOM) by default on the server-side and use JSON data to update elements that can change from Ajax callbacks" would make more sense IMO

Definitely advantages to that in principle, but in practice (unless you're running Node.js and can share code, and even then) it doesn't lead to good code (reduplication, strange cross-language coupling, etc).

-- 

@rtibbles and I have built a fully client-side app together (to the extreme), and have seen the pain points in that, so we're not blind fanatics. But we see a bright future with a judicious and structured approach to getting the "best of both worlds". Don't worry, we'll show you the beauty waiting around the corner! ;)

Ok, I'm *literally* going to show myself home now...

jamalex:
> I challenge you to a Django template vs precompiled Handlebars template speed race! :racehorse:

:horse_racing: Ok, I couldn't help it; I raced myself. The results: https://gist.github.com/jamalex/a0ce3b213b9362c8ba46

EdDixon:
DEBUG=True
MIND=BLOWN

Results=Amazed!

benjaoming:
@jamalex 

Nice benchmark :) That looks hard to bargain with!!!

But I guess it would not be the same if it wasn't for Handlebars' precompiling, which can only happen server side (otherwise loading a template 100 times on a client would also mean compiling it 100 times).

I also agree on the basic condition that you've mentioned: It's worth moving computations into client side if the server in terms of hardware resources is very far from of the accumulated resources of the clients, moreover that it won't be able to serve its clients.

But servers can do so much. For instance, they can serve a cache for all the clients.

Btw... Jinja is in Django 1.8, it will speed up template rendering vastly... but doubtfully as much as to compete with your benchmark.

Isn't the current way that we're using Handlebars *not* precompiling anything?

See the chapter here: https://bitbucket.org/chrisv/django-handlebars/

rtibbles:
We are not precompiling Handlebars templates currently, but that is going
to have to change, as our Javascript unit tests are useless without this.
Also because not precompiling was always a fudge.

benjaoming:
This one seems to do it...

https://bitbucket.org/chrisv/django-handlebars/

rtibbles:
Yes, I noticed, and as a quick fix, that would probably work, as those
files would then be available for django-compressor to pick up. Just a
question of whether it is easier to push ahead with the node driven build
server or to implement this for now.

jamalex:
> But I guess it would not be the same if it wasn't for Handlebars' precompiling, which can only happen server side (otherwise loading a template 100 times on a client would also mean compiling it 100 times).

In the benchmark, I wasn't using the "pre-compile" feature of Handlebars, just the runtime "compile" method. However, I was running "compile" method only once, and then using it many times. This is closer to how we're currently using it, except that we're likely using each template fewer than 100K times within a single page load. :)

The "pre-compile" feature actually pre-processes the template server-side and produces some pure JS that can be served to the client and used to directly render templates without a compile stage. I tested this as well, and the benchmark speeds were identical to the HB benchmark speeds here. So, yes, it can be used effectively/efficiently on the client-side (as per the design).

> But servers can do so much. For instance, they can serve a cache for all the clients.

Yep! Hence the "best of both worlds" approach, and judiciously choosing what should happen where. We did use page-level caching extensively within KA Lite, on the topic and content pages, and it helped somewhat (but still orders of magnitude slower than the sidebar nav). Also, it used quite a bit of disk space. And there were some issues with the invalidation logic. And we couldn't put user-specific data in the page (since we were doing full-page caching, that is), so we would load it separately and inject it into the standard pages (similar to the server-side-render-and-client-side-update approach you described above).

> Btw... Jinja is in Django 1.8, it will speed up template rendering vastly... but doubtfully as much as to compete with your benchmark.

Awesome! Yes, it's definitely faster than Django's current template engine.

> Just a question of whether it is easier to push ahead with the node driven build server or to implement this for now

Note that we'll need node.js in either case, since both LESS compiling and HB precompiling require the node.js modules. The django-handlebars and django-less apps then just call those binaries. So it's probably more a question of what build system controller will be the most elegant, concise, and flexible.

MCGallaspy:
Curiously, I've discovered the logout link doesn't work.

MCGallaspy:
Nope, I cleverly broke the logout by changing an id from "nav_logout" to "nav-logout".

PR #3523, opened 2015-04-15
Added ARIA landmarks

radinamatic:
Diving into WCAG from the outward - added main ARIA landmark roles.
http://www.w3.org/TR/WCAG20-TECHS/ARIA11

ka_aria_landmarks


radinamatic:
Just realized that a change from another branch that I'm still working on "sneaked" into this PR. Feel free to cherry-pick-it-out if necessary.

radinamatic:
BTW, would it be possible to have a proper "a11y" label for issues and PRs?  
Should I post an issue for that? :P

I'm unsure if I could actually use the label myself, but it should definitely be available.

radinamatic:
@aronasorman, thanks for the label! 
I don't know if the color was intentional, but I like it... :)

rtibbles:
Och, this is going to get messy as @jtamiace and I are backbonifying the homepage.

Be aware of this as you build, @jtamiace!

rtibbles:
All good. Merging.

radinamatic:
@rtibbles In case I have to work again on homepage, anything I should keep in mind regarding the backbonifying?

rtibbles:
No - you just keep on going with what you're doing. Just shout at us if we happen to revert any of your changes!

radinamatic:
Gottit! :)

PR #3519, opened 2015-04-14
A new setup.py installer using setuptools - Fix for #3396

benjaoming:
Running KA Lite can now happen in 2 different ways... for a discussion of the approach, see #3396

# 1. Directly from source folder

Navigating to a source folder and running `bin/kalite` will invoke KA Lite and its data files directly from the source tree. Furthermore, all media and database files are maintained in the source tree.

In other words: **exactly like before** !!

# 2. As an installed python package

An example with virtualenvwrapper:

```
mkvirtualenv --system-site-packages my_kalite_env
workon my_kalite_env
cd /path/to/my/kalite/git/repo
```

Now you can install KA Lite in 3 different ways as a developer:

## 2.1. Dynamically link the source code to your site-packages:

    pip install -e .

## 2.2. Install directly as a source package

    python setup.py install

## 2.3. Simulate the full packaging+installation process

This method is really good for testing that data files are correctly listed in MANIFEST.in and setup.py and thus packaged and distributed...

    python setup.py sdist
    pip install dist/kalite-X.X.X.tar.gz

# Commit contents

This is only possible by re-arranging some data files and refactoring the way that settings were loaded directly from filesystem paths.

Squashed commit of the following:

commit 6ee7eca9e539cd33d552f4502ad95d36f5c8fea2
Author: Benjamin Bach <benjaoming@gmail.com>
Date:   Mon Apr 13 22:13:55 2015 +0200

    assessment items: print out downloaded size before starting

commit c1e47be779a1f3cec5737d9e9c33dd0322393604
Author: Benjamin Bach <benjaoming@gmail.com>
Date:   Mon Apr 13 20:07:03 2015 +0200

    add *handlebars files

commit e274e2a6ecfaabefd6d0538339b50fcb42cb895e
Merge: 7bee016 c15b07d
Author: Benjamin Bach <benjaoming@gmail.com>
Date:   Mon Apr 13 18:57:33 2015 +0200

    Merge branch 'develop' into setuptools-installer

commit 7bee016ccc1537e7a4485a2705afed10803cc044
Author: Benjamin Bach <benjaoming@gmail.com>
Date:   Wed Apr 8 01:07:01 2015 +0200

    remove a bunch of PROJECT_PATH and replace with USER_DATA_ROOT

commit 7ff6d6ec86b160b8c63437368f0d993dffb7f2dd
Author: Benjamin Bach <benjaoming@gmail.com>
Date:   Wed Apr 8 01:06:22 2015 +0200

[...]

PR #3526, opened 2015-04-16
Remove the test for the `securesync` module.

### LINKED ISSUE #3525: Basetests are failing for `kalite`.
**Branch** Develop
**Target** 0.14.x
**Current Behavior**
Base tests are failing because we also test `securesync` which now has a different version.

**Expected Behavior**
Base tests must pass.

**To reproduce**
1. Go inside your vagrant box at `/home/vagrant/ka-lite/`
1. Make sure ka-lite is not running.
1. Run `python kalite/basetests/tests.py`.
1. Notie the following output:

    ...
    ...importing securesync...need version 1.0, found 0.14.dev... FAIL!
    ...
    FAIL: test_required_packages_and_versions (__main__.PackagesTests)
      File "kalite/basetests/tests.py", line 279, in test_required_packages_and_versions
        self._fail("\n...Result: %s required Python package/s failed import..." % fail_count)
      File "kalite/basetests/tests.py", line 83, in _fail
        self.fail(msg)
    AssertionError:
    ...Result: 1 required Python package/s failed import...

    Ran 13 tests in 3.727s

    FAILED (failures=1)
    ...

**Possible Solution**
As per @aronasorman, we can remove the check for the `securesync` package because it's already part of the `ka-lite` source.

cpauya:
Hi @aronasorman - this fixes #3525 - Basetests are failing for `kalite`.

I've removed the check for `securesync` module version as discussed.

See attached screenshot for result of running `python kalite/basetests/tests.py`.
screenshot 2015-04-16 09 18 17

### DIFF
--- a/kalite/basetests/tests.py
+++ b/kalite/basetests/tests.py
@@ -226,7 +226,6 @@ class PackagesTests(DependenciesTests):
         "pytz": "2012d",
         "requests": "0.14.2",
         "rsa": "3.1.1",
-        "securesync": "1.0",
         "smmap": "0.8.2",
         "south": "0.8.4",
         "tastypie": "0.11.0",


benjaoming:
This was also fixed in #3519

https://github.com/learningequality/ka-lite/commit/dddb48f54301e96e2c1dc56ca5cc98d4acdcad8f

benjaoming:
@aronasorman Okay but the other fix actually fixes it without disabling the test... maybe you could be a bit more consequential and remove ALL the version tests instead then? :)

aronasorman:
The main reason for these version tests is that some systems might have bundled certain python packages which act differently across versions. The package that comes to mind is `psutil` -- certain OS X and Ubuntu versions come with the 1.x version, while the more recent releases of said OSes come with 2.x (and downloading from PyPI automatically gives you the 2.x version too). There were some subtle breaking changes between the two versions, and the version tests were created in order to ensure we're using the version we are expecting.

Also, I expect to remove these version tests soon once we've eliminated python-packages and use `pip` for all of our dependencies, but until then it'd be best to keep this for the above reasons.

PR #3527, opened 2015-04-16
Set the html language attribute required by screen readers, fixes #3350  

### LINKED ISSUE #3350: Accessibility: Document language missing, screen reader uses incorrect language
Branch: develop

Expected Behavior: When ka-lite language is set to a specific language (in my case Spanish/Castillano), screen reader programs should start reading it in that language.

__WCAG Guidelines: 3.1.1 Language of Page (**Level A**)__

Current Behavior: Tested with JAWS & NVDA, both start in English, because of the missing document language attribute (<html lang="es">).

Steps to reproduce:

1. Download & install language pack
2. Change the default KA lite language
3. Start the screen reader

doc-lang-missing

radinamatic:
One small change for KA Lite code, one giant leap for screen reader users.

Didn't like the language codes inside the messages for users downloading and deleting language packs, so I added the change to show the full language names instead.


rtibbles:
Looks good to me!

PR #3528, opened 2015-04-16
0.13.x into develop

MCGallaspy:
Let's keep an eye on this.


aronasorman:
Merging this now, as this has the new assessment item URLs, plus the assessment item extraction fix for Windows.

PR #3531, opened 2015-04-17
Remove deprecated scripts

MCGallaspy:
They're just a source of confusion, now.


rtibbles:
Do any of these get used by the windows installer at all?

rtibbles:
@ruimalheiro ^

ruimalheiro:
@rtibbles, start/stop and add/remove_systemstart_task are still related with with both the installer and the gui. They will not be necessary once the binary starts being used.

aronasorman:
Closing this, as it's too late for 0.13.x now. Can you merge this against develop instead?

PR #3533, opened 2015-04-17
Added ContentRecommenderResource and urls

sialan:
First pass, work-in-progress.


rtibbles:
Tests pass. Does nothing worse than a no-op for now. Merging.

PR #3534, opened 2015-04-17
Adds BDD tests for Modal Login

rtibbles:
Summary of changes:
* Test that facilities are properly hidden or shown depending whether > 1 or not
* Test error highlighting is done properly for incorrect username and password input
* Test that logging in actually results in (currently) a page reload.

### DIFF
--- a/kalite/distributed/features/steps/login.py
+++ b/kalite/distributed/features/steps/login.py
@@ -92,6 +92,12 @@ def impl(context):
 def impl(context):
     assert check_single_popover(context, "password")
 
+@then('the page should reload')
+def impl(context):
+    wait_elem = context.browser.find_element_by_tag_name("body")
+    assert WebDriverWait(context.browser, TIMEOUT).until(
+        EC.staleness_of(wait_elem)
+    )
 
 def fill_field(context, text, field_id):
     field = find_id_with_wait(context, field_id)
--- a/kalite/distributed/features/steps/login.py
+++ b/kalite/distributed/features/steps/login.py
@@ -7,7 +7,7 @@ from kalite.testing.behave_helpers import *
 
 from kalite.facility.models import Facility
 
-from kalite.testing.mixins.facility_mixins import CreateFacilityMixin
+from kalite.testing.mixins.facility_mixins import CreateFacilityMixin, CreateStudentMixin
 
 TIMEOUT = 3
 
@@ -25,7 +25,7 @@ def step_impl(context):
         context.facility = CreateFacilityMixin.create_facility(name=str(name))
         name += 1
 
-@given("that I am on the homepage")
+@given("I am on the homepage")
 def step_impl(context):
     go_to_homepage(context)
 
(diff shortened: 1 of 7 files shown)


kalite/distributed/features/steps/login.py
```diff
@@ -0,0 +1,118 @@
...
+
+def fill_username(context, text):
+    fill_field(context, text, "id_username")
+
+def fill_password(context, text):
+    fill_field(context, text, "id_password")
+
+def check_highlight(context, item):
+    highlight = find_css_class_with_wait(context, "has-error")
+    return "id_{item}-container".format(item=item) == highlight.get_attribute("id")
```
MCGallaspy: Regarding the failing test, could it be that there is another unintended element with the "has-error" class being found? What is the value of `highlight.get_attribute("id")`?

rtibbles: Yes, this is possible. In which case it is a good test failure, as it is flagging that the username is also being highlighted, I would hope. Will debug.

rtibbles: This is not happening locally, can only assume this is a timing issue.

MCGallaspy: Okay, I'll look into it today.

kalite/distributed/features/steps/login.py
```diff
@@ -0,0 +1,118 @@
...
+    fill_username(context, "notarealusernameIswear")
+
+@when('I enter my username correctly')
+def impl(context):
+    fill_username(context, context.user.username)
+
+@when('I click the login button')
+def impl(context):
+    login_button = find_css_class_with_wait(context, "login-btn")
+    login_button.click()
```
MCGallaspy: Okay, upon closer inspection I think one possibility is that the step which follows this one (`@then('the page should reload')`) is subject to a race condition if the new page loads before Selenium grabs an element from the current page.

kalite/distributed/features/steps/login.py
```diff
@@ -0,0 +1,118 @@
...
+
+def fill_password(context, text):
+    fill_field(context, text, "id_password")
+
+def check_highlight(context, item):
+    highlight = find_css_class_with_wait(context, "has-error")
+    return "id_{item}-container".format(item=item) == highlight.get_attribute("id")
+
+def check_single_popover(context, item):
+    popover = find_xpath_with_wait(context, "//*[@id='id_{item}-container']/div/div".format(item=item))
```
MCGallaspy: Any way to use a more robust method of locating the element besides xpath? I think it would be preferable to add an id or class name to the element.

kalite/testing/behave_helpers.py
```diff
@@ -70,6 +70,18 @@ def click_and_wait_for_page_load(context, elem, wait_time=MAX_PAGE_LOAD_TIME):
...
 
+def click_and_wait_for_id_to_appear(context, elem_click, elem_wait, wait_time=MAX_WAIT_TIME):
+    """ Click an element and then wait for another element to appear.
+    context: a behave context
+    elem_click: a WebElement to click.
+    elem_wait: an ID selector to wait for.
+    wait_time: Optional. Has a default value.
+    """
+    elem_click.click()
+    id_shown_with_wait(context, elem_wait, wait_time=wait_time)
```
MCGallaspy: What if the element with the given id already exists on the page? Should we try to grab a reference and assert its staleness (if it exists) after the click, in addition to asserting that a *good* reference exists, or shall we say *caveat pragrammor*?

kalite/distributed/features/steps/login.py
```diff
@@ -0,0 +1,118 @@
...
+def impl(context):
+    assert check_highlight(context, "username")
+
+@then('a tooltip should appear on the password box only')
+def impl(context):
+    assert check_single_popover(context, "password")
+
+@then('the page should reload')
+def impl(context):
+    wait_elem = context.browser.find_element_by_tag_name("body")
```
MCGallaspy: By the way, this combined with a preceding step essentially reproduce the `click_and_wait_for_page_load` method I wrote. Is there a way to rephrase the steps to use that method (for DRYness)?

kalite/distributed/features/steps/login.py
rtibbles: Yep. Added an id to the popover.

kalite/testing/behave_helpers.py
rtibbles: I think I will say *caveat programmer*.

MCGallaspy:
Error was with npm.

rtibbles:
Comments addressed, hopefully!

rtibbles:
Finally...

PR #3535, opened 2015-04-17
Con alg team

almgong:
This PR involves creating a new file content_recommendation.py in the topic tools directory. The purpose of this file is to contain all of the logic for generating and retrieving the different types of recommendations that will be requested by the client, via the new API call that will be/has been made by Alan Si. 

This PR will not affect the user interface, nor is it targeted to fix any bugs.

Generous commenting in the code for easy understanding :).


rtibbles:
Hrm. I assume it must be an artifact of the multiple merge commits here - but git seems to reckon that 60 or so files have been deleted in this PR.

Would you feel confident creating a fresh branch from latest develop and cherry picking just your and @mwong21's commits to that branch for a clean PR?

almgong:
Yep! After the recent refactoring, all the new code we wrote is in its separate file so it shouldn't be too hard.
 
Will do tomorrow and make a new PR.

rtibbles:
Closing for now, this has been reopened elsewhere.

PR #3537, opened 2015-04-21
Smaller sidebar tab on small viewports

Antrikshy:
This is a fix for #3315.

screen shot 2015-04-21 at 1 23 46 pm

This is what the tab looks like now (on narrow viewports). When the navbar dropdown is toggled, it quickly fades in and out (115ms, just like almost every other animation in KA Lite). This way, there is no overlap with anything.


MCGallaspy:
Slick, but starting with a small viewport, clicking the hamburger (causing the the tab to fade) and then expanding the page causes the tab to remain hidden. Instead of fading the tab, can you reposition it? @jtamiace @rtibbles wanna weigh in?

Antrikshy:
Oh. Probably affects few users, but I think I can fix that too.

Antrikshy:
Tab now appears when the navbar is uncollapsed and the screen is resized. But it does not disappear when the screen is collapsed again, unless the navbar is collapsed and uncollapsed once. I can't find a way to check if the navbar is collapsed or not. Should not be a huge problem. It does not break anything major and I don't think too many people will change their screen sizes so much and be affected by this anymore.

MCGallaspy:
Let me make sure I have the intended behavior straight. Is this right?

                    | small screen     |   large screen
--------------------------- | -------------------- | ---------------------
hamburger collapsed | small tab visible | large tab visible
hamburger expanded | no tab visible    | large tab visible

I think if you want to ensure this behavior, you can look at two conditions.
1. Window resize: above or below a certain threshold, switch between showing the large tab or the small tab
2. Hamburger toggle: Did someone click the hamburger? Toggle visibility if the window is below the threshold. Otherwise ensure it's visible. You can keep track of the visibility state yourself.

Perhaps you can wrap this logic up into one debounced function (`render_tab` or something) and set it as the callback for both window resize and hamburger toggle events. It's a small feature, but if it's easy to do right we might as well do it right, right? :)

Antrikshy:
Done. All bases should be covered now. I track when the navbar is collapsed and uncollapsed in `SidebarView`, showing and hiding tab whenever necessary on screen size change. Make sure there are no edge cases breaking. Pretty sure it's solid.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -65,9 +66,30 @@ window.SidebarView = BaseView.extend({
             if (Math.floor(newWindowWidth/100) != Math.floor(windowWidth/100)) {
                 self.resize_sidebar();
                 windowWidth = $(window).width();
+
+                if ($(window).width() > 768) {
+                    self.show_sidebar_tab();
+                }
+
+                if ($(window).width() < 768) {
```
MCGallaspy: What about when `$(window).width()` is exactly 768?

MCGallaspy:
On chrome 42.0.2311.135, by making the window small, expanding the navbar by clicking the hamburger and gradually increasing the window size, I can achieve a state (>768px window width) where the sidebar tab isn't visible when it apparently should be. It's also possible to decrease the window size to <768px window width and have the sidebar tab visible (while the menu is expanded).

Here it is at 796px without the sidebar tab:
ss

And at 721px with the sidebar tab:
ss

MCGallaspy:
I think the solution is to have one callback for both the window resize event and the navbar toggle events that determines whether the sidebar tab should be visible (and fades it in or out as appropriate), instead of different callbacks for all those events. And again, make sure it's debounced so that it is only called once when the events stop rolling in.

Antrikshy:
Hmm. I'll look into this. I am reusing the window size change listener from the responsive sidebar thing.

Antrikshy:
I see what the issue was. I was reusing a media query from earlier, which is why it was not resizing at 768px (where Bootstrap switches the navbar). Another mistake on my part was nesting some show/hide function calls inside a condition that only triggered in screen size changes of 100px increments (this was implemented for sidebar responsiveness, because recalculating the sidebar every few pixels was expensive). Much smoother now. See if this works.

Antrikshy:
I also added a tiny bit of CSS in this same PR to fix #3558.

screen shot 2015-05-11 at 11 39 16 am

rtibbles:
@Antrikshy I think this is the wrong fix for the login modal issue. It was caused by some changes to the Handlebars template itself, the CSS should be fine. @radinamatic should be able to give input on this.

MCGallaspy:
Right, probably better to let that be a separate PR.

Antrikshy:
Ooh. I'll remove that from this PR then. I wasn't following the modal issue.

MCGallaspy:
I'm okay with this. Waiting for tests to pass.

radinamatic:
#3647 fixes the margins for the login modal.

PR #3540, opened 2015-04-22
Successfully adding a coach takes you to the "Add a new coach" form

### LINKED ISSUE #2970: Successfully adding a coach takes you to the "Add a new coach" form
On `develop`, adding a new coach takes you back to the same "Add a new coach" form. Now this is up for debate, but shouldn't successfully submitting that form redirect one back to the associated facility page?

To reproduce, log in as an administrator, click on the Manage tab, choose a facility, click "Add a new coach" link at the bottom of the list of coaches, create a new coach, and then get redirected back to the same form.

amodia:
Hi @MCGallaspy this fixes #2970 

ka-lite-a

### DIFF
--- a/kalite/facility/views.py
+++ b/kalite/facility/views.py
@@ -157,7 +157,7 @@ def _facility_user(request, facility, title, is_teacher=False, new_user=False, u
             # New user created by admin
             elif request.is_admin or request.is_django_user:
                 messages.success(request, _("You successfully created user '%(username)s'") % {"username": form.instance.get_name()})
-                return HttpResponseRedirect(next)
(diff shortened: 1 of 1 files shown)


MCGallaspy:
@rtibbles @aronasorman re: out initial discussion on the issue, do we want to incorporate this? How desirable is it to streamline adding a lot of coaches in a row?

aronasorman:
I favor consistency for now. Let's change all of them if we do decide that this is how we're gonna add coaches in bulk.

MCGallaspy:
@aron so favoring consistency means you favor incorporating this change?

aronasorman:
Yes. I believe the issue before was that  groups and student forms redirect to the group or student list page once you create them. This makes the coaches form fall in line with that behavior.

Issue #3541, opened 2015-04-23
Silent MemoryErrors


aronasorman:
Branch: develop

Debugging https://github.com/learningequality/ka-lite/pull/3519 has led to strange failures at the very end of the KA Lite startup process, inside a Vagrant Virtualbox VM with 495 MB available (the default for trusty32). It turns out that at the very end of our startup sequence, `kalite`'s memory usage spikes up to around 508 MB, making the OS kill it, but somehow not leading python to print out a `MemoryError`. 

This leads me to believe that somewhere in our code is an `except` clause which catches all errors, including `MemoryError`s, and that's where the spike is happening. 

To replicate:
1.  Download the `Vagrantfile` version [here](https://github.com/aronasorman/ka-lite/blob/4ebd58ccd43a51376f8f25895483c4d803c6dad0/Vagrantfile). Notice that it doesn't increase the memory of the VM whatsoever. So it defaults to 495 MB.
2. `vagrant ssh` into your vagrant machine to run the following commands inside your VM.
3. Start the server and download `fr` and `xh`.
4. Run `kalite restart`.

**Result**
At the very end, it would show a successful startup message:
```
INFO:kalite:Preloading content data for language xh.
INFO:kalite:Preloading topic tree data for language xh.

0 static files copied, 2828 unmodified.
To access KA Lite from another connected computer, try the following address(es):
	http://10.0.2.15:8008/
	http://192.168.33.10:8008/
	http://127.0.0.1:8008/
```

But then doing a `curl http://127.0.0.1:8008` would return `connection refused`, indicating that our server did not start up.

benjaoming:
@aronasorman is this fixed now then? Or is the issue more generally about the silent failure of the cherrypy daemon? Because I was able to quite reliably make the cherrypy server die and the main process continue thinking that it had been started... and I'm not quite sure how to fix it in a nice way because this happens after `become_daemon()` in `django_cherrypy_wsgiserver.cherrypyserver`.

MCGallaspy:
For this, I'm going to specify any unqualified except clauses. That should make it easier to hunt down future errors.

MCGallaspy:
It turns out tracking down fishy `except` blocks is tricky. I also can't reproduce this error as described; any tips?

MCGallaspy:
I'm guessing part of my trouble in reproducing it was that we reduced memory use, too :)

benjaoming:
[...]

PR #3542, opened 2015-04-23
increase memory for vagrant virtualbox instances.

aronasorman:
Workaround for #3541. We increase the memory so we wouldn't hit the
silent MemoryError, and so people can still continue using Vagrant for development.


provision.sh
```diff
@@ -7,5 +7,6 @@ curl -sL https://deb.nodesource.com/setup_0.12 | bash -
 apt-get install -y nodejs
 pip install -r /home/vagrant/ka-lite/requirements_dev.txt
 cd /home/vagrant/ka-lite
+pip install -e .
```
aronasorman: We now also install the `kalite` executable globally! So within the VM you can call `kalite` anywhere and it will work.

PR #3549, opened 2015-04-23
Made fresh branch of current develop and added my changes to that

almgong:
Same as previous PR for content recommendation. Other PR involved some behind the scenes git magic which led to multiple files being deleted (reportedly). The changes here are all localized in a single file, and this file contains all of the logic for generating and passing back recommended content.


kalite/topic_tools/content_recommendation.py
```diff
@@ -0,0 +1,506 @@
...
+    result = {
+        "resume":{},        #not yet completed exercises, but have been started
+        "next_steps":{},    #things user is struggling on, next relevant items
+        "explore":{}        #less relevant exercises, for exploration
+    }
+
+    #preliminary user check
+    if not user:
+        #return 'Whoa there buddy, you did not specify a user!'
+        user = 'Yamira Jones (Facility: Wilson Elementary (#1829))'
```
rtibbles: The user should probably be either:
a) An id for a FacilityUser model
or b) A FacilityUser model

kalite/topic_tools/content_recommendation.py
```diff
@@ -0,0 +1,506 @@
...
+# of your content recommendation needs (i.e. in the API call).
+#
+# @param user: the unique user value of the current user (id/name I believe?)
+# @param current_subtopic: optional latest subtopic id that the user has worked on/completed. This
+#                          will act as a starting point for some of the algorithms (nearest neighbor, etc.)
+#
+# @return: Still deciding on this, but I assume that it will be a dictionary in the 
+#          form: { 'next_steps':[ ... ], 'explore':[ ... ], etc. } 
+###
+def get_recommendations(user=None, current_subtopic=None):
```
rtibbles: I wonder how much useful work this wrapper function is doing. It might be better to just define your lower level functions that you have below, and let this level be handled by @sialan's API.

The reason I suggest this, is that if we are going to allow filtering in the API to *only* retrieve 'resume' or 'next_steps' then it would be helpful to call those functions in a more granular way in the API.

almgong: I agree. I felt like there would also be cases where one would rather not have to completely grab all 3 types of recommendations every time. 

I will comment out the larger function and continue with the next level of functions (the ones called inside of the wrapper).

rtibbles: OK - could you make that change and push it to this branch, and I'll merge this so @sialan has access to the functions at least to use in his API.

almgong: Done ~

kalite/topic_tools/content_recommendation.py
almgong: OK - using FacilityUser model as input

sialan: Looks good! I'm just wrapping up some of the API work tonight. Can we set
up some time Thursday or Friday to reconnect? Anytime before 11am PT

Al

rtibbles:
@almgong Are you happy for this to be merged, or do you want me to hold off?

almgong:
I think some changes aren't pushed from my local yet, but I am fine with this being merged. Are there any considerations for merging it in this state that I am overlooking?

If not feel free to!

rtibbles:
Unless @sialan needs this merged desperately for his endpoint work, we are probably fine with just leaving this open as you polish stuff up.

kalite/topic_tools/content_recommendation.py
```diff
@@ -0,0 +1,613 @@
...
+
+###
+# Returns a dictionary with each subtopic and their related
+# topics.
+#
+###
+def generate_recommendation_data():
+
+    #hardcoded data, each subtopic is the key with its related subtopics and current courses as the values. Not currently in use.
+    data_hardcoded = {
```
rtibbles: We should probably put this into the data/khan folder as a json blob that we can load in for this.

Issue #3548, opened 2015-04-23
Error while upgrading language packs: No such file or directory: '...ka-lite\\kalite\\legacy\\data\\video_file_sizes.json'


radinamatic:
Branch: develop (after fetching today's upstream)

Expected Behavior: Updating language packs should not report any errors (it never did for me until now).

Current Behavior: Updating language packs starts without problems, seems to be done correctly (state is "Up to date" for the selected language), but at the end of the process there is an error message:

Error during update: Error: [Errno 2] No such file or directory: 'D:\\github15\\ka-lite\\kalite\\legacy\\data\\video_file_sizes.json'

Steps to reproduce:
1. Log in as the either administrator or coach.
2. Choose "Manage > Language".
3. Press the "Upgrade" button for the desired language.

![error_subtitle_upgrade](https://cloud.githubusercontent.com/assets/1457929/7306161/7a87a872-ea03-11e4-88c9-ab46521b9ae1.png)

benjaoming:
@radinamatic thanks for the great and perfectly detailed report.

We have cleaned up a lot of locations of data for the project and unfortunately our test system didn't catch all mistakes.

Will fix this now...

radinamatic:
Ohhh, the pleasures of contributing to bug-catching... :)

aronasorman:
Managed to replicate this error on develop. Merged Ben's PR, which now fixes this.

radinamatic:
Confirming a successful language pack upgrade with this patch! :)

![success_lang_pack_upgrade](https://cloud.githubusercontent.com/assets/1457929/7323813/acd76de0-eab1-11e4-996f-d23636c66f48.png)

benjaoming:
Thanks for checking back in with a confirmation @radinamatic :)

PR #3551, opened 2015-04-24
Fix UPDATES_DATA_PATH #3548

benjaoming:
#3548

### DIFF
--- a/kalite/legacy/updates_settings.py
+++ b/kalite/legacy/updates_settings.py
@@ -66,7 +66,15 @@ CONTENT_URL = getattr(local_settings, "CONTENT_URL", "/content/")
 BACKUP_VIDEO_SOURCE = getattr(local_settings, "BACKUP_VIDEO_SOURCE", None)
 BACKUP_THUMBNAIL_SOURCE = getattr(local_settings, "BACKUP_THUMBNAIL_SOURCE", None)
 
-UPDATES_DATA_PATH = os.path.join(os.path.dirname(__file__), "data")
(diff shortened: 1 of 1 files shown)


kalite/legacy/updates_settings.py
```diff
@@ -66,7 +66,15 @@
...
-UPDATES_DATA_PATH = os.path.join(os.path.dirname(__file__), "data")
+# This is the standard method... but doesn't work because we cannot load
+# kalite.i18n while loading settings because of its __init__.py
+# from pkgutil import get_data
+# I18N_DATA_PATH = get_data("kalite.i18n", "data")
+
+# Use resource_filename instead of get_data because it does not try to open
+# a file and does not complain that its a directory
+from pkg_resources import resource_filename
+UPDATES_DATA_PATH = resource_filename("kalite", "updates/data")
```
aronasorman: Interesting. When should we use `resource_filename` vs. `get_data` vs. using `os.path`?

aronasorman:
Better PR descriptions next time.

aronasorman:
Tested. Working. Merging.

kalite/legacy/updates_settings.py
benjaoming: Never use `os.path`, if you need data files, let pkgutils find them for you.

`get_data` would be conventional, but the problem is that it resolves module paths using import mechanisms, and because of circularity, we cannot resolve a lot of imports in the the `settings` package.. so the commented out patter might be the future.

Overall, though: We should always set paths in settings modules, and not leave them spread all over the codebase! :)

PR #3564, opened 2015-04-27
Allow subtitle_urls through the TastyPie firewall

### LINKED ISSUE #3561: No way to switch subtitles
Branch: 0.13.x
Browser: Firefox 37

Alright, so I have subtitles downloaded for both English and French:
image

But there's no way in the interface to activate the subtitles for the given videos:
image

rtibbles:
Fixes #3561.

Summary of changes:
* Counterintuitively, removes 'subtitle_urls' from the list of explicit Content model fields, so that it gets parsed as an 'extra_field' and then properly transmitted to the client side.

@aronasorman This is the fix you've been looking for. It is a very minor change, that I think would be best to include in 0.13.0.

(Note: subtitles will not work with DEBUG=True as we are currently copying them directly into the static folder which is not served).

### DIFF
--- a/kalite/main/api_resources.py
+++ b/kalite/main/api_resources.py
@@ -240,7 +240,6 @@ class Content:
             "kind",
             "content_urls",
             "selected_language",
-            "subtitle_urls",
             "available",
         ]
 


aronasorman:
Tests pass for 2.7 Let's merge this.

PR #3566, opened 2015-04-28
Hotfix: Factor in software version when checking whether a langpack is new.

jamalex:
Without this, it's not possible to download updated language packs after a software version upgrade (like right now), as the `language_pack_version` is [reset after every version upgrade](https://github.com/fle-internal/ka-lite-central/blob/92d5cdafa5117679c5260dbdb1307757a2a19f2a/centralserver/i18n/management/commands/update_language_packs.py#L755).

### DIFF
--- a/kalite/updates/static/js/updates/update_languages.js
+++ b/kalite/updates/static/js/updates/update_languages.js
@@ -2,6 +2,21 @@ var installable_languages = [];
 var installed_languages = [];
 var downloading = false;
 
+function version_comparison(v1, v2) {
+    // compare two version strings and return 1 if the first is higher than the second,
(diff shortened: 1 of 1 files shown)


jamalex:
Wanted to get this merged in before people started installing 0.13.0, and so that the demo server won't be borked

aronasorman:
Can you make the 0.13.1 tag too @jamalex?

jamalex:
Wilco.

jamalex:
@aronasorman Can you also check that what I did looks reasonable, or perhaps test it out further? My testing showed it worked in the cases I tried.

PR #3569, opened 2015-04-28
fix vagrant box name

jkirkell:
==> default: Adding box 'trusty64' (v0) for provider: virtualbox
    default: Downloading: trusty64
An error occurred while downloading the remote file. The error
message, if any, is reproduced below. Please fix this error and try
again.

Couldn't open file /Development/ka-lite/trusty64

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,7 +7,7 @@ VAGRANTFILE_API_VERSION = "2"
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Every Vagrant virtual environment requires a box to build off of.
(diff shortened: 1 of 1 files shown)


aronasorman:
Hi @jkirkell, thanks for making a PR! We actually have this fixed on develop: https://github.com/learningequality/ka-lite/blob/develop/Vagrantfile

Is there a specific purpose why you would like to have vagrant working in master? We currently use develop as our bleeding edge branch, and master should only be for release. Ideally you should be able to just launch KA Lite using `bin/kalite` without any further dependencies.

aronasorman:
Superseded by #3575.

PR #3576, opened 2015-04-28
Add edit links

MCGallaspy:
Extends a builtin theme to include a link to the github source for the files. Had to include a some static files from the sphinx project. The branch targeted by the link is set by an option in `theme.conf` -- an ansible playbook script should probably override this option.

ss


aronasorman:
I trust that this is working as advertised.

Issue #3552, opened 2015-04-24
Adding to INSTALLED_APPS in local_settings.py deletes all existing INSTALLED_APPS


rtibbles:
Branch: develop

Expected Behaviour: Adding INSTALLED_APPS in local_settings.py should concatenate it to the tuple of INSTALLED_APPS.

Current Behaviour: Adding INSTALLED_APPS in local_settings.py overwrites the tuple of INSTALLED_APPS.

@benjaoming this seems to have changed with your latest changes in develop, but I can't for the life of me see where this might be happening. Any ideas?

benjaoming:
First, I'll say have to things the harsh way: Of course that shouldn't concatenate anything!! That would be pure black magic!!

But I can understand that there may be project settings that would expect this. What's your take on this @rtibbles ?

benjaoming:
I wasn't aware that this was a convention, but do we want to keep it? I don't mind breaking bad conventions at some point, but I wouldn't wanna break them without introducing a proper alternative... so since the whole project refactor is not in place yet, it might be better to keep it...

benjaoming:
Sorry, I didn't answer the last part :) Where things are happening is in `kalite.settings.base`. At the very bottom.

rtibbles:
@benjaoming I am quite happy for bad conventions to be broken, I assume in future the way we would add something like this would be through the plugin architecture you had previously proposed?

benjaoming:
Not the plugin architecture per say, but the settings arch that would be
part of splitting project logic from app logic... so I'm all for keeping a
`settings.local` pattern where you can put local settings there.. it can be
quite convenient. In the longer run, it's easier though to use perhaps a
few common options for the `kalite` command or environment variables with
one's own project such as "DJANGO_SETTINGS_MODULE=my_project.settings
kalite start".

We could even create a helper function for kalite to bootstrap a custom
django project environment `kalite create_project` or something.

The difference in `settings.local` from the old local_settings.py method
will be that settings can be imported from `settings.local` without
circularity, so you could just do this in `settings.local` :

    from settings.base import INSTALLED_APPS
    INSTALLED_APPS= list(INSTALLED_APPS) + ["windows_xp"]

benjaoming:
Not the plugin architecture per say, but the settings arch that would be
part of splitting project logic from app logic... so I'm all for keeping a
`settings.local` pattern where you can put local settings there.. it can be
quite convenient. In the longer run, it's easier though to use perhaps a
[...]

PR #3579, opened 2015-04-29
Deprecate local settings (fix for #3552)

benjaoming:
#3552


kalite/settings/base.py
```diff
@@ -13,8 +15,15 @@
...
     # fixed later.
-    from kalite.local_settings import *  # @UnusedWildImport
     from kalite import local_settings
+    # This is not a DeprecationWarning by purpose, because those are
+    # ignored during settings module load time
+    warnings.warn(
+        "Using local_settings is deprecated, please write your own settings "
+        "module, import kalite.settings.base, and point your "
+        "DJANGO_SETTINGS_MODULE to your customized settings or run kalite "
+        "COMMAND --settings=my.settings",
```
aronasorman: This might be too terse and cryptic for normal admins, as they may not be aware of all the Django variables and python's import model. Here's a suggestion for the deprecation message:

> We will be deprecating the old way of statically importing custom settings, in favor of a more flexible way. The easiest way to update your installation is to the new way is to add an import statement above your `local_settings.py`:
> `import settings.base`
> and then call `kalite start` with additional arguments: `kalite start --settings=kalite.local_settings`

Some derivative of that. The goal is to have concrete steps for admins to use. Although ideally, the installers would be able to hide a lot of these details behind their shiny interfaces :)

aronasorman:
@benjaoming this is a big change on how settings will function. Please put in a description.

MCGallaspy:
Could this PR have any bearing on [this docs entry](https://github.com/learningequality/ka-lite/pull/3588/files#diff-b7247cb1b0de3924fd85b04f25c5097dR584)?

kalite/settings/base.py
benjaoming: That's circular :) If local_settings is supposed to deprecated and still usable in the "old way", we have to tell people to put their customized settings in a different module in the future... but making another suggestion based on that.

aronasorman: Sure. My only concern is to outline concrete actions towards moving from the old way to the new way.

benjaoming:
Can we have this merged and look at the documentation in another issue? I think we need to rewrite stuff on how to configure things, anyways.

benjaoming:
Put an issue about docs in #3611 -- think this one is ready.. @aronasorman ? @MCGallaspy ?

MCGallaspy:
@benjaoming I'll take a look at this one.

kalite/contentload/__init__.py
```diff
@@ -0,0 +1 @@
+# DO NOT PUT ANYTHING HERE
```
aronasorman: Why are we not putting anything here? There's a dearth of explanations on this PR.

benjaoming: Because it's an __init__.py file so anything you put in there will be
loaded when importing sub-modules and that's really annoying and has caused
lots of head aches already.. so I would really recommend not to do this
until we're done cleaning up the codebase :) Which we will some day :)

aronasorman: doubleplusgood :+1:

aronasorman:
Looks good to me. We need an entry in the release notes section of the docs, though, as per what @MCGallaspy said :) Mike said he'll take a look at this to though, so I'll leave it up to him to merge.

PR #3580, opened 2015-04-29
Vagrant fix

MCGallaspy:
Someone added the Vagrant file, but provisioning fails because the provisioning script doesn't exist. Also remove selenium, since it's been added to `requirements_dev.txt`


aronasorman:
Don't think the windows installer (and 0.13 in general) supports pip installed packages yet, so can you update the provision script to reflect that, as well as restore selenium, and delete requirements.txt?

PR #3582, opened 2015-04-29
0.13.x

### LINKED ISSUE #3536: Some questions do not contain the needed content to answer them.
Trying to make this as vague as possible because I do not want to point to a specific cause (I have no idea what that cause is) so I will leave this picture of the issue encountered? I think it may be relatd to how the equations are rendered in our code as the same question at the khan academy offical site works fine I will include both pics for reference?

MCGallaspy:
Hopefully resolves #3536.


MCGallaspy:
This mysteriously does not fix #3536.

MCGallaspy:
The conflicts were only `version.py` and `dev_requirements.txt`.

rtibbles:
Curious.

PR #3585, opened 2015-04-30
enable to specify khan(perseus images) folder from local_settings

66eli77:
This is especially useful when run kalite on Android, where user may put video content on SD card and don't need to move the khan folder(super huge)
Also added TARGET_OS flag in local_settings for kalitectl.py to solve the unrooted privilege issue on android

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -80,7 +80,13 @@ if os.name == "nt":
 # Necessary for loading default settings from kalite
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kalite.settings")
 
-KALITE_HOME = os.path.join(os.path.expanduser("~"), ".kalite")
+from django.conf import settings
+if getattr(settings, "TARGET_OS", None) == "Android":
+    #puting this dir in the current working dir solves the unrooted privilege issue on Android
+    KALITE_HOME = os.path.join(os.getcwd(), ".kalite")
+else:
+    KALITE_HOME = os.path.join(os.path.expanduser("~"), ".kalite")
+
 if not os.path.isdir(KALITE_HOME):
     os.mkdir(KALITE_HOME)
 PID_FILE = os.path.join(KALITE_HOME, 'kalite.pid')
@@ -219,7 +225,6 @@ def get_pid():
         raise NotRunning(7)  # Unclean shutdown
 
     # TODO: why is the port in django settings!? :) /benjaoming
-    from django.conf import settings
     listen_port = getattr(settings, "CHERRYPY_PORT", LISTEN_PORT)
 
     # Timeout is 1 second, we don't want the status command to be slow
--- a/kalite/distributed/urls.py
+++ b/kalite/distributed/urls.py
@@ -43,6 +43,9 @@ urlpatterns += patterns('',
 # serving it otherwise. Cherrypy is currently serving it through modifications
 # in kalite/django_cherrypy_wsgiserver/cherrypyserver.py
 urlpatterns += patterns('',
+    url(r'^%s/khan(?P<path>.*)$' % settings.CONTENT_URL[1:], 'django.views.static.serve', {
+        'document_root': settings.CONTENT_ROOT_KHAN,
+    }),
     url(r'^%s(?P<path>.*)$' % settings.CONTENT_URL[1:], 'django.views.static.serve', {
         'document_root': settings.CONTENT_ROOT,
     }),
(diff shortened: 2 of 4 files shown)


66eli77:
@benjaoming  can you review this PR since the many of the changes are made on your code.

kalitectl.py
```diff
@@ -80,7 +80,13 @@
 # Necessary for loading default settings from kalite
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kalite.settings")
 
-KALITE_HOME = os.path.join(os.path.expanduser("~"), ".kalite")
+from django.conf import settings
```
benjaoming: You can't do this here :) Well, maybe in practice it will work :) ...but we should separate loading the whole django stack from the main scope of kalitectl, otherwise we are taking a functional django env. for granted which will not be the case when using commands that configure stuff.

kalitectl.py
```diff
@@ -219,7 +225,6 @@ def get_pid():
         raise NotRunning(7)  # Unclean shutdown
 
     # TODO: why is the port in django settings!? :) /benjaoming
-    from django.conf import settings
```
benjaoming: So the TODO was left in there together with the import statement because we're trying to avoid loading the settings module unnecessarily... but also because we should find a more configurable approach to configuring which port the server runs on.

66eli77: I'm have problem with settings object cannot find the correct source code. Do you think it has anything to do with this change?  for error detail, please check hipchat

benjaoming:
Sorry but I have to question this new `settings.CONTENT_ROOT_KHAN` :)

Is it necessary?

It's basically saying "this url path content/khan should be served statically from a different location".

If you need a customized location, would a symlink be sufficient? Try it out and see how cherrypy likes that :) Or maybe Android can't symlink from the application to an SD card?

And why not just customize the whole `content/` folder (`CONTENT_ROOT`), why only the "khan" part?

Furthermore, the `khan` folder name should be treated as a variable coming from that specific channel.. not a hard-coded string.

Is there an easier way to get around it without adding more to our legacy settings code base? Could it perhaps be that we bookmark it as a separate issue until we can treat the whole Android-specific setup in its own settings structure?

66eli77:
you are right, cannot do symlink on SD card(FAT32 don't support symlink)
I just need a way to read khan folder(perseus image) from a location I specified
and I'm happy to do it in current practice instead of using "legacy", if you can give me some pointers to show how the current practice works.

benjaoming:
@66eli77 there's no current practice implemented :)

But would it work to simply switch the location of `CONTENT_ROOT` ?

66eli77:
the goal is to decouple khan folder and content folder, so that khan folder(perseus images) can be packed into the app, and user can put the content(videos, mp3 ... ) on a SD card.
In our current implementation, the khan has to be inside the content folder(CONTENT_ROOT).

kalitectl.py
benjaoming: @66eli77 unless you have an alternative solution, I'm going to make a PR that makes it possible to use an env. variable for this. You just can't import django settings here.

kalitectl.py
benjaoming: Can't find anything on hipchat?

benjaoming: Could you post the error here?

66eli77: Traceback (most recent call last):
  File "kalitectl.py", line 507, in <module>
    args=arguments['DJANGO_OPTIONS']
  File "kalitectl.py", line 369, in start
    manage('kaserve', args)
  File "kalitectl.py", line 300, in manage
    utility.execute()
  File "python-packages/django/core/management/__init__.py", line 392, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "python-packages/django/core/management/__init__.py", line 272, in fetch_command
    klass = load_command_class(app_name, subcommand)
  File "python-packages/django/core/management/__init__.py", line 77, in load_command_class
    module = import_module('%s.management.commands.%s' % (app_name, name))
  File "python-packages/django/utils/importlib.py", line 36, in import_module
    __import__(name)
  File "/data/data/com.android.kalite27/files/kalite/distributed/management/commands/kaserve.py", line 18, in <module>
File "/data/data/com.android.kalite27/files/kalite/caching/__init__.py", line 27, in <module>
    from kalite import i18n, topic_tools
  File "/data/data/com.android.kalite27/files/kalite/i18n/__init__.py", line 42, in <module>
    DUBBED_VIDEOS_MAPPING_FILEPATH = os.path.join(settings.I18N_DATA_PATH, "dubbed_video_mappings.json")
  File "python-packages/django/conf/__init__.py", line 54, in __getattr__
    return getattr(self._wrapped, name)
AttributeError: 'Settings' object has no attribute 'I18N_DATA_PATH'

66eli77:
this PR is irrelevant anymore, we changed the approach to retrieve Perseus image from khan folder on Android

benjaoming:
Okay, but I think that #3599 is still relevant as you will probably run into other similar issues.

PR #3591, opened 2015-05-01
Propagate ALL django settings to their respective commands, also fixes t...

benjaoming:
Propagate ALL django settings to their respective commands, also fixes the kaserve command which also didn't do this. This introduces the option of kalite start --settings=my_settings_module and also support for --traceback

Needed for deprecating local_settings #3579

Update:

The description of this PR was a bit hard to understand.. but for instance, this didn't work before:

    kalite start --settings=tralalaa

Because the `--settings` option was never passed on to the next django management commands.

Furthermore, this didn't work:

    kalite start --traceback

We should've had tracebacks in case of startup failures... we will now :)

### DIFF
--- a/kalite/distributed/settings.py
+++ b/kalite/distributed/settings.py
@@ -19,7 +19,6 @@ except ImportError:
 
 DEBUG = getattr(local_settings, "DEBUG", False)
 
-
 ########################
 # Functions, for support
 ########################
@@ -37,6 +36,7 @@ def USER_FACING_PORT():
 # TODO(bcipolli): change these to "login" and "logout", respectively, if/when
 #  we migrate to a newer version of Django.  Older versions require these
 #  to be set if using the login_required decorator.
+# TODO(benjaoming): Use reverse_lazy for this sort of stuff
 LOGIN_URL = "/securesync/login/"
 LOGOUT_URL = "/securesync/logout/"
 
--- a/kalite/distributed/management/commands/kaserve.py
+++ b/kalite/distributed/management/commands/kaserve.py
@@ -98,8 +98,12 @@ class Command(BaseCommand):
 
 
     def handle(self, *args, **options):
-        # Eliminate irrelevant settings
+        # Store base django settings and remove them from the options list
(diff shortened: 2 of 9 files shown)


benjaoming:
I'm including a fix for @66eli77 's problems on Android because basically this is as critical to get merged as everything else in here :)

kalite/distributed/settings.py
```diff
@@ -37,6 +36,7 @@ def USER_FACING_PORT():
 # TODO(bcipolli): change these to "login" and "logout", respectively, if/when
 #  we migrate to a newer version of Django.  Older versions require these
 #  to be set if using the login_required decorator.
+# TODO(benjaoming): Use reverse_lazy for this sort of stuff
 LOGIN_URL = "/securesync/login/"
 LOGOUT_URL = "/securesync/logout/"
```
MCGallaspy: @rtibbles I don't think these are valid URLs anymore, right?

rtibbles: No, but I was unsure of their exact function as settings and didn't know
how to migrate them.

benjaoming: When you use for instance the `@login_required` decorator, you'll be redirected to settings.LOGIN_URL when the user isn't logged in.. this is the default URLs that django.contrib.auth uses except in the django admin.

MCGallaspy: Is it possible to override that redirect?

benjaoming: Dunno, but settings.LOGIN_URL works for everything except django.contrib.admin which always redirects to the admin backend's login page.

benjaoming:
@aronasorman I think this is pretty merge-able!

MCGallaspy:
Hmm, test says `Unknown command: 'kaserve'`, but it works ok for me.

MCGallaspy:
Passed now. I'm happy with it, too.

benjaoming:
I know what causes the spontaneous "unknown command", will dig into it.

PR #3600, opened 2015-05-04
Add python version warning

### LINKED ISSUE #3593: Python quits and ka-lite server is killed when trying to download videos (on Macbook)
Branch: master (0.13.1)
Expected Behavior: Upon clicking the download button in the video management section the download should start
Current Behavior: Upon clicking the download button the download seems to start in the background (page is updated) but Python quits immediately with an error message and shuts down the ka-lite server
Steps to reproduce: Every time I try to download videos via the user interface a ka-lite server running on the Macbook Air 2011. Newest version of ka-lite and python are being used.
 
screenshot 2015-05-03 19 29 18

MCGallaspy:
Fixes #3593. Adds a warning about the included version of python and a link to download 2.7.9.

### DIFF
--- a/sphinx-docs/installguide/install_mac.rst
+++ b/sphinx-docs/installguide/install_mac.rst
@@ -3,23 +3,24 @@ Mac Installation Guide
 
 Macintosh installation uses the command line, but if you follow the instructions carefully, you should have no problem installing KA Lite. KA Lite uses the `git` program to distribute its software to developers and other users.
 
-#. Download and install either Python2.6_ or Python2.7_ for Mac
-	.. _Python2.6: https://www.python.org/download/releases/2.6/
-	.. _Python2.7: https://www.python.org/download/releases/2.7/
-#. You need to open the Terminal app. There are many ways to do this. We suggest the following:
+1. Download and install `Python 2.7.9 <https://www.python.org/downloads/mac-osx/>`_ for Mac
+
+.. warning:: Even if Python is already installed on your Mac by default, it is *highly* recommended that you ensure you are using the latest version. Using older versions of Python with KA Lite can result in errors.
+
+2. You need to open the Terminal app. There are many ways to do this. We suggest the following:
 	* Click the magnifier icon on the upper right corner and search "terminal".
 	* Press *'Enter'* with the "Terminal" app selected and it should open.
-#. In the Terminal, enter *Python -V* and press *'Enter'*. **The version number should reflect the Python version you just installed.**
-#. Create the folder where you would like to install KA Lite.	
+3. In the Terminal, enter *Python -V* and press *'Enter'*. **The version number should reflect the Python version you just installed.**
+4. Create the folder where you would like to install KA Lite.
 	* In the Terminal, type *cd* and then drag the folder you created into the Terminal.
 	* This will copy the path name into the Terminal window.
 	* Press *'Enter'*.
-#. Now you can copy and paste this command into the terminal and press *'Enter'*.
+5. Now you can copy and paste this command into the terminal and press *'Enter'*.
 	* *git clone https://github.com/learningequality/ka-lite.git*
-#. Once it finishes downloading, enter *cd ka-lite* to move into the folder that was just downloaded.
-#. To begin installation enter *./setup_mac.command*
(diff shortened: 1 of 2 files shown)


benjaoming:
@MCGallaspy Is there a way to detect this problem programmatically? If so, I would recommend (after you've had a chance to merge my local_settings deprecation ;) ) to have a look at `kalite.shared.warnings` and perhaps include a warning targeting this.

sphinx-docs/installguide/install_mac.rst
```diff
@@ -3,23 +3,24 @@ Mac Installation Guide
 
 Macintosh installation uses the command line, but if you follow the instructions carefully, you should have no problem installing KA Lite. KA Lite uses the `git` program to distribute its software to developers and other users.
 
-#. Download and install either Python2.6_ or Python2.7_ for Mac
-	.. _Python2.6: https://www.python.org/download/releases/2.6/
-	.. _Python2.7: https://www.python.org/download/releases/2.7/
-#. You need to open the Terminal app. There are many ways to do this. We suggest the following:
+1. Download and install `Python 2.7.9 <https://www.python.org/downloads/mac-osx/>`_ for Mac
```
aronasorman: Note: This won't be necessary once we have an OS X installer, as we have a python interpreter bundled in there.

aronasorman:
I agree, there should be a way to issue a warning when `sys.platform` is OS X and python version < 2.7.

aronasorman:
`sys.platform` and `python.version_info`, respectively.

benjaoming:
But is that all there is to it or are there certain versions of 2.7 on OSX
that also reveal that it's the OSX shipped one? And what exactly is the
problem with this one? Missing libraries?

aronasorman:
The problem was that OS X's builtin python is hitting a segmentation fault when we download too many videos. So I don't think it's a missing library, but rather a bug in the interpreter itself.

My guess here is that this specific bug is fixed in the latter micro releases (2.7.7+), as [2.7.7 includes a lot of segmentation fault fixes](https://hg.python.org/cpython/raw-file/f89216059edf/Misc/NEWS). Yosemite has 2.7.6, so it might've not been fixed by then. Using homebrew's python fixes this.

Or I dunno, could be something with MacPython specifically, as we don't encounter this anywhere else.

benjaoming:
Would there be a point about issuing warnings targetting python 2.7.0 - 2.7.7 on all platforms?

2.6 needs a warning because it's getting deprecated no matter what, right?

aronasorman:
> Would there be a point about issuing warnings targetting python 2.7.0 - 2.7.7 on all platforms?

That's a good point, I just read through the `2.7.7` changelog and there's just too many bugfixes in there :)

> 2.6 needs a warning because it's getting deprecated no matter what, right?

It might be good to emit the same warning as above when we see that they're running 2.6, for "better stability and security".

MCGallaspy:
I'm happy to add a warning somewhere. Just to be clear, are we only supporting python on Mac above version 2.7.7? Or are earlier minor versions (2.7.0-2.7.7) problematic on all OSes? Might as well be explicit about which combos of OS/Python interpreter are supported.

benjaoming:
Okay great, I think the earlier 2.7 versions are problematic on all systems. But after having read that Ubuntu 12.04 which is still supported and doesn't seem to cause issues, is running Python 2.7.3, I think it might be a bit over the top to warn all OS'es about it.

Maybe:

1. Warn all users of 2.6 that we're dropping it?
1. Warn OSX users if using > 2.6 and < 2.7.7

Btw... allow me to pitch the new `kalite.shared.warnings` once again :)

aronasorman:
@benjaoming I'd actually prefer to keep it simple and warn all users using versions < 2.7.7 :) I think it's just one less thing to worry about, on the off-chance that their system will crash we can say we already warned them about older versions of python, and should just find a way to upgrade.

aronasorman:
P.S. I myself am running 2.7.6 (the debug version).

MCGallaspy:
Logout link test! It's been crushed on `develop` at least.

aronasorman:
Looks good to me. Is there anything else we need for this PR? If not, feel free to merge.

benjaoming:
Looks god to me, too :+1:

benjaoming:
Also tests fail with usual timeouts.

Issue #3562, opened 2015-04-27
Move version.py version metadata to YAML


aronasorman:
I've had a couple of stupid commits before amending my mistakes when editing this file directly. If only there was a more human-friendly version of JSON, meant for humans to use...

Oh cool there is one: http://yaml.org/spec/

So my suggestion for here is to create a new pre-commit hook that will convert the yaml file to JSON before committing. That JSON file will be the one that's read by `version.py` on app startup.

aronasorman:
@ruimalheiro wanna take this on? This should be pretty simple.

jamalex:
> convert the yaml file to JSON before committing

I would recommend just having the conversion to JSON (which is just for efficiency, I'm assuming) happen during the build process, since:
- we can't rely on everyone having these precommit hooks
- we can just have version.py load the YAML directly during development

It could look for the JSON file, and fall back to the YAML if the JSON didn't exist (might be good as a general-purpose function, as well -- `load_json_or_yaml`?)

aronasorman:
@jamalex sounds good.

benjaoming:
Isn't the problem that we _aren't_ putting things in kalite.version and isn't that actually more a problem about not doing things conventional and accessible enough for devs?

I would suggest maintaining a CHANGELOG.md for conventionality. If we desire to show it to users, either dynamically render it to HTML with python-markdown (but don't think it's available) or automatically build an HTML version every time we change CHANGELOG.md

The change log IMO is mainly for people who are really interested in following the development... hiding it somewhere in the UI and in an unconventional python module or YAML file won't get us where we should be. The rest of the audience will read it on the blog or in a newsletter.

jamalex:
> The change log IMO is mainly for people who are really interested in following the development... hiding it somewhere in the UI and in an unconventional python module or YAML file won't get us where we should be

In the web-based update process, we've experimented with displaying the incoming changes to the admin, so they can choose whether to update, and know what they'll be getting. So, a machine-readable format is useful for things like that. But the idea is just to have very high-level items in there, not a full changelog.

benjaoming:
I'm not sure I understand correctly? If you're running a local 0.13, it won't have the YAML / kalite.version of 0.14 so you can only display the changes after upgrading? Or you mean that people should see them by calling an external online webservice?

ruimalheiro:
@aronasorman I will take a look on this! :+1:

@jamalex and @benjaoming thanks for the insights!

[...]

PR #3602, opened 2015-05-05
Version.py to YAML

ruimalheiro:
This PR will address issue #3562 

Notes regarding the "version.yaml" file:
- Python's "None" is represented as:
```
fieldname:   
```
- Empty lists are represented as:
``` 
[]
```
- A list is represented as:
```
fieldname:
                - item 1
                - item 2
                ...
```
- Empty string represented as:
```
fieldname: ""
```

IMPORTANT - yamp parser seems to have trouble when you have spaces and tabs ("\t"). We can avoid any trouble by simply converting all tabs to spaces.

TODO - Generalise the solution to load YAML or JSON. A better way to get "data" path.


kalite/version.py
```diff
@@ -5,386 +9,30 @@
...
+    Raises:
+        IOError: If the file doesn't exist or can't be opened.
+    """
+    try:
+        file_stream = open(file_name, "r")
+        version_object = yaml.load(file_stream)
+        version_object[current_version] = version_object[yaml_file_version_mapping]
+        del version_object[yaml_file_version_mapping]
+        return version_object
+    except IOError:
```
aronasorman: Rather than trying to catch this error, why not just let this bubble up the stack?

ruimalheiro: Hey @aronasorman !

By "why not just let this bubble up the stack", do you mean, remove de try - except?

benjaoming:
@ruimalheiro great to have `version.VERSION_INFO` removed, but have you tested where this variable might be used elsewhere?

ruimalheiro:
@benjaoming there variable was not removed. It is still there!

benjaoming:
@ruimalheiro ah, now I see, sorry Github really made a mess of the diff for this one :)

But regarding the placement of the .yml file, could you have a look around and find a better place for it? Otherwise we have to track it as an explicit data file for installations. Also, you can't look it up relative to version.py, please refer to the way that data files are generally located. Can't give out more hints this week, but I think there's sufficient proof of concept in the current codebase :)

benjaoming:
If the file is supposed to be writable by a user process (for web updates), you might wanna distribute it as a global data file (like the locales) that's automatically copied into user writable directories when it doesn't exist.

ruimalheiro:
haha np :+1: 

Yep, that sounds good! Thanks for the hints @benjaoming ! I will take a look on it.

kalite/version.py
aronasorman: Yup!

kalite/version.yaml
```diff
@@ -0,0 +1,397 @@
+
+"VERSION" :
```
aronasorman: Hmm can you try removing the quotes in here and just have it as:

```yaml
0.13.0:
    - release_date:
    - git_commit: b9caa
```

If that works, I prefer that. It looks cleaner :) 

Also, I think you should remove the VERSION entry, since yaml doesn't do any substitution.

ruimalheiro: @aronasorman, 

Yep, I think that works! Will try once I arrive home :+1: 

I'm doing the substitution once I load the YAML file. (So it uses the VERSION value). If you think it is not necessary, sounds good to me and then, we just write the version directly on the yaml file! :+1:

aronasorman: Sounds good!

aronasorman:
Oh, and please resolve the merge conflict :)

kalite/version.py
```diff
@@ -5,416 +9,18 @@
...
-        },
-        "bugs_fixed": {
-            "all": [],
-            "students": [],
-            "coaches": [],
-            "admins": [],
-        },
-    },
-}
+VERSION_INFO = load_yaml(os.path.join(os.path.dirname(__file__), "../data/version.yml"))
```
aronasorman: couple of notes here:

- don't use os-specific path separators, use the `os.path.join` fully instead :) so change this to something like `os.path.join(dirname(file), "..", "data", "version.yml")`.

- @benjaoming mentioned that looking for the version.yml relative to this file is a bad idea, but right now we import version.py inside settings, so importing settings (where DATA_PATH is defined) would cause a circular import. Not sure what's the best solution, but I'm willing to merge this in for now until we find a better way.

kalite/version.py
```diff
@@ -5,416 +9,18 @@
...
-        },
-        "bugs_fixed": {
-            "all": [],
-            "students": [],
-            "coaches": [],
-            "admins": [],
-        },
-    },
-}
+VERSION_INFO = load_yaml(os.path.join(os.path.dirname(__file__), "..", "data", "version.yml"))
```
benjaoming: @aronasorman @ruimalheiro you still can't do this... as stated in my previous comment.. data files can't just be loaded from a relative path.. see how they're loaded elsewhere.

ruimalheiro:
@benjaoming is the data path defined on settings (base.py or __init__.py)?

If so, is it possible to use any of thoses variables in version.py? What would be the best approach to follow then?

benjaoming:
Hi @ruimalheiro 

Thinking about this again... in `kalite/__init__.py`, we have this:

    from version import *

This means that it's pretty hard to avoid the `version` module being loaded before other things. But loading it means you also load and parse a data file.

This is bad because:

1. Inevitably loading and parsing a file no matter what is bad. That makes the `kalite` package slower for all purposes, and accessing detailed version data is not something that we wanna do more than just a little once in a while.
1. You are right. Determining where data files should go happens in settings, if anything we should be able to fetch the application's version numbers without relying on the full settings stack being parsed before. Version numbers are pretty fundamental.

So I think you should deprecate `VERSION_INFO` and burn it down completely. If the `.yml` data should be accessed, we have to call the function `load_yaml` and that function can be responsible for fetching the `.yml` file's path by importing the Django settings stack `from django.conf import settings`.. but you do this in the `load_yaml` function space.

@aronasorman would you agree?

benjaoming:
@ruimalheiro in case you lack clarity about what I mean, consider the following:

```
➜  ~  mkdir lalala
➜  ~  echo "print 'module 1'" > lalala/__init__.py
➜  ~  mkdir lalala/bababa
➜  ~  echo "print 'module 2'" > lalala/bababa/__init__.py
➜  ~  mkdir lalala/bababa/hahaha
➜  ~  echo "print 'module 3'" > lalala/bababa/hahaha/__init__.py
➜  ~  python
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
>>> import lalala.bababa.hahaha
module 1
module 2
module 3
```

What you see is that I'm creating the package `lalala.bababa.hahaha` and loading that package will trigger execution of all `__init__.py` files.

That's why it's not recommendable to put stuff in those files, because the consequences are hard to deal with.

PR #3612, opened 2015-05-06
Topic tools backend

MCGallaspy:
Just want to get the ball rolling on the discussion here. The main action is in `topic_tools/models.py` and `kalite/topic_tools/management/commands/init_topic_tree.py`.

Points of discussion include: 

* What kind of attributes do we want our topic tree models to have?
* How should we handle different node types (topic, exercise, etc)?
  * My first thought is to use a polymorphic foreign key on nodes to another model which will represent subtypes -- so a Topic model, an Exercise model, or YourCoolContentType model.
  * Perhaps it might be better to create a polymorphic tree, i.e. all topic tree nodes extend a base node type. Not sure if it's supported in django, but it would be desirable if the queryset for the base class included all the subclasses.
* How to import data from KA? It's a little tricky when it's not clear what attributes are available.

By the way, I get the following error when I run the `init_topic_tree` management command, which is trying to bootstrap the DB from the json stuff.
```sh
  File "python-packages/django/core/management/base.py", line 224, in run_from_argv
  File "python-packages/django/core/management/base.py", line 263, in execute
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 61, in handle
    recurse_nodes(raw_topics)
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 60, in recurse_nodes
    recurse_nodes(child, node)
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 60, in recurse_nodes
    recurse_nodes(child, node)
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 60, in recurse_nodes
    recurse_nodes(child, node)
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 60, in recurse_nodes
    recurse_nodes(child, node)
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 60, in recurse_nodes
    recurse_nodes(child, node)
  File "/home/vagrant/ka-lite/kalite/topic_tools/management/commands/init_topic_tree.py", line 54, in recurse_nodes
    node.save()
  File "/usr/local/lib/python2.7/dist-packages/mptt/models.py", line 956, in save
    self.insert_at(parent, position='last-child', allow_existing_pk=True)
  File "/usr/local/lib/python2.7/dist-packages/mptt/models.py", line 702, in insert_at
    self, target, position, save, allow_existing_pk=allow_existing_pk, refresh_target=refresh_target)
  File "/usr/local/lib/python2.7/dist-packages/mptt/managers.py", line 519, in insert_node
    self._post_insert_update_cached_parent_right(parent, right_shift)
  File "/usr/local/lib/python2.7/dist-packages/mptt/managers.py", line 667, in _post_insert_update_cached_parent_right
    raise InvalidMove
InvalidMove
```
@benjaoming as an django-mptt contributor do you have any special insight here? I suspect in-memory MPTTModel attributes are not being updated when new nodes are inserted in the tree.

Some benchmark data incoming.

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -266,6 +266,7 @@ INSTALLED_APPS = (
     "django.contrib.sessions",
     "kalite.distributed",
     "compressor",
+    "mptt",
 )
 
 if not BUILT:
--- a/kalite/topic_tools/management/__init__.py
+++ b/kalite/topic_tools/management/__init__.py
@@ -1,0 +1,1 @@
+__author__ = 'Mike'
--- a/kalite/topic_tools/management/commands/__init__.py
+++ b/kalite/topic_tools/management/commands/__init__.py
@@ -1,0 +1,1 @@
+__author__ = 'Mike'
--- a/kalite/topic_tools/management/commands/init_topic_tree.py
+++ b/kalite/topic_tools/management/commands/init_topic_tree.py
@@ -40,8 +40,8 @@ class Command(BaseCommand):
             """
             # In general title or description could be absent, but it's probably an error
             # if the other fields are.
-            subtype = NodeSubtype()
-            subtype.save()
+            # Just use one object to avoid proliferation, until this field becomes meaningful
+            subtype, found = NodeSubtype.objects.get_or_create(pk=1)
             node = TopicTreeNode(
                 title=_(raw_node.get("title", "")),
                 description=_(raw_node.get("description", "")) if raw_node.get("description") else "",
@@ -52,6 +52,10 @@ class Command(BaseCommand):
                 node_subtype=subtype
             )
             node.save()
+            # Have to reload the parent from the database in order to ensure tree
+            # attributes in memory are properly updated.
+            if parent:
+                parent = TopicTreeNode.objects.get(pk=parent.pk)
             for child in raw_node.get("children", []):
                 recurse_nodes(child, node)
         recurse_nodes(raw_topics)
--- a/kalite/topic_tools/management/commands/init_topic_tree.py
+++ b/kalite/topic_tools/management/commands/init_topic_tree.py
@@ -35,7 +35,7 @@ class Command(BaseCommand):
         def recurse_nodes(raw_node, parent=None):
             """
             Recurse through the raw_topics and create TopicTreeNodes
-            :param node: a node from raw_topics
+            :param raw_node: a node from raw_topics
             :param parent: a saved topic_tools.models.TopicTreeNode
             """
             # In general title or description could be absent, but it's probably an error
--- a/kalite/topic_tools/management/commands/init_topic_tree.py
+++ b/kalite/topic_tools/management/commands/init_topic_tree.py
@@ -1,0 +1,57 @@
+"""
+A management command to populate the topic tree from our various json files.
+This is a limited-use command meant to transition from the json magic in
+topic_tools/__init__.py to using the Django backend. Specifically, it assumes
+there's one channel and it's set by settings.CHANNEL.
+
+This management command should be idempotent if the underlying data is the same, which just means
+that running twice should be a no-op.
+"""
+import os
+from django.core.management.base import BaseCommand
+from django.utils.translation import gettext as _
+
+from fle_utils.general import softload_json
+
+from django.conf import settings; logging = settings.LOG
+TOPICS_FILEPATHS = {
+    settings.CHANNEL: os.path.join(settings.CHANNEL_DATA_PATH, "topics.json")
+}
+EXERCISES_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "exercises.json")
+ASSESSMENT_ITEMS_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "assessmentitems.json")
+CONTENT_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "contents.json")
+
+from kalite.topic_tools.models import TopicTreeNode, NodeSubtype, Channel
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        # Create the one channel.
+        khan_ch, found = Channel.objects.get_or_create(language="en", title="khan")
+
+        raw_topics = softload_json(TOPICS_FILEPATHS.get(settings.CHANNEL), logger=logging.debug, raises=False)
+        # raw topics is the actual tree structure... have to parse it a node at a time
+        def recurse_nodes(raw_node, parent=None):
+            """
+            Recurse through the raw_topics and create TopicTreeNodes
+            :param node: a node from raw_topics
+            :param parent: a saved topic_tools.models.TopicTreeNode
+            """
+            # In general title or description could be absent, but it's probably an error
+            # if the other fields are.
+            subtype = NodeSubtype()
+            subtype.save()
+            node = TopicTreeNode(
+                title=_(raw_node.get("title", "")),
+                description=_(raw_node.get("description", "")) if raw_node.get("description") else "",
+                slug=raw_node["slug"],
+                path=raw_node["path"],
+                channel=khan_ch,
+                parent=parent,
+                node_subtype=subtype
+            )
+            node.save()
+            for child in raw_node.get("children", []):
+                recurse_nodes(child, node)
+        recurse_nodes(raw_topics)
--- a/kalite/topic_tools/migrations/0001_initial.py
+++ b/kalite/topic_tools/migrations/0001_initial.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding model 'Channel'
+        db.create_table(u'topic_tools_channel', (
+            (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
+            ('language', self.gf('django.db.models.fields.CharField')(max_length=8)),
+            ('title', self.gf('django.db.models.fields.CharField')(max_length=50)),
+        ))
+        db.send_create_signal(u'topic_tools', ['Channel'])
+
+        # Adding model 'TopicTreeNode'
+        db.create_table(u'topic_tools_topictreenode', (
+            ('title', self.gf('django.db.models.fields.CharField')(max_length=50)),
+            ('description', self.gf('django.db.models.fields.CharField')(max_length=50)),
+            ('slug', self.gf('django.db.models.fields.SlugField')(unique=True, max_length=255, primary_key=True)),
+            ('path', self.gf('django.db.models.fields.URLField')(unique=True, max_length=200)),
+            ('channel', self.gf('django.db.models.fields.related.ForeignKey')(to=orm['topic_tools.Channel'])),
+            ('content_type', self.gf('django.db.models.fields.related.ForeignKey')(to=orm['contenttypes.ContentType'])),
+            ('object_id', self.gf('django.db.models.fields.PositiveIntegerField')()),
+            (u'lft', self.gf('django.db.models.fields.PositiveIntegerField')(db_index=True)),
+            (u'rght', self.gf('django.db.models.fields.PositiveIntegerField')(db_index=True)),
+            (u'tree_id', self.gf('django.db.models.fields.PositiveIntegerField')(db_index=True)),
+            (u'level', self.gf('django.db.models.fields.PositiveIntegerField')(db_index=True)),
+        ))
+        db.send_create_signal(u'topic_tools', ['TopicTreeNode'])
+
+        # Adding model 'NodeSubtype'
+        db.create_table(u'topic_tools_nodesubtype', (
+            (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
+        ))
+        db.send_create_signal(u'topic_tools', ['NodeSubtype'])
+
+
+    def backwards(self, orm):
+        # Deleting model 'Channel'
+        db.delete_table(u'topic_tools_channel')
+
+        # Deleting model 'TopicTreeNode'
+        db.delete_table(u'topic_tools_topictreenode')
+
+        # Deleting model 'NodeSubtype'
+        db.delete_table(u'topic_tools_nodesubtype')
+
+
+    models = {
+        u'contenttypes.contenttype': {
+            'Meta': {'ordering': "('name',)", 'unique_together': "(('app_label', 'model'),)", 'object_name': 'ContentType', 'db_table': "'django_content_type'"},
+            'app_label': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            u'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'model': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            'name': ('django.db.models.fields.CharField', [], {'max_length': '100'})
+        },
+        u'topic_tools.channel': {
+            'Meta': {'object_name': 'Channel'},
+            u'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'language': ('django.db.models.fields.CharField', [], {'max_length': '8'}),
+            'title': ('django.db.models.fields.CharField', [], {'max_length': '50'})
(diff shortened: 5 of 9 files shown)


MCGallaspy:
Regarding the error, looks like it's being raised because I'm creating a cycle. :(

MCGallaspy:
Which probably means that the pk I'm using (the slug) is not unique. @rtibbles is that expected?

benjaoming:
@MCGallaspy I was going to throw in this comment as a general remark... maybe somehow you can forward it to others? We really have to stop putting all these migration files... one release should max. have one migration per application. The others can just be reverted and squashed

South can do this in three steps... don't know if there's an even easier way...

1. Roll back your local migrations: `kalite manage migrate application 0001  # or whatever was the latest released migration`
1. Delete all the new migration files
1. Recreate everything as a single migration: `kalite manage schemamigration application --auto`
1. Re-run the new migration `kalite manage migrate application`

benjaoming:
Here's the content models I created back in January, trying to get that guy with the two ice coffees to continue.

```python

"""
Models for storing content in structured trees with topic nodes.

Contents are editable by users, and thus there is a Draft inheritor for
different node types which is detached from the tree.

Current limitations:
 - A user cannot draft a different location in the tree.
 - Previous tree attachments are not maintained automatically, however
   leaving a copy of a node in a previous location with deleted=True
   is possible

Initially, this is thought to store all content as we start to only use JSON
data for importing and exporting. This is a long-term goal.

YAGNI (You Aren't Going to Need It): Much meta data is possible, please do not
add fields before a specific use case has occurred.
"""

from django.db import models
from django.utils.translation import ugettext_lazy as _

from mptt.models import MPTTModel, TreeForeignKey


class TopicTree(MPTTModel):
    """Base model for all trees / channels"""
    name = models.CharField(
        max_length=255,
        verbose_name=_("name / title for node"),
        help_text=_("Displayed to the user"),
    )
    editors = models.ManyToManyField(
        'auth.User',
        verbose_name=_("editors"),
        help_text=_("User accounts with full rights to edit tree"),
    )
    root_node = models.ForeignKey(
        'Node',
        verbose_name=_("root node"),
        help_text=_(
            "The starting point for the tree, the title of it is the "
            "title shown in the menu"
        ),
    )
    
    class Meta:
        verbose_name = _("Topic tree")
        verbose_name_plural = _("Topic trees")


class NodeData(models.Model):
    """
    Abstract model of fields belonging to all nodes.
    
    This model is there, because nodes can be removed from their and thus
    not belong to a tree (in draft mode etc.)
    """

    created = models.DateTimeField(auto_now_add=True, verbose_name=_("created"))
    modified = models.DateTimeField(auto_now=True, verbose_name=_("modified"))

    name = models.CharField(
        max_length=50,
        verbose_name=_("name"),
        help_text=_("Name of node to be displayed to the user in the menu"),
    )
    icon = models.CharField(
        max_length=24,
        choices=[  # Made up choices for non-existing icon set
            ('book', _('book')),
            ('school', _('school')),
            ('video', _('video')),
            ('math', _('math symbol')),
            ('chemistry', _('chemistry/science symbol')),
            ('astronomy', _('astronomy symbol')),
            ('picture-frame', _('picture frame')),
            ('check-mark', _('exercise icon')),
        ],
        verbose_name=_("icon"),
        help_text=_("Icon is next to the name in the menu"),
    )

    class Meta:
        abstract = True


class Node(MPTTModel, NodeData):
    """
    By default, all nodes have a title and can be used as a topic.
    """
    published = models.BooleanField(
        default=False,
        verbose_name=_("Published"),
        help_text=_("If published, students can access this item"),
    )
    deleted = models.BooleanField(
        default=False,
        verbose_name=_("Deleted"),
        help_text=_(
            "Indicates that the node has been deleted, and should only "
            "be retrievable through the admin backend"
        ),
    )
    sort_order = models.FloatField(
        max_length=50,
        unique=True,
        default=0,
        verbose_name=_("sort order"),
        help_text=_("Ascending, lowest number shown first"),
    )
    parent = TreeForeignKey('self', null=True, blank=True, related_name='children')
    
    @property
    def has_draft(self):
        return self.draft_set.all().exists()
    
    @property
    def get_draft(self):
        """
        NB! By contract, only one draft should always exist per node, this is
        enforced by the OneToOneField relation.
        :raises: Draft.DoesNotExist and Draft.MultipleObjectsReturned
        """
        return Draft.objects.get(publish_in=self)
    
    class MPTTMeta:
        order_insertion_by = ['sort_order']

    class Meta:
        verbose_name = _("Topic")
        verbose_name_plural = _("Topics")
        # Do not allow two nodes with the same name on the same level
        unique_together = ('parent', 'name')


class Draft(models.Model):
    """
    This model is not abstract in order to have a generic way to access
    drafts and attach them to nodes
    """
    publish_in = models.OneToOneField(
        'Node',
        verbose_name=_("Node to attach to if published"),
    )


class Content(models.Model):
    """
    Abstract model for content data, fields that are shared across videos,
    exercises, and books...
    """
    
    author = models.CharField(
        max_length=255,
        verbose_name=_("author(s)"),
        help_text=_("Name of the author(s) of book/movie/exercise"),
    )
    license_owner = models.CharField(
        max_length=255,
        blank=True,
        null=True,
        verbose_name=_("license owner(s)"),
        help_text=_("Organization of person who holds the essential rights"),
    )
    published_on = models.DateField(
        null=True,
        blank=True,
        verbose_name=_("when published"),
        help_text=_(
            "If applicable, state when this work was first published (not on "
            "this platform, but for its original publication)."
        ),
    )
    retrieved_on = models.DateTimeField(
        null=True,
        blank=True,
        verbose_name=_("when retrieved/downloaded"),
        help_text=_(
            "Should be automatically filled in when an item is downloaded ",
            "from its source of origin, either manually by user or "
            "automatically by script."
        ),
    )
    license = models.ForeignKey(
        'ContentLicense',
        verbose_name=_("license"),
        help_text=_("License under which the work is distributed"),
    )
    thumbnail = models.ImageField(
        upload_to='contents/video/thumbnails/',
        verbose_name=_("video thumbnail"),
        help_text=_("Automatically created when new video is uploaded"),
    )
    
    class Meta:
        abstract = True


class ContentVideo(Content):
    """
    Abstract model for video data
    """
    
    video_file = models.FileField(
        blank=True,
        null=True,
        upload_to='contents/video/thumbnails/',
        verbose_name=_("video file"),
        help_text=_("Upload video here"),
    )
    
    class Meta:
        abstract = True


class ContentVideoNode(Node, ContentVideo):
    
    class Meta:
        verbose_name = _("Video node")
        verbose_name_plural = _("Video node")


class ContentVideoDraft(Draft, ContentVideo):

    class Meta:
        verbose_name = _("Video draft")
        verbose_name_plural = _("Video drafts")


class ContentPDF(Content):
    """
    Abstract model for video data
    """
    
    pdf_file = models.FileField(
        blank=True,
        null=True,
        upload_to='contents/video/thumbnails/',
        verbose_name=_("video file"),
        help_text=_("Upload video here"),
    )

    class Meta:
        abstract = True


class ContentPDFNode(Node, ContentPDF):

    class Meta:
        verbose_name = _("PDF node")
        verbose_name_plural = _("PDF nodes")


class ContentPDFDraft(Draft, ContentPDF):

    class Meta:
        verbose_name = _("PDF draft")
        verbose_name_plural = _("PDF drafts")


class Exercise(Content):
    """
    Abstract class for Exercise data. Is used both as a node in a
    Topic Tree
    """
    
    class Meta:
        abstract = True


class ExerciseNode(Node, Exercise):

    class Meta:
        verbose_name = _("Exercise node")
        verbose_name_plural = _("Exercise nodes")


class ExerciseDraft(Draft, Exercise):
    
    class Meta:
        verbose_name = _("Exercise draft")
        verbose_name_plural = _("Exercise drafts")


class ContentLicense(models.Model):
    
    name = models.CharField(
        max_length=255,
        verbose_name=_("name"),
        help_text=_("Name of license, e.g. 'Creative Commons Share-Alike 2.0'")
    )
```

benjaoming:
@MCGallaspy having seen the work done in models.py here, I think the thoughts and ideas are about aligned, but maybe you'll find that the above is a bit more detailed?

I discussed this tree structure a lot with Jamie.

Also, the intention was to create "one content application to rule them all", so I put it in a new application, `kalite.content`.

The intention was furthermore to also include exercises as content data.

MCGallaspy:
> We really have to stop putting all these migration files

Noted. Will not commit migration files.

MCGallaspy:
@benjaoming but yeah, this is certainly helpful. Didn't realize I was duplicating your efforts here.

MCGallaspy:
So how should I proceed? It seems like there's been discussion already about how content should be handled (which unfortunately I didn't follow at the time). I think there is an immediate goal to stop loading the whole topic tree (and pals) into memory in order to reduce server start time and total memory use. It was hoped that using the Django ORM would help that. It's also clear that there's a longer term goal of having a more general content architecture.

Seems like the best way to avoid reduplicating efforts is to use @benjaoming's models here. Is it granted that this is the schema we'd like to use for content? @benjaoming can you elaborate on the use of these models a little bit? Correct me if I'm wrong, but:
* Instances of `TopicTree` represents the root node of different topic trees. These models don't prescribe any use, but down the road the intent is to have different topic trees associated with different channel & language combinations.
* Models that inherit from `Node` can have `parent`s of arbitrary classes as long as the `parent`'s class inherits from `MPTTModel`. Therefore, a topic tree is polymorphic in the sense that:
  * The root node has class `TopicTree`.
  * Children of the root node subclass `Node`, but for example an `ExerciseNode` can have a parent that's a `ContentPDFNode`.
  * Topics (nodes that are neither content nor exercises) are just plain `Node`s, suggesting that a topic has no metadata that's not common to all nodes.

Furthermore, the strategy for porting KA topics and exercises to this schema would be to simply take some attributes verbatim and simply discard the stuff we don't use.

rtibbles:
I don't think we would want anything other than a topic node having a child
node.

This is essentially what we do with KA topic data now. We might just be
leaving in a bit too much information.

MCGallaspy:
By the way, I did in fact determine that slugs are not unique -- in particular there are two "american-civics" -- so not a good pk.

benjaoming:
Are slugs important actually? We are building an offline site so SEO is not really a worry :) We could do /topic/123/random-ignored-text-slug/ and then only do lookups by ID, so the slugs can remain non-unique?

Fun fact: In Denmark, we call moustaches "slugs" :)

> @benjaoming can you elaborate on the use of these models a little bit?

Well, I think your understanding looks correct. Maybe if you and @jamalex can get together around a Whiteboard, Jamie can just share the structures of the tree with you? It's a bit perculiar... but key elements are: An element can contain other elements and those can be both content, exercises, and other topics.

Like:

```
            "Christianity"
                   / \
"Copticism"     "bible.pdf"
       |
    "video.mpeg"
```

MCGallaspy:
Alright, using the slug as the pk was pretty arbitrary in retrospect. Will remove and see how it fares. Still, I think I might close this PR and open a new one incorporating the models you've made here.

benjaoming:
Always use numeric values as the main PK!!! Anything else will backfire :)

MCGallaspy:
My initial rationale is that the tastypie API looks up nodes in the topic tree through their slug -- but in that case it seems that it also implicitly finds nodes by their parents. So, for instance, if one wanted to find a node it would be by looking at `all/the-slugs/in-a-sequence/to-the-last-one`, where `to-the-last-one` is the slug of the node being found, and all the `/`-separated slugs are the parents. A drop-in alternative using the ORM would need to be able to look up nodes similarly -- is it efficient using django-mptt to follow hierarchical structure like this @benjaoming? Meaning, in order to retrieve the example node, first one would need to retrieve the node with slug `all` (presumably the root, so a special reference could be held) and then search its children for `the-slugs`, and then search it's children and so on until the terminal node is reached. Obviously being able to retrieve a node by its pk is ideal, but if traversing the tree is also efficient then that, as they say, is that.

benjaoming:
@MCGallaspy yes that's what mptt is for :) It has reverse relations, so you could just do `SomeModel.objects.filter(id=3, parent__id=2, parent__parent__id=1)` for resolving something that is in "1/2/3".

rtibbles:
I believe it actually looks things up by id, which in the case of exercises
happen to look suspiciously like slugs. We should probably change this
practice.

jamalex:
> > We really have to stop putting all these migration files

> Noted. Will not commit migration files.

I believe we still need to commit them during the development process, right @benjaoming? Otherwise the code will fail when other devs update and run? The idea is that prior to release, we collapse down all new ones into a single migration per app?

jamalex:
Note that @Antrikshy and I have been playing with a simpler (but similar) set of models for now, for the content curation work:
https://github.com/fle-internal/content-curation/pull/1/files

But for @MCGallaspy's work here, I think the first critical thing to do is still some basic benchmarking, to see how loading these data structures from SQLite (instead of JSON) impacts performance.

MCGallaspy:
> But for @MCGallaspy's work here, I think the first critical thing to do is still some basic benchmarking, to see how loading these data structures from SQLite (instead of JSON) impacts performance.

It's not going to be useful to benchmark memory use unless the json stuff is replaced, in which case it makes sense I think to use models that we intend to keep. If we'd just like to benchmark, say, time for API requests then this could be approximately useful.

MCGallaspy:
At any rate @jamalex, I was trying to come up with meaningful comparisons. No real surprises. I created a tastypie endpoint for the TopicTreeNode resource and profiled `obj_get_list` vs `get_topic_tree`. The speeds are basically incomparable for the first time either one is called -- in the case of `get_topic_tree`, since it needs to load the whole topic tree into memory first, the ORM endpoint blows it out of the water. Subsequent requests are comparable, with the existing JSON stuff being marginally faster. It seems a little premature to measure memory use at this point, but given that right now we simply load *all* the entries of the topic tree into memory, I expect using the ORM can only improve it.

benjaoming:
Creating intermediate models will mean that later updates have to do both schema and data migration.. if we can keep things in the right structure, we don't have to migrate data, we can just expand the schema.. I think that's better.

Great you did benchmarks, @MCGallaspy... they may sound pretty obvious, but it's nice to be able to communicate the improvement to observers of the project :)

MCGallaspy:
>  if we can keep things in the right structure

@benjaoming what do you mean by that exactly? Do you mean we should limit ourselves to only adding fields to models?

benjaoming:
@MCGallaspy by expanding the schema, I mean adding new fields and models... but not changing existing fields and models, thus having to migrate data... this might not even be trivial when you have a tree full of special fields for maintaining tree structures.

jamalex:
> I created a tastypie endpoint for the TopicTreeNode resource and profiled obj_get_list vs get_topic_tree. The speeds are basically incomparable for the first time either one is called -- in the case of get_topic_tree, since it needs to load the whole topic tree into memory first, the ORM endpoint blows it out of the water. Subsequent requests are comparable, with the existing JSON stuff being marginally faster.

Sounds like you're comparing speeds for retrieving the entire topic tree, is that right?

I think a useful benchmark would just be comparing the exercise data endpoint (http://demo.learningequality.org/api/exercise/division_1/)
with a similar endpoint, drawing from the ORM. No need to do anything with MPTT or fancy schemas for this benchmark; I just want to make sure we're not shooting ourselves in the feet with this approach (in terms of API load times), so we can consider other options before we get too deep.

> profiled `obj_get_list`

Note that this won't be including the serialization steps, and, depending on how you implemented this, it might not even be retrieving the actual data from the database (since Django ORM querysets ae lazy). So I think it's *important to do a simple, but full-stack (API curl speeds), benchmark of loading single entries from the JSON vs from the DB*.

Example of how `obj_get_list` loads a lazy queryset, without actually looking up the data:

```
    def get_object_list(self, request):
        user_id = request.GET.get('user_id')
        result = PlaylistProgress.user_progress(user_id=user_id)
        return result

    def obj_get_list(self, bundle, **kwargs):
        self.permission_check(bundle.request)
        return self.get_object_list(bundle.request)
```

MCGallaspy:
> Sounds like you're comparing speeds for retrieving the entire topic tree, is that right?

Right.

> No need to do anything with MPTT or fancy schemas for this benchmark; I just want to make sure we're not shooting ourselves in the feet with this approach (in terms of API load times), so we can consider other options before we get too deep.

Noted, I'll create something real quick.

Closing this PR, as we're not likely to use these models.

PR #3614, opened 2015-05-06
Create a management command for setting up in-context translations mode.

aronasorman:
This does a few things:
1. Downloads our pseudo-language
2. Adds it to our locale dir
3. Compiles the po files into mo files.

Will only run when IN_CONTEXT_LOCALIZED = True.


aronasorman:
Internal tool, and only meant to be run on our in-context translations site. Hence the lack of tests.

aronasorman:
Passes on travis. Will merge now, so I can redeploy the in-context translations site.

PR #3628, opened 2015-05-11
turn VERSION_INFO into a function.

### LINKED ISSUE #3562: Move version.py version metadata to YAML

aronasorman:
Fixes #3562. So apparently upon installation, data files (things in
data/) and py files are placed in separate directories. So importing
version.yaml relative to version.py would break things upon install.

The solution is to turn VERSION_INFO into a function, and call
django.conf.settings when VERSION_INFO's value is needed. That would
avoid circular imports and allow us to use settings.CONTENT_DATA_PATH,
which is where the data files are defined.

### DIFF
--- a/kalite/control_panel/views.py
+++ b/kalite/control_panel/views.py
@@ -586,7 +586,7 @@ def local_install_context(request):
 
     return {
         "software_version": current_version,
-        "software_release_date": VERSION_INFO[current_version]["release_date"],
+        "software_release_date": VERSION_INFO()[current_version]["release_date"],
         "install_dir": settings.SOURCE_DIR if settings.IS_SOURCE else "Not applicable (not a source installation)",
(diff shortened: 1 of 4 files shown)


benjaoming:
Yes, looking great... gonna wait for Travis...

benjaoming:
Oh Travis failed, that's because of the non-thread safe Django management utility... I have a fix for that.. gonna make a PR soon and then we can merge that and see if it works :)

PR #3629, opened 2015-05-11
Minimal benchmark stuff

MCGallaspy:
Just a proof of concept that getting assessment items from the ORM is not crazy slow. Is this what you were getting at @jamalex? How to test this out:

1. Migrate your db, then run the `add_random_db_stuff` management command to initialize the db.
2. Visit exercise pages and look at the response time for the `api/assessment_item/` endpoint -- generally between 15 and 30ms for me.
3. Visit various `api/orm_assessment_item/<pk>`s, where `<pk>` is just an integer between 0 and 4999. Note the load time again; generally between 20 and 50ms for me.

So the ORM (as expected) is marginally slower but probably not enough to be the bottleneck on any given page. Should I be a little more systematic about this, or do we agree that we're not shooting ourselves in the foot?

Also, don't merge this. It's for discussion only.

### DIFF
--- a/data/version.yml
+++ b/data/version.yml
@@ -1,4 +1,4 @@
-0.14.dev:
+0.14.dev0:
   release_date:
   git_commit: ""
   new_features:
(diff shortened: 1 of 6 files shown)


MCGallaspy:
This has served its purpose.

PR #3630, opened 2015-05-11
Installer test

### LINKED ISSUE #3616: "Unknown command: kaserve"
Django 1.5 (don't know about later versions) unfortunately sets a global variable `_commands` which is causing problems when running cronograph as a thread. You can read a bit of background in the below code snippet.

I'm figuring out a solution after having disabled another piece of code that avoided the race condition in the first place.

```python
def manage(command, args=[], in_background=False):
    """
    Run a django command on the kalite project

    :param command: The django command string identifier, e.g. 'runserver'
    :param args: List of options to parse to the django management command
    :param in_background: Creates a thread for the command
    """
    # TODO(benjaoming): Why was this necessary? Had to be commented out because
    # get_commands imports django.conf.settings which should not be imported
    # outside of the command itself as DJANGO_SETTINGS_MODULE could have been
    # set from command line options.
    #
    # Follow-up May 6th: Apparently there was a kind race-condition because the
    # list of possible commands isn't thread safe... so once we have started
    # a background thread for command 1, and command 2 runs on top of that,
    # command 2 may fail saying for instance "Unknown command: 'kaserve'"
    # because the list of command was not ready-built.
    #
    # Best solution: Have this call not return until first entry thread has
    # finished building list of commands with get_commands()
    #
    # Original comment:
    # Ensure that django.core.management's global _command variable is set
    # before calling commands, especially the ones that run in the background
    #
    # get_commands()
```

@MCGallaspy I just wanted to loop you in on this, but I'll give a solution to this as soon as I'm back in the chair.

To get around it quickly, you can comment back in get_commands, but be warned that it will break when parsing a custom `--settings` option.

benjaoming:
This PR adds installer testing for Travis and removes the threaded execution of cronserver, being replaced by a subprocess.Popen, same on all platforms.

This in turn fixes #3616

@aronasorman do you known why the previous call for Windows contained `creationflags=CREATE_NEW_PROCESS_GROUP` ? I removed it because I couldn't find anything on Google explaining why it was needed (it doesn't work on other platforms, so I wanted something that could be similar on all).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,18 @@ before_install:
 install:
 -  npm install
 -  npm install -g grunt-cli
+-  pip install .
 -  pip install -r requirements_dev.txt
 -  pip install coveralls
 script:
+# Test that we can run from source directory
 - bin/kalite start --traceback -v 2
 - bin/kalite status
 - bin/kalite stop --traceback -v 2
+# Test that we can run as the installed script
+- kalite start --traceback -v 2
+- kalite status
+- kalite stop --traceback -v 2
 - python python-packages/fle_utils/tests.py
 - coverage run --source=kalite --omit="kalite/testing/*,*/tests/*" manage.py test --traceback -v 2
 - grunt jshint
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ before_install:
 install:
 -  npm install
 -  npm install -g grunt-cli
--  pip install .
+-  pip install -r requirements.txt
+-  python setup.py install
+#-  pip install .
 -  pip install -r requirements_dev.txt
 -  pip install coveralls
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,7 @@ before_install:
 install:
 -  npm install
 -  npm install -g grunt-cli
--  pip install -r requirements.txt
--  python setup.py install
-#-  pip install .
+-  pip install .
 -  pip install -r requirements_dev.txt
 -  pip install coveralls
 script:
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -326,7 +326,7 @@ def manage(command, args=[], in_background=False):
 
     :param command: The django command string identifier, e.g. 'runserver'
     :param args: List of options to parse to the django management command
-    :param in_background: Creates a thread for the command
+    :param in_background: Creates a new process for the command
     """
     
     if not in_background:
@@ -336,11 +336,18 @@ def manage(command, args=[], in_background=False):
         utility.prog_name = 'kalite manage'
         utility.execute()
     else:
-        # Create a new subprocess, it will die together with the main process
-        # which is fine.
+        # Create a new subprocess, beware that it won't die with the parent
+        # so you have to kill it in another fashion
+        
+        # If we're on windows, we need to create a new process group, otherwise
+        # the newborn will be murdered when the parent becomes a daemon
+        if os.name == "nt":
+            kwargs = {'creationflags': subprocess.CREATE_NEW_PROCESS_GROUP}
+        else:
+            kwargs = {}
         subprocess.Popen(
             [sys.executable, os.path.abspath(sys.argv[0]), "manage", command] + args,
-            # creationflags=CREATE_NEW_PROCESS_GROUP
+            **kwargs
         )
 
 
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -53,8 +53,6 @@ from __future__ import print_function
 # DO NOT IMPORT BEFORE THIS LIKE
 import sys
 import os
-import atexit
-import signal
 
 # KALITE_DIR set, so probably called from bin/kalite
 if 'KALITE_DIR' in os.environ:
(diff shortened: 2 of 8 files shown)


MCGallaspy:
> do you known why the previous call for Windows contained creationflags=CREATE_NEW_PROCESS_GROUP

I added it. I found that killing the `cronserver` process in Windows caused all processes in the process group to be killed, and during startup the `cronserver` process would inevitably be killed for some reason. See [this article on CREATE_NEW_PROCESS_GROUP](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682083(v=vs.85).aspx). Unfortunately, I think we have to anticipate some OS specific behavior here. :(

.travis.yml
```diff
@@ -10,12 +10,18 @@ before_install:
...
 -  npm install -g grunt-cli
+-  pip install .
 -  pip install -r requirements_dev.txt
 -  pip install coveralls
 script:
+# Test that we can run from source directory
 - bin/kalite start --traceback -v 2
 - bin/kalite status
 - bin/kalite stop --traceback -v 2
+# Test that we can run as the installed script
```
MCGallaspy: What is this testing? Instead of calling all three commands again, can we just call one command (like `status`) since `start` is pretty slow.

kalitectl.py
```diff
@@ -345,12 +336,12 @@ def manage(command, args=[], in_background=False):
         utility.prog_name = 'kalite manage'
         utility.execute()
     else:
-        if os.name != "nt":
-            thread = ManageThread(command, args=args, name=" ".join([command] + args))
-            thread.start()
-        else:
-            # TODO (aron): for versions > 0.13, see if we can just have everyone spawn another process (Popen vs. ManageThread)
-            Popen([sys.executable, os.path.abspath(sys.argv[0]), "manage", command] + args, creationflags=CREATE_NEW_PROCESS_GROUP)
+        # Create a new subprocess, it will die together with the main process
```
MCGallaspy: What do you mean, `it will die together with the main process`?

MCGallaspy: As opposed to what? Is this in reference to `CREATE_NEW_PROCESS_GROUP`? See my earlier comments on why I feel it's necessary in Windows.

.travis.yml
benjaoming: It's a test that they don't return with a non-zero failure. I'm contemplating doing a Tox.ini to have more clarity. Is it the word "test" that annoys you because it's not a TestCase instance? :)

kalitectl.py
benjaoming: Sorry, that was an old comment from an experiment with `atexit.register` which unfortunately made it die with the wrong parent (after forking) so I had to drop it. I'll fix the comment...

.travis.yml
MCGallaspy: No, I wonder why you're calling the same commands twice, especially given that `start` is particularly slow.

MCGallaspy: If you just want to test the the script is installed (and thus executable from any directory) why not just try one (fast) command? Unless you have reason to believe that it may or may not start depending on how its called.

MCGallaspy:
Though I was skeptical that `bin/kalite start` and `kalite start` would have different results, that is indeed the case here:
```sh
  File "/home/travis/virtualenv/python2.7.9/share/kalite/python-packages/django/core/management/base.py", line 224, in run_from_argv
  File "/home/travis/virtualenv/python2.7.9/share/kalite/python-packages/django/core/management/base.py", line 263, in execute
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/kalite/distributed/management/commands/kaserve.py", line 127, in handle
    self.setup_server_if_needed()
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/kalite/distributed/management/commands/kaserve.py", line 82, in setup_server_if_needed
  File "/home/travis/virtualenv/python2.7.9/share/kalite/python-packages/django/core/management/__init__.py", line 161, in call_command
    return klass.execute(*args, **defaults)
  File "/home/travis/virtualenv/python2.7.9/share/kalite/python-packages/django/core/management/base.py", line 263, in execute
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/kalite/distributed/management/commands/setup.py", line 324, in handle
    shutil.copyfile(os.path.join(src_dir, script_file), os.path.join(dest_dir, script_file))
  File "/opt/python/2.7.9/lib/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/home/travis/.kalite/../scripts/start.sh'
The command "kalite start --traceback -v 2" exited with 1.
0.17s$ kalite status
Failed to start (check logs) (6)
```

benjaoming:
Yup, that's the point here... we have a very young setup.py and installation is very shaky so let's better run a few extra commands. The difference between running `kalite start` and `bin/kalite start` is that one will run from the system installation and the other will run directly from the source tree.

benjaoming:
Apparently the new failure is about one of these deprecated scripts... I'll have a look at that...

benjaoming:
@MCGallaspy I thought a bit more about it... if we want the tests to be faster, maybe we should actually test the installation and running commands from the distributed source, `kalite` not `bin/kalite` -- because devs will use `bin/kalite` and forget to use the distributed command?

MCGallaspy:
@benjaoming I see your point, but I think I was just plain wrong earlier.

> Yup, that's the point here... we have a very young setup.py and installation is very shaky so let's better run a few extra commands.

Your earlier point above makes the most sense to me, let's keep both until we're reasonably confident that one or the other is very stable.

MCGallaspy:
Anyway, I'm happy to merge this. Is this supposed to get in for 0.14?

aronasorman:
@MCGallaspy I believe so, since this fixes a race condition in Django's initialization.

benjaoming:
Great, @MCGallaspy agreed, let's keep this until someone thinks "this is slow and has never exposed any failures" :)

PR #3631, opened 2015-05-12
Update version.yml

MCGallaspy:
So that it's in line with what's expected in version.py.


MCGallaspy:
I wanted there to be a record of this.

PR #3632, opened 2015-05-12
Preload loop

### LINKED ISSUE #3621: Caught in 'preloading loop' when accessing content from external HDD
I came across another unintended behaviour of kalite I wanted to point out:

Branch: master (0.13.1)

When pointing kalite to search for the educational material in a custom location via creating a local_settings.py file in the 'kalite' subfolder kalite sometimes gets stuck in a 'preloading loop' upon trying to start the server. (See attached screenshot -> it keeps re-loading the content forever.)

screenshot 2015-05-08 16 01 42

So far this only happened to me when pointing kalite to an external USB hard drive. Never had any issues with internal memory or USB sticks. But with an old 160GB Seagate disk I can reproduce this behaviour repeatedly. 

What is more: As soon as kalite got stuck in this boot loop once, it subsequently will always get stuck even if the local_settings.py is removed or points back to the standard content folder. Reinstalling the server and keeping the database (kalite manage setup, keep database) does not fix the issue but only a clean install of the database will (kalite manage setup, delete database). 

Further it seems to make a difference whether kalite is pointed to the custom folder during the first start of a new installation already or only at a later point. When inserting the custom locale_settings.py before firing the server up on a new installation for the first time it always worked so far. When starting it once with the standard folder, then stopping the server, inserting/changing the locale_settings.py to external HDD and starting the server again it is likely to get stuck. It also indicates the successful identification of xyz new videos still (so the folder and its contents are read correctly) but then gets stuck in the loop again.

Not a big deal for me since I don't usually change the content location anymore once a setup is completed. But could create trouble for some since it seems to render the database unusable after the loop occurred once.

(Running on a MAC with Yosemite 10.3.3, Python 2.7.9 and kalite 0.13.1)

MCGallaspy:
Fixes #3621. Removes duplicate definitions in `kalite.caching` and `kalite.updates`. Disconnects `pre_delete` callback for the duration of the offending section in `videoscan` command, where potentially a lot of deletions are made. Ensures that `caching.invalidate_all_caches` _is_ called once if a video is deleted. Plus a test!

### DIFF
--- a/kalite/updates/management/commands/videoscan.py
+++ b/kalite/updates/management/commands/videoscan.py
@@ -4,6 +4,7 @@ import glob
 import os
 from optparse import make_option
 
+from django.db.models.signals import pre_delete
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 
@@ -12,6 +13,7 @@ from ...models import VideoFile
 from fle_utils.chronograph.management.croncommand import CronCommand
 from fle_utils.general import break_into_chunks
 from kalite import caching, i18n
+from kalite import updates
 
 
 class Command(CronCommand):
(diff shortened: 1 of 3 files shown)


MCGallaspy:
@rtibbles was assigned arbitrarily. Feel free to pass the torch.

kalite/updates/management/commands/videoscan.py
```diff
@@ -90,13 +92,17 @@ def delete_objects_for_missing_videos(youtube_ids_in_filesystem, videos_marked_a
...
             for chunk in videos_needing_model_deletion_chunked:
                 video_files_needing_model_deletion = VideoFile.objects.filter(youtube_id__in=chunk)
+                deleted_video_ids += [video_file.youtube_id for video_file in video_files_needing_model_deletion]
                 video_files_needing_model_deletion.delete()
-                deleted_video_ids += [video_file.video_id for video_file in video_files_needing_model_deletion]
             if deleted_video_ids:
+                caching.invalidate_all_caches()
                 self.stdout.write("Deleted %d VideoFile models (because the videos didn't exist in the filesystem)\n" % len(deleted_video_ids))
             return deleted_video_ids
+            pre_delete.connect(receiver=updates.invalidate_on_video_delete, sender=VideoFile)
```
aronasorman: That's strange, so signals get triggered when you attach to them after you call `delete`?

MCGallaspy: Do you mean do the signals get queued up and then called? No.

aronasorman:
Looks good enough to me.

MCGallaspy:
This will conflict with #3636, though I think that PR's changes should be preferred. However I won't just close this since it also cleans up unused code elsewhere.

MCGallaspy:
Oops, disregard, I was thinking of a different PR.

PR #3636, opened 2015-05-13
Performance improvements

rtibbles:
Fixes to improve performance, both of load time of KA Lite, and first render of the sidebar.

Summary of changes:
* get_topic_tree now returns a flat topic tree as a list, with each node possessing a parent attribute.
* Topic tree data is now lazily loaded as needed by the client
* Fixes a bunch of auxiliary functions that relied on get_topic_tree returning a hierarchical object
* During exercise and content cache loading, rather than repeatedly pounding the disk, just read the disk once and check the items against an in memory cache of the file system.

@MCGallaspy this may be relevant to your interests.

PR #3639, opened 2015-05-13
Assessmemt item backend

MCGallaspy:
Load the assessment items into the backend instead of loading all of them into memory. Memory profiling to come.

What's inside:
* A new AssessmentItem model in `main` app (and a migration)
* Refactor tastypie resource and topic_tools functions to use the model.
* A management command (`init_assessment_items`) to bootstrap the AssessmentItem table.

How to test it out:
* Run migrations for the `main` app.
* Run the `init_assessment_items` management command. Pretty slow!
* Have fun using KA Lite.

### DIFF
--- a/kalite/main/api_resources.py
+++ b/kalite/main/api_resources.py
@@ -162,11 +162,6 @@ class ExerciseResource(Resource):
 
 
 class AssessmentItemResource(ModelResource):
-
-    item_data = fields.CharField(attribute='item_data')
-    author_names = fields.CharField(attribute='author_names')
-    id = fields.CharField(attribute='id')
-
     class Meta:
         resource_name = 'assessment_item'
         queryset = AssessmentItem.objects.all()
--- a/kalite/main/api_resources.py
+++ b/kalite/main/api_resources.py
@@ -5,7 +5,7 @@ from tastypie.exceptions import NotFound
 from django.conf.urls import url
 from django.conf import settings
 
-from .models import VideoLog, ExerciseLog, AttemptLog, ContentLog
+from .models import VideoLog, ExerciseLog, AttemptLog, ContentLog, AssessmentItem
 
 from kalite.distributed.api_views import get_messages_for_api_calls
 from kalite.topic_tools import get_exercise_data, get_assessment_item_data, get_content_data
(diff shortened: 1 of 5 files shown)


MCGallaspy:
I'm just going to report some findings here.

__Method__:
Set a breakpoint in `kalite.distributed.views.learn`, where I use `guppy.hpy` to inspect the heap.

__Results__:
In `develop` after starting the server and navigating to the learn tab, total memory use was about 137 MB, with the vast majority of that (sorted `byrcs` -- see http://smira.ru/wp-content/uploads/2011/08/heapy.html) unowned `dict`s. Very likely our big cached variables.

Results for this branch incoming.

rtibbles:
I think I may have cleared up some of these other cached vars in my PR #3636. Due to the now flat topic tree structure.

kalite/main/api_resources.py
```diff
@@ -161,27 +161,15 @@ def rollback(self, bundles):
         raise NotImplemented("Operation not implemented yet for exercises.")
 
 
-class AssessmentItem():
+class AssessmentItemResource(ModelResource):
```
rtibbles: Singing "Bye, bye, non-ORM Tastypie..."

kalite/main/api_resources.py
```diff
@@ -161,27 +161,15 @@ def rollback(self, bundles):
...
-        self.sha = kwargs.get('sha')
-
-
-class AssessmentItemResource(Resource):
-
-    kind = fields.CharField(attribute='kind')
     item_data = fields.CharField(attribute='item_data')
     author_names = fields.CharField(attribute='author_names')
-    sha = fields.CharField(attribute='sha')
     id = fields.CharField(attribute='id')
```
rtibbles: I forget - do we even need any of these fields if they are already defined on the Model?

kalite/topic_tools/__init__.py
```diff
@@ -151,7 +152,7 @@ def get_exercise_cache(force=False, language=settings.LANGUAGE_CODE):
                 items = []
                 for item in exercise.get("all_assessment_items","[]"):
                     item = json.loads(item)
-                    if assessmentitems.get(item.get("id")):
+                    if main_models.AssessmentItem.objects.filter(id=item.get("id")).exists():
```
rtibbles: Could we use get_assessment_item_data function here so that everything flows through one function?

kalite/main/api_resources.py
MCGallaspy: The fields determine how the model is serialized, I think. Will double check the docs.

rtibbles:
A few comments. Once addressed, do you have any blockers on merging this?

kalite/topic_tools/__init__.py
MCGallaspy: Sure, it was my hope that using `exists` might be faster, but I think it typically performs the same query as actually getting the row anyway.

kalite/main/api_resources.py
MCGallaspy: It appears non-relational Model fields are found automatically by ModelResources.

MCGallaspy:
@rtibbles I think this is good to go.

PR #3642, opened 2015-05-13
Wraps env['KA_LITE'] with a string

jtamiace:


aronasorman:
To further explain this issue:

On windows, somehow `KALITE_DIR` gets turned into a unicode string instead of a normal string somewhere. `subprocess` doesn't like unicode strings in its env variables, and errors out.

bin/kalite
```diff
@@ -71,7 +71,7 @@ kalite_path = os.path.abspath(kalite_path)
 
 env = {}
 env.update(os.environ.copy())
-env['KALITE_DIR'] = kalite_path
+env['KALITE_DIR'] = str(kalite_path)
```
benjaoming: thanks! could you add a comment, what @aronasorman explained?

aronasorman:
@jtamiace I just noticed that you have other commits in here. We can just make this the PR for what you're doing, or we can cherry pick those out.

aronasorman:
NVM, "Files Changed" says it's only two changes anyway.

benjaoming:
yeah, i don't get it... 6 commits and only 2 small changes... is something misbehaving?

aronasorman:
@benjaoming develop might've been rebased with these changes before, thus changing the commit hash but keeping the set of files changed.

aronasorman:
Oh, why was this closed @jtamiace?

jtamiace:
oops, i had created two different branches for this issue and intended to delete the one i wasn't working off of....... but deleted this one accidentally

benjaoming:
:+1: for the comment

PR #3644, opened 2015-05-14
Buildserver-oriented memory profiling

aronasorman:
The goal of this patch is to introduce lightweight memory profiling that can be run in the build server/in the background. The data will be used to track how our memory usage grows over time as measured by the benchmarking suite. This isn't designed to help pinpoint where memory leaks happen, see `guppy-heapy` and `memory_profiler` for that.

Oh, and with this change I also introduce a more substantial CircleCI config file.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -14,6 +14,7 @@ test:
     - bin/kalite status
     - bin/kalite stop --traceback -v 2
     - python python-packages/fle_utils/tests.py
+    - bin/kalite manage test
     - grunt jshint
     - bin/kalite manage bundleassets
     - ./node_modules/karma/bin/karma start --single-run --browsers PhantomJS
--- a/circle.yml
+++ b/circle.yml
@@ -1,8 +1,6 @@
 machine:
   python:
-    version:
-      - 2.6.8
-      - pypy-2.4.0
+    version: 2.6.8
 
 dependencies:
   override:
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,21 @@
+machine:
+  python:
+    version:
+      - 2.6.8
+      - pypy-2.4.0
+
+dependencies:
+  override:
+    - pip install -r requirements_dev.txt
+    - npm install
+    - npm install -g grunt-cli
+
+test:
+  override:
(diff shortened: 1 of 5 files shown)


aronasorman:
The CircleCI test failures are mostly related to query tests. I hope to remove those soon, and have them measured in a similar way to this (tracked in a CSV rather than hard test failures).

circle.yml
```diff
@@ -0,0 +1,20 @@
+machine:
+  python:
+    version: 2.6.8
```
MCGallaspy: Why not 2.7.9?

kalitectl.py
```diff
@@ -129,16 +132,16 @@ def __init__(self, status_code):
...
     """
     Takes a list of default arguments and overwrites the defaults with
     contents of updates.
-    
+
     e.g.:
-    
-    udpate_default_args(["--somearg=default"], ["--somearg=overwritten"])
+
+    update_default_args(["--somearg=default"], ["--somearg=overwritten"])
```
MCGallaspy: udpate is a perfectly cromulent word!

circle.yml
aronasorman: Unfortunately testing multiple python versions in CircleCI is a bit more involved than Travis, so I went with the lowest version we support for now.

MCGallaspy:
Please add some documentation, regarding intended use and output.

aronasorman:
@MCGallaspy where's the best place to put such documentation? Docstring, dev docs, github wiki?

MCGallaspy:
Probably here: https://github.com/learningequality/ka-lite/tree/develop/sphinx-docs/developer_docs

Be sure to add an entry to `index.rst` as well.

aronasorman:
I should also mention that this will only work for Unix systems.

aronasorman:
@MCGallaspy I created a new file called `profiling.rst`.

MCGallaspy:
Looks good! How seriously should we take failing tests on `circleci` right now?

aronasorman:
Not serious. We should migrate to it once we move away from query tests.

Issue #3584, opened 2015-04-30
Streak bar behaves erratically when logged in as an admin


jamalex:
Branch: master

Expected behavior: Admins should either see the same thing in the streakbar as other users, or not see a streakbar at all. Preferably the latter.

Observed: When logged in as an admin: after answering the first question on an exercise, the streak bar first expands a tiny bit as normal, then after a little bit continues to grow. Later, when you answer another question, it shrinks way back. The checks/X'es also appear and disappear in strange ways.

Causes: Admin users aren't FacilityUsers, so they can't save ExerciseLogs etc. They also have different permissions. We don't save an exercise log for them, but we do try retrieving /api/exerciselog/. As they have global permissions, this returns *all* exercise logs in the system. Very slow, and not what we want.

Recommended fix: We should disable all exerciselog/attemptlog stuff for admins, and hide the progress bar area.

jamalex:
@EdDixon I think this is likely what you were observing during earlier testing, with erratic points stuff.

EdDixon:
@jamalex very possible in my effort to test the installers in so many different environments I did not take the time to generate our test data set or in a lot of cases even take the time to create student and coach accounts (I did some but not all) so this is a very likely cause. I will confirm that real quick here almost done moving everything to the new setup here! Thanks,

EdDixon:
@jamalex this is now confirmed logged in as student we do have the expected behavior and as admin we do not.

aronasorman:
@Antrikshy or @jtamiace would you be able to create a fix for this? We basically just want to hide the streak bar when you're logged in as an admin.

aronasorman:
Darn. Didn't notice that @jtamiace targeted that fix to develop. @jtamiace can you make another PR for 0.13.x?

jtamiace:
oops. sorry didn't know. working on it right now!

PR #3646, opened 2015-05-15
Fixes issue #3584: hides exercise streak bar from super users

jtamiace:
Fix for https://github.com/learningequality/ka-lite/issues/3584

Adds an if conditional to prevent the ExerciseProgressView from initializing for admins.


aronasorman:
Can confirm that there is no streakbar when logged in as admin: 

screen shot 2015-05-15 at 9 51 26 am

Nondetermistic test failures for python 2.6. Merging.

Issue #3558, opened 2015-04-25
Margin issue with the fields at the login modal dialog.


cpauya:
**Branch** develop
**Target** 0.14.x
**Browser** Firefox v36.0.4
**Expected Behavior** The username and password fields should have left and right margins.
**Current Behavior** Not sure if this is intended behavior but currently there's no left and right margins for the fields.  Please see attached screenshot.

![screenshot 2015-04-25 14 18 12](https://cloud.githubusercontent.com/assets/175580/7331879/393a3462-eb56-11e4-860e-acce0481014b.png)

Antrikshy:
I was thinking of taking a look at this myself, unless someone else has grabbed it.

PR #3647, opened 2015-05-15
Finishing what was started in #3459

radinamatic:
Fixed the margins, toned back the has-error state and toned up the focus one as a bonus.

SR popover problems are another can of worms. Turns out NVDA isn't reading them properly, but JAWS is, so it's more of the AT provider issue than markup itself. Have an inkling that 'tooltip' instead of 'popover' might have a better acceptance, but will have to test further. 

Should fix the #3558 also.

### DIFF
--- a/kalite/distributed/static/css/distributed/khan-lite.css
+++ b/kalite/distributed/static/css/distributed/khan-lite.css
@@ -896,11 +896,24 @@ to items you want to have the icon appear with.
     box-shadow: -4px 0 9px 5px #ffa500;
 }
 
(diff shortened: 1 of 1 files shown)


rtibbles:
Tests passed. @radinamatic can you add a quick screenshot here and I'll merge?

radinamatic:
```
SIMPLIFIED_LOGIN = True
```
simplified-login-true

```
SIMPLIFIED_LOGIN = False
```
simplified-login-false

Issue #1739, opened 2014-03-17
Feature Proposal: Redesign of coach reports


rtibbles:
Feature Name: Redesign of coach reports

Team members: 
@rtibbles
@christinepham

Need more!

PR reviewer: @jtamiace or @66eli77 for front end code, @MCGallaspy  or @benjaoming for back end code.

Description:

Refactor and reimplement coach reports to provide a more immediate _and_ deeper experience for the user.

Key features will include:
- Having a consistent user interface and flow through all views onto student data.
- Giving coaches immediately useful information upon landing on their landing page.
- Flagging recent learner activity events through heuristics in order to give coaches data that is most likely to be relevant.
- Leverage attempt logs to track recent user growth as well as overall achievement.
- Leverage attempt logs to provide teachers with the ability to do a deep diagnostic dive into student performance.
- Enable push updating of coach reports for 'live' mode.

This work will begin with a refactor of the existing tabular coach report to be rendered in Backbone/Handlebars and driven by an API endpoint.

The next step will be to implement a new landing page based on the work of @sethachoi and Mikaela Oen. This will be the place for the aggregated recent activity data to be displayed.

Every display of learner data will be amenable to clicking to produce a more detailed display of the user data shown (whether it is learner activity on the aggregate page, learner progress on the tabular report, or items on the real time display).

Milestones:
April 10 - Refined mockups of each of the new interfaces.
April 17 - Submit refactoring of tabular report for PR.
April 24 - PR for implementation of activity aggregation endpoint. First pass implementation of learner data detail endpoint.
May 1 - PR for Coach landing page/aggregate view. Embeddable view for learner data detail.
May 8 - PR for first pass implementation of push API.
May 15 - Opening PRs for live interface.
May 23 - Efficiency testing of aggregation engine complete - target: Run well on a Raspberry Pi.
May 26 - PRs merged.

PR #3649, opened 2015-05-15
Coach report burn down and refactor

rtibbles:
A first pass at fixing #1739.

Summary of changes:
* Refactors tabular view from Django template to Backbone View driven by an API call.
* Adds detail view to allow expanded information on student engagement within tabular view.
* Simplifies and removes all other coach report code (leaves individual student progress reports intact).
* Deletes knowledgemap app.
* Adds handlebars helper to render django URLs in handlebars templates (accepts arguments in sequence, as per django-js-reverse function).

More to come.


ka-lite
```diff
@@ -0,0 +1 @@
+Subproject commit 1c9703a4032410d91a515a1eda913728b0d83ac1
```
aronasorman: What's this file for? A submodule?

rtibbles: No idea. Will remove.

aronasorman:
Can you give a summary of why we burned down coachreports, and what this PR replaces it with?

rtibbles:
I will. As I said, this is up for early perusal, pending completion. Will
have it described at end of day today.

python-packages/fle_utils/handlebars/static/js/handlebars/handlebars-helpers.js
```diff
@@ -58,4 +58,38 @@ Handlebars.registerHelper ('truncate', function (str, len) {
...
+});
+
+// from http://stackoverflow.com/questions/19646244/handlebars-js-access-object-value-with-a-variable-key
+Handlebars.registerHelper('withItem', function(object, options) {
+    return options.fn(object[options.hash.key]);
+});
+
+// A little bit of magic to let us use Django JS Reverse directly from inside a Handlebars template
+// Simply pass any arguments that you might otherwise use in order
+Handlebars.registerHelper('url', function(url_name) {
```
rtibbles: @jamalex I think you expressed an interest in this helper being created.

kalite/coachreports/api_views.py
```diff
@@ -1,382 +1,87 @@
...
-        for data_type in (data_types if not hasattr(data_types, "lower") else [data_types]):  # convert list from string, if necessary
-            if data_type in data[data.keys()[0]]:  # if the first user has it, then all do; no need to calc again.
-                continue
+    if page*limit < len(learners):
 
-            #
-            # These are summary stats: you only get one per user
-            #
-            if data_type == "pct_mastery":
+        learners = learners[(page - 1)*limit: page*limit]
```
MCGallaspy: What's the intent here? If the requested elements exceed the length of the list, then you reduce the page count... but in principle isn't it possible that the expression `(page - 1)*limit` will still exceed the length of the `learners` list?
If the intent is to return the last page, then perhaps you should use negative list indices in your slice.

kalite/distributed/baktests/__init__.py
```diff
@@ -0,0 +1,5 @@
+from ajax_messages_test import *
```
MCGallaspy: What are `baktests`?

rtibbles: Ugh, some files seem to have snuck in. Will have to do a clean up.

kalite/coachreports/api_views.py
```diff
@@ -1,382 +1,87 @@
...
-    {"key": "ex:attempts",        "name": _("Attempts"),   "type": "number", "description": _("Number of times submitting an answer to an exercise.")},
-    {"key": "ex:streak_progress", "name": _("Streak"),     "type": "number", "description": _("Maximum number of consecutive correct answers on an exercise.")},
-    {"key": "ex:points",          "name": _("Exercise points"),    "type": "number", "description": _("[Pointless at the moment; tracks mastery linearly]")},
-    { "key": "ex:completion_timestamp", "name": _("Time exercise completed"),"type": "datetime", "description": _("Day/time the exercise was completed.") },
-    {"key": "vid:points",          "name": _("Video points"),      "type": "number", "description": _("Points earned while watching a video (750 max / video).")},
-    { "key": "vid:total_seconds_watched","name": _("Video time"),   "type": "number", "description": _("Total seconds spent watching a video.") },
-    { "key": "vid:completion_timestamp", "name": _("Time video completed"),"type": "datetime", "description": _("Day/time the video was completed.") },
-]
+@require_admin
+def learner_logs(request):
```
MCGallaspy: Tests would be good! I think you can use the ClientTestCase here without having to use all the selenium machinery.

kalite/coachreports/static/js/coachreports/coach_reports/models.js
```diff
@@ -0,0 +1,32 @@
...
+var FacilityModel = Backbone.Model.extend();
+
+var GroupModel = Backbone.Model.extend();
+
+var FacilityCollection = Backbone.Collection.extend({
+    url: FACILITY_RESOURCE_URL
+});
+
+var GroupCollection = Backbone.Collection.extend({
+    url: GROUP_RESOURCE_URL
```
MCGallaspy: django-js-reverse is in `develop` now, yes? Can you use it here?

kalite/coachreports/static/js/coachreports/coach_reports/router.js
```diff
@@ -0,0 +1,39 @@
+CoachReportRouter = Backbone.Router.extend({
+    initialize: function(options) {
+        _.bindAll(this);
+        this.state_model = new Backbone.Model();
+        this.listenTo(this.state_model, "change", this.set_url);
+    },
+
+    routes: {
+        "(:facility/)(:group/)(:report/)":    "report"
```
MCGallaspy: What's the intent here? The documentation suggests to me that, if `:facility/` is absent, then a url like `somegroup/somereport/` will result in `facility == "somegroup"` and `group == "somereport"`.

Can optional parameters be nested? If so I think `(:facility(/:group(/:report/)))` might be what's intended. If not, you can use [regular expressions in routes](http://backbonejs.org/#Router-route) to achieve the same thing.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+        this.group_select_view = new GroupSelectView({model: this.model});
+
+        this.render();
+    },
+
+    render: function() {
+        this.$el.html(this.template());
+        this.$('#group-select-container').append(this.group_select_view.$el);
+        this.$('#facility-select-container').append(this.facility_select_view.$el);
+        this.tabular_view = new TabularReportView({model: this.model});
```
MCGallaspy: Why do you instantiate the other two subviews in `initialize` but instantiate this one here?

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+            tagName: 'td',
+            model: this.model,
+            content_item: this.content_item,
+            attributes: {colspan: this.contents_length + 1}
+        });
+        this.$el.append(this.detail_view.el);
+    }
+})
+        
+var detailsPanelView = BaseView.extend({
```
MCGallaspy: `detailsPanelView` -> `DetailsPanelView`?

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+    },
+    
+    render: function() {
+        this.$el.html(this.template({
+            collection: this.collection.to_objects()
+        }));
+    }
+});
+
+var TabularReportView = BaseView.extend({
```
MCGallaspy: Lots of different views here, so for clarity I think it would be useful to explicitly spell out the hierarchy of views in this document, like
```yml
CoachReportView:
  - FacilitySelectView
  - GroupSelectView
  - TabularReportView:
    - [subviews go here]
```

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+});
+
+var FacilitySelectView = Backbone.View.extend({
+    template: HB.template('coach_nav/facility-select'),
+
+    initialize: function() {
+        _.bindAll(this);
+        this.facility_list = new FacilityCollection();
+        this.facility_list.fetch();
+        this.listenTo(this.facility_list, 'sync', this.render);
```
MCGallaspy: What happens in the case of an `error` event (the other possibility from `.fetch()`)? Is there a sensible default in the template, since this element won't be rendered in that case?

MCGallaspy: Also, you might wish to register the listener prior to `fetch`ing.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+    }
+});
+
+var GroupSelectView = Backbone.View.extend({
+    template: HB.template('coach_nav/group-select'),
+
+    initialize: function() {
+        this.group_list = new GroupCollection();
+        this.fetch_by_facility();
+        this.listenTo(this.group_list, 'sync', this.render);
```
MCGallaspy: Same as above, you should register the listener before `fetch`ing, and what happens on the `error` event?

kalite/coachreports/static/js/coachreports/coach_reports/views.js
MCGallaspy: Actually, since you call `render` anyway, it's probably irrelevant when you add this listener.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+        }));
+    }
+});
+
+var TabularReportView = BaseView.extend({
+
+    template: HB.template("tabular_reports/tabular-view"),
+
+    initialize: function() {
+        _.bindAll(this);
```
MCGallaspy: If we ever update underscore, then this function [requires another argument](http://underscorejs.org/#bindAll).

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+            facility: this.model.get("facility"),
+            group: this.model.get("group")
+        });
+        if (this.model.get("facility")) {
+            this.data_model.fetch().then(function() {
+                self.learners = new Backbone.Collection(self.data_model.get("learners"));
+                self.contents = new Backbone.Collection(self.data_model.get("contents"));
+                for (var i = 0; i < self.learners.length; i++) {
+                    self.learners.models[i].set("logs", _.object(
+                        _.map(_.filter(self.data_model.get("logs"), function(log) {
```
MCGallaspy: Such functional js! :+1:

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+
+        this.contents = options.contents;
+        this.render();
+    },
+
+    render: function() {
+        this.$el.html(this.template(this.model.attributes));
+
+        var cell_views = [];
+        for (var i = 0; i < this.contents.length; i++) {
```
MCGallaspy: Just a matter of taste, but you could use (_.each)[http://underscorejs.org/#each] here, I think it's even a Collection method.

kalite/coachreports/api_views.py
rtibbles: If you over count the end of an index range (i.e. if page*limit is bigger than the length of the list) it just returns everything to the end of the list. For the sake of simplicity, I left it like this so that we could use the same logic regardless of which page was being requested.

kalite/coachreports/api_views.py
rtibbles: Yes, tests need to be created or modified for pretty much this whole PR.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
```diff
@@ -0,0 +1,343 @@
...
+            var new_view = this.add_subview(TabularReportRowCellView, {model: new Backbone.Model(data)});
+            cell_views.push(new_view);
+            this.listenTo(new_view, "detail_view", this.show_detail_view);
+        }
+        this.append_views(cell_views);
+    },
+
+    show_detail_view: function(model) {
+        if (this.detail_view) {
+            // TODO (rtibbles): Implement Models properly here to reflect server side id attributes
```
MCGallaspy: What do you mean by this comment?

kalite/coachreports/static/js/coachreports/coach_reports/models.js
rtibbles: It is. Unfortunately, these are TastyPie URLs which do not play nicely with django-reverse-js :(

@jamalex Another argument for Django RESTful (assuming it plays nicely with Django-reverse-js).

kalite/coachreports/static/js/coachreports/coach_reports/router.js
rtibbles: Given that the arguments are entirely positional, I am not sure what the url "somegroup/somereport/" would really mean. ```(:facility(/:group(/:report/)))``` is ultimately what is intended, but that is how I read the Backbone Router documentation, that ```(:facility/)(:group/)(:report/)``` means just that, precisely because it parses arguments positionally using '/'.

Absent some identifier, and then using regex to match them, would work, but it would make the route matching unnecessarily complicated. The only reason for making facility optional is so that we can direct to the default facility without it being specified in the referring URL.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Because, ultimately, the TabularReportView is not intended to be created on pageload, but only after a user interaction.

@sethachoi is due to add these additional elements today.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: But yes, adding the listener first is probably better. This code is all moved from another file, which got lost in the translation.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Ditto. Not my code, but I get the blame now, and will fix.

kalite/coachreports/static/js/coachreports/coach_reports/router.js
MCGallaspy: Can you add some examples of valid urls? Including one where the `:facility` parameter isn't included?

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Yep, we would probably want to deal with this for the browserify refactor, put a first pass attempt showing the way is visible here:
https://github.com/rtibbles/ka-lite/commit/9148ed54bd2d653015df02d37b92b5af180fad3c

As the computer science platform from Khan Academy uses an updated version of underscore.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: I think this would be a good place to be .each-ing in fact.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
MCGallaspy: Okay, so does that mean that `render` is not intended to be called from `initialize` here? I'm assuming the CoachReportView is created at page load, and since `render` is called from `initialize` then `render` would be called at page load as well.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Almost certainly cleaner. After too much time spent with async.js I had it in my head that this did not deterministically happen in order.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: I am referencing .cid here for the model client id (i.e. the id of the model as set up on the client).

When you define Backbone Models, you can define an ```idAttribute``` property that allows you to set a particular property of the setting object to act as the id that is then used in the Collection.

If we did this with the Django Model id, we could then dynamically update these models on the client side when they changed on the server, by only pushing up the items that had changed. Because they would then be matched on their id property, the Collection could intelligently update the models that changed.

This would then allow us to set up push notifications to the client, making the coach report interface seamlessly realtime.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: To answer your original question, in case of an error, it gets handled by the messages code and nothing displays. It is showing a list of items, so I am not sure there is much meaningful to display otherwise.

kalite/coachreports/static/js/coachreports/coach_reports/router.js
rtibbles: Valid urls:
""
"[some facility_id]/"
"[some_facility_id]/[some_group_id]/"
"[some facility_id]/[some_group_id]/[some_report_type]/"

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Render is intended to be called at pageload, but the TabularReportView instantiation is not ultimately intended to be in the render method.

This should happen as a result of user interaction.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Yep.

kalite/coachreports/static/js/coachreports/coach_reports/views.js
rtibbles: Moar comments needed, definitely.

kalite/coachreports/api_views.py
MCGallaspy: My contention is that `(page - 1)*limit` could exceed the end of the list, in which case this will return the empty list. Is there some guarantee that it won't?

rtibbles: It certainly could if someone asked for a page that did not exist. Would you rather it errored?

MCGallaspy: > Would you rather it errored?

No, I guess I read it that a request which exceeds the page limit should always return the last page. This doesn't, instead it potentially returns nothing. That's OK in my book.

kalite/coachreports/static/js/coachreports/tabular_report/views.js
```diff
@@ -0,0 +1,14 @@
+var TabularReportView = BaseView.extend({
```
MCGallaspy: Duplicate view? Toss this one out?

kalite/include/python2.7
```diff
@@ -0,0 +1 @@
+/usr/include/python2.7
```
MCGallaspy: que?

kalite/coachreports/static/js/coachreports/tabular_report/views.js
rtibbles: Oooh, merge conflictorama.

MCGallaspy:
I'd like to see some tests, and it seems there are a few changes being cooked up even now. Plus some relatively minor comments from me. But otherwise I'm ready to merge this and start bug-bashing.

kalite/coachreports/static/js/coachreports/coach_reports/router.js
MCGallaspy: Okay, you're right. If the only valid route without a facility is the blank route, this is fine.

ka-lite
rtibbles: This doesn't appear to be a file? It's not on my local file system. And it's not 'viewable' here on github.

kalite/coachreports/api_resources.py
```diff
@@ -54,9 +60,9 @@ class PlaylistProgressResource(PlaylistParentResource):
     ex_pct_struggling = fields.IntegerField(attribute='ex_pct_struggling')
     ex_pct_incomplete = fields.IntegerField(attribute='ex_pct_incomplete')
     ex_status = fields.CharField(attribute='ex_status')
-    quiz_exists = fields.BooleanField(attribute='quiz_exists')
-    quiz_status = fields.CharField(attribute='quiz_status')
-    quiz_pct_score = fields.IntegerField(attribute='quiz_pct_score')
+    # quiz_exists = fields.BooleanField(attribute='quiz_exists')
+    # quiz_status = fields.CharField(attribute='quiz_status')
+    # quiz_pct_score = fields.IntegerField(attribute='quiz_pct_score')
```
MCGallaspy: Extra fields? Is there a reason to keep these around?

rtibbles: If we ever want to reinstate the quiz feature in the future. Which, if we are going to explore different mastery models, and use the 'quiz' as the basis for giving people mixed sets of materials they have previously practiced, it could be useful.

rtibbles:
Righto - will split the endpoint off into a new one.

rtibbles:
Comments addressed, apart from tests. Will attempt to get the tests in order in the very nearish future.

rtibbles:
Modified existing tests (including removing some), and added new, very basic API tests. Still needs behave tests, but this is definitely on the way.

PR #3650, opened 2015-05-15
Silence collectstatic so we can read test output easier

benjaoming:
So eventhough it should be a healthy annoyance/punishment to have to read all those static files that are being copied, I kind of can't see test failures without scrolling around for several minutes...

Would you agree?

### DIFF
--- a/kalite/distributed/management/commands/kaserve.py
+++ b/kalite/distributed/management/commands/kaserve.py
@@ -135,7 +135,10 @@ class Command(BaseCommand):
         # them to be started up again as needed.
         Job.objects.update(is_running=False)
 
-        call_command("collectstatic", interactive=False)
+        # Copy static media, one reason for not symlinking: It is not cross-platform and can cause permission issues
+        # with many webservers
+        logging.info("Copying static media")
+        call_command("collectstatic", interactive=False, verbosity=0)
 
         if options['startuplock']:
             os.unlink(options['startuplock'])


kalite/distributed/management/commands/kaserve.py
```diff
@@ -135,7 +135,10 @@ def handle(self, *args, **options):
         # them to be started up again as needed.
         Job.objects.update(is_running=False)
 
-        call_command("collectstatic", interactive=False)
+        # Copy static media, one reason for not symlinking: It is not cross-platform and can cause permission issues
+        # with many webservers
+        logging.info("Copying static media")
+        call_command("collectstatic", interactive=False, verbosity=0)
```
aronasorman: Instead of silencing it for all cases, would it be better to just downgrade the logging level to DEBUG?

benjaoming: This happens in the `kaserve` command, I don't think that users running that command via `kalite start` are expecting this sort of output ever... if you want it, you have the choice of explicitly running `kalite manage collectstatic` but really I think this output is only ever useful for devs.

`collectstatic` is an internal Django command, so we would have to patch Django to control it. But it actually doesn't use the logging system :( It just does stdout.write and checks its own verbosity setting. The stupid thing here is that apparently `call_command` doesn't properly forward this setting so I have to find another solution...

For the part of changing log level, I think we have a different task ahead of us regarding cleaning up the way that we log things so we can introduce

benjaoming: Disregard, it actually worked but `kalite setup` also does collectstatic. I'm wondering if we should remove it from `kaserve`... but that could go in a separate issue.

PR #3556, opened 2015-04-24
Django js reverse

rtibbles:
Adds Django JS Reverse to expose a Urls object in client side code for access to Django URLs in the browser.

Summary of changes:
* Adds django_js_reverse app
* Ports as many Javascript referenced URLs as possible over to the Urls object.

Note: Tastypie messes it all up. Again.

PR #3662, opened 2015-05-18
Inline

tk21:
Inline help system functional. Backbone view tied to "show me how" button that starts the inline tutorial, placed at bottom left of any page that has a narrative_id (currently, this is only going to be on the "Manage Tab" page when signed in as an admin). Backbone model makes request to endpoint which is tied to a django view that spits out JSON of attributes to be selected on a page for the tutorial.

Not sure how to ID the narratives -- use a narrative_id in the templates, or use relative URL as the id...?

TO DO: 
- will change naming conventions to match ka_lites_naming_convention
- figure out why the circle + number of each modal is not sized correctly
- figure out button placement

screen shot 2015-05-18 at 11 08 16 am

screen shot 2015-05-18 at 11 08 20 am

screen shot 2015-05-18 at 11 08 32 am

screen shot 2015-05-18 at 11 08 23 am


aronasorman:
Those screenshots look cool!

kalite/distributed/static/css/distributed/introjs.css
```diff
@@ -0,0 +1,343 @@
+.introjs-overlay {
```
aronasorman: @jtamiace can you review the CSS stuff here?

kalite/distributed/static/js/distributed/exercises/api_resources.py
```diff
@@ -0,0 +1,31 @@
+from django.db import models
+from django.http import HttpResponse
+
+from tastypie import fields
+from tastypie.resources import ModelResource
```
aronasorman: Please follow the import structure outlined [here](https://github.com/learningequality/ka-lite/wiki/Coding-guidelines-and-conventions#imports)

kalite/distributed/static/js/distributed/exercises/api_resources.py
```diff
@@ -0,0 +1,31 @@
...
+
+import yaml
```
aronasorman: After a bunch of research, looks like `PyYAML` can run as a pure python module, given that it's not compiled with `libyaml` bindings. We can force that with `pip install pyyaml --global-option="--without-libyaml"`.

kalite/distributed/static/js/distributed/exercises/api_resources.py
```diff
@@ -0,0 +1,31 @@
...
+    
+    class Meta:
+        resource_name = 'narrative'
+
+    # Fetches an individual object, this is called when
+    # the URL endpoint is matched
+    def obj_get(self, bundle, **kwargs):
+        print bundle.request
+        
+        narr_yaml = open('managetab.yaml')
```
aronasorman: Please turn this into a call inside a `with` statement.

kalite/distributed/static/js/distributed/inline/views.js
```diff
@@ -0,0 +1,95 @@
...
+                var intro = introJs();
+                intro.setOption('tooltipPosition', 'auto');
+                intro.setOption('positionPrecedence', ['left', 'right', 'bottom', 'top']);
+                intro.setOptions(options);
+
+                intro.start();
+            },
+
+            error: function(model, response, options) {
+                console.log("You got an error because this sucks kbye");
```
aronasorman: More informative error string :)

tk21:
@aronasorman  Your comments have been addressed above! A lot of the comments were towards api_resources.py file, which has been deleted.

kalite/distributed/static/css/distributed/introjs.css
```diff
@@ -0,0 +1,343 @@
...
+  text-align: center;
+  text-shadow: 1px 1px 1px rgba(0,0,0,.3);
+  background: #ff3019; /* Old browsers */
+  background: -webkit-linear-gradient(top, #ff3019 0%, #cf0404 100%); /* Chrome10+,Safari5.1+ */
+  background: -webkit-gradient(linear, left top, left bottom, color-stop(0%, #ff3019), color-stop(100%, #cf0404)); /* Chrome,Safari4+ */
+  background:    -moz-linear-gradient(top, #ff3019 0%, #cf0404 100%); /* FF3.6+ */
+  background:     -ms-linear-gradient(top, #ff3019 0%, #cf0404 100%); /* IE10+ */
+  background:      -o-linear-gradient(top, #ff3019 0%, #cf0404 100%); /* Opera 11.10+ */
+  background:         linear-gradient(to bottom, #ff3019 0%, #cf0404 100%);  /* W3C */
+  width: 20px;
```
jtamiace: Style bug with current code:
inline number

can be fixed by changing width/height to 25px:
inline number fix

kalite/distributed/static/js/distributed/exercises/api_resources.py
aronasorman: We discussed this in the dev room. We're gonna make a function/command that turns all yml files into json files, and then read those json files when they're available.

kalite/distributed/static/css/distributed/introjs.css
jtamiace: but awesome job otherwise looks good functions really nicely! :D

kalite/inline/api_urls.py
```diff
@@ -0,0 +1,5 @@
+from django.conf.urls import include, patterns, url
+
+urlpatterns = patterns (__package__ + '.api_views',   
+    url(r'^(?P<narrative_id>\w+)', 'narrative_view', name="narrative-view")
```
MCGallaspy: I'm not sure what you decided for the `pk` of narratives, but if you decided to use relative urls then I don't think this regular expression will match them.

MCGallaspy: Check out this neat site: http://pythex.org/

In particular, `^(?P<narrative_id>[\w+/])` should match relative urls, since it includes the `/` character.

kalite/distributed/features/inline-help.feature
```diff
@@ -0,0 +1,37 @@
...
+        When I click the "next" button
+        Then I see a modal with step number 2
+        When I click the "back" button
+        Then I see a modal with step number 1
+
+    Scenario: The page has no intro! Oh no!
+        Given I'm on a page with no intro
+        Then I should not see the starting point
+
+    Scenario: I click outside the modal
```
MCGallaspy: Wow, who wrote these great features?

kalite/distributed/static/css/distributed/introjs.css
MCGallaspy: These are 3rd party definitions. If we want to override them we should add our own styles elsewhere.

kalite/distributed/static/css/distributed/introjs.css
MCGallaspy: Probably best to set this style in our own stylesheet (as long as it's included after this one!) since this is a 3rd party asset.

kalite/distributed/static/js/distributed/exercises/api_resources.py
MCGallaspy: [Like this!](https://docs.python.org/2/library/stdtypes.html#file.close)

MCGallaspy: But it seems you meant to remove this file, right?

kalite/distributed/static/js/distributed/exercises/api_resources.py
MCGallaspy: So who is writing this function? @tk21? @aronasorman? me?

kalite/distributed/features/inline-help.feature
benjaoming: Yeah, this is great stuff, should be a school example of how to write BDD! :+1:

kalite/distributed/static/js/distributed/inline/models.js
```diff
@@ -0,0 +1,9 @@
+// Model for the inline tutorial system that fetches from tastypie resource "NarrativeResource"
+
+var NarrativeModel = Backbone.Model.extend({
+    // urlRoot since we are using a model outside of a backbone collection,
+    // enables the url function to generate URLs based on the model id.
+    urlRoot: function() {
+        return '/api/inline/narrative';
```
MCGallaspy: @rtibbles has an open PR to expose django urls in js files by their name (#3556). Once that's merged we should use it here, in case the url changes down the line. More robust that way!

kalite/distributed/static/js/distributed/exercises/api_resources.py
aronasorman: I'll write them tomorrow. Feel free to do so if you need it earlier.

kalite/distributed/static/js/distributed/inline/models.js
MCGallaspy: Okay, I just merged it. If you pull the latest `develop` here you can use the `window.Urls`, which has callable attributes corresponding to url names (which you can inspect by using the `show_urls` management command). Is that right @rtibbles?

kalite/distributed/static/js/distributed/inline/views.js
```diff
@@ -0,0 +1,95 @@
...
+                key = key[0];
+                var value = attribute[key];
+                var newkey = null;
+                var newvalue = null;
+
+                if (key === "text") {
+                    newkey = "intro";
+                    newvalue = value;
+                }
+                else if (key === "position") {
```
MCGallaspy: I was looking at the [spec](https://docs.google.com/document/d/11wY1EViZrCKXZkus-7Pf4RMCbEYuaTNSCRa5YCglMMU/edit):

__Description of intro YAML format:__
Each url pattern has a list of jQuery-style selectors. Modals will be attached to each selector. Each selector has a list of attributes which determine the content and presentation of the modal. Valid attributes include:
* step: An integer. Should be sequential. Required.
* text: The text presented in the modal. Required.
* before-showing: Optional. User actions taken to make an element visible. A list of key-value pairs. Supposed to be extensible, in case we find we need more actions. Valid keys are:
  * click: Clicks an object specified by the value. Value is a jQuery-style selector.
  * scroll-to: An integer. Value is x-position on the page that should be scrolled to.

Which of the attributes we discussed corresponds to `position` here? We should also implement `before-showing`. I can't recall if we had another discussion about this and I failed to update the spec, if that's the case please update the spec.

kalite/distributed/static/js/distributed/inline/views.js
```diff
@@ -0,0 +1,95 @@
...
+        });
+
+        options["steps"] = steps;
+        console.log(options);
+        return options;
+    },
+
+    render: function() {
+        this.$el.html(this.template());
+        $("body").append(this.el);
```
MCGallaspy: Like we discussed, the element should be placed somewhere else. You could potentially create a block in the base template with an element to attach the button to. Then the block can be placed elsewhere in other templates if necessary.

I think @jamalex also suggested that the button should be invisible or deactivated if no narrative is found, right? In that case, one would need to fetch the model right away (like in `initialize`) in order to determine the state of the button. And we should probably add a test for that feature.

kalite/distributed/static/js/distributed/inline/views.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+});
+
+// Only load button and narrative if there is one defined for page
+$(function() {
+    if (window.NARRATIVE_ID) {
```
MCGallaspy: I don't favor this, since it requires the dev to add cruft to the template among other things. Is there a technical problem with using `window.location` (or whatever)?

kalite/distributed/static/js/distributed/intro-injector.js
```diff
@@ -0,0 +1,174 @@
+//on page load
```
MCGallaspy: Remove this file. It's superseded by `inline/views.js` and `inline/models.js`.

kalite/distributed/static/js/distributed/intro.js
```diff
@@ -0,0 +1,1263 @@
+/**
```
MCGallaspy: Would be good to replace this (and the corresponding stylesheet) with a minified version.

kalite/inline/local_settings.py
```diff
@@ -0,0 +1 @@
+DEBUG=True
```
MCGallaspy: Don't commit this.

kalite/inline/api_views.py
```diff
@@ -0,0 +1,18 @@
...
+
+def narrative_view(request, narrative_id):
+    f = open("kalite/inline/narratives.yaml", 'r')
+    narrative_json = yaml.safe_load(f)
+    f.close()
+
+
+    print "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!API_VIEWS!!!!!"
+    print narrative_json
+    return JsonResponse(narrative_json)
```
MCGallaspy: Some work needs to be done here, but it depends on what you choose to use as the `pk` for narratives. The original design suggested that the keys of the narratives data would be patterns against which the url should be matched. For example, one key could be `management/zone/.*`, which would match ` `management/zone/None`,  `management/zone/YourCoolDeviceID`, and others.

MCGallaspy: At any rate, if you choose to implement a different scheme this view will need to pick the appropriate narrative using the `narrative_id` parameter.

kalite/inline/managetab.yaml
```diff
@@ -0,0 +1,19 @@
+intro:
```
MCGallaspy: Let's consolidate all the narratives into one file.

kalite/distributed/features/steps/inline-help.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@then("the modal disappears")
+def step_impl(context):
+    assert elem_is_invisible_with_wait(context, context.modal), "The modal should not be visible."
+
+
+@then("the back button is disabled")
+def step_impl(context):
+    assert not modal.back_button.is_enabled(), "The back button should be disabled!"
```
MCGallaspy: Looking at the tests failures, I made some oopsies. This line should be:
`assert not context.modal.back_button.is_enabled(), "The back button should be disabled!"`

kalite/distributed/features/steps/inline-help.py
```diff
@@ -0,0 +1,118 @@
...
+    try:
+        context.browser.find_element_by_id(STARTING_POINT_ID)
+        assert False, "There should not be a starting point element for the intro!"
+    except NoSuchElementException:
+        pass
+
+
+@when("I click outside the modal")
+def step_impl(context):
+    elem = find_css_class_with_wait(STEP_NUMBER_CLASS)
```
MCGallaspy: This should be `elem = find_css_class_with_wait(context, STEP_NUMBER_CLASS)`

MCGallaspy:
The comments here have been addressed in #3717 so this PR is truly superseded, although this feature is not complete as per the specification. More info found in #3717.

PR #3664, opened 2015-05-18
Created backbone view for the navigation bar

jtamiace:
Looks the same as the current one, so no screenshots necessary.

The view for the nav bar listens to the statusModel for they type of user (student, teacher, guest, etc.) and displays the appropriate tabs with appropriate activation according to which page the user is on. 
URL's are currently hardcoded.

Will be adding a few commits shortly for cleaning up the no-longer-needed django blocks on templates that used to activate the tabs, but just wanted the put this PR now since the code is functional.

### DIFF
--- a/kalite/distributed/static/js/distributed/distributed-server.js
+++ b/kalite/distributed/static/js/distributed/distributed-server.js
@@ -194,6 +194,20 @@ var StatusModel = Backbone.Model.extend({
         points = points || 0;
         // add the points that existed at page load and the points earned since page load, to get the total current points
         this.set("points", this.get("points") + points);
+    },
+
+    pageType: function() {
+
+        if ( window.location.pathname.search(Urls.coach_reports()) > -1 ) {
+            return "teachPage";
+        } 
+        if ( window.location.pathname.search(Urls.learn()) > -1 ) {
+            return "learnPage";
+        } 
+        if ( window.location.pathname.search(Urls.zone_redirect()) > -1 || window.location.pathname.search("/update/") > -1 ) {
+            return "managePage";
+        }
+        
     }
 
 });
--- a/kalite/distributed/static/js/distributed/distributed-server.js
+++ b/kalite/distributed/static/js/distributed/distributed-server.js
@@ -5,7 +5,7 @@
 
 // Functions related to loading the page
 
-function toggle_state(state, status){
+function toggle_state(state, status) {
     $("." + (status ? "not-" : "") + state + "-only").hide();
     $("." + (!status ? "not-" : "") + state + "-only").show();
     // Use display block setting instead of inline to prevent misalignment of navbar items.
@@ -188,7 +188,6 @@ var StatusModel = Backbone.Model.extend({
             toggle_state("admin", self.get("is_admin")); // combination of teachers & super-users
             $('.navbar-right').show();
         });
-
     },
 
     update_total_points: function(points) {
@@ -425,3 +424,19 @@ $(function() {
     });
 
 });
+
+
+/*This function addresses Bootstrap's limitation of having a dropdown menu in an already collapsed menu*/
+function collapsedNav() {
+    var data_toggle = document.getElementById("user-name-a");
+    var menu = document.getElementById("user-name");
+        if ( $('body').innerWidth() <= 750 ) {
+            data_toggle.removeAttribute("data-toggle");
+            menu.classList.add("open");
+      } else {
+            data_toggle.setAttribute("data-toggle", "dropdown");
+            menu.classList.remove("open");
+      }
+};
+window.addEventListener("resize", collapsedNav);
+window.addEventListener("pageshow", collapsedNav);
--- a/kalite/distributed/static/js/distributed/distributed-server.js
+++ b/kalite/distributed/static/js/distributed/distributed-server.js
@@ -441,7 +441,7 @@ $(function() {
 
 
 /*This function addresses Bootstrap's limitation of having a dropdown menu in an already collapsed menu*/
-/*function collapsedNav() {
+function collapsedNav() {
     var data_toggle = document.getElementById("user-name-a");
     var menu = document.getElementById("user-name");
         if ( $('body').innerWidth() <= 750 ) {
@@ -452,5 +452,3 @@ $(function() {
             menu.classList.remove("open");
       }
 };
-window.addEventListener("resize", collapsedNav);
-window.addEventListener("pageshow", collapsedNav);*/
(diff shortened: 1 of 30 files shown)


kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -425,3 +424,19 @@ $(function() {
     });
 
 });
+
+
+/*This function addresses Bootstrap's limitation of having a dropdown menu in an already collapsed menu*/
```
rtibbles: Is this being used? It seems not, so could it just be deleted?

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -314,4 +314,64 @@ window.UserView = BaseView.extend({
...
+
+    render: function() {
+        
+        this.facility_id = {"facility_id": this.model.get("facility_id")};
+        this.$el.html(this.template(this.facility_id));
+        
+        this.userView = new UserView({ model: this.model, el: "#topnav" });
+        $("#topnav").prepend(new AutoCompleteView().$el);
+        
+        // GUEST
```
rtibbles: I feel like all of this could be collapsed into the handlebars template.

Instead of toggling 'display-none' classes, you can just use {{#if is_logged_in}} {{/if}} around navbar items, and just not include them if not needed.

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -314,4 +314,64 @@ window.UserView = BaseView.extend({
...
+        // TEACHER
+        if ( !this.model.get("is_django_user") && this.model.get("is_admin") ) {
+            this.$(".teacher").removeClass("display-none");
+        }
+        // ADMINS
+        if ( this.model.get("is_django_user") ) {
+            this.$(".admin").removeClass("display-none");
+        }
+
+        // ACTIVATES NAV TAB ACCORDING TO CURRENT PAGE
```
rtibbles: I think this should be an attribute on the status model (and then could be handled in the template as above).

Doing it this way means you won't have to worry about a race condition on rendering - whereby if this renders before your ```var onTeachPage``` is set, or whichever, then you might end up with no active tab.

kalite/distributed/templates/distributed/base.html
```diff
@@ -89,22 +89,6 @@
         $(function() {
             FastClick.attach(document.body);
         });
-
-         /* TODO(jtamiace): Backbonify this function and this page... when the time comes.
```
rtibbles: So you Backbonified it by putting it in another file and commenting it out?

rtibbles:
Does the job - but could be more cleanly implemented by better leveraging handlebars. Something also seems to have been lost in translation with regards the collapsing nav stuff.

kalite/distributed/static/js/distributed/user/views.js
jtamiace: Took your suggestion and it works! But in order to get the handlebars template to read conditions like "is_logged_in" I had to get the view to get that info from the statusModel (renamed this.facility_id to this.user_info):

this.user_info = {
            "facility_id": this.model.get("facility_id"),
            "logged_in": this.model.get("is_logged_in"),
            "admin": this.model.get("is_admin"),
            "super_user": this.model.get("is_django_user")
        };

Is this necessary or is there a way to make the if statements without doing this?

rtibbles: Easiest thing to do is just to pass window.statusModel.attributes directly to the template (and use the model's property names in the template).

window.statusModel.attributes gives you a Javascript object, just like you are defining with this.user_info.

kalite/distributed/static/js/distributed/user/views.js
jtamiace: Am I taking the right approach on this? I defined this on the statusModel:

pageType: function() {
        if ( window.location.pathname == "/coachreports/table/" ) {
            return "teachPage";
        } 
        if ( window.location.pathname == "/learn/khan/") {
            return "learnPage";
        } 
        if ( window.location.pathname == "/management/zone/None/" ) {
            return "managePage";
        }

    }

and rewrote this in the ToggleNavbarView:
if ( this.model.pageType() == "teachPage" ) {
            this.$(".teach-tab").addClass("active");
        } else if ( this.model.pageType() == "learnPage" ) {
            this.$(".learn-tab").addClass("active");
        } else if ( this.model.pageType() == "managePage" ) {
            this.$(".manage-tab").addClass("active");   
        }

It works, but this hardcoding doesn't work so well for the manage tab...

rtibbles: This seems like a decent approach.

Instead of using ==, though, you could just check that the start of the URL contains the relevant url:

```
pageType: function() {
if ( window.location.pathname.search("/coachreports/table/") > -1 ) {
return "teachPage";
} 
if ( window.location.pathname.search("/learn/khan/") > -1) {
return "learnPage";
} 
if ( window.location.pathname.search("/management/zone/None/") > -1 ) {
return "managePage";
}
```

However, as you rightly suggest, this hardcoding doesn't work for the management page, (it is also vulnerable to URL changes on the server side). Instead you could use the newly added Urls object to generate the URL dynamically:

```
pageType: function() {
if ( window.location.pathname.search(Urls.coach_reports()) > -1 ) {
return "teachPage";
} 
if ( window.location.pathname.search(Urls.learn()) > -1) {
return "learnPage";
} 
if ( window.location.pathname.search(Urls.zone_redirect()) > -1 ) {
return "managePage";
}
```

jtamiace:
....................... What have I done........ I was just trying to fix a merge conflict I got while trying to push up my finished work D:

kalite/distributed/static/js/distributed/distributed-server.js
jtamiace: I... don't think I'll have time during this cycle to finish this *shamefully uncomments*

kalite/distributed/static/js/distributed/distributed-server.js
```diff
@@ -188,13 +188,26 @@ var StatusModel = Backbone.Model.extend({
...
+
+    pageType: function() {
+
+        if ( window.location.pathname.search(Urls.coach_reports()) > -1 ) {
+            return "teachPage";
+        } 
+        if ( window.location.pathname.search(Urls.learn()) > -1 ) {
+            return "learnPage";
+        } 
+        if ( window.location.pathname.search(Urls.zone_redirect()) > -1 || window.location.pathname.search("/update/") > -1 ) {
```
rtibbles: :+1:

rtibbles:
Test failures are non-sensical/trivial - but you should fix this JS Hint error:
```
   kalite/distributed/static/js/distributed/distributed-server.js
    454 |};
          ^ Unnecessary semicolon.
```

rtibbles:
Good to merge once JS Hint is satisfied.

rtibbles:
Will fix JS Hint error post hoc.

PR #3666, opened 2015-05-18
Sidebar initial size fix

MCGallaspy:
Ensures the sidebar's size is calculated correctly even if the user doesn't resize or scroll the window.


aronasorman:
I couldn't replicate the bug being addressed here. But I'll merge this in the hope that it does the right thing™

MCGallaspy:
Maybe Xmonad fires off a window resize event because of the tiling? Dunno.

PR #3667, opened 2015-05-19
Content ORMing

MCGallaspy:
This is, in fact, unacceptably slow. Probably the `annotations` function on the model. Here we can discuss alternate strategies.


jamalex:
How are speeds without the annotation stuff? Can you post a couple
comparisons with this vs the old approach?

MCGallaspy:
@jamalex starting the server takes forever because in `get_topic_tree` all the nodes are accessed -- which for content nodes in particular means a disk access to check availability. The way we were doing it before was OK, but fetching a row from the DB adds enough overhead to make it very noticeable. I'm going to merge @rtibbles changes (since they're relevant here) and see if that makes it more manageable. Otherwise, we'll need to come up with an alternative to checking the availability of every content node.

In `topic_tools/__init__.py`:
```python
                # By default this is very charitable, assuming if something has not been annotated
                # it is available.
                if node.get("kind") == "Exercise":
                    cache_node = exercise_cache.get(node.get("id"), {})
                else:
                    cache_node = get_content_data(node.get("id")) or {}
                node["available"] = cache_node.get("available", True)
```

MCGallaspy:
Takes about 1s per call to `get_content_data` right now.

jamalex:
Ok, thanks for the details. Could you try running without the availability stamping stuff, too, just to see how much of it came from there and how much came from the DB querying?

But yeah, @rtibbles' stuff for retrieving one topic at a time should be helpful here.

rtibbles:
Note: Accessing the top node still accesses all other nodes on first load,
due to availability calculations. I think @MCGallaspy is referring to the
speed up benefits conferred by only checking the disk once when annotating
all the content.

kalite/topic_tools/__init__.py
```diff
@@ -515,6 +414,9 @@ def get_content_data(request, content_id=None):
         elif not request.is_logged_in:
             messages.warning(request, _("This content was not found! You must login as an admin/coach to download the content."))
 
+    annotations = content.annotations()
+    content.update(annotations)
```
MCGallaspy: With these lines `get_content_data` takes about 1.7s/execution, and without it's still about 0.7s/execution. I counted about 7377 content items, so reducing execution time to about 1ms would put us in the 7s range for `get_topic_tree`.

MCGallaspy:
I'm not positive we'll be able to get node access times down to the ~1ms level that we need for `get_topic_tree` to run in a reasonable time. I think the best approach is to ORM-ify the topic tree and determine availability at setup time, and then update availability on a row-by-row basis as needed, in order to minimize the number of node accesses. In other words,

1. Decrease node access time as much as possible.
2. Decrease number of node accesses as much as possible.

MCGallaspy:
Stale

benjaoming:
What happened to this exactly? Are we never gonna do content ORMing or did we already do it?

rtibbles:
It's now been done in #4261.

PR #3668, opened 2015-05-19
enable user to create super user from browser

66eli77:
the users will be forced to create a superuser because there is no way for them to dismiss this modal till they finish it.
We will check if superuser exists on homepage, if not, this modal will be shown.
I use ajax to POST the superuser creating form, and there is no redirect after submission.


aronasorman:
Just to clarify what we need here: we need Feature tests (Gherkin) and miscellaneous selenium tests.

MCGallaspy:
I'm taking a closer look at this, but just at a glance there's a lot of behave tests for what appears to be a simple feature. It's really good to have comprehensive test coverage, but if possible we should write non-browser based tests to supplement behave tests, since the former are much faster and have less potential for race conditions than the latter. I recommend writing behave tests before development (instead of afterwards) in order to avoid behave test proliferation.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+from django.contrib.auth.models import User
+
+@given("there is superuser")
+def step_impl(context):
+    if User.objects.exists():
+        pass
+    else:
+        User.objects.create_superuser(username='superusername', password='superpassword', email='super@email.com')
```
MCGallaspy: > password='superpassword'

How did you know my password?!

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+
+@then("there should be no modal displayed")
+def step_impl(context):
+    assert not find_id_with_wait(context, "superusercreate-container")
```
MCGallaspy: Please add a helpful assertion message.

MCGallaspy: I also recommend pulling the id `"superusercreate-container"` into a variable, for easy changing.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+@then("the modal will dismiss")
+def impl(context):
+    assert not find_id_with_wait(context, "superusercreate-container")
+    assert User.objects.exists()
+
+
+
+def fill_field(context, text, field_id):
+    field = find_id_with_wait(context, field_id)
+    field.send_keys(text)
```
MCGallaspy: `find_id_with_wait` returns `None` if the element isn't found, right? I think it makes sense to assert that `field is not None` here (with an appropriate message), otherwise an error will be raised by line 131 that will be somewhat less than helpful.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+    assert find_id_with_wait(context, "superusercreate-container")
+
+@given("the username is empty")
+def step_impl(context):
+    fill_username(context, "")
+
+@when('I click the create button')
+def impl(context):
+    create_button = find_css_class_with_wait(context, "create-btn")
+    create_button.click()
```
MCGallaspy: What's the expected outcome of the button click? A page load? A style change? There is a potential race condition if in a subsequent step you don't add a wait for the expected outcome.

If there's more than one possible outcome (like a success or failure) I recommend waiting for one or the other to occur and then asserting the desired outcome, in order to differentiate between success, failure, and unexpected outcomes like timeouts in which neither success nor failure occurs.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+
+@then('the username border will turn red')
+def impl(context):
+    is_border_red(context, "super_username")
+
+@given("I enter a username longer than 30 letters")
+def step_impl(context):
+    fill_username(context, "123456789123456789123456789123456789")
+
+@when('I click the create button')
```
MCGallaspy: You only need to define a step once.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+def impl(context):
+    create_button = find_css_class_with_wait(context, "create-btn")
+    create_button.click()
+
+@then("the modal won't dismiss")
+def impl(context):
+    assert find_id_with_wait(context, "superusercreate-container")
```
MCGallaspy: If the modal is dismissed, is it removed from the DOM? Or is it simple made invisible?

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,145 @@
...
+    create_button = find_css_class_with_wait(context, "create-btn")
+    create_button.click()
+
+@then('the password border will turn red')
+def impl(context):
+    is_border_red(context, "super_password")
+
+@given("I enter a password longer than 30 letters")
+def step_impl(context):
+    fill_password(context, "123456789123456789123456789123456789")
```
MCGallaspy: `"x" * 31`?

kalite/distributed/views.py
```diff
@@ -41,8 +41,9 @@ def check_setup_status_wrapper_fn(request, *args, **kwargs):
             logging.error("Key 'registered' not defined in session, but should be by now.")
 
         #Fix for #2047: prompt user to create an admin account if none exists
```
MCGallaspy: Is the comment still relevant?

kalite/distributed/views.py
```diff
@@ -41,8 +41,9 @@ def check_setup_status_wrapper_fn(request, *args, **kwargs):
...
-        if not User.objects.exists():
-            messages.warning(request, _("No administrator account detected. Please run 'kalite manage createsuperuser' from the terminal to create one."))
+        if User.objects.exists():
+            request.has_superuser = True
+            # messages.warning(request, _("No administrator account detected. Please run 'kalite manage createsuperuser' from the terminal to create one."))
```
MCGallaspy: Down with dead code!

kalite/distributed/urls.py
```diff
@@ -101,6 +101,9 @@
     url(r'^store/', include(kalite.store.urls)),
     # the following pattern is a catch-all, so keep it last:
 
+    # Let user to create the super user from browser
+    url(r'^create_superuser_from_browser/$', 'create_superuser'),
```
MCGallaspy: Can we choose a name more consistent with our other API endpoint names?

MCGallaspy: In fact, it might be beyond the scope of this PR (how urgent is it?) but this could be an opportunity to create more general RESTful API endpoints for creating & modifying users (or at least Django users if not also FacilityUsers).

kalite/distributed/views.py
```diff
@@ -212,6 +213,21 @@ def search(request):
         'category': category,
     }
 
+def create_superuser(request):
```
MCGallaspy: I don't know how we feel about TastyPie these days (@rtibbles, @jamalex), but given that `User` is a Model it should be really easy to create a `UserResource` that handles this kind of thing automatically.

kalite/distributed/features/superuser_create.feature
```diff
@@ -0,0 +1,51 @@
...
+    Scenario: Create superuser with email address longer than 30
+        Given I enter a email address longer than 30 letters
+        When I click the create button
+        Then the modal won't dismiss
+
+    Scenario: Create superuser with correct username and password and email address
+        Given I enter everything correctly
+        When I click the create button
+        Then the modal will dismiss
+        And a superuser is created
```
MCGallaspy: No corresponding step implementation for this?

kalite/distributed/features/superuser_create.feature
```diff
@@ -0,0 +1,51 @@
...
+        Then I should see a modal
+        And I will not be able to dismiss the modal until a superuser is created
+
+    Scenario: Create superuser with empty username
+        Given the username is empty
+        When I click the create button
+        Then the username border will turn red
+        And the modal won't dismiss
+
+    Scenario: Create superuser with string longer than 30
```
MCGallaspy: It's probably good to create one behave test for invalid input conditions, but for the specifics (i.e., password length requirements, valid emails) we can test the view function directly without the extra browser test overhead.

kalite/distributed/features/steps/superuser_create.py
66eli77: haha, as long as we are not using the same superuser name

kalite/distributed/features/superuser_create.feature
66eli77: yes there is one

kalite/distributed/urls.py
66eli77: I have a bad sense of consistent names, any suggestions?

kalite/distributed/views.py
66eli77: do you mean the validation part when you talk about automation? I will refactor this with Django forms which handles the validation for us.

kalite/distributed/urls.py
MCGallaspy: Maybe something like `api/django_user`? In general it seems to be something like `api/<resource_name>` or `api/<app>/<resource_name>`.

kalite/distributed/views.py
MCGallaspy: No, I mean adding urls/views that handle actions like creating and updating models.

kalite/distributed/features/steps/superuser_create.py
66eli77: can you give an example how to add assertion message?

MCGallaspy: You give `assert` a tuple, like this:
```python
assert not find_id_with_wait(context, "superusercreate-container"), "My cool assertion message!"
```

66eli77:
@MCGallaspy I refactored the modal, now using Django forms for better validation. And the feature test has been fixed. The superuser creating modal may cause problem for other feature tests since it will block access to UI if there are no superuser exist. A simple fix is to create a superuser at the beginning of any effected tests.

kalite/distributed/features/superuser_create.feature
```diff
@@ -0,0 +1,53 @@
...
+        When I click the create button
+        Then the username border will turn red
+        And the modal won't dismiss
+
+    Scenario: Create superuser with string longer than 40
+        Given I enter a username longer than 40 letters
+        When I click the create button
+        Then the modal won't dismiss
+
+    Scenario: Create superuser with empty password
```
MCGallaspy: Given that the form has its own validation, I think it should be sufficient to only have one test with bad input. Every bad input scenario ends in the same result (`Then the modal won't dismiss`), so in my opinion not much is gained from testing every possible error condition vs the cost of running a lot more relatively slow Selenium tests.

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -1,6 +1,124 @@
...
+        this.render();
+        $("body").append(this.el);
+    },
+
+    render: function() {
+        this.$el.html(this.template());
+        _.defer(this.addSuperUserForm);
+    },
+
+    addSuperUserForm: function() {
```
MCGallaspy: Let's use a consistent naming scheme (looks like the rest of it is snake cased) so `add_super_user_form`

kalite/distributed/views.py
```diff
@@ -212,6 +213,29 @@ def search(request):
...
+    if request.method == 'POST':
+        intent = request.REQUEST.get('intent', None)
+        if intent == 'addform':
```
MCGallaspy: I'm opposed to the idea of endpoints with different "intents" like this. Makes them harder to test and reason about. Let the intent be determined by the request method (GET, POST, etc.) only. In this case, I'd say create one endpoint to return form html (though I wonder if it can't be put into a handlebars template?) and another endpoint that is the target of the form and ONLY creates a superuser.

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -1,6 +1,124 @@
...
+    close_modal: function() {
+        $("#superUserCreateModal").modal('hide');
+    },
+
+    show_modal: function() {
+        $("#superUserCreateModal").modal('show');
+    },
+
+    key_user: function(event) {
+        if (event.which == 13) {
```
MCGallaspy: Which keypress does this correspond to? "Enter" right? Can you do something like 

```js
this.ENTER_KEY = 13;
...
if (event.which === this.ENTER_KEY ) {
...
```

kalite/distributed/features/superuser_create.feature
66eli77: those tests also check the front end UI behavior, like check if the target input border turn red, ect.. Django forms mainly for server side validation.

kalite/distributed/views.py
66eli77: what's the advantage of put the html template to handlebars template?

MCGallaspy:
Looks like some existing tests are failing, probably because of the new behavior on the landing page:

```sh
  Scenario: Logging in with the wrong username           # kalite/distributed/features/login.feature:26
    Given I have an account                              # kalite/distributed/features/steps/login.py:42
Given I have an account                              # kalite/distributed/features/steps/login.py:42 0.249s
    And I am on the homepage                             # kalite/distributed/features/steps/generic.py:4
And I am on the homepage                             # kalite/distributed/features/steps/generic.py:4 0.817s
    When I click log in                                  # kalite/distributed/features/steps/login.py:28
When I click log in                                  # kalite/distributed/features/steps/login.py:28 5.544s
    And I enter my username incorrectly                  # kalite/distributed/features/steps/login.py:55
And I enter my username incorrectly                  # kalite/distributed/features/steps/login.py:55 0.088s
...
```

@rtibbles do we need to add another `given` to these tests?

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,124 @@
...
+    assert not User.objects.exists(), "superuser not deleted successfully!"
+
+@then("I should see a modal")
+def step_impl(context):
+    assert find_id_with_wait(context, modal_container).is_displayed(), "modal not displayed!"
+
+@then("I will not be able to dismiss the modal until a superuser is created")
+def step_impl(context):
+    # don't know how to test this
+    pass
```
MCGallaspy: :( don't just `pass` it! That's a false positive.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,124 @@
...
+def step_impl(context):
+    fill_email(context, "correct@email.com")
+
+@then("the modal will dismiss")
+def impl(context):
+    assert not find_id_with_wait(context, modal_container).is_displayed(), "modal not dismissed!"
+
+@then("a superuser is created")
+def impl(context):
+    context.browser.implicitly_wait(1)
```
MCGallaspy: This sets a wait time for the whole browser instance. You can use [WebDriverWait.until](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.support.wait.WebDriverWait.until).

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -0,0 +1,124 @@
...
+
+@given("the username is empty")
+def step_impl(context):
+    fill_username(context, "")
+
+@when('I click the create button')
+def step_impl(context):
+    create_button = find_css_class_with_wait(context, "create-btn")
+    create_button.click()
+    context.browser.implicitly_wait(1)
```
MCGallaspy: This doesn't actually wait for anything, it just sets an implicit wait time for future calls to find elements.

66eli77:
I think this PR is ready to be merged if no further comments added

MCGallaspy:
I'd like to look at it more; I don't think there's any rush given that it's tagged as a 0.15 milestone.

MCGallaspy:
@66eli77 Can you resolve the merge conflicts here?

MCGallaspy:
Note: @benjaoming and @66eli77 suggested this is important to merge in for 0.14, so unless there are strong objections that's what I'll do.

benjaoming:
I'm not seeing intruding changes here, can @66eli77 merge in current develop and have a look at any possible test failures?

benjaoming:
@66eli77 I meant to merge develop into the PR and see what Travis says about the tests, you can't just merge a PR without green lights from Travis....

PR #3680, opened 2015-05-19
Multi dbs

MCGallaspy:
Create a separate DB for topic_tools models. Move previously committed stuff from the `main` app to the `topic_tools` app in order to avoid tricky migration stuff. Delete a migration from that previous stuff after it was determined to be more trouble than it was worth.

In order to create and bootstrap this db, one should run the following commands:
```sh
> bin/kalite manage syncdb --database="topic_tools"
> bin/kalite manage init_assessment_items
```

The result is the file `kalite/database/topic_tools.sqlite`.


kalite/router.py
```diff
@@ -0,0 +1,51 @@
...
+        else:
+            return None
+
+    def allow_relation(self, obj1, obj2, **hints):
+        """
+        Allow relations between the given models only.
+        """
+        if obj1.__name__ in self.MODELS and obj2.__name__ in self.MODELS:
+            return True
+        return None
```
aronasorman: I'm assuming that if it's `None` then we're gonna use the `default` database. Correct?

MCGallaspy: That's right, `None` indicates no opinion and so the default router will be used.

MCGallaspy:
The error is seemingly because a step needs to be added to the setup in travis.yml.

kalite/router.py
aronasorman: How does returning `True` tell Django to use the `topic_tools` DB though? The documentation says you have to put in the right DB name: https://docs.djangoproject.com/en/1.5/topics/db/multi-db/#db_for_read

aronasorman: Oh NVM. I was looking at `allow_relation`.

aronasorman:
Should we add that additional syncdb step to `manage setup` too?

aronasorman:
Looks like it's not needed anymore.

aronasorman:
After pulling in your code and running `kalite restart` some migrations ran, and then I get this error:

```
INFO:kalite:Preloading exercise data for language en.
DatabaseError: no such table: topic_tools_assessmentitem
[658000 refs]
[22842 refs]
~/src/github.com/aronasorman/ka-lite $ 
```

aronasorman:
so we might need to run syncdb --db=topic_tools in setup.

MCGallaspy:
Sure, we can add it to `setup`, otherwise it's an extra step for devs. But if we plan on distributing this db, then the end user won't have to run `syncdb` (since the db already exists).

MCGallaspy:
Otherwise I think your issue can be resolved by running the `syncdb` command I indicated in the top comment.

aronasorman:
Gotcha.

MCGallaspy:
Race condition in travis setup? It complains that `topic_tools_assessmentitem` table wasn't found, yet also says that `topic_tools_assessmentitem` table was created.

MCGallaspy:
Problem seems to be the travis setup:
```yml
install:
-  npm install
-  npm install -g grunt-cli
-  pip install .
-  pip install -r requirements_dev.txt
-  pip install coveralls
-  bin/kalite manage syncdb --database="topic_tools"
```

This works fine:
```sh
> bin/kalite manage setup
> bin/kalite manage syncdb --database="topic_tools"
```

MCGallaspy:
@aronasorman @benjaoming am I missing something here? How can we integrate the extra step needed to create the database into the build process?

aronasorman:
It seems like we implicitly assume that `setup` will be run on `kalite start`. Would adding `setup` before `syncdb` fix the problem?

MCGallaspy:
Probably, yes.

MCGallaspy:
Or should I add the extra syncdb to the setup command?

aronasorman:
I prefer the latter.

benjaoming:
The flow is like this, which I gather from the comments, you already are aware of :)

`kalite start` => runs `kalite setup` if it hasn't been run => runs syncdb if necessary

So you need to look at the setup management script and do syncdb on all relevant databases.

But setup.py isn't part of this equation. It's responsible for distributing the application itself... the databases are part of user-generated data, hence not something that we should distribute as part of the app.

There's also an update process to consider (also run through `manage setup`) that usually prompts people if they want to overwrite an existing database file. That's worth considering, too, in case users should have control of when this file is recreated... but it would be a shame to prompt users about a database file that's kind of hard to explain ("do you want to overwrite the existing database file containing non-user data blah blah") ?

I think it's cool we're doing this with syncdb in the setup process. That way, it's not manda

kalite/settings/base.py
```diff
@@ -248,9 +255,18 @@
         "OPTIONS": {
             "timeout": 60,
         },
+    },
+    "topic_tools": {
+        "ENGINE": "django.db.backends.sqlite3",
+        "NAME": TT_DATABASE_PATH,
+        "OPTIONS": {
+            "timeout": 60,
```
benjaoming: Curious... what's this? Could it go with a comment? I see that it's also in the above database.

MCGallaspy: Dunno, I just figured it seemed reasonable to have the same options as the other database.

benjaoming: I think it's worth removing and then find out later why it's there :)

MCGallaspy: Sets cache invalidation timeout: https://docs.djangoproject.com/en/1.5/ref/settings/#timeout

I think it's probably safe to remove from `topic_tools` database at least.

aronasorman:
Tests passed! Are we waiting for anything else for this PR

MCGallaspy:
I guess that's up to you guys. @jamalex was there anything else I can add here to help incorporate it into our build/release process?

aronasorman:
Let's merge this now.

benjaoming:
*fat lady signing*

kalite/distributed/management/commands/setup.py
```diff
@@ -272,6 +272,8 @@ def handle(self, *args, **options):
         # Migrate the database
         call_command("syncdb", interactive=False, verbosity=options.get("verbosity"))
         call_command("migrate", merge=True, verbosity=options.get("verbosity"))
+        # Create *.json and friends database
+        call_command("syncdb", interactive=False, verbosity=options.get("verbosity"), database="topic_tools")
```
jamalex: This shouldn't be done during setup (or anywhere, for the end user). We'll be shipping a fully configured assessment item DB in the assessments.zip.

MCGallaspy: It should be a no-op if the DB exists.

jamalex: A 50ms no-op, on my fast desktop computer. So, not too slow, but also unnecessary, no?

kalite/topic_tools/__init__.py
```diff
@@ -32,7 +32,7 @@
 from fle_utils.general import softload_json, json_ascii_decoder
 from kalite import i18n
 
-from kalite.main import models as main_models
+from kalite.topic_tools import models as main_models
```
jamalex: `main_models` seems like a misnomer now, and perhaps we can make this a relative import, also

MCGallaspy: Granted.

kalite/settings/base.py
```diff
@@ -215,6 +216,12 @@
         os.mkdir(DEFAULT_DATABASE_PATH)
     
     DEFAULT_DATABASE_PATH = os.path.join(DEFAULT_DATABASE_PATH, 'default.sqlite')
+
+    TT_DATABASE_PATH = os.path.join(USER_DATA_ROOT, "database",)
```
jamalex: I'll be moving this to the /content/khan/ folder, since it's getting downloaded along with the /content/khan/ images, and hence will get moved around with them if the folder is moved -- seems like the best way to ensure the existence of the populated DB aligns with the existence of the corresponding images.

kalite/distributed/management/commands/setup.py
jamalex: Scratch that -- discussed in HipChat, and this will at least avoid errors for devs when no assessment items have been downloaded yet.

PR #3686, opened 2015-05-20
Fix erred output by manage initdconfig

benjaoming:
@aronasorman I found a problem with this, please do cherry-pick to 0.13 as well.

### DIFF
--- a/kalite/distributed/management/commands/initdconfig.py
+++ b/kalite/distributed/management/commands/initdconfig.py
@@ -5,11 +5,12 @@ import sys
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
+import getpass
 
 script_template = """
-#! /bin/sh
+#!/bin/sh
 
-# Author: Jamie Alexandre, 2012
+# Author: Foundation for Learning Equality
 #
 # /etc/init.d/kalite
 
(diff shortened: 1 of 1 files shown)


kalite/distributed/management/commands/initdconfig.py
```diff
@@ -5,10 +5,12 @@
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
+import getpass
 
-script_template = """#! /bin/sh
+script_template = """
+#!/bin/sh
```
MCGallaspy: This will create a script with a newline character as the first character, like this:
```

#!/bin/sh
...
```
Which I think is technically incorrect, as the `#!` should come first, right?

kalite/distributed/management/commands/initdconfig.py
```diff
@@ -5,10 +5,12 @@
...
 
-# Author: Jamie Alexandre, 2012
+# Author: Foundation for Learning Equality
```
MCGallaspy: Don't deny the man his copyright claim.

benjaoming: It's not a copyright claim, and the script isn't from 2012 :)

kalite/distributed/management/commands/initdconfig.py
benjaoming: Done! This would only be relevant in case output is piped directly into the file... but I guess docs might point out that method.

PR #3695, opened 2015-05-21
Windows kalite command

benjaoming:


kalite/version.py
```diff
@@ -3,7 +3,7 @@
 # THIS IS USED BY settings.py.  NEVER import settings.py here; hard-codes only!
 MAJOR_VERSION = "0"
 MINOR_VERSION = "14"
-PATCH_VERSION = "dev2"
+PATCH_VERSION = "dev3"
```
aronasorman: I'm curious on why these versions are being incremented.

aronasorman:
I don't know why you're not putting a description man :(

benjaoming:
I don't know what happened, I wasn't posting this, I think there's something with my ENTER key :) I just have lots of tabs and windows.

setup.py
```diff
@@ -20,19 +22,26 @@
...
+    req_pattern = re.compile(r'^\s*([^\#]+)')
+
     m = req_pattern.search(req)
     if m:
         return m.group(0).replace(" ", "")
 
+
+# Filter out comments from requirements
 DIST_REQUIREMENTS = map(filter_requirement_statements, DIST_REQUIREMENTS)
 DIST_REQUIREMENTS = filter(lambda x: bool(x), DIST_REQUIREMENTS)
```
aronasorman: Much functional!

kalite/version.py
benjaoming: Because you can't overwrite existing uploads to PyPi, and I want to add more tests.

benjaoming:
Merging because it only touches setup.py, a file that fellow coders will become more familiar with over time... but not just before code freeze :)

aronasorman:
I concur.

PR #3699, opened 2015-05-22
Load either the json or yml version file.

### LINKED ISSUE #3562: Move version.py version metadata to YAML

aronasorman:
Fixes #3562. Summary of `open_json_or_yml` algorithm:

Attempt to load a json version of the file first, by changing the file extension to `json`. If it's not present, load the `yml` version by changing the extension to `yml`.

~~todo: a management command that "compiles" all the yaml files into json.~~

You can compile all relevant yml files into json by running `kalite manage compileymltojson`. For now it just compiles `version.yml`.

### DIFF
--- a/kalite/distributed/management/commands/compileymltojson.py
+++ b/kalite/distributed/management/commands/compileymltojson.py
@@ -1,0 +1,41 @@
+from __future__ import print_function
+
+import json
+import os
+import yaml
+
+from django.core.management.base import NoArgsCommand
+
+BLACKLISTED_YAML_FILES = [
+    ".travis.yml",
+    "circle.yml"
+]
+
+
+class Command(NoArgsCommand):
+
+    def handle_noargs(*args, **kwargs):
+        # avoid using django.conf.settings, for that sweet fast startup
+        # time
+        project_root = os.environ.get("KALITE_DIR")
+        for root, _, files in os.walk(project_root):
+            for f in files:
+                full_name = os.path.join(root, f)
+                if (full_name.endswith(".yml") and
+                    "node_modules" not in root and
+                    f not in BLACKLISTED_YAML_FILES):
+                    print(full_name)
+                    yml_to_json(full_name)
+
+
+def yml_to_json(filename):
+    """Convert a .yml file into a json file and save it to the same
+    directory.
+    """
+    jsonfilename = "{0}.json".format(*os.path.splitext(filename))
+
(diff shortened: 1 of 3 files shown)


benjaoming:
Does this mean we can move pyyaml to requirements_dev.txt ?

aronasorman:
yes :)

benjaoming:
Wonderful, and which command needs to be run to convert all yaml files to json? Is this the responsibility of the YML contributor or the packaging team?

benjaoming:
( i don't think it should be on the packaging team unless it's a very simple management command )

aronasorman:
@benjaoming packaging team. Contributors must take care not to import `yaml` anywhere else though.

benjaoming:
But @aronasorman how do I convert all the yaml files to json?

aronasorman:
I am in the process of writing such a management command (see the todo section of my description)>

benjaoming:
Pardon my ignorance :)

aronasorman:
Management command created!

kalite/distributed/management/commands/compileymltojson.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import print_function
```
benjaoming: Ready for the future!

kalite/distributed/management/commands/compileymltojson.py
```diff
@@ -0,0 +1,41 @@
...
+import json
+import os
+import yaml
+
+from django.core.management.base import NoArgsCommand
+
+BLACKLISTED_YAML_FILES = [
+    ".travis.yml",
+    "circle.yml"
+]
```
benjaoming: I agree on this approach, I was thinking the same... like let's not have a management command where we need to add new yml files every time we add one.

kalite/shared/utils.py
```diff
@@ -0,0 +1,28 @@
+import os
+
+def open_json_or_yml(file_name):
```
MCGallaspy: @tk21 you can use this function.

PR #3700, opened 2015-05-22
For inclusion: add scandir dependency.

aronasorman:
scandir claims to be faster than os.walk. I have not tested this claim
myself, but I trust reddit.

Oh, and it's also part of python 3.5. Surely we can't say no to that??

PR created for discussion.

### DIFF
--- a/kalite/contentload/tests/unpack_assessment_zip.py
+++ b/kalite/contentload/tests/unpack_assessment_zip.py
@@ -2,6 +2,7 @@ import StringIO
 import json
 import os
 import requests
+import scandir
 import tempfile
 import zipfile
 from mock import patch, MagicMock, mock_open
@@ -27,7 +28,7 @@ class UnpackAssessmentZipCommandTests(KALiteTestCase):
         _, self.zipfile_path = tempfile.mkstemp()
         with open(self.zipfile_path, "w") as f:
             zf = zipfile.ZipFile(f, "w")
-            for dirpath, _, filenames in os.walk(os.path.join(os.path.dirname(__file__), "fixtures")):
+            for dirpath, _, filenames in scandir.walk(os.path.join(os.path.dirname(__file__), "fixtures")):
                 # this toplevel for loop should only do one loop, but
                 # it's in a for loop nonetheless since it's more idiomatic
                 for filename in filenames:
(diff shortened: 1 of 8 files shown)


aronasorman:
Pros:

Faster. 

Cons:

Additional dependency. Will slow down startup time.

aronasorman:
I'm merging this after standup if no one has any complaints, as I want to use this for #3699.

aronasorman:
`scandir` has its (pseudo) C extension, but if that's not present it will drop down to its pure python implementation.

MCGallaspy:
Is it a drop-in for `os.walk`? Can you replace existing uses?

aronasorman:
@MCGallaspy yes, I believe it's a drop-in for `os.walk` if you're just looking at files and filenames. If you're inspecting the attributes given by `os.stat`, then it won't work for that.

MCGallaspy:
> The API for scandir.walk() is exactly the same as os.walk()

Looks like it's good to go.

rtibbles:
As part of this PR, could you change, say, the use of os.walk in the content availability stamping (where it is used to add subtitle URLs) and then we could do a direct speed comparison?

benjaoming:
@aronasorman are you sure that the python-only version of scandir has chances of being faster than os.walk?

benjaoming:
I'm in the middle of cleaning up *unused* dependencies so I would rather not like to merge this... it's an unused dep :)

If you schedule a full refactor of the entire project with respect to os.walk, it'd be fine... perhaps an issue fir 0.15 ?

aronasorman:
@benjaoming I've done my best to refactor our use of `os.walk` to use `scandir.walk` in my latest commit. So by this PR it should be a used dependency :)

However, I feel like introducing a dependency so close to the code freeze may be walking a fine line. @benjaoming also brought up a good point of benchmarking the *pure python* version of this module first to see if there are any significant speedups.

aronasorman:
I'll post benchmarks.

aronasorman:
No time to do this. 0.15.

MCGallaspy:
Let's see some comparisons.

benjaoming:
@aronasorman could you resolve the conflicts for this?

aronasorman:
@benjaoming do you want to profile this yourself? If not I don't think this
is important enough for 0.15.

benjaoming:
@aronasorman no, we might differ a bit on issues of bringing about new dependencies ;) that's what deters me from these sort of changes at the moment.. so I think it's fine that we wait.

aronasorman:
Stale. @gwong89 please take a look at this PR, then open another one.

PR #3702, opened 2015-05-22
Make setup.py understand stdeb commands

benjaoming:


aronasorman:
"In this town, we don't need no code review!" - cowboy ben

setup.py
```diff
@@ -42,7 +42,7 @@ def filter_requirement_statements(req):
 STATIC_REQUIREMENTS = []
 
 # Decide if the invoked command is a request to do building
-DIST_BUILDING_COMMAND = any([x in sys.argv for x in ("bdist", "sdist", "bdist_wheel")])
+DIST_BUILDING_COMMAND = any([x in sys.argv for x in ("bdist", "sdist", "bdist_wheel", "bdist_deb", "sdist_dsc")])
```
aronasorman: What does this do?

benjaoming: It makes setup.py aware of whether it's building or installing... when it's building, it can import "pip" but when it's installing, pip might not be available... only setuptools.

benjaoming: I guess there are are other cases further down in the code where there's build-specific logic.

benjaoming:
Yes, watch me as I break everything with my unreviewed oneliner :)

PR #3703, opened 2015-05-22
What is this file doing?

benjaoming:
I don't see comments or find anything that's using it...


jamalex:
@cpauya, looks like you added this?

cpauya:
Ah yes, I remember adding that for unit testing before but it can be
removed now.

benjaoming:
There we go, pardon the silly commit message, went a bit fast :)

PR #3708, opened 2015-05-23
Understand an stdeb option that cancels building and uses prebuilt ta…

benjaoming:
…rball

necessary for build debs


benjaoming:
2.6 test fails because of travis timeouts

PR #3709, opened 2015-05-23
Empty facility __init__.py file which was loading models

benjaoming:
This caused hours of work in debugging a debian distribution... !?

So please, it's really really costly to leave these odd imports messing around :(


jamalex:
This seems pretty straightforward. There's a test failing, but only for 2.6, and it looks like a stochastic one. Merging!

PR #3710, opened 2015-05-23
stop trying to import settings from everything in INSTALLED_APPS

benjaoming:
It's not cool to have a function that looks at everything in INSTALLED_APPS and just imports `settings` from it.


benjaoming:
** importing `settings` also triggers loading the application's __init__.py

jamalex:
Turning :imp: into :smiling_imp:, one step at a time...

jamalex:
This time it was 2.7 that stochastically failed...

PR #3711, opened 2015-05-23
Only complain about missing --static argument if it's an install comm…

benjaoming:
…and in source or if it's a build command

Sorry about the many PRs, but working in a different repo makes me want to get things merged quickly into ka-lite.


benjaoming:
One-liner in setup.py, merging :)

PR #3712, opened 2015-05-24
Fix remaining circular module init in chronograph/securesync

benjaoming:
This was causing previously unseen issues in the .deb installation. Who knows, these issues could have cause lots of other unexplained breakage through time.

PR also removes a lot of unused imports.


python-packages/securesync/devices/decorators.py
```diff
@@ -0,0 +1,30 @@
+"""
+This sub-module deals with Zones--how they're defined, permissions related to them,
+and the functions to allow joining them.
+
+The "engine" sub-module deals with sharing data within a zone.
+"""
```
jamalex: This docstring looks like it belongs back in the `__init__` from whence it came? (it describes the module, not the function here)

python-packages/securesync/engine/__init__.py
```diff
@@ -1,352 +0,0 @@
-"""
```
jamalex: This is likely also going to break imports on the ka-lite-central repo. We'll need to fix things up there too.

jamalex:
A lot of churn here, but seems important to get merged in for the code freeze, since it's a blocker for the .deb, and possibly causing problems elsewhere.

PR #3713, opened 2015-05-24
remove unused imports

benjaoming:
not comfortable that kalite.main.settings imports kalite

### DIFF
--- a/kalite/main/settings.py
+++ b/kalite/main/settings.py
@@ -1,8 +1,3 @@
-import os
-import json
-
-from kalite import ROOT_DATA_PATH
-
 try:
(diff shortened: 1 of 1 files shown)


jamalex:
Still seems to get imported [here](https://github.com/learningequality/ka-lite/blob/develop/kalite/settings/base.py#L7); is that concerning also?

benjaoming:
that should be okay, I just don't feel comfortable about current settings.py files importing anything at all, because it seems to always break later... it's this dynamic importing in fle_utils.settingshelper that causes my concern.. but it was unused, anyways, so a freebie this time :)

PR #3714, opened 2015-05-25
Distribute dist packages + move pyyaml to requirements_dev

benjaoming:


setup.py
```diff
@@ -120,10 +120,20 @@ def gen_data_files(*dirs):
...
     results = []
-
+    
+    filter_illegal_extensions = lambda f: os.path.splitext(f)[1] != ".pyc"
+    
     for src_dir in dirs:
         for root, dirs, files in os.walk(src_dir):
-            results.append((root, map(lambda f: os.path.join(root, f), files)))
+            results.append(
+                (
```
aronasorman: much parens, such Lisp

aronasorman:
There's an error regarding the `six` module.

setup.py
benjaoming: Yes, unfortunately not an ideal scenario that it's solving :( Basically, it has to list tuples containing all files (listed recursively) in all src_dirs, but the files have to filter so there are no pyc files... if the problem is ugly, the solution is, too.

rtibbles:
Note: This error appears to be extant on develop, as both my PRs (after pulling in latest develop) also have the same issue in the Travis builds.

benjaoming:
This error is due to using an old version of six... i.e. the host system's version and not the one supplied in python-packages.

So I have to wonder, why this is happening all of a sudden.. it wasn't showing up in any of the previously merged PRs.

aronasorman:
@benjaoming looks like Django has its own six module bundled in: https://docs.djangoproject.com/en/1.5/topics/python3/#django-customized-version-of-six

It might be that Django is expecting the bundled
 version, whereas we import the latest one?

benjaoming:
that module wouldn't be available outside django.. so i suspect something
weird to be happening in the path environments.. will see if I can recreate
locally.

benjaoming:
>>> import six
>>> six.__file__
'/media/archive/code/ka-lite/python-packages/six.pyc'
>>> from six.moves import range
  File "<console>", line 1, in <module>
ImportError: cannot import name range

benjaoming:
it's our six that's outdated and somehow this shows up in the tests now, suggesting to unbundle it from python-packages

benjaoming:
Okay, for the record this was the problem:

requests bundled six 1.1.0 which included the following line of horror:

    moves = sys.modules["six.moves"] = _MovedItems("moves")

This means that when requests loaded its own six, then all other future loaded six modules would be overwritten. Because load order had changed, the error showed up.

benjaoming:
Also, the build is now failing with the usual Behave tests, so it's ready to merge IMO.

I will be making an additional PRs for unbundling requests.

rtibbles:
Actually, not the usual behave failures. Ones that I caused by creating a features folder with no ssteps folder in coachreports. Will fix up.

aronasorman:
The behave and JSHint errors are known issues on develop. This should be good to go.

MCGallaspy:
> Also, the build is now failing with the usual Behave tests

I'm hoping there aren't any _usual_ behave tests.

PR #3717, opened 2015-05-25
Inline

MCGallaspy:
Supersedes #3662. Just adds the option of enabling or disabling loading of inline js assets by using the INLINE_HELP setting; truthy values mean assets are included, otherwise not.


MCGallaspy:
Though it would be useful to preserve the comments from #3662, I think that PR will probably be automatically closed if this one is merged.

kalite/distributed/static/js/distributed/exercises/api_resources.py
```diff
@@ -0,0 +1,31 @@
+from django.db import models
+from django.http import HttpResponse
+
+from tastypie import fields
+from tastypie.resources import ModelResource
+
+import yaml
```
aronasorman: Use `kalite.shared.utils.open_json_or_yml` rather than importing `yaml` directly.

kalite/distributed/templates/distributed/base_manage.html
```diff
@@ -9,6 +9,13 @@
     {% block manage_active_admins %}active{% endblock manage_active_admins %}
 {% endblock i18n_do_not_translate %}
 
+<!-- {% block headjs %}
+    <script>var NARRATIVE_ID = "manage_tab";</script>
```
aronasorman: Why is this commented out? If it's not used, can you delete it?

kalite/inline/api_views.py
```diff
@@ -0,0 +1,16 @@
+"""
+Views accessible as an API endpoint for inline.  All should return JsonResponses.
+"""
+import yaml
```
aronasorman: use `open_json_or_yml` here.

MCGallaspy:
[This](https://github.com/learningequality/ka-lite/pull/3662#discussion_r30656197) still needs to be addressed.

MCGallaspy:
@aronasorman shall I merge this? Or did you still wish to review it?

Issue #3718, opened 2015-05-25
Unbundle requests


benjaoming:
So we just got rid of a big problem caused by requests' bundled old six 1.1.0 and we should unbundle requests and depend on a later version that either doesn't include six or bundles a later version.

benjaoming:
```
Traceback (most recent call last):

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/core/management/base.py", line 224, in run_from_argv

    self.execute(*args, **options.__dict__)

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/core/management/base.py", line 262, in execute

    self.validate()

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/core/management/base.py", line 288, in validate

    num_errors = get_validation_errors(s, app)

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/core/management/validation.py", line 35, in get_validation_errors

    for (app_name, error) in get_app_errors().items():

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/db/models/loading.py", line 166, in get_app_errors

    self._populate()

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/db/models/loading.py", line 72, in _populate

    self.load_app(app_name, True)

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/db/models/loading.py", line 96, in load_app

    models = import_module('.models', app_name)

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/utils/importlib.py", line 35, in import_module

    __import__(name)

  File "/home/travis/build/learningequality/ka-lite/python-packages/tastypie/models.py", line 32, in <module>

    from tastypie.compat import AUTH_USER_MODEL

  File "/home/travis/build/learningequality/ka-lite/python-packages/tastypie/compat.py", line 14, in <module>

    User = get_user_model()

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/contrib/auth/__init__.py", line 125, in get_user_model

    user_model = get_model(app_label, model_name)

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/db/models/loading.py", line 230, in get_model

    self._populate()

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/db/models/loading.py", line 72, in _populate

    self.load_app(app_name, True)

  File "/home/travis/build/learningequality/ka-lite/python-packages/django/db/models/loading.py", line 96, in load_app

[...]

PR #3719, opened 2015-05-25
Unbundle requests #3718

benjaoming:
#3718


benjaoming:
Not going to bother with this now

Issue #3722, opened 2015-05-25
New problems for settings import


benjaoming:
Currently, `kalite.distributed.settings` raise an ImportError when `kalite.settings` tries to import it.

```
>>> from kalite import settings
kalite.distributed
ImportError <type 'exceptions.ImportError'>
Traceback (most recent call last):
  File ".virtualenvs/kalite_pip/share/kalite/python-packages/fle_utils/settingshelper.py", line 65, in import_installed_app_settings
    app_settings = __import__(app + ".settings")
  File ".virtualenvs/kalite_pip/local/lib/python2.7/site-packages/kalite/distributed/settings.py", line 164, in <module>
    version_info = version.VERSION_INFO()
  File ".virtualenvs/kalite_pip/local/lib/python2.7/site-packages/kalite/version.py", line 29, in VERSION_INFO
    return open_json_or_yml(os.path.join(settings.CONTENT_DATA_PATH, "version.yml"))
  File ".virtualenvs/kalite_pip/local/lib/python2.7/site-packages/kalite/shared/utils.py", line 24, in open_json_or_yml
    import yaml
ImportError: No module named yaml
```

This is because we want to use version data in our settings:

https://github.com/learningequality/ka-lite/blob/develop/kalite/distributed/settings.py#L162

It would be great to burn this down, since we don't actually need to be using a git commit hash for cache key prefixes.

benjaoming:
First part fixed in bcf2863

MCGallaspy:
How do I reproduce this? `from kalite import settings` works without error for me, using the `shell_plus` command.

benjaoming:
@MCGallaspy the fix was already merged into develop at the time whe you tested, perhaps that's the problem. No matter what, you'd need to be running without pyyaml.

I'll have a second look at develop to make sure that there's no overlapping or brainfarts from my side.

aronasorman:
Looks like it's fixed for the current 0.14.x.

PR #3726, opened 2015-05-26
Prefix caching with version number and do something weird to generate…

benjaoming:
… install_location_hash based on version instead of perceived location of user data, which was useless anyways #3722


kalite/distributed/settings.py
```diff
@@ -160,17 +159,13 @@ def USER_FACING_PORT():
...
 #   EXCEPT: if CACHE_TIME=0
 if CACHE_TIME != 0:  # None can mean infinite caching to some functions
-    # We can bump the version and forget about version info
-    version_info = version.VERSION_INFO()
-    if version.VERSION in version_info:
-        KEY_PREFIX = version_info[version.VERSION]["git_commit"][0:6]  # new cache for every build
-    else:
-        KEY_PREFIX = "unknown_version"
+    # When we change versions, cache changes, too
+    KEY_PREFIX = ".".join(version.VERSION)
```
MCGallaspy: For anyone wondering like me, here's what [KEY_PREFIX](https://docs.djangoproject.com/en/1.5/ref/settings/#key-prefix) does. [See also](https://docs.djangoproject.com/en/1.5/topics/cache/#cache-key-prefixing).

MCGallaspy:
Your test failures here should now be resolved in `develop`.

MCGallaspy:
Otherwise looks ok, just merge `develop` back in.

aronasorman:
I'm encountering a similar problem due to VERSION_INFO getting loaded. Will merge this now to get it into the central server.

Issue #3676, opened 2015-05-19
Unbundle playground


benjaoming:
The package `playground` is unused:

```
➜  ka-lite git:(dist-packages) ack-grep --python playground kalite
kalite/basetests/tests.py
222:        "playground": NO_VERSION,
➜  ka-lite git:(dist-packages) ack-grep --python playground python-packages
python-packages/fle_utils/build/management/commands/generate_blacklist.py
67:            "python-packages/playground",

python-packages/playground/test_tools/mount_branch.py
11:import playground
12:from playground.docker.docker import Docker, PersistentDocker
13:from playground.utils.testing import lexec, get_open_ports
626:            self.dockers[server_type].run_command("cd /playground", wait_time=0.1)
632:            docker.run_command("export PYTHONPATH=${PYTHONPATH}:/playground", wait_time=0.1)
633:            docker.stream_command("python /playground/test_tools/mount_branch_on_docker.py %s %s %s %d %s" % (self.git_user, self.repo_branch, server_type, self.docker_port, self.git_repo), wait_time=wait_time)
691:#        self.docker.run_command("/playground/test_tools/mount_docker_branch.sh %s %s %s" % (self.git_user, self.git_repo, self.repo_branch))
802:    kap.mount_project(server_types=server_types, host="playground.learningequality.org")
```

PR #3727, opened 2015-05-26
Unbundle unused playground #3676

benjaoming:
#3676


benjaoming:
Only two completely unrelated Behave test failures. Would be nice to have these failing tests removed soon!!!

MCGallaspy:
One would hope that a failing test is truly an error instead of random failure; if it's the latter I'll open an issue and try to fix it. In the future if a test fails and you think it's not an actual regression, please open an issue.

MCGallaspy:
Indeed, a legitimate error! But fixed by https://github.com/learningequality/ka-lite/commit/9434bcfcfa327727df4fee880b99cdb47996730c.

benjaoming:
@MCGallaspy I'd normally do that but yesterdays meeting made me confused about whether these are tests that are expected to fail (and also completely unrelated to the contents of this PR) and I thought they would be disabled.

Indeed, failing tests are totally annoying.

benjaoming:
@MCGallaspy seems that travis is finally happy :)

Issue #3544, opened 2015-04-23
Register device link redirect to page not found.


amodia:
Branch: `Develop`
Expected behavior: `register device link` must not redirect to page not found.
Current behavior: `register device link` redirect to page not found
Steps to reproduce: log in as coach and click the `register device link`.

![ka-lite-a](https://cloud.githubusercontent.com/assets/8663934/7291451/49b96194-e9c3-11e4-9677-cba514877c87.gif)

MCGallaspy:
@rtibbles the issue here appears to be that a user without sufficient permissions is redirected based on the setting (in the distributed app) `LOGIN_URL`, which is no longer a valid url -- hence the 404. We should probably change the way permissions like this are handled. Is there a GET parameter which will cause the login modal to be displayed?

rtibbles:
There will be once #3729 is merged (I believe I had changed LOGIN_URL to this at some point to, but that seems to have got lost in translation at some point).

MCGallaspy:
Cool, do you expect that PR to resolve this issue then?

rtibbles:
No. But I can update it to change the LOGIN_URL, as that seems the best approach.

rtibbles:
Will fix up shortly.

MCGallaspy:
Agreed, then I think this will be fixed by #3729.

rtibbles:
Fixed by #3729

PR #3729, opened 2015-05-26
Fixes redirect for modal login.

### LINKED ISSUE #3538: Getting logged out refreshes the page, but doesn't bring up the modal.
When I restart the server, it logs me out (expected), but it simply leads me to this:
login

And doesn't show the modal anywhere. In develop, but maybe it's extant in 0.13.x too.

rtibbles:
Fixes #3538

Summary of changes:
* Use next GET param to determine redirect location.
* If no next GET param, then use any supplied redirect from the API response
* Alter the 403 handler view to add login=true to the GET param to trigger the modal.

### DIFF
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -315,10 +315,8 @@ window.UserView = BaseView.extend({
             var options = {};
             var next = getParamValue("next");
             var login = getParamValue("login");
-            if (login || this.login_start_open) {
+            if (login) {
                 options.start_open = true;
-
-                delete this.login_start_open;
             }
             if (next) {
                 options.next = next;
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -4,8 +4,9 @@
 window.LoginModalView = BaseView.extend({
     template: HB.template("user/loginmodal"),
 
-    initialize: function() {
+    initialize: function(options) {
         _.bindAll(this);
+        this.set_options(options);
         this.render();
         $("body").append(this.el);
     },
@@ -18,8 +19,9 @@ window.LoginModalView = BaseView.extend({
     addLoginView: function() {
         if (this.loginView) {
             this.loginView.render();
+            this.loginView.next = this.next;
         } else {
-            this.loginView = new LoginView({model: this.model, el: "#login-container"});
+            this.loginView = new LoginView({model: this.model, el: "#login-container", next: this.next});
             this.listenTo(this.loginView, "login_success", this.close_modal);
         }
         if (this.start_open) {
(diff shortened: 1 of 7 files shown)


kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -299,15 +312,25 @@ window.UserView = BaseView.extend({
                 this.totalPointViewXs = new TotalPointView({model: this.model, el: "#points-xs"});
             }
         } else {
+            var options = {};
+            var next = getParamValue("next");
+            var login = getParamValue("login");
+            if (login || this.login_start_open) {
+                options.start_open = true;
+
+                delete this.login_start_open;
```
jamalex: Some comments in here (and a few more, in general) could be helpful -- not clear to me what this variable (`login_start_open`) means, or why it's being deleted rather than set to `false`, etc.

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -89,6 +97,11 @@ window.LoginView = BaseView.extend({
     handle_login: function(response) {
         if (response.status == 200) {
             this.trigger("login_success");
```
MCGallaspy: Does this set up the potential for a race condition? Maybe "login_success" should only be triggered if a redirect doesn't occur. Otherwise it's possible someone might come to expect a side effect from the "login_success" event that occurs before the page reloads, which isn't a safe expectation.

rtibbles: I am not sure what that side effect might be. This is being used to close the modal upon success, which is a reasonable UI expectation regardless of whether the page subsequently redirects (from a user point of view the modal dismissing is a good indicator of the login success, even if momentarily).

Unless this event was being used to trigger some server side logging, I don't know how this could possibly result in an unsafe expectation.

MCGallaspy: > I don't know how this could possibly result in an unsafe expectation.

It doesn't right now, but my point is that it could in the future if someone decides to add another listener. Since this _could_ trigger a page load (yet _may_ not), any listener will have to finish what it's doing in a hurry or potentially be cut short. If all you wish to do is close the modal, why not just call `close_modal` directly?

rtibbles: Because the modal logic is separated from the interior login view.

MCGallaspy: Okay, point taken. It's unlikely this will be used in an unsafe way.

kalite/distributed/static/js/distributed/user/views.js
rtibbles: I... think this got left in from another approach I took to this. I will delete this flag.

rtibbles:
Comments addressed. Rebased to latest develop to hopefully get tests to pass.

MCGallaspy:
It appears there are some test failures here because the redirect logic has changed. I wonder if `LoginRequiredTestCase` can be removed? I'll take a look.

MCGallaspy:
Huh, weird, just noticed it's a django test. Why are we even running that?

rtibbles:
That is odd. Even more odd, is the fact that previously the URLs were just bare URLs, and so would have failed the reverse lookup that is causing the failure here too...

rtibbles:
Should also fix #3544

rtibbles:
Tests pass, merging.

PR #3733, opened 2015-05-26
Scaffolding for loading overlays

Antrikshy:
We intended to make a loading overlay for the sidebar, but turns out we don't need that anymore right now. However, I have this bit of scaffolding for this really [plainOverlay](http://anseki.github.io/jquery-plainoverlay/) library. It would be nice to use this with the exercises, as @jamalex suggested.

Anything that extends `BaseView` should be able to call `show_loading_overlay` and `hide_loading_overlay` to show and hide a cool overlay and disable the element while it is loading. Currently, the functions only show the default spinner overlay that comes with the library. But maybe we can replace it with a simple "Loading..." text message and/or a logo (how about a pulsing Kolibri logo later!?).

I also add the `.is-loading` class, in case something else needs to keep track of loading status. Also, the show function can be called several times (ex. when different async events are fired) and the overlay will only disappear when the hide function is called an equal number of times (ex. when they all trigger callbacks).

There is a caveat to this overlay though. The dimensions of an element need to be finalized before the show function is called. Otherwise, the overlay will not appear unless the screen is resized a little.

Anyone expanding on this (and putting it to use) should look at the documentation on its website.


jamalex:
Very cool! Could you add some short inline comments (drawn from your description above) to help later readers understand how this works?

If you can see a good place for it, a quick example of it in use would also be handy.

MCGallaspy:
Does this violate the code freeze? Or are we trying to merge this for 0.14?

Antrikshy:
Not sure. We can put it in the next version. It's not being used anywhere yet.

aronasorman:
> Does this violate the code freeze?

This seems like a gray area, as it shouldn't affect anything else. However since we're only ~2 weeks away from release, and exercises shouldn't be get modified heavily enough to cause a merge conflict with this, it might be good to hold off merging this 'til 0.14 is out.

jamalex:
Since its functionality isn't being used yet, and it's not likely to lead to merge conflicts, let's merge post-0.14-fork, targeting for 0.15.

rtibbles:
Unfortunately, this did not get merged prior to refactor.

@Antrikshy would you have any time to fit this into the current framework, or would you like me to fix it up?

rtibbles:
Superseded by #4414

PR #3734, opened 2015-05-26
Clarify kaserve message

### LINKED ISSUE #3732: "Access from another computer message" showing 127.0.0.1
When, for instance, you run `kalite start`, you might see the following message:
>To access KA Lite from another connected computer, try the following
address(es):
    http://127.0.0.1:8008/

Which won't work from _another_ computer, as suggested.

MCGallaspy:
Just a slight rewording of the message in the `kaserve` command. Fixes #3732.

### DIFF
--- a/kalite/distributed/management/commands/kaserve.py
+++ b/kalite/distributed/management/commands/kaserve.py
@@ -150,12 +150,10 @@ class Command(BaseCommand):
         else:
             del options["production"]
             addresses = get_ip_addresses(include_loopback=False)
-            loopbacks = set(get_ip_addresses(include_loopback=True)) - set(addresses)
             sys.stdout.write("To access KA Lite from another connected computer, try the following address(es):\n")
             for addr in addresses:
                 sys.stdout.write("\thttp://%s:%s/\n" % (addr, settings.USER_FACING_PORT()))
-            sys.stdout.write("To access KA Lite from this machine, try the following address(es):\n")
-            for addr in loopbacks:
-                sys.stdout.write("\thttp://%s:%s/\n" % (addr, settings.USER_FACING_PORT()))
+            sys.stdout.write("To access KA Lite from this machine, try the following address:\n")
+            sys.stdout.write("\thttp://127.0.0.1:%s/\n" % settings.USER_FACING_PORT())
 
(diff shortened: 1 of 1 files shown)


kalite/distributed/management/commands/kaserve.py
```diff
@@ -149,7 +149,13 @@ def handle(self, *args, **options):
             call_command("runserver", "%s:%s" % (options["host"], options["port"]))
         else:
             del options["production"]
+            addresses = get_ip_addresses(include_loopback=False)
+            loopbacks = set(get_ip_addresses(include_loopback=True)) - set(addresses)
```
aronasorman: So we only want the loopback address? I think it's always `127.0.0.1`. If not then they're using a strange computer.

benjaoming:
:+1:

kalite/distributed/management/commands/kaserve.py
```diff
@@ -149,7 +149,11 @@ def handle(self, *args, **options):
             call_command("runserver", "%s:%s" % (options["host"], options["port"]))
         else:
             del options["production"]
+            addresses = get_ip_addresses(include_loopback=False)
             sys.stdout.write("To access KA Lite from another connected computer, try the following address(es):\n")
-            for addr in get_ip_addresses():
+            for addr in addresses:
                 sys.stdout.write("\thttp://%s:%s/\n" % (addr, settings.USER_FACING_PORT()))
+            sys.stdout.write("To access KA Lite from this machine, try the following address:\n")
+            sys.stdout.write("\thttp://127.0.0.1:%s/\n" % settings.USER_FACING_PORT())
```
benjaoming: There should kind of be a TODO on this, I don't believe that 127.0.0.1 is always the loopback... there could be one or more, and they could be different.

benjaoming: But like, just leave it, it's a stupid detail

aronasorman: Yeah... If loopback isn't 127.0.0.1, then they are either using a very esoteric system, or have modded their system heavily.

MCGallaspy: I just can't win here: https://github.com/learningequality/ka-lite/pull/3734#discussion_r31078541

PR #3735, opened 2015-05-26
fix #3732 - remove mention of "another computer".

### LINKED ISSUE #3732: "Access from another computer message" showing 127.0.0.1

aronasorman:
Since we always display 127.0.0.1:8008, we should remove mentions of
"another computer" since they won't be able to access ka lite through localhost.


aronasorman:
Duplicate of #3734.

Issue #2908, opened 2015-01-29
Cleanup unused management commands


benjaoming:
Suggestions:
- [x] `distributed -> katest` currently has no references
- [x] `distributed -> initdconfig` should be a task for installers
- [x] `distributed -> nginxconfig` should be a task for installers / documentation
- [x] `distributed -> apacheconfig` should be a task for installers / documentation
- [x] `distributed -> todolist` let's encourage Github issues
- [x] `distributed -> setup` rename to `kasetup` **wontfix**
- [x] `i18n -> i18nize_templates` is a great tool, but it's already a thirdparty and not really something we wanna distribute

aronasorman:
Assigned you @benjaoming. Feel free to start this on a feature branch, then merge it in once the feature window for 0.14 is open (if it's done by that time).

aronasorman:
I'll work on deleting `initdconfig`, that should force our debian installer to implement that themselves ;)

radinamatic:
@benjaoming You did a major cleanup for `kalite` command in 0.17, so this could be closed, correct?

benjaoming:
@radinamatic correct! Even the `i18nize_templates` was removed.

PR #3736, opened 2015-05-26
delete initdconfig.

aronasorman:
Partially solves #2908. This is an OS-specific feature -- sounds like
something an OS-specific installer should do.

### DIFF
--- a/kalite/distributed/management/commands/initdconfig.py
+++ b/kalite/distributed/management/commands/initdconfig.py
@@ -1,11 +1,14 @@
 """
 """
+import getpass
 import os
 import sys
+import warnings
 
 from django.conf import settings
 from django.core.management.base import BaseCommand
-import getpass
+
+from kalite.shared.warnings import RemovedInKALite_v015_Warning
 
 script_template = """#!/bin/sh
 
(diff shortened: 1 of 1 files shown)


aronasorman:
I removed the `runatboot.sh` script, but not the windows start at boot scripts, since the Windows installer already depends on them, and priority one for Windows is to get it working.

benjaoming:
If you target this for 0.14, I have to implement it in the debian installer, and @cpauya has to confirm that the Mac installer also has this feature without depending on the initdconfig command?

aronasorman:
@benjaoming I believe the mac installer doesn't use this, and has its own initialization script that's specific to OS X (as it doesn't use init.d).

aronasorman:
Of course we can always target this to 0.15.

benjaoming:
@aronasorman the deleted initdconfig management command also contained Mac startup code!

aronasorman:
Oh, I didn't expect that. Haha. Will file another issue on the installers repo. Guess that means we're bumping this to 0.15 as well.

benjaoming:
DeprecationWarning?

(it will be printed to stderr so commands can still use this one to pipe file output using stdout, e.g. `manage initdconfig > /etc/init.d/blah` will spill work!

aronasorman:
@benjaoming restored `initdconfig` and added the deprecation warning.

benjaoming:
I'll try to achieve a .deb that doesn't use this command

jamalex:
Since all that remains in this PR is a deprecation warning, we should probably merge it for 0.14?

benjaoming:
@jamalex it already has the deprecation warning so I'm not sure if anything is remaining :)

The .deb won't use it.

PR #3737, opened 2015-05-26
0.13.x into develop

MCGallaspy:
I've got another 0.13.x bugfix incoming. But after this, perhaps 0.13.x bugfixes should target 0.14.


aronasorman:
I agree. 0.13.x should only be for critical bugfixes and documentation tweaks.

MCGallaspy:
There should be no new commits on this one; once tests pass it's ready to merge.

MCGallaspy:
Looks ok.

PR #3738, opened 2015-05-26
Correct explanation in learner docs

### LINKED ISSUE #3665: Incorrect information in learner-facing user manual
The documentation section for learners on [how to use the exercises](https://learningequality.org/docs/usermanual/userman_learner.html#how-to-practice-exercises) is not very fleshed out (it could use some screenshots), and also contains inaccuracies (describing how things worked in 0.12):

> The bar on the top right will display your progress as you go through the exercises. After 10 correct answers in a row, your points for the exercise will be added to our total and you will have mastered the exercise. If you get an answer wrong or ask for a hint, all the points you have earned for this exercise will be lost.

In 0.13, you only need to get 8 out of the last 10, and you never suddenly lose all your points for an exercise. See the blog post: https://learningequality.org/blog/2015/ka-lite-013-released/

MCGallaspy:
Fixes #3665, almost. What should be done about exercise screenshots in the docs? One would need to ensure the machine building the docs has the exercise referenced by the screenshot directive, otherwise it will fail to build the docs. @aronasorman ?

### DIFF
--- a/sphinx-docs/usermanual/userman_learner.rst
+++ b/sphinx-docs/usermanual/userman_learner.rst
@@ -59,7 +59,7 @@ To answer a question, enter the answer in the box on the top right of the module
 
 After answering a question, the exercise module will not advance to the next question until you click the "Next Question..." button on the right hand side of the module.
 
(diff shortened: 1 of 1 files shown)


aronasorman:
@MCGallaspy I don't think there's any way around having to have the exercise in the docs server. We just have to quality check the docs before releasing them, and amending any screenshots which are off. We should be doing that anyway!

benjaoming:
Thanks for clarifying that @coveralls :)

Issue #3550, opened 2015-04-23
develop branch (.14) leaks processes on linux.


kyconquers:
it seems 'kalite start' creates a new process and 'kalite stop', stops the web server, but doesn't stop the process from 'kalite start'

<user>+ 25712  0.0  1.7 237600 36452 ?        Sl   16:04   0:00 /usr/bin/python2 /usr/src/ka-lite/kalitectl.py start
<user>+ 25885  0.0  1.8 165568 38028 ?        S    16:06   0:00 /usr/bin/python2 /usr/src/ka-lite/kalitectl.py start
<user>+ 26077  0.0  1.8 164932 37500 ?        S    16:13   0:00 /usr/bin/python2 /usr/src/ka-lite/kalitectl.py start
<user>+ 26082  1.7 13.1 1776516 271132 ?      Sl   16:13   0:04 /usr/bin/python2 /usr/src/ka-lite/kalitectl.py start

aronasorman:
Very interesting! @benjaoming

benjaoming:
Thanks for the report @kyconquers 

I concur that there are issues about stopping kalite processes, but I'm not sure they're as consequent as implied here... it normally works fine for me, but if I start another KA lite on top of a running one, things start going wrong...

I'll investigate further!

benjaoming:
Does anyone have any pointers as how to recreate this issue on current develop? @kyconquers ?

benjaoming:
Not sure how to recreate this... but now there's a test to see that `kalite stop` doesn't leave any processes with the string `kalite` in their name.

benjaoming:
In addition to this, here is a scenario that's very easy to recreate:

    kalite start
    [CTRL+C]  # Just a seconds after
    kalite stop

This will effectively leak a cronserver process and `kalite stop` won't remove it.

aronasorman:
FYI I encountered this as well whenever `kalitectl.py` exits prematurely.

benjaoming:
Confirming that a CTRL+C during startup no longer leaves behind a leaked process.

Because of the order in which we start things (now calling `initialize_kalite` before spawning any threads or forked processes), CTRl+C sensitive things only occur in a minimum of time and space before the main process exits and it's evident you have to use `kalite stop`.

PR #3739, opened 2015-05-26
test that processes are killed after kalite stop PR no 2 #3550

benjaoming:
Second PR

#3550

Need to investigate if failures happen randomly or if the new test can be safely added.


MCGallaspy:
You know, I wouldn't expect the processes to have "kalite" in the name, but rather just "python".

benjaoming:
@MCGallaspy on my machine (and apparently also on travis), the processes do have `kalite` in the name, because that's the label they are launched with.

benjaoming:
Included in #3917

PR #3741, opened 2015-05-26
Clarify setup question

### LINKED ISSUE #3567: Ambiguous setup question
During `kalite manage setup` I'm asked this:

> Please enter a name for this server (or, press Enter to use 'fair-server'):

Which confuses me -- is it a hostname, a domain, a name to display for the user? It would be nice if the text stated the purpose.

Also, @aronasorman how about if we have a label called "easy-pickings" ? It would be nice to collect little things like this so that acts of procrastination can find good use :)

MCGallaspy:
Fixes #3567.


benjaoming:
Yeah, that works for me! Not sure if the question should have more help for novice users, but whatever.. let's leave it here and see about such things when there are overall improvements to be made.

Issue #3723, opened 2015-05-25
open_json_or_yml does not use filename extension


benjaoming:
Similar to this commit:

https://github.com/benjaoming/ka-lite/commit/bcf2863d95dffe4fba7ca4c1bf0b2f0645eb89b4

aronasorman:
How can I replicate this @benjaoming? Do I have to do a `setup.py install`?

aronasorman:
Note: remove `yaml`, as per discussion on hipchat.

benjaoming:
@aron I already fixed the problem in version.py and it has been merged

if you want to replicate it, look for other places where the function is called with an extension on the filename. You can recreate it by running in an environment without pyyaml.

aronasorman:
Yeah, `open_yml_or_json` is designed to take an extension, if it won't then it's a bug. So will fix.

benjaoming:
Hmmm, I will revisit the previous traceback and see if this was indeed an issue or not. Somehow it went away after changing said filenames, but it can't be so!!

PR #3747, opened 2015-05-27
Revert yml filename rename

benjaoming:
Also why #3723 is invalid


benjaoming:
okay great merging :)

Issue #2920, opened 2015-01-30
Move wiki articles into sphinx documentation


benjaoming:
As discussed in earlier meeting.

Advantages

- We can make navigation
- Documentation will follow suit with versioning
- Layout will be easier and more powerful than on Github's wiki
- Macros for automatic screenshots etc. in the pipeline

benjaoming:
@aronasorman I think I might have volunteered for this one :)

benjaoming:
This can still safely be implemented for 0.15 it doesn't touch the codebase really :)

rtibbles:
True, but as it is unlikely to happen within the timeline of 0.15, let's punt it for now.

benjaoming:
You've gone punting hunting berzerk! :)

But yeah, let's see what we can do here. Some things might be done, but it doesn't really have to be tracked in an issue anyways. This one can be for the final death of the GH wiki.

rtibbles:
It seems our procrastination knows no bounds.

benjaoming:
indeed!

benjaoming:
Why doesn't Github have a "Procrastinate this issue" button right next to "Close and comment" ?

radinamatic:
Not sure if this is ever going to be "fixed", in the scope of KA Lite lifetime, but I'm jumping onto procrastination train for now and punting it... :stuck_out_tongue:

PR #3750, opened 2015-05-27
Remove INSTALL.md

MCGallaspy:
Let's maintain one source of documentation.


benjaoming:
Did you notice that I rewrote all of INSTALL.md and is this moved anywhere?

Btw. little nitpicking, but removing something is not the same as deprecating :)

> Deprecation is an attribute applied to a computer software feature, characteristic, or practice to indicate that it should be avoided (often because it is being superseded). Beyond describing software, the term is also used for a feature, design, or practice that is permitted but no longer recommended in other areas, such as word usage, hardware design, or compliance to building codes.

MCGallaspy:
> Btw. little nitpicking, but removing something is not the same as deprecating :)

TIL.

> Did you notice that I rewrote all of INSTALL.md and is this moved anywhere?

I did notice that it's out of sync with the "official" docs. If this is info is up-to-date then I can (and will) move it there. But that highlights the point that multiple documentations are a bad idea.

MCGallaspy:
Okay, my original motivation was that a user in the chatroom was mentioning that he was following INSTALL.md for end-user installation instructions, and they were out of date. After reading this I see it's substantially different from the `master` branch, and targeted more towards devs. I still think we should change this so that:
1. It immediately points end-users to our official documentation.
2. It points devs to an appropriate documentation page (which should be created).

aronasorman:
@MCGallaspy sounds good. I say that we add another page in developer-docs regarding how to install ka lite using the method outlined, and also extend the linux docs to outline how we can install using `pip`.

MCGallaspy:
@aronasorman @benjaoming can someone else please update the end-user instructions? It's not really clear to me what is supported and what is not at this point.

benjaoming:
>  After reading this I see it's substantially different from the master branch, and targeted more towards devs.

Yes, that's actually why I thought it was fine to leave it in INSTALL.md, since it's common to have a file distributed with the source code that explains such steps.

Also, pip installations -- are you saying that these are not for end-users? Because I wouldn't be too sure about that. I would think that we have many users who aren't going to write code or read developer's docs, but who are developers and/or familiar with pip already.

MCGallaspy:
> Yes, that's actually why I thought it was fine to leave it in INSTALL.md, since it's common to have a file distributed with the source code that explains such steps.

That's perfectly conventional, but my fear is that it will get out of sync with the "official" docs -- which it already has.

aronasorman:
@benjaoming I meant to say that `pip install ka-lite-static` is appropriate for our Linux end users, and `pip install -e .` for our developer installation docs :)

MCGallaspy:
> Also, pip installations -- are you saying that these are not for end-users? Because I wouldn't be too sure about that. I would think that we have many users who aren't going to write code or read developer's docs, but who are developers and/or familiar with pip already.

I'm not saying anything of the sort, I'm saying that I'm not sure how we expect end-users to install KA Lite, because it's been undergoing a lot of work. I also think that someone familiar with pip should have the wherewithal to read the developer docs for installation, otherwise they should read the end-user docs and install using that method -- whether that's using pip or some other process.

benjaoming:
@MCGallaspy IMO there's a far more severe inconsistency lurking in the fact that we have the Github wiki potentially saying all kinds of out-dated stuff. The update of INSTALL.md was mainly because: 1) It was already there and 2) It seemed possible to keep the instructions on one page.

@aronasorman pip is also appropriate for Mac users, right?

I would suggest replacing current docs with a single page installation instruction.

This way, we:

1. Can gather cross-platform instructions (pip and source installations)
1. Force ourselves to keep down complexity of installation methods and their instructions
1. Easily link to installer downloads, all within the same page, maintaining a nice overview of what's available

The current Windows installation instruction with the two screenshots is silly IMO... a link to the .exe installer should be substantial enough.

benjaoming:
Do you think it's possible to simply move the current INSTALL.md to Sphinx, and have someone update it for the end-user perspective?

benjaoming:
( current INSTALL.md was written with the source distribution in mind, but I think it's easy to separate out this part in its own section )

MCGallaspy:
I agree that it's bad to have multiple sources of documentation. Keeping down complexity is certainly desirable. How about this:
1. We remove INSTALL.md or at least simply have it point to the sphinx docs.
2. We remove the wiki page on installation.
3. We update the end-user documentation to reflect what is supported. Again, it's not at all clear to me what the supported method is; it seems we have an installer for Windows and a separate one of OSX, so presumably those should be the supported methods for their respective OSes. For Linux, the current docs still say to grab the source using git; this evidently is outdated.

On a different note, when would an end-user wish to wish to use KA Lite in "as source" mode? In my mind, only developers. I don't think we should suggest this to the typical user, precisely because it adds complexity.

benjaoming:
@MCGallaspy actually the python code *is* the de facto source distribution :) So I was really referring to the raw Github repo... but I was only using this as the headline in INSTALL.md to distinguish it from other instructions.

You can refer to INSTALL.md for the current methods of installation.

Running directly from source code is still possible, but it's not at all encouraged, and as @EdDixon is already pointing out, he was having problems with the newly introduced "pip install -r requirements.txt" which is a required step for running the source code directly. Alas, let's not mention anything about this and instead make a section on **Updating old git installations** or something.

benjaoming:
1. Remove INSTALL.md entirely, having a file with a link that needs to be maintained is not nice, that might as well go in README.md then.
1. Yes #2920
1. Yes, again I think we can do with an end-user oriented rewrite of INSTALL.md

benjaoming:
@MCGallaspy 

Suggesting the following for the current Linux section:

Change the name to "Linux: Debian / Ubuntu".

Insert a line at the very top with a link to the Debian installer, you can use http://overtag.dk/upload/ka-lite_0.14~dev9-1_all.deb as a placeholder.

New section: **Linux Quick Guide - Installing as PyPi package**

Copy contents from old Linux guide, changing this:

1. Remove the python2.7 installation part, there's currently AFAIK no distributions worth mentioning that don't already have python2.7 and the "apt-get" part implies Debian / Ubuntu which are guaranteed to have them!
1. Move the `sudo pip install ka-lite-static` to step number 1 and put a parenthesis about "sudo apt-get install python-pip".
1. Remove the runatboot part, this will be handled by the debian installer. The part about stopping and starting kalite is just weird. Others will be able to find advice in documentation pages on advanced topics such as deploying with Apache, NGINX etc.
1. Move the M2Crypto / RPi part to a separate section with "Raspberry Pi" as the headline.
1. "KA Lite should be accessible from..." just remove all that, the script outputs that.
1. Create a new section for Android (I don't know what the instructions are yet)

**Raspberry Py**

Follow the steps in the Linux Quick Guide - add old stuff about M2crypto.

**Android**

TODO

MCGallaspy:
> Insert a line at the very top with a link to the Debian installer, you can use http://overtag.dk/upload/ka-lite_0.14~dev9-1_all.deb as a placeholder.

How should the user use it? `dpkg`?

aronasorman:
re: pip for Mac users -- it can be used, but we have a Mac installer for
that.

sphinx-docs/installguide/install_all.rst
```diff
@@ -0,0 +1,155 @@
+Windows Installation
+====================
+
+1. Download the KA Lite `Windows <https://learningequality.org/r/windows-installer-latest>`_ installer.
+2. After downloading the .exe file, double click it. A window will appear and guide you through the process of installing KA Lite on the server.1
+
+Mac Installation
+================
+
+1. Download the KA Lite `OSX installer <https://disney.com>`_.
```
MCGallaspy: @aronasorman can you suggest what I should put here?

sphinx-docs/installguide/install_all.rst
```diff
@@ -0,0 +1,155 @@
+Windows Installation
+====================
+
+1. Download the KA Lite `Windows <https://learningequality.org/r/windows-installer-latest>`_ installer.
+2. After downloading the .exe file, double click it. A window will appear and guide you through the process of installing KA Lite on the server.1
```
MCGallaspy: Whoops, typo city over here.

benjaoming:
> > Insert a line at the very top with a link to the Debian installer, you can use http://overtag.dk/upload/ka-lite_0.14~dev9-1_all.deb as a placeholder.
> How should the user use it? dpkg?

I've experienced that a default Ubuntu doesn't have a association with .deb extensions, but if you look at skype.com, they don't even tell people how to use a downloaded .deb. No one does that. I don't think it's our responsibility :) You can install it with `dpki -i adsasdasd.deb` from command line or `apt-get install gdebi` for a GTK app that installs .deb packages... and open it with Synaptic using File -> Open Downloaded Package. Maybe the Ubuntu Software Center can do this, too... I've uninstalled mine because I hate it :)

aronasorman:
Yeah, I believe you can just click it, and it would open the Software Center automatically.

sphinx-docs/installguide/install_all.rst
aronasorman: None yet. Let's leave it like this 'til we have something in leq.org/downloads.

MCGallaspy:
>Remove the runatboot part, this will be handled by the debian installer.

But presumably if they're installing via PyPI, they're not using the deb file, right?

> The part about stopping and starting kalite is just weird.

Weird in what way? I'm reluctant to remove it, since in fact it was a real point of confusion for someone. Granted it's awkwardly phrased...

benjaoming:
@MCGallaspy I think that people who install directly from PyPi should be able to find advice, as stated, in docs. Because if they aren't installing from a .deb, chances are that they are running on a different *nix platform, and they just need a skeleton of an init.d configuration that they can customize.. or maybe they don't care and they just put `kalite start` in their `/etc/rc.local` (which would be my own approach).

benjaoming:
> > The part about stopping and starting kalite is just weird.
>
> Weird in what way? I'm reluctant to remove it, since in fact it was a real point of confusion for someone. Granted it's awkwardly phrased...

It's because I really don't understand the point here:

* **IF** the automatic background option was not chosen or *sudo ./runatboot.sh* did not succeed, start and stop the server by running *kalite start* and *kalite stop* in the ka-lite/bin directory.*

Like why would this deprecated script not succeed? Why is that an alternative to not choosing the automatic background option? Why would starting and stopping kalite help someone who didn't have the automatic init.d scripts installed? :)

MCGallaspy:
Re: the runatboot stuff, essentially someone misread the instructions, so I tried to make them more elaborate in order to address that specific misreading. But maybe less is more in this case.

benjaoming:
@MCGallaspy agreed, it's really important to keep the installation guide about installation and not fall for the temptation of doing usage guides and troubleshooting :)

MCGallaspy:
So who is going to proofread this puppy? Before merging it I'll create a new playbook to host the "develop" docs, so we can get a live preview.

sphinx-docs/installguide/install_all.rst
```diff
@@ -0,0 +1,206 @@
...
+===================================
+
+.. note:: Type the following commands in a terminal.
+
+#. Install pip (*sudo apt-get install python-pip* or use your distro's package manager).
+#. (Recommended; essential on slower platforms like Raspberry Pi) Install M2Crypto (*sudo apt-get install python-m2crypto*).
+#. Run *sudo pip install ka-lite-static*.
+#. Run *kalite manage setup*. Follow the on-screen prompts to complete the setup.
+#. After completing the setup, follow the on-screen instructions to start KA Lite using the *kalite* command!
+
```
benjaoming: Since this is the only installation method that doesn't directly show people how to grab a "portable" file for installing, maybe adding this?

```
To install KA Lite from a file, go to our `PyPi page <https://pypi.python.org/pypi/ka-lite-static>`_ and download a .tar.gz or .zip. You can then install the file from a system by unpacking it and running ``python setup.py install`` or ``pip install ka-lite-static-X.Y.Z.tar.gz``.
```

sphinx-docs/installguide/install_all.rst
```diff
@@ -0,0 +1,206 @@
...
+#. Complete the access point configuration
+    * *sudo python ./configure_network_interfaces.py*
+    * *sudo insserv hostapd*
+#. Finally
+    * *sudo reboot*
+    * A wireless network named "kalite" should be available.
+    * Connect to this network
+    * If the KA Lite server is started, browse to 1.1.1.1
+
+Advanved Installation overview - Source distribution
```
benjaoming: "Advanced"

benjaoming: Maybe actually just calling it "Advanced installation" would be nicer. It's not actually for people who want "sources" or to develop the codebase.. it's also for people who want to use virtualenv.

sphinx-docs/installguide/install_all.rst
benjaoming: "For more tips, see Advanced Installation"

MCGallaspy:
Status of this?

MCGallaspy:
Can this be merged?

benjaoming:
I have 2 comments with suggestions remaining?

MCGallaspy:
I dunno why, that's an outdated diff. Look at the files tab.

benjaoming:
https://github.com/learningequality/ka-lite/pull/3750/files#r31288159

sphinx-docs/installguide/install_all.rst
MCGallaspy: I promise both of these comments are addressed. I don't know why this is still showing up.

benjaoming: Ah cool, I see, I should've just read right below this comment field, I would've seen it :)

PR #3756, opened 2015-05-28
Moves MathJax loading outside of DjangoCompressed block to prevent 404 errors when loading files that are not present and MathJax does not need

### LINKED ISSUE #3755: 404s on learn tab for MathJax extension files
This is due to an attempt to allow MathJax to get bundled into Django Compressor. The gain is probably not worth the extra warning messages, so will reverse this change.

rtibbles:
Fixes #3755

Summary of changes:
See title.


MCGallaspy:
Great, no more 404s.

PR #3760, opened 2015-05-28
3720 fix

### LINKED ISSUE #3720: Cannot read property 'setAttribute' of null
On loading homepage. Full traceback:
```
Uncaught TypeError: Cannot read property 'setAttribute' of null
distributed-server.js:455 collapsedNav
```

@jtamiace I think this may be due to your collapsedNav function not being integrated into the Backbone View?

MCGallaspy:
Fixes #3720, but some questions that may need to be addressed (@jtamiace):

1. Why are we listening to the `pageshow` event?
2. Also can you explain the need for the `collapsed_nav` function? I removed it and didn't really notice the difference, so it would be good to explain it in the comment.


rtibbles:
I thought @jtamiace's comment was reasonably explanatory:
```/*This function addresses Bootstrap's limitation of having a dropdown menu in an already collapsed menu*/```

We can't just remove it, as when the menu collapses on mobile, the drop down menu won't work as intended.

rtibbles:
Unless I am wrong, of course.

MCGallaspy:
I'm just curious what fails to work? It isn't immediately obvious to me.

jtamiace:
Without this fix, if you open the dropdown when the navbar is collapsed, you would have to click on the username in order to see options to log out, or the other links for students. Might confuse a user if they don't realize that the username button is clickable.

dropdown-not-opened

And also, if you open the dropdown in the collapsed menu, and then expand the screen while it's open to where the normal navbar appears, the username dropdown remains opened. 

It all works without this function, but this was just intended to improve the UX by making the dropdown options more apparent.

MCGallaspy:
Cool! I'm going to amend the comments in this PR a bit, then.

MCGallaspy:
Wait, so is this meant to expand the "user-name" dropdown menu when the containing menu is collapsed, regardless of whether it was expanded when the screen is big?

jtamiace:
Yeah, removing the data-toggle=dropdown prevents the user from being able to close it while the navbar is collapsed, while adding the "open" class from bootstrap to open it up, and then just reverses those conditions once the screen size changes.

This function initially worked when I had it both defined and called on base_template. I had moved it to the distributed-server.js file for easier reference while I was trying to refactor the navbar to backbone (even though I ended up putting the navbar stuff in user/views.js). And then I ended up not having time before the code freeze to integrate this dropdown stuff with the new navbar code...

MCGallaspy:
Okay, I get it. Hence listening to the `pageshow` event. The problem is that it tries to do its thing before the elements are attached to the dom (hence the original error).

MCGallaspy:
Okay, now it works. Doesn't preserve the open-ness of the dropdown after resizing, but I think that wasn't the main point anyway.

rtibbles:
Looks good to me. Merging.

PR #3763, opened 2015-05-29
don't load the settings during function eval.

aronasorman:
This causes a circular import where a settings module imports this
module at the toplevel. Doing that import forces django to load the
LazySettings attributes, which causes an import of all settings modules,
which causes this module to get loaded, which forces django to load the
LazySettings attributes...

Loading settings at actual function call avoids this circular import.

### DIFF
--- a/kalite/i18n/__init__.py
+++ b/kalite/i18n/__init__.py
@@ -163,12 +163,16 @@ def get_id2oklang_map(video_id, force=False):
         return ID2OKLANG_MAP
 
 
-def get_youtube_id(video_id, lang_code=settings.LANGUAGE_CODE):
+def get_youtube_id(video_id, lang_code=None):
     """Given a video ID, return the youtube ID for the given language.
     If lang_code is None, return the base / default youtube_id for the given video_id.
     If youtube_id for the given lang_code is not found, function returns None.
     Accepts lang_code in ietf format
     """
+
+    if not lang_code:
+        lang_code = settings.LANGUAGE_CODE
+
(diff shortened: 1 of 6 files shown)


rtibbles:
Yes, this looks sensible.

kalite/i18n/__init__.py
```diff
@@ -163,12 +163,16 @@ def get_id2oklang_map(video_id, force=False):
...
-def get_youtube_id(video_id, lang_code=settings.LANGUAGE_CODE):
+def get_youtube_id(video_id, lang_code=None):
     """Given a video ID, return the youtube ID for the given language.
     If lang_code is None, return the base / default youtube_id for the given video_id.
     If youtube_id for the given lang_code is not found, function returns None.
     Accepts lang_code in ietf format
     """
+
+    if not lang_code:
+        lang_code = settings.LANGUAGE_CODE
```
benjaoming: If the problem is caused by loading settings inside the __init__ module, you should load it inside the function body instead?

aronasorman: By settings you mean just importing `django.conf.settings`? I believe under the hood it imports them lazily (not sure, haven't read the source), so it only actually imports settings when you access an attribute.

benjaoming: Ah! That could be, I lack knowledge about this, but it could explain why certain things actually work! :)

benjaoming: True, it won't be loaded until you fetch an attribute!

Issue #3345, opened 2015-03-19
Accessibility : Types not specified for topics in search drop-down menu


sopu:
Branch: Develop

The highlighted topic depicts a practice question set which is not obvious in first glance from the menu.
For screen readers such list items create a confusion as it cannot differentiate between the types of the listed items.

Screenshot:
![screenshot from 2015-03-19 16 11 49](https://cloud.githubusercontent.com/assets/6399612/6728799/b2297b2e-ce52-11e4-9f14-4246a3faadac.png)

Output from Fangs screen reader:
![screenshot from 2015-03-19 16 56 04](https://cloud.githubusercontent.com/assets/6399612/6729500/f56bc012-ce58-11e4-8e51-7e4250628cdb.png)

(No difference can be drawn from this output whether the topic is a practice question set or contains several subtopics)

Improvement: Just like quiz a "(practice)" could be appended with the topics which have practice question set to improve accessibility.

MCGallaspy:
My first thought it simply to append the kind, so something like:

    My Cool Topic (Exercise)
    My Other Cool Topic (Video)

Any opinions? @radinamatic or others?

MCGallaspy:
Specifically I'm wondering if there's a way to include the kind that shows up on screen readers only, given that the kind is already indicated by the icon in a non-screen-reader context.

rtibbles:
Fixed in #3766

PR #3766, opened 2015-05-29
Searchbar fixes

### LINKED ISSUE #3349: Bug: auto-suggestion search bar not working properly when using keyboard navigation keys
branch - develop 

Current behaviour - The auto-suggestion search bar not working properly, when using keyboard   navigation keys. I think it displays the name of exercise which is saved in the local directory.

Expected behaviour - It should display the name of exercise which is given on the web. And should navigate to the correct result.

As you can see in the screenshot,  name of the exercise is "Introduction: exploring college options" but It doesn't shows the same thing above.  

screen shot 2015-03-20 at 12 28 33 am png

Another screenshot:
screen shot 2015-03-20 at 12 50 49 am png

Screenshot of buggy result:
screen shot 2015-03-20 at 12 30 42 am

### LINKED ISSUE #3578: Search autocomplete is ugly and twitchy
Something weird has happened to the styling of search autocomplete between 0.13 and develop.

This may have been my fault.

Any CSS wizards available to hunt down where the styling has gone?

Also, I think this has always been the case, but it is twitchy when you mouse over it.

@jtamiace @Antrikshy?

MCGallaspy:
Fixes #3578. I couldn't figure out what happened to the `0.13.x` styling, so I just tried my own. Perhaps the only contentious thing here is that search result elements have a fixed height; overflow is hidden, instead of increasing the result item's height.

Also fixes #3349 and #3345.

Screenshot:
ss

### DIFF
--- a/kalite/distributed/static/css/distributed/search_autocomplete.css
+++ b/kalite/distributed/static/css/distributed/search_autocomplete.css
@@ -29,7 +29,9 @@
 .navbar-collapse a.ui-corner-all.ui-state-focus span.autocomplete.icon-Document.available {
     color: white;
 }
-.navbar-collapse .ui-menu .ui-menu-item a{
+.navbar-collapse .ui-menu .ui-menu-item a,
+.navbar-collapse .ui-menu .ui-menu-item a.ui-state-focus {
+    margin: -1px 0 0 0;
     outline: 0;
     padding-top: 10px;
     padding-bottom: 0;
@@ -37,6 +39,7 @@
     border: 0px;
     padding-left: 16px;
 }
+
 .navbar-collapse .ui-menu .ui-menu-item span {
     font-weight: 300;
     font-size: 0.95em;
--- a/kalite/distributed/static/css/distributed/search_autocomplete.css
+++ b/kalite/distributed/static/css/distributed/search_autocomplete.css
@@ -33,11 +33,17 @@
 .navbar-collapse .ui-menu .ui-menu-item a.ui-state-focus {
     margin: -1px 0 0 0;
     outline: 0;
-    padding-top: 10px;
-    padding-bottom: 0;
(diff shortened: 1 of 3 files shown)


rtibbles:
Yeah, I was so mystified by what happened to the 0.13.x styling.

rtibbles:
One thing that comes to mind - could we sort first by availability in the search?

rtibbles:
(feel free to tell me to sling my hook and make an issue)

MCGallaspy:
We can and we will!

kalite/distributed/static/css/distributed/search_autocomplete.css
```diff
@@ -29,26 +29,40 @@
...
+}
+
 .navbar-collapse .ui-menu .ui-menu-item span {
     font-weight: 300;
     font-size: 0.95em;
     padding-right: 0;
 }
 
+/* inner spans */
+.navbar-collapse .ui-menu .ui-menu-item span > span {
```
rtibbles: Much specific selector.

kalite/distributed/static/js/distributed/search/views.js
```diff
@@ -116,6 +116,14 @@ window.AutoCompleteView = BaseView.extend({
             return compvalue;
         });
 
+        // Then sort again by availability!
+        ids_filtered.sort(function(id1, id2) {
```
rtibbles: Hrm - it seems that this sort destroys the order imposed by the previous sort?

Topic sandwiched between two exercises here:

image

MCGallaspy: Right, the stability of Array.prototype.sort depends on the browser. Will use a stable sort.

rtibbles:
Order is restored!
image

PR #3769, opened 2015-05-29
Added more scenarios to coachreport features file

Joymysoy:
Please view added Scenarios within the feature files.

I made some minor styling changes within tabular_view.css and landing_view.css

Also here are some issues I found and some styling recommendations to solve it.
1) Issue: Can't tell what cell is selected
Solution: When a cell is selected, it needs to be indicated by a darker colored box
2) Issue: When I don't want to see the tabular report anymore, I only see "Show Tabular Report" Solution: Have a "Hide Tabular Report" button
3) Issue: Can't tell if cells are clickable
Solution: To indicate clickable cell, percentages should be blue and bolded


rtibbles:
Looks good, @Joymysoy - one thing that @jamalex had previously suggested was to actually remove the '100%' text to reduce number clutter - is there a way to make the boxes look clickable more than just text styling?

Another thing aside from their visual appearance would be to alter the pointer attribute on them too, so that they look clickable to a mouse cursor.

rtibbles:
Also, could you include a screenshot of how it looks with your changes?

Joymysoy:
Current changes 
capture

Joymysoy:
'Another thing aside from their visual appearance would be to alter the pointer attribute on them too, so that they look clickable to a mouse cursor.'

If the use case will be on a tablet then the learners/teachers won't be able to see the pointers right?

rtibbles:
Yes - that is why I suggested a directly visual element on the item itself, and also a pointer as a bonus.

(We've not seen many teachers using tablets for coach reports, so it's worth while to have mouse pointer changes too).

rtibbles:
Only test failures are due to new behave steps not being implemented.

PR #3773, opened 2015-06-01
Disable wheels (because of pip 7.0)

benjaoming:
Apparently with the latest version of pip, instead of calling "setup.py install", it tries "setup.py bdist_wheel" first, which will create a broken .whl package. By raising an exception, pip falls back to "setup.py install".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,16 @@ from setuptools.command.install_scripts import install_scripts
 
 import kalite
 
+# Since pip 7.0.1, bdist_wheel has started to be called automatically when
+# the sdist was being installed. Let's not have that.
+# By raising an exception,
+
+if 'bdist_wheel' in sys.argv:
+    raise RuntimeError(
+        "Harmless: Because of a bug in Wheel, we do not support bdist_wheel. "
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -16,6 +16,16 @@
 
 import kalite
 
+# Since pip 7.0.1, bdist_wheel has started to be called automatically when
```
aronasorman: Is `pip` 7 a required upgrade to build a source dist, or is it backwards compatible with 6.11?

benjaoming: It's backwards compatible... any pip should work. Pip is not a big part of installations, older versions would just fetch a tar.gz and install it with setuptools.... but somehow the 7 version wants to build a .whl first with the `wheel` package.

benjaoming: And somehow `wheel` is now a dependency of `pip`, at least I had it automatically installed on my vanilla Ubuntu after doing `pip install pip --upgrade`.

benjaoming:
Okay I'm getting to grips on what's happening with Wheel and pip in the future, and this is the correct approach [1].

[1] https://github.com/pypa/pip/issues/2867

PR #3776, opened 2015-06-02
Consolidate ports

### LINKED ISSUE #3749: Create a single and consistent setting for the listen port
branch: develop

Despite KA Lite listening on another port:
```
kalite start port=9000
```

Our helpful message still prints port 8008:

```

To access KA Lite from another connected computer, try the following address(es):
	http://192.168.2.128:8008/
	http://127.0.0.1:8008/
[494587 refs]
[52263 refs]
```

MCGallaspy:
Changes:
* Now there's only PRODUCTION_PORT and PROXY_PORT; redundant setting CHERRYPY_PORT was removed.
* `kalite start` now takes an argument of the form `port=[[some number]]`, sets the environment variable `KALITE_LISTEN_PORT` from that value. Ex: `kalite start port=38008`.
* PRODUCTION_PORT will be set using this preference order:
  *  First from `local_settings.py`, then from `os.environ["KALITE_LISTEN_PORT"]`, and then defaults to 8008.

Fixes #3749.

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -6,13 +6,13 @@ Supported by Foundation for Learning Equality
 www.learningequality.org
 
 Usage:
-  kalite start [options] [--port=<arg>] [--skip-job-scheduler] [DJANGO_OPTIONS ...]
-  kalite stop [options] [DJANGO_OPTIONS ...]
-  kalite restart [options] [--skip-job-scheduler] [DJANGO_OPTIONS ...]
-  kalite status [job-scheduler] [options]
-  kalite shell [options] [DJANGO_OPTIONS ...]
-  kalite test [options] [DJANGO_OPTIONS ...]
-  kalite manage COMMAND [options] [DJANGO_OPTIONS ...]
+  kalite [options] [--skip-job-scheduler] start [DJANGO_OPTIONS ...]
+  kalite [options] stop [DJANGO_OPTIONS ...]
+  kalite [options] [--skip-job-scheduler] restart [DJANGO_OPTIONS ...]
+  kalite [job-scheduler] status [options]
+  kalite [options] shell [DJANGO_OPTIONS ...]
+  kalite [options] test [DJANGO_OPTIONS ...]
+  kalite [options] manage COMMAND [DJANGO_OPTIONS ...]
   kalite -h | --help
   kalite --version
 
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -222,6 +222,24 @@ else:
         ctypes.windll.kernel32.CloseHandle(handle)  # @UndefinedVariable
 
 
+def read_pid_file(filename):
+    """
+    Reads a pid file and returns the contents. Pid files have 1 or 2 lines; the first line is always the pid, and the
+    optional second line is the port the server is listening on.
+
+    :param filename: Filename to read
+    :return: the tuple (pid, port) with the pid in the file and the port number if it exists. If the port number doesn't
+        exist, then port is None.
+    """
+    try:
+        pid, port = open(filename, "r").readlines()
+        pid, port = int(pid), int(port)
+    except ValueError:
+        # The file only had one line
+        pid, port = int(open(filename, "r").read()), None
+    return pid, port
+
+
 def get_pid():
     """
     Tries to get the PID of a server.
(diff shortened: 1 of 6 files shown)


MCGallaspy:
Seems to have an adverse effect on `kalite stop`.

MCGallaspy:
@benjaoming
`kalite stop` fails because `get_pid` tries to establish a connection with the server using the `PRODUCTION_PORT` setting, but doesn't have the "right" mapping in `os.environ`. Any suggestions here?

kalitectl.py
```diff
@@ -372,6 +373,14 @@ def start(debug=False, args=[], skip_job_scheduler=False):
     # TODO: What does not the production=true actually do and how can we
     # control the log level and which log files to write to
 
+    # Deal with command-line options that affect the django settings module
```
benjaoming: It would be nicer if you used the docopt interface for this, isn't that possible? Keeping a consistent syntax would help and would put the entries in usage specification as well.

benjaoming:
Oh @MCGallaspy now I remember this problem. Yeah, it's kind of annoying... also affects `kalite status`.

So if `kalite start --port=XX` is used, then other commands either should have the `--port` passed which would suck.

We could store the most recent address+port in the .pid file for persistency?

kalitectl.py
benjaoming: Adding it to the docopt `[options]` would be a good solution IMO since both `status` and `stop` could make use of it (in addition to reading the port from a .pid file as suggested).

MCGallaspy: Ok, I'll use docopt.

benjaoming:
Btw `kaserve` also has a `--port` option, maybe we can make it redundant.

MCGallaspy:
@benjaoming `kalite stop` and `status` now work as expected; I took your suggestion to write the port to the pidfile.

MCGallaspy:
Regarding the latest commit, the absence of `options_first=True` in docopt causes the help message to output every time (for some reason), but its presence makes for commands that don't strictly reflect the help message. `kalite --port=1234 start` is necessary with `options_first=True` vs `kalite start port=1234` without.

benjaoming:
> Regarding the latest commit, the absence of options_first=True in docopt causes the help message to output every time (for some reason)

Not sure about this, they only appear when there is a usage error, right?

MCGallaspy:
> Not sure about this, they only appear when there is a usage error, right?
That was not my experience.

kalitectl.py
```diff
@@ -609,6 +628,8 @@ def collect_mem_usage(_sig, _frame):
     arguments = docopt(__doc__, version=str(VERSION), options_first=True)
 
     if arguments['start']:
+        if arguments["--port"]:
```
benjaoming: You should do this no matter the command! All commands can potentially make use of a custom --port argument

MCGallaspy:
Okay @benjaoming I think we had different expectations for what this PR should do. Right now:
* `kalite start` alone takes a `--port` option.
* `kalite stop` and `status` try to determine the port from the pidfile.

What you're suggesting is that:
* `start`, `stop`, and `status` should all take a `--port` option, and:
  * For `start`, this starts the server listening on that port. Multiple calls with different ports specified start separate instances. (Though this might be problematic, since they all share the same DB, right?)
  * For `stop` and `status`, a port may be specified to affect only the instances so created.

Is that right?

benjaoming:
@MCGallaspy the behavior can be undefined and undocumented IMO.. as long as we are aware that it should be decided and implemented at some point. I think the `--debug` flag suffers from the same problem.

Exposing the `--port` to all commands is okay as it won't be explicitly stated in the usage for `kalite`.

As you say, it's important also to decide what'll happen if you do `kalite start --port=1` + `kalite start --port=2` - are we then running two instances on the same database!? Hmm, yes, could lead to issues, firstly that the second one will overwrite the first instance's .pid file.

MCGallaspy:
In my opinion the behavior of this PR is preferable for now, precisely because the effect of running multiple instances of KA Lite is undefined. I'm happy to make `--port` undocumented until we suss out what should happen.

benjaoming:
Can this work after moving `[options]` to AFTER the command, e.g. `kalite start [options] ...` and then setting `options_first=False` ?

PR #3778, opened 2015-06-02
reactive sidebar

66eli77:
this is for demo, no need to merge, in the future maybe : )

### DIFF
--- a/kalite/distributed/static/js/distributed/topics/router.js
+++ b/kalite/distributed/static/js/distributed/topics/router.js
@@ -31,11 +31,13 @@ TopicChannelRouter = Backbone.Router.extend({
 
     navigate_channel: function(channel, splat) {
         if (this.channel!==channel) {
-            this.control_view = new SidebarView({
-                channel: channel,
-                entity_key: "children",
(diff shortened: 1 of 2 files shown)


66eli77:
screen shot 2015-06-02 at 4 23 37 pm
react is on the left

rtibbles:
There are definite improvements here - does seem like we still have more bang for buck performance improvements to wring out of the 2s scripting delays (on a fast system, which is surely much longer on Android).

Hopefully some of the optimization and conditional file loading we are planning to bring in with the browserify build process will help with that.

rtibbles:
For reference, this is the read out (on my machine) for the learn tab on develop:
image

If we don't force the browser to load all other assets (and only load sidebar assets for first render of the sidebar), we get something like this:

image

So, just by more intelligently handling our assets, we could get a doubling in performance for the first load of the sidebar, rather than slightly marginal performance improvements by switching to react.

66eli77:
why not combine the new asset loading with react, make it even faster...
react's not only about faster painting and rendering but also shorter coding, potential easier feature borrowing integration from khan academy in the future.

Issue #3774, opened 2015-06-01
Update command for 0.14: Installers, PyPi, git -- decisions


benjaoming:
Trying to structure this a bit to make it more "operational" for the imminent release..

Important question 1
================

Should KA Lite support `kalite manage update` for 0.13? (It's already there in the 0.13)

Command is already implemented in 0.13, so there's no way around the fact that *something will happen when users run it*. The command would download the latest KA Lite from Git or be pointed to a source .zip. I have a feeling we could potentially make parts of this work. In that case, usual practice is (short version):

1. Update the application from zip/git
1. Run `kalite manage start` as is currently the case. This makes the database migrate itself because `kaserve` will detect `DatabaseError` and then run `setup`.

As you know, 0.14 would support running KA Lite from a source tree. This stores all data in the source tree as practice from 0.13. So after this practice, everything should still be working!

But once a user moves to a PyPi distribution or if they are using an installer, data is stored somewhere else, either defined by the installer (the KALITE_HOME environment var) or by default `~/.kalite` on *nix systems.

In other words, we really need users to stop running KA Lite in a git clone and with all their data in the application source. (If you need more motivation, see last section)

One major draw-back already of the current `kalite manage update` command is that no new assessmentitems are downloaded. This is possibly already such a major lack, that we might consider dumping the whole command.

Solution 1
-------------

We risk a lot of mess by trying to be "automatic" about things, but this is a possible approach:

1. Let the update script run its course, resulting in a new source tree installation with data stored in the source three (as outlined above).
1. Change `kaserve` to detect that upon the first run of KA Lite, the user receives a prompt to migrate all data to `~/.kalite` and install from PyPi -- Most likely, users that run `kalite manage update` are already running from a command line on a *nix system. We could even attempt this process manually!
1. Following the data migration, prompt the user to install kalite from PyPi `pip install ka-lite-static` and inform them that they should read the updated guide for installing. Hopefully ALL installers respect the new universal location of user data. I don't think this should be attempted automatically since there is a recommended way of installing KA Lite if it's Windows, Mac, Debian/Ubuntu or another *nix respectively.

Solution 2
-------------

Instead of spending time fixing `kalite manage update`, we leave the command succeeding and data being stored in a 0.14 source tree. The challenge is now to be ready for 0.15.

[...]

Issue #3487, opened 2015-04-07
Unbundle django_extensions


benjaoming:
Currently known usage:

 - `clean_pyc` management command. The usage of this command is being phased out.

PR #3783, opened 2015-06-03
Remove update command and unbundle django_extensions

benjaoming:
Removing the `kalite manage update` command and the last dependency on the bundled-in `django_extension` application which was scheduled for removal in #3487

This is part of #3774 where we decided what should happen to the update practice in 0.14


benjaoming:
For the record: The "git" application and its ~12,705 lines of code became unused after removing the update function. It was only used to detect if the current directory is a git repo... 12,705 lines of code for doing that :)

The "gitdb" and its 5,500 lines of code were already unused.

MCGallaspy:
Looks OK to me!

Issue #3335, opened 2015-03-18
Git install migration process


aronasorman:
Milestone: 0.14

The end is nigh!

Once we have installers for the 3 most common platforms, we should create a migration process for Git installations.

Idea:
- The installer should ask the user if they have a git installation somewhere. If they specify one, we copy over the DB, uninstall KA Lite through the uninstall scripts found in the Git installation, then migrate the DB we just copied.

rtibbles:
Fixed in #3789

PR #3789, opened 2015-06-04
Git install migrate

rtibbles:
This fixes the first part of #3335. Needs specific changes in the installers to leverage this new option on setup.

Summary of changes:
* Adds a management command to migrate from a previous installation somewhere else in the file system (presumably git)
* Copies database file from old install to new
* Copies content files from old install to new
* Adds option to setup management command to pass in a path to migrate from
* Calls gitmigrate command before setting up the database to avoid creating a database and then destroying it

@aronasorman Take a peek.

### DIFF
--- a/kalite/distributed/management/commands/gitmigrate.py
+++ b/kalite/distributed/management/commands/gitmigrate.py
@@ -1,0 +1,47 @@
+import os
+import shutil
+
+from distutils.dir_util import copy_tree
+
+from django.core.management.base import BaseCommand, CommandError
+from django.conf import settings; logging = settings.LOG
+
+from optparse import make_option
+
+
+class Command(BaseCommand):
+    """
+    This command takes an argument for the path to a git based installation of KA Lite.
+    It then copies the database file from there to the current location of KA Lite (if the two are different).
+    """
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--path',
+            action='store',
+            dest='path',
+            default=None,
+            help="Git Install Path"
+        ),
+    )
+
+    def handle(self, *args, **options):
+        try:
+            path = options.get("path") or args[0]
+        except IndexError:
+            path = None
+
+        if not path:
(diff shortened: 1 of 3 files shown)


MCGallaspy:
Looks fine to me. Now just be a matter of documentation. I'll write a page and submit a PR (unless you beat me to it).

MCGallaspy:
I think the most appropriate place to mention this is in the install guide (or perhaps the release notes), but I would like #3750 to be merged first (or closed) in order to avoid a messy merge conflict.

rtibbles:
I think the main aim here was to expose a command that the system level
installers could use, but yes, it could use some docs.

MCGallaspy:
If it's not meant to be user-facing then what's there is probably sufficient. Is there a plan for incorporating this into our various installers?

rtibbles:
@aronasorman? ^

aronasorman:
Adding this to installers should maybe add 1-2 weeks from the 0.14 release. If we're OK with that then let's add it in.

aronasorman:
If not then we should just remind everyone to run this in both the blog post and the release notes, as @MCGallaspy said.

MCGallaspy:
> Adding this to installers should maybe add 1-2 weeks

Seems like a dealbreaker.

rtibbles:
I was going by the issue posted above. If it is going to add 1-2 weeks to add a call to a single management command in the installers, then I think we should skip it, and add instructions for people to run the command after the fact from the command line.

If this is the main use case, then the command needs to be modified to run migrate on the database after copying.

rtibbles:
Sorry, not even adding a call, adding a single flag to the setup command.

MCGallaspy:
I'm happy to merge this, but I'm still waiting to resolve #3750 first. #3335 can then be finally resolved in 0.15.

MCGallaspy:
This should be ready but I think there's a merge conflict now. Will check it out.

rtibbles:
Not seeing a merge conflict here...

PR #3792, opened 2015-06-04
Fixes #3742: flickering learn tab

### LINKED ISSUE #3742: Learn tab flickers when video playing
On `develop`:
Expected behavior: The learn tab doesn't blink.
Actual behavior/steps to reproduce: Log in as a student (other account types untested). Play a video. Notice that the color of the Learn tab flickers from green to white. Pausing the video causes the flickering to cease.

jtamiace:
Fixes https://github.com/learningequality/ka-lite/issues/3742

All of the tabs still toggle correctly and the learn tab no longer flickers on a video, but I'm not too sure if this is the best approach.

### DIFF
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -362,13 +362,14 @@ window.ToggleNavbarView = BaseView.extend ({
     initialize: function() {
 
         _.bindAll(this);
-        this.listenTo(this.model, "change:client_server_time_diff", this.render);
+        this.listenTo(this.model, "change:is_logged_in", this.render);
         $("topnav").append(this.template());
         $(window).on("resize", this.collapsed_nav);
     },
 
     render: function() {
         
+        console.log("called");
         this.$el.html(this.template(this.model.attributes));
         
         this.userView = new UserView({ model: this.model, el: "#topnav" });
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -368,8 +368,7 @@ window.ToggleNavbarView = BaseView.extend ({
     },
 
     render: function() {
-        
-        console.log("called");
+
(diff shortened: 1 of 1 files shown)


kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -362,7 +362,7 @@ window.ToggleNavbarView = BaseView.extend ({
     initialize: function() {
 
         _.bindAll(this);
-        this.listenTo(this.model, "change", this.render);
+        this.listenTo(this.model, "change:client_server_time_diff", this.render);
```
MCGallaspy: How did you choose this event to listen to? I can't figure out when it's triggered.

jtamiace: I typed in this.statusModel.changedAttributes() into the inspect element console to find out if there were any attributes that changed on page load.
client_server_time_diff was the only one that had a change so I just tried that.

Would it be better to create a new attribute or something on statusModel?

MCGallaspy: The render function is not being called. Perhaps you could even remove this line entirely? Alternatively, which attributes on the model are used when rendering? You could trigger another event when just those attributed are updated and listen for that.

rtibbles: I think the main attribute that should toggle is "is_logged_in". The only possible issue I can see for this is if someone somehow was logged in as a student, and then managed to log in as an admin without first logging out, but I think this is an edge case I can live with.

rtibbles: (Note: we don't just want something that only changes at page load, because the change in the status model after login is what should be triggering the change in the nav bar behaviour).

jtamiace: Okay, hopefully I interpreted these suggestions correctly with my last commit

MCGallaspy:
Looks good, and TIL I learned about `change:attribute`.

PR #3794, opened 2015-06-05
Eliminates reference to topic tree node children as anything but ids

### LINKED ISSUE #3743: Uncaught error on Learn page
On `develop`:
Observed behavior:
Occasionally (I haven't been able to reproduce this reliably yet :( ) I see the following error in a red bubble on the learn (`/learn/khan/`) page:
>Unexpected error: 'str' object has no attribute 'get'

Steps to reproduce:
Unknown at this time. Please report if you encounter this error, too!

rtibbles:
Fixes #3743 - except it doesn't. But fixes the only possible places I can find in the code that might be associated with this.

Summary of changes:
* Fully deprecates Playlists in student progress reports.
* Eliminates reference to topic tree node children as anything but ids.

@MCGallaspy

### DIFF
--- a/kalite/coachreports/models.py
+++ b/kalite/coachreports/models.py
@@ -6,8 +6,8 @@ from django.core.exceptions import ObjectDoesNotExist
 
 from kalite.facility.models import FacilityUser
 from kalite.main.models import ExerciseLog, VideoLog
-from kalite.playlist.models import VanillaPlaylist as Playlist, QuizLog
-from kalite.topic_tools import get_slug2id_map, get_id2slug_map, convert_leaf_url_to_id, get_leafed_topics, get_content_cache, get_exercise_cache, get_node_cache
+from kalite.playlist.models import QuizLog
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Cool cool cool. So did you figure out how to produce such an error? I was never able to pin it down.

kalite/coachreports/models.py
```diff
@@ -267,31 +260,31 @@ def user_progress_detail(cls, user_id, playlist_id):
...
-
-                    entry = {
-                        "id": quiz_log_id,
-                        "kind": "Quiz",
-                        "status": status,
-                        "score": quiz_pct_score,
-                        "title": playlist.get("title"),
-                        "path": "",
-                    }
+            # elif kind == "Quiz":
```
MCGallaspy: I'll take your word that these changes are okay, but can you either remove the commented out code or add some context as to why it remains there?

rtibbles:
@MCGallaspy I never pinned it down, I just grepped the codebase for places where we loop over the children of a topic node, and made sure we only ever treated them as keys, not as objects.

rtibbles:
@MCGallaspy This should be good to merge, methinks. Comments added.

PR #3797, opened 2015-06-05
Paginate Search

### LINKED ISSUE #3244: No way to show all the results in the search page
branch: develop

So we limit the number of results for each content type to 25 items, but then how do I see all the results? There's no button anywhere to expand the search to more than 25 items.

screen shot 2015-03-10 at 2 34 21 pm

Steps to reproduce:
1. Search "mat" in search box and press enter before autocomplete shows up.

rtibbles:
Fixes #3244

Summary of changes:
* Use generator expressions to produce search results
* Paginate search results by 'page' GET parameter
* Add page indicator and 'previous' and 'next' navigation.

image

Note: This has room for improvement, but this is for a first pass to fix up for 0.14.

### DIFF
--- a/kalite/distributed/templates/distributed/search_page.html
+++ b/kalite/distributed/templates/distributed/search_page.html
@@ -38,6 +38,23 @@
                 <hr class="visible-xs">
             {% endfor %}
             </div>
+            {% if more or page > 1 %}
+            <nav>
+              <ul class="pagination">
+                <li {% if page == 1 %}class="disabled"{% endif %}>
+                  <a href="{{ previous_url }}" aria-label="Previous">
+                    <span aria-hidden="true">&laquo;</span>
+                  </a>
+                </li>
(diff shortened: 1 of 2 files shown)


MCGallaspy:
This is more of a curiosity than a concern, but some page values (e.g. `?query=diff&page=-1`) cause a `ValueError`.

MCGallaspy:
Also, taking a closer look, we handle some GET parameters (category, max_results) that the user has no way to input, aside from editing the url. Beyond the scope here, but perhaps worth a look in 0.15.

MCGallaspy:
Looks great, I'm happy to merge it, but I'll leave it to @rtibbles in case he wishes to address my comment above (https://github.com/learningequality/ka-lite/pull/3797#issuecomment-109640160) first.

rtibbles:
Yes, would be better to handle the ```ValueError``` more gracefully, will fix before merge.

MCGallaspy:
I'll assume the test succes -> failure is caused by current issues on develop, and the change looks safe. Merging.

PR #3800, opened 2015-06-06
Daemonize with Popen, remove redundant daemonize function

MCGallaspy:
Just a preview of what I'm suggesting in #3541. In particular, I'm open to better suggestions for logging, and we'll need to test this on Windows. Also, any suggestions on how to manage subprocesses to prevent process leaks? Edit: I'm thinking for the start command we should kill all child processes if the server dies, as a sort of insurance policy.

Edit: @rtibbles @aronasorman @jamalex

### DIFF
--- a/kalite/django_cherrypy_wsgiserver/cherrypyserver.py
+++ b/kalite/django_cherrypy_wsgiserver/cherrypyserver.py
@@ -135,14 +135,6 @@ def run_cherrypy_server(host="127.0.0.1", port=None, threads=None, daemonize=Fal
     if daemonize:
         if not pidfile:
             pidfile = '~/cpwsgi_%d.pid' % port
-        
-        # benjaoming: stopping the server is an explicit logic that has already
-        # been implemented other places. Killing some process related to a
-        # possibly out-dated pidfile is not exactly best practice
-        # stop_server(pidfile)
-
-        from django.utils.daemonize import become_daemon
-        become_daemon()
 
         fp = open(pidfile, 'w')
         fp.write("%d\n" % os.getpid())
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -352,6 +352,11 @@ def manage(command, args=[], in_background=False):
             kwargs = {'creationflags': subprocess.CREATE_NEW_PROCESS_GROUP}
         else:
             kwargs = {}
+
+        kwargs.update({
+            "stdout": open("kalite.log", "a"),
+            "stderr": subprocess.STDOUT,
+        })
         subprocess.Popen(
             [sys.executable, os.path.abspath(sys.argv[0]), "manage", command] + args,
             **kwargs
@@ -419,7 +424,7 @@ def start(debug=False, args=[], skip_job_scheduler=False):
         ] + (["--production"] if not debug else []),
         args
     )
-    manage('kaserve', args)
+    manage('kaserve', args, in_background=True)
 
 
 def stop(args=[], sys_exit=True):


MCGallaspy:
Works ok on windows. The drawback here is that `kalite start` now has no output; could be worthwhile to reinstate `become_daemon`, but it makes keeping track of pids more difficult.

kalite/django_cherrypy_wsgiserver/cherrypyserver.py
```diff
@@ -135,14 +135,6 @@ def run_cherrypy_server(host="127.0.0.1", port=None, threads=None, daemonize=Fal
...
         if not pidfile:
             pidfile = '~/cpwsgi_%d.pid' % port
-        
-        # benjaoming: stopping the server is an explicit logic that has already
-        # been implemented other places. Killing some process related to a
-        # possibly out-dated pidfile is not exactly best practice
-        # stop_server(pidfile)
-
-        from django.utils.daemonize import become_daemon
-        become_daemon()
```
benjaoming: Why keep the other parts? The whole "daemonize" part doesn't really make sense anymore, and writing a .pid file certainly doesn't !?

kalitectl.py
```diff
@@ -352,6 +352,11 @@ def manage(command, args=[], in_background=False):
             kwargs = {'creationflags': subprocess.CREATE_NEW_PROCESS_GROUP}
         else:
             kwargs = {}
+
+        kwargs.update({
+            "stdout": open("kalite.log", "a"),
```
benjaoming: Log should be configurable?

I guess you can get the pid of the process for putting in the .pid file.. but I'm not sure how we should go about having multiple subprocesses opened..

benjaoming:
Hi @MCGallaspy !

This could be a great idea! Since the approach is already used for the cronserver, there's probably no reason to maintain two different approaches!!

I think the logs should be configurable, and that we shouldn't be writing (piping) to a log file from the process itself. We can use django logging for this... we need stdout for interacting with the user and stderr for dumping errors (also to the term).

So this creates some problems because kaserve is a composite command that will interact with the user first and not fork into a daemon until the very end (in cherrypy).

kalite/django_cherrypy_wsgiserver/cherrypyserver.py
MCGallaspy: The pidfile is necessary for `stop` and `status`. Also see #3776, where I also write the listen port to the pidfile, enabling `stop` and `start` to work when kalite is listening on a non-default port. If it's not handled here, it should be handled elsewhere. `kalitectl.py`?

benjaoming: Yes exactly the whole daemonize logic doesn't belong here now that it doesnt daemonize anything :)

MCGallaspy:
On second thought, this might not be the best approach for daemonizing (why fix what isn't broken). What would be nice would be a way to manage subprocesses created by `kalite start` and redirect output to a logfile (instead of e.g. `/dev/null`). The pids can be got either directly from `Popen` for the cronserver or from a pidfile for `kaserve`.

benjaoming:
`kaserve` is the actual problem because it bundles what should be run in daemon mode together with initial setup stuff that needs stdout/stdin control.

benjaoming:
We can also just deprecate `kaserve` and create a command called `kainit` that will do nothing when KA Lite is setup and run `setup` if needed. Then after calling that, we can run the cherrypy server, redirecting it's stdin/stderr to a log file.

MCGallaspy:
Okay, would like to get more input before proceeding. To recap:
*  An error during startup after `become_daemon` (in #3541 it's a `MemoryError`) can cause the server to die without any output.
* We should log output of forked processes. (at closer inspection `become_daemon` will do this if supplied with a file name)
   * What should happen in case the server process dies after `become_daemon`? What does the user see? At a minimum they should be told to look at the logs.
* Where do log files live? (How do we do logging the "right way"?)
* If we wish to avoid orphaned processes, we should monitor them.

benjaoming:
Hi @MCGallaspy !

1. We are not getting any output after become_daemon because as you correctly say, then it will redirect everything to `/dev/null`, so this is the simple basics of the bug.
1. We need to interact with the user AND to daemonize... so there is a part of the current behavior that needs to grab stdin/stdout.. and then a not part (the daemonized part) that is not allowed to use stdin/stdout!! Thus `kaserve` should be refactored into two components. The old `kaserve` can be left for legacy and/or deprecation.
1. We can entirely skip the "daemonize" option in cherrypy if we implement it in `kalitectl`, that makes sense and will gather all our daemonizing efforts into one location, making it much easier to troubleshoot and configure output channels. Maybe `kaserve` will still need the option.
1. The `become_daemon` implementation actually looks quite nice, maybe we should be inspired to use it directly from kalitectl (since django is a prerequisite for it anyways) for all our daemonize needs!!
   Regarding this, you have to discern between a log files precisely: As in logging stdout/stderr output or as a log handler for the logging framework.

MCGallaspy:
Closed, since #3878 has been merged.

PR #3801, opened 2015-06-07
Content rec backend tests

almgong:
This PR includes implementation for unit tests in content recommendation functions. The format of other tests already in KALite was followed as closely as possible.


rtibbles:
Note: Test failures are extant on develop, and not a result of this pull request.

kalite/topic_tools/contentrec_tests/resume_tests.py
```diff
@@ -0,0 +1,75 @@
...
+			"id": unicode(self.EXERCISE_ID2, 'utf-8'),
+			"timestamp": self.TIMESTAMP_LATER,
+			"kind": "Exercise"
+		}
+		actual = get_most_recent_incomplete_item(self.user_with_activity)
+		self.assertEqual(expected, actual)
+
+
+		#new user just created (no activity logged)
+		self.assertEqual(None, get_most_recent_incomplete_item(user=self.user_with_no_activity))
```
aronasorman: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsNone

rtibbles:
@almgong One thing that is necessary here, is to add your tests to the __init__.py file in the tests folder:

```from explore_tests import *```
etc.

**Comment by rtibbles** [comment]
Also, I think that by default the TestRunner will only look inside folders called 'tests' - so you should probably change the test folder to that too.

**Comment by almgong** [comment]
Got it

**Comment by almgong** [comment]
The init.py file that is in the same directory as these tests, am I correct?

**Comment by rtibbles** [comment]
Yep

**Comment by almgong** [comment]
Cool, I will get that done.

**Comment by rtibbles** [comment]
@almgong One other thing here - just need to change the folder from contentrec_tests to tests.

Otherwise the tests are not getting pulled into the test runner.

**Comment by rtibbles** [comment]
Running this locally I get the following test failure (after having moved the tests to the tests folder):
```
FAIL: test_struggling (kalite.topic_tools.tests.next_tests.TestNextMethods)
get_struggling_exercises()
  File "/home/richard/github/ka-lite/kalite/topic_tools/tests/next_tests.py", line 105, in test_struggling
    self.assertSequenceEqual(expected, actual, "Struggling return incorrect.")
AssertionError: Sequences differ: [u'counting-out-1-20-objects'] != [u'radius_diameter_and_circumference', u'num[35 chars]cts']

First differing element 0:
counting-out-1-20-objects
radius_diameter_and_circumference

Second sequence contains 2 additional elements.
First extra element 1:
number_line

- [u'counting-out-1-20-objects']
+ [u'radius_diameter_and_circumference', u'number_line', u'counting-out-1-20-objects'] : Struggling return incorrect.

```

**Comment by almgong** [comment]
Oh that is odd... I ran into the problem of not being able to set the struggling attribute for a new exerciselog record as true after the instantiation line, and so I tried to do so in the constructor. Let me try the tests again locally and see if I get the same issue (it worked before I changed the superclass to be kalitetestcase).

**Comment by almgong** [comment]
I just retried it locally, and I am still stuck on a missing module error for 'selenium'. Running the tests using unittest.TestCase as the superclass instead of KALiteTestCase gives me the green light, however.

**Comment by rtibbles** [comment]
@almgong to set 'struggling' = to true, just set attempts = 20 or more, and set streak_progress to less than 100, during the save method, struggling will be set to true.

Need to install Selenium with ```pip install -r requirements_dev.txt```. Also, still need to put these tests in topic_tools/tests not contentrec_tests, as I don't think these tests are getting picked up by the testrunner.

**Comment by almgong** [comment]
Ah, I see. I will get that done tonight!

**Comment by MCGallaspy** [comment]
@rtibbles @almgong correction, I think you need to run `pip install -r requirements_test.txt`.

**Comment by almgong** [comment]
I will do that as well!

On Tue, Jun 16, 2015 at 11:44 AM, Michael Gallaspy <notifications@github.com

**Comment by rtibbles** [comment]
1 test failure, @almgong:
```
FAIL: test_struggling (kalite.topic_tools.tests.next_tests.TestNextMethods)
get_struggling_exercises()
  File "/home/travis/build/learningequality/ka-lite/kalite/topic_tools/tests/next_tests.py", line 106, in test_struggling
    self.assertSequenceEqual(expected, actual, "Struggling return incorrect.")
AssertionError: Sequences differ: [u'counting-out-1-20-objects'] != [u'radius_diameter_and_circumference', u'num[35 chars]cts']
First differing element 0:
counting-out-1-20-objects
radius_diameter_and_circumference
Second sequence contains 2 additional elements.
First extra element 1:
number_line
- [u'counting-out-1-20-objects']
+ [u'radius_diameter_and_circumference', u'number_line', u'counting-out-1-20-objects'] : Struggling return incorrect.
```

**Comment by almgong** [comment]
How can I run the kalite tester on my machine? Perhaps I should have debugged using that.

**Comment by rtibbles** [comment]
bin/kalite test topic_tools

(doing it without topic_tools will run *all* the tests)

**Comment by MCGallaspy** [comment]
There's an issue right now where extra arguments (in particular test labels like `topic_tools` here) aren't handled properly. So either you'll have to run all the tests, are you can temporarily checkout the `kalitectl.py` from `master` to use test labels again, using these commands (assuming you named the leq repo upstream):
```s
git fetch upstream master
git checkout upstream/master kalitectl.py
```

almgong:
I was wondering why it seemed like front end tests were running. This did
the trick, thanks Michael!

On Wed, Jun 17, 2015 at 11:42 AM, Michael Gallaspy <notifications@github.com

almgong:
Ok - the tests run correctly on my machine, @rtibbles can you confirm the fix?

rtibbles:
Travis seems to think so. Will just check that the latest merge passes, and then I think this is good to go!

rtibbles:
Travis says yes!

almgong:
Yay!

Issue #3803, opened 2015-06-08
Serve html docs if they exist


MCGallaspy:
If the html docs exist at a particular location, then we'll serve them. This is so that built docs can be bundled with the app.

MCGallaspy:
During the bundling process, we'll have to make sure static resources (js, css, etc) are somwhere that Django can find them.

PR #3806, opened 2015-06-08
Serve built docs within app

MCGallaspy:
What's inside:
* In `kalite.settings.base`, the `STATICFILES_DIRS` setting includes `os.path.join(_data_path, "sphinx-docs", "_build", "html")` if it exists. For me, this resolved to `/usr/share/kalite/sphinx-docs/_build/html`. This is so the included static files in that directory are served by Django.
* A new view and url that serves documentation pages (html files only) found in the directory `[[ROOT_DATA_PATH]]/sphinx-docs/build/html`, or serves a 404 page if the file doesn't exist. The url matches e.g. `documentation/`, `documentation/usermanual/userman_admin.html`.
* A management command (`sphinx_docs_preprocess`) to preprocess bundled docs during the build process. This is necessary so that `img` tags and references to css and js files resolve correctly. This is meant to be used in the build process, after building or otherwise obtaining docs.

Todo: Include an access point to the docs somewhere in the app itself... suggestions?

### DIFF
--- a/kalite/distributed/management/commands/sphinx_docs_preprocess.py
+++ b/kalite/distributed/management/commands/sphinx_docs_preprocess.py
@@ -1,0 +1,26 @@
+import os
+import re
+
+from django.core.management.base import NoArgsCommand
+
+from kalite.settings import USER_DATA_ROOT
+from os import walk
+
+class Command(NoArgsCommand):
+    """
+    Pre-process html files in sphinx-docs.
+    At the moment, this just means making static file references (css, js) use an absolute href.
+    """
+
+    def handle_noargs(*args, **kwargs):
+        expr = re.compile(r'(href|src)="(\.\./)*(_static|_images)/')
+        repl = r'\1="/static/\3/'
+        docs_path = os.path.join(USER_DATA_ROOT, "sphinx-docs", "_build", "html")
+        for dirpath, dirnames, filenames in walk(docs_path):
+            for filename in filenames:
+                if ".html" in filename:
+                    with open(os.path.join(dirpath, filename), "r+") as f:
+                        preprocess = f.read()
+                        postprocess = expr.sub(repl, preprocess)
+                        f.seek(0)
+                        f.write(postprocess)
--- a/kalite/distributed/views.py
+++ b/kalite/distributed/views.py
@@ -5,6 +5,7 @@ Views for the KA Lite app are wide-ranging, and include:
 * Administrative pages
 and more!
 """
+import os
 import sys
 from annoying.decorators import render_to
 from annoying.functions import get_object_or_None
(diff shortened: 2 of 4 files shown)


MCGallaspy:
Fixes #3803.

kalite/distributed/management/commands/sphinx_docs_preprocess.py
```diff
@@ -0,0 +1,26 @@
...
+    def handle_noargs(*args, **kwargs):
+        expr = re.compile(r'(href|src)="(\.\./)*(_static|_images)/')
+        repl = r'\1="/static/\3/'
+        docs_path = os.path.join(USER_DATA_ROOT, "sphinx-docs", "_build", "html")
+        for dirpath, dirnames, filenames in walk(docs_path):
+            for filename in filenames:
+                if ".html" in filename:
+                    with open(os.path.join(dirpath, filename), "r+") as f:
+                        preprocess = f.read()
+                        postprocess = expr.sub(repl, preprocess)
```
aronasorman: This is kinda opaque to me. Can you move this substition logic to its own function? Move the regex as well.

kalite/distributed/management/commands/sphinx_docs_preprocess.py
```diff
@@ -0,0 +1,26 @@
+import os
+import re
+
+from django.core.management.base import NoArgsCommand
+
+from kalite.settings import USER_DATA_ROOT
```
aronasorman: Can you move this import inside `handle`? This will mess up the central server imports, as this will force all settings to load.

kalite/distributed/views.py
```diff
@@ -249,3 +250,15 @@ def handler_500(request):
         "value": unicode(value),
     }
     return HttpResponseServerError(render_to_string("distributed/500.html", context, context_instance=RequestContext(request)))
+
+def sphinx_docs(request, doc_url):
```
aronasorman: Have you tried just serving the html files statically? Or do we really need to have a separate view?

kalite/distributed/management/commands/sphinx_docs_preprocess.py
MCGallaspy: You got it.

kalite/distributed/management/commands/sphinx_docs_preprocess.py
MCGallaspy: Sure thing.

aronasorman:
re: access point -- maybe in the dropdown box for the username?

aronasorman:
best place for me though is to have a "Help" section in the navbar. But knowing the effort @jtamiace gave to make the navbar showing and hiding play nicely, I don't think we want that for 0.14.

kalite/distributed/views.py
MCGallaspy: I have not tried serving them statically. What I hoped to achieve was to serve the docs from an arbitrary url. But statically might be simpler.

Issue #3757, opened 2015-05-28
Settings strategy and introduction of new methods in place of the deprecated one


benjaoming:
There are several ways of injecting your own settings:

1. By editing the source code **DISCOURAGED**. However, I think a lot of people might just do that.
1. By placing one's own local_settings.py in the system's `kalite` package **DEPRECATED**
1. Creating your own settings module and using the command line switch to point KA Lite and Django to it, for instance `kalite start --settings=my_module`. **NEW**
1. Same as above but using DJANGO_SETTINGS_MODULE to point django.conf to the settings module.

I would propose the following for 0.14:

1. To have `kalite.project.settings` created and functional
1. To make any references for `kalite.settings` raise a `DeprecationWarning`
1. Create command line option for most commonly used settings: `--listen-port`, `--debug`, `--content-folder`.
1. As far as possible, to get rid of `fle_utils.settingshelper`, not expecting this work to be conclusive yet.

Then, in 0.15, we can start cleaning up settings and do remaining work in case `fle_utils.settingshelper` wasn't fully dealt with.

One of the big issues lately has been how "PRODUCTION_PORT undefined blah blah" exceptions show up, but are actually caused by exceptions in settings modules that are magically imported by `fle_utils.settingshelper`. If this happens to end-users, we will have no way of helping them, because their exceptions will seem quite meaningless, as @EdDixon is currently experiencing with testing.

There's also an issue on writing documentation for this: #3611
And an issue for the over-all idea of making kalite a "normal" Django application: #2952

MCGallaspy:
I'm most interested in points 1 and 3. By creating `kalite.project.settings` you essentially mean defining a bunch of default settings that can be imported, yes? What's preventing this right now, in `mycoolsettings.py`:
```python
import kalite.settings.base
MYCOOLSETTING = 42
```

benjaoming:
@MCGallaspy nothing! The original idea was also to say (as I think the current deprecation warning message says) that you can do that.

But since settings are normally nested in the project scope, I'd like the move to consist of a refactor so we don't promote `from kalite.settings.base import *` but start out with the nicest possible practice, which I believe is to create a package for everything that's related to the project and not the app. For internal practices especially, having `settings` in `kalite` encourages that applications should interfere with settings, which they should never do, they should *read* settings, not try to access them and change them runtime.

[...]

PR #3809, opened 2015-06-09
Add new project layout and settings location

benjaoming:
This is for #3757

TODO

* [x] Create new project layout
* [x] Use `kalite.project.settings`
* [x] Create deprecation warnings when using deprecated location of settings
* [x] Remove `fle_utils.settingshelper`
* [x] Use settings modules for central server and docs building as proof of concept.
* [x] Remove `i18n.settings`, `cache.settings`, `topic_tools.settings`, and `updates.settings` because they cannot be loaded from their respective packages due to circularity.


MCGallaspy:
Is the milestone correct? Please change if not.

benjaoming:
Yeah, it's 0.14 alright! Thanks!

kalite/project/wsgi.py
```diff
@@ -0,0 +1,34 @@
+"""
+WSGI config for kalite.project
+
+It exposes the WSGI callable as a module-level variable named ``application``.
+
+For more information on this file, see
+https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/
```
MCGallaspy: Remember we're using Django 1.5 at the moment. Is this page accurate?

benjaoming: Oh sorry, this has already been correct, let me push that change now...

benjaoming: No wait yeah so I just left that module comment from Django 1.8's wsgi.py template, I didn't actually write it... but if you look below it makes sense!

benjaoming:
For the record, after cleaning up settings, `LOGIN_URL` is now `"/?login=true"` and not `'/securesync/login/'`, this seems to be correct also.

python-packages/fle_utils/settingshelper.py
```diff
@@ -60,14 +60,17 @@ def import_installed_app_settings(installed_apps, global_vars, cur_app="__root__
...
                 app_settings = __import__(app + ".settings")
                 # The module that we are importing is stored as an attribute
                 # of app_settings... i.e. "kalite.x.y.z.settings" becomes
                 # "app_settings.x.y.z.settings"
                 for x in app.split(".")[1:]:
                     app_settings = getattr(app_settings, x)
+                print "from %s.settings import *" % app
                 app_settings = getattr(app_settings, "settings")
         except ImportError:
+            print "FAILED"
```
aronasorman: debug logging statement?

aronasorman:
@benjaoming does this remove the need for `import_installed_app_settings`? Because it's what's blocking progress on the central server.

aronasorman:
I'd hold off on merging this before the develop tests get fixed though.

benjaoming:
@aronasorman we agreed yesterday that we could move forward with this and fix the central server after, you should only need to put some additional conditional statements on the settings modules in `kalite.settings.base` (look at the bottom of the file)

python-packages/fle_utils/settingshelper.py
benjaoming: oops yes, removing that... but actually the whole module can be ditched :)

benjaoming:
HELP! All my builds time out here:

```
test_user_interface (kalite.control_panel.tests.control_panel.CSVExportBrowserTests) ... 

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

But I can run the test locally without problems !?

MCGallaspy:
I'll see if I can figure out something.

benjaoming:
This issue PR somewhat critical, but I'm very curious as to why the tests are timing out... merging in develop to see if there's any change...

benjaoming:
@MCGallaspy the same test still times out, would you be up for investigating this since you're in closer contact with the ones who do the tests?

MCGallaspy:
@benjaoming yes, adding it to my to-do list.

MCGallaspy:
@benjaoming 
Looks like you managed to resolve your earlier test error, right? Sorry I didn't get to it earlier.

MCGallaspy:
Let me make sure I'm getting this straight:
* For development, settings should now be imported from `kalite.project.settings`
* `kalite.project.settings` explicitly imports the settings from various apps
* `kalite.project.settings` represents the "base" settings, and settings in included apps could potentially _overwrite_ (but not modify) the "base" settings
* Altering base settings in an app is illegal (or at least immoral) and this PR removes instances where base settings are _modified_ (like `INSTALLED_APPS`) by consolidating them in one place.

@benjaoming correct?

benjaoming:
@MCGallaspy 
This is the soft middle-step. Point is that users can start writing their own settings modules like this:

    MY_SETTING = True

Notice the `.base`.

Thus, we can **deprecate** the need for local_settings.

Very importantly, I also removed settingshelper which made exceptions completely misguiding. I had to clean up how settings are loaded and make this a lot more explicit in order to get rid of it... maybe a bit more refactoring than what's in good taste so close to the freeze, but if we shipped without this, we would risk getting user feedback with useless tracebacks.

The future restrictions on how kalite apps should manage their own settings is still pending, but it's discussed over in #3757 -- for now, as you point out, they no longer modify INSTALLED_APPS, MIDDLEWARE_CLASSES and TEMPLATE_CONTEXT_PROCESSORS which was part of the crazy way that settingshelper worked by magically appending to these. Thus, the other issues of having the same applications repeated over and over and over again in INSTALLED_APPS, and settings modules loaded again and again and again. There's actually a considerable speed-up just loading `kalite manage help`due to this.

benjaoming:
To keep the future refactor easy, I suggest putting shared settings in `kalite.settings.base` and app-specific settings in `kalite.some_app.settings`.

kalite/settings/__init__.py
```diff
@@ -17,20 +16,13 @@ def package_selected(package_name):
...
-#   config packages to override settings.
-########################
 
-# Here is all the old magic crap that should go... /benjaoming
+CHERRYPY_PORT = getattr(local_settings, "CHERRYPY_PORT", PRODUCTION_PORT)
 
-import_installed_app_settings(INSTALLED_APPS, globals())
+if DEBUG:
+    from kalite.testing.loadtesting.settings import *
+    TEST_RUNNER = KALITE_TEST_RUNNER
```
MCGallaspy: This is the culprit for your failing test, methinks.

MCGallaspy:
Okay, now I __really__ am looking at the test failures here.

MCGallaspy:
I think the test failures you're experiencing are due to race conditions.

MCGallaspy:
@rtibbles @benjaoming @aronasorman 
Again, running these locally I don't get the same test failures. Leads me to believe it's an issue with the test environment on travis, as in #3817.

benjaoming:
Will resolve the remaining 3 tasks, sorry if it seemed because of the open tasks that the PR wasn't ready.. that was a miscommunication from my side, The only thing IMO that stops a merge is the weird test failures.

benjaoming:
@MCGallaspy I've added a dump of the Firefox version used, says:

    Successfully setup Firefox 31.0

Wondering if this is related in any way to the failures. I will try to run ALL tests on my local machine now to see if that yields any differences, but I'm not downgrading my FF :)

benjaoming:
Am recreating all the failures locally with FF 38.0. Also I'm getting this:

```
FAIL: test_query_status_teacher (kalite.distributed.tests.browser_tests.query_tests.QueryTest)
  File "/media/archive/code/ka-lite/kalite/distributed/tests/browser_tests/query_tests.py", line 72, in test_query_status_teacher
    self.test_query_login_teacher()
  File "/media/archive/code/ka-lite/kalite/distributed/tests/browser_tests/query_tests.py", line 51, in test_query_login_teacher
    self.browser_login_teacher("t1", passwd, self.facility)
  File "/media/archive/code/ka-lite/python-packages/django/test/testcases.py", line 196, in __exit__
    executed, self.num
AssertionError: 61 != [28..46] : 61 queries executed, [28..46] expected
```

benjaoming:
@MCGallaspy @rtibbles 

`http://localhost:8000/api/coachreports/summary/?facility_id=XXX` takes ~10 seconds to load on my local machine, hence also the failing behave tests in this PR, I suspect...

kalite/distributed/features/steps/login.py
```diff
@@ -33,7 +33,14 @@ def step_impl(context):
 
 @then("there should be no facility drop down")
 def step_impl(context):
-    assert not find_id_with_wait(context, "id_facility-container")
+    try:
+        # TODO(benjaoming): This is a crazy test... waiting for a timeout in
```
MCGallaspy: One might wish to avoid the case where the element doesn't appear in 1 sec, but appears after > 1s. Obviously we can't wait forever, but 1 sec is probably reasonable until it actually causes an error. Also, `find_id_with_wait` shouldn't raise a TimeoutException, but instead just return None if no element is found within the specified time.

kalite/testing/behave_helpers.py
```diff
@@ -201,14 +201,11 @@ def _find_elem_with_wait(context, by, wait_time=MAX_WAIT_TIME):
...
-    Returns the element if found or None
+    Returns the element if found or raises TimeoutException
     """
-    try:
-        return WebDriverWait(context.browser, wait_time).until(
-            EC.presence_of_element_located(by)
-        )
-    except TimeoutException:
-        return None
+    return WebDriverWait(context.browser, wait_time).until(
```
MCGallaspy: Oh, I see. Disregard my earlier comment. This is probably the better way to do it, anyway.

kalite/distributed/features/steps/login.py
benjaoming: Oh but I changed that :) ...because it was cause illogical test failures elsewhere... if a timeout occurs, you'd want to know and not have the test continue with a None value

benjaoming: The 1 second wait time is a way of skipping the test until someone feels like coming up with a better strategy for testing this... I mean testing that something doesn't happen is a pretty weird scenario... what if the modal comes up after 30 minutes?

kalite/testing/behave_helpers.py
benjaoming: Ah, then disregard my comment, too :D

benjaoming:
@rtibbles I think that hitting the said URL which calls `api_view.aggregate_learner_logs` is the first instance of the test where `topic_tools.get_topic_leaves, get_exercise_cache, get_content_cache` are called... would that sound reasonable? I'm thinking it's easier to ask than to profile it :)

benjaoming:
@rtibbles yeah it's definitely the first call to get_content_cache!

Start: 2015-06-22 23:48:12.610007
Finish: 2015-06-22 23:48:28.978419

16 seconds

benjaoming:
@rtibbles @MCGallaspy okay I see that another test case calls get_exercise_cache before the behave tests, and when running locally, my coachreport behave tests pass without problems and without taking much time.

benjaoming:
Tested manually on Firefox 31, didn't cause any issues on the coach report page.

MCGallaspy:
I think we should merge this. Both you and I have confirmed that these tests pass locally, so the only thing we're learning from these test failures is that KA Lite runs very slowly on travis.

benjaoming:
@MCGallaspy agreed

I'm leaving the BDD tests DOM dump for now as I haven't debugged it yet.. it's harmless anyways.

Hoping that the develop branch is in good shape when you guys start your working day...

Issue #3810, opened 2015-06-09
settingshelper loads the same settings lots of times


benjaoming:
Dump from before:

```
➜  ka-lite git:(project-layout) ✗ bin/kalite shell
from kalite.distributed.settings import *
from kalite.django_cherrypy_wsgiserver.settings import *
from kalite.coachreports.settings import *
from kalite.testing.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.facility.settings import *
from kalite.main.settings import *
from kalite.distributed.settings import *
from kalite.contentload.settings import *
from kalite.main.settings import *
from kalite.contentload.settings import *
from kalite.testing.loadtesting.settings import *
from kalite.contentload.settings import *
from kalite.facility.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.main.settings import *
from kalite.contentload.settings import *
from kalite.facility.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.main.settings import *
from kalite.main.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.facility.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.main.settings import *
from kalite.main.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.main.settings import *
from kalite.testing.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.facility.settings import *
from kalite.testing.loadtesting.settings import *
from kalite.contentload.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.facility.settings import *
from kalite.contentload.settings import *
from kalite.facility.settings import *
from kalite.facility.settings import *
from kalite.playlist.settings import *
from kalite.contentload.settings import *
from kalite.testing.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.facility.settings import *
from kalite.testing.loadtesting.settings import *
from kalite.facility.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.facility.settings import *
from kalite.testing.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from kalite.testing.loadtesting.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
from securesync.settings import *
from fle_utils.chronograph.settings import *
[...]

PR #3811, opened 2015-06-09
remove a giant problem in settingshelper, it was loading the same set…

benjaoming:
…tings many times, assertion doesnt work

#3810

### DIFF
--- a/python-packages/fle_utils/settingshelper.py
+++ b/python-packages/fle_utils/settingshelper.py
@@ -40,7 +40,7 @@ def import_installed_app_settings(installed_apps, global_vars, cur_app="__root__
     assert not set(processed_apps).intersection(set(installed_apps)), "Should never process the same app twice."
 
     this_filepath = global_vars.get("__file__")  # this would be the project settings file
-
+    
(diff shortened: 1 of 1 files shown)


benjaoming:
This PR is really not very important, because I will burn settingshelper down shortly.

benjaoming:
@MCGallaspy this should be easy to merge, I just wanted you to be aware of the issue being solved (which has been an issue forever, I made a backport to 0.13.x as well)

I think the case here is that the way `set.union` works is by reference, and set.add works by value... well, that's my theory, I didn't look it up because it empirically just works for me.

MCGallaspy:
While I think I understand what you're trying to accomplish with this PR *and* while I also acknowledge that `develop` is kind of a mess with failing tests anyway, the fact that this causes new failures is a red flag for me. If, as you say, this will become irrelevant anyway, then I suggest we ignore it and simply wait for the irrelevant-making PR.

benjaoming:
Just put all efforts in #3809 :) It completely removes the dreaded settingshelper

PR #3814, opened 2015-06-10
Serve built docs within app

### LINKED ISSUE #3803: Serve html docs if they exist

MCGallaspy:
Fixes #3803. Supersedes #3806.

Serves the docs statically, if they exist. Adds a conditional item to the user dropdown menu, with a link to the statically served docs. Adds an new mapping to the status model.


kalite/facility/api_resources.py
```diff
@@ -177,6 +178,8 @@ def generate_status(self, request, **kwargs):
             "version": version.VERSION,
             "facilities": request.session.get("facilities"),
             "simplified_login": settings.SIMPLIFIED_LOGIN,
+            "docs_exist": True if any(map(lambda x: os.path.join("sphinx-docs", "_build") in x,
```
aronasorman: I believe you can remove the `if` statement and just use `any`, which returns a boolean.

MCGallaspy: You are correct.

aronasorman:
Looks good to me. I'm waiting for the develop tests to get fixed. Once that's done then we can merge this.

aronasorman:
Tests pass. Merging.

PR #3815, opened 2015-06-10
0.13.x into develop

MCGallaspy:
Just to keep track.


aronasorman:
Do we still want this for develop? We've sunsetted the 0.13.x branch at this point.

PR #3817, opened 2015-06-10
Helenas doc and testing

howdyitshelena:
Changed the user manual for learners to accommodate for content recommendation addition

### DIFF
--- a/kalite/testing/base_environment.py
+++ b/kalite/testing/base_environment.py
@@ -11,7 +11,7 @@ from selenium import webdriver
 from urlparse import urljoin
 from django.contrib.auth.models import User
 
-from kalite.testing.behave_helpers import login_as_admin, login_as_coach, logout
+from kalite.testing.behave_helpers import login_as_admin, login_as_coach, logout, login_as_learner
 
 def before_all(context):
     browser = context.browser = webdriver.Firefox()
@@ -36,6 +36,9 @@ def before_feature(context, feature):
     elif "as_coach" in feature.tags:
         context.logged_in = True
         login_as_coach(context)
+    elif "as_learner" in feature.tags:
+        context.logged_in = True
+        login_as_learner(context)
 
 def after_feature(context, feature):
     if context.logged_in:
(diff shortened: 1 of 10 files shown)


rtibbles:
Also fixes #3884 now.

MCGallaspy:
@rtibbles I can't reproduce these test failures on my machine, though I do get a different test failure in the `Scenario: Logging in with the wrong username`. I discussed it with @aronasorman and we think it might be a problem with the test environment, namely that the server is just too dang slow. Trying to find alternate build servers.

kalite/testing/base_environment.py
```diff
@@ -11,20 +11,19 @@
...
-    try:
-        context.browser.quit()
-    except CannotSendRequest:
-        pass
+    pass
 
 def before_feature(context, feature):
+    browser = context.browser = webdriver.Firefox()
+    # ensure the window is reasonably sized.
+    browser.set_window_size(1024, 768)
```
MCGallaspy: Re: this https://travis-ci.org/learningequality/ka-lite/jobs/67872651#L2345

The element may simply be outside of the visible area. It might help to increase the window size.

MCGallaspy: Wait, disregard, it's that the page isn't loading.

rtibbles: I have no idea why.

rtibbles: Trying adding extra wait time, as, still, it is passing perfectly on mine. Really quite frustrating.

MCGallaspy: Just a few thoughts:
* What happens if the content isn't available? I would still expect the page to load, but I have encountered issues with tests that passed locally but failed on travis because I had content where travis doesn't.
* Which long-running functions are triggered first by the learn page __views__, if any? Could something be causing the learn page to hang because Django is taking a long time to come up with a response?

MCGallaspy:
I think we should merge this. As in #3809 it seems that the only thing we're getting from these test failures is that the app runs poorly on travis.

rtibbles:
Yeah, I am certainly learning nothing from these test failures.

MCGallaspy:
Suuuuuperseded by #3932.

PR #3818, opened 2015-06-11
create superuser for other tests

66eli77:
as long as the test uses Given "I am on the homepage" from generic.py 
a superuser will be created if none exist.


kalite/distributed/features/steps/superuser_create.py
```diff
@@ -46,6 +52,8 @@ def step_impl(context):
 
 @then("the modal won't dismiss")
 def step_impl(context):
+    driver = webdriver.Firefox()
+    driver.implicitly_wait(2)
```
MCGallaspy: `implicitly_wait` is not a wait. It sets a timeout for other actions, and it only needs to be called once per WebDriver instance. See [the docs](http://selenium-python.readthedocs.org/api.html#selenium.webdriver.remote.webdriver.WebDriver.implicitly_wait)! The upshot is don't use it here if you expect it to wait 2 seconds (it won't, and that kind of behavior is awfully susceptible to race conditions), but instead use behave's [environmental controls](https://pythonhosted.org/behave/tutorial.html#environmental-controls) to set it once for the whole feature/test suite.

kalite/distributed/features/steps/generic.py
```diff
@@ -1,6 +1,9 @@
 from behave import *
 from kalite.testing.behave_helpers import *
+from django.contrib.auth.models import User
 
 @given("I am on the homepage")
 def step_impl(context):
-    go_to_homepage(context)
\ No newline at end of file
+	if not User.objects.exists():
```
MCGallaspy: I consider this unexpected behavior for this step. I would create a new step like this:
```python
@given("There is a superuser")
def step_impl(context):
    if not User.objects.exists():
        User.objects.create_superuser(username='superusername', password='superpassword', email='super@email.com')
```

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -5,6 +5,10 @@
 
 modal_container = "superusercreate-container"
 
+@given("I go to homepage")
```
MCGallaspy: You don't need to define a new step here, considering that `@given("I am on the homepage")` already exists and does the same thing. If you change the wording of your .feature file to match that step, it should use it automatically.

kalite/distributed/features/steps/superuser_create.py
```diff
@@ -46,6 +52,8 @@ def step_impl(context):
 
 @then("the modal won't dismiss")
 def step_impl(context):
+    driver = webdriver.Firefox()
```
MCGallaspy: Also, if you use [from kalite.testing.base_environment import *](https://github.com/learningequality/ka-lite/blob/develop/kalite/testing/base_environment.py) it defines a `before_all` function that adds a WebDriver instance to the context as `context.browser`.

kalite/distributed/features/steps/superuser_create.py
66eli77: I intentionally avoid using the @given("I am on the homepage") because it will create a super user which I don't want.

kalite/distributed/features/steps/generic.py
66eli77: if we create a new step like this, we need to manually update all the affected tests. Seems anyone working in other feature don't want to be blocked by superuser modal, making superuser always exists so that they don't even need to know to include @given("There is a superuser")

kalite/distributed/features/steps/superuser_create.py
MCGallaspy: All the more reason not to create a superuser in that step, and instead make a new @given.

kalite/distributed/features/steps/generic.py
MCGallaspy: I understand what you're saying, but my contention is that adding unexpected side effects to an existing step violates the principle of least surprise. Imagine that a breaking change is made down the road, and someone discovers that the `I am on the homepage` step fails not because they failed to be on the homepage, but because a superuser wasn't created. Therefore I would indeed like to manually update all the affected tests, since it explicitly spells out the fact that many existing features *now* depend on the fact that a superuser exists.

benjaoming: @MCGallaspy understand the point and I like the implementation, but is there a way to make that kind of test not have to feature for ALL behave tests? I mean, the given pre-condition should be describing user behavior, too, right? So we could say that tests that aren't supposed to care about superuser creation can get away with a simpler interface?

....and then make the superuser-creation an automatic step that other behave tests automatically feature so we don't have to worry about it?

kalite/distributed/features/steps/superuser_create.py
benjaoming: Isn't it also problematic to create a new web driver for a single test method? Can't these instances somehow be shared across tests, since the tests are sequentially executed anyways?

MCGallaspy: They can be shared, it's defined in the environment functions (I made a comment about it below).

kalite/distributed/features/steps/generic.py
MCGallaspy: > So we could say that tests that aren't supposed to care about superuser creation can get away with a simpler interface?

Maybe the compromise solution is to use tags in the `.feature` files (something like `@superuser_required`, and then add a snippet in [kalite.testing.base_environment.before_feature](https://github.com/learningequality/ka-lite/blob/develop/kalite/testing/base_environment.py#L26) to create a superuser if one doesn't exist.

kalite/distributed/features/steps/superuser_create.py
benjaoming: Have to agree with @MCGallaspy, it's important not to create two similar-worded givens, they will create confusion for everyone :)

"there is superuser" should perhaps be called "there exists a superuser account".

kalite/distributed/features/steps/generic.py
66eli77: yeah, this is a good location to create superuser.

MCGallaspy:
I'm waiting with bated breath to merge this. Made some additional changes for @66eli77 to merge (PR on his repo). Test results [here](https://travis-ci.org/MCGallaspy/ka-lite/builds/66626259). Should fix all remaining failing tests on `develop`.

PR #3823, opened 2015-06-12
Docs build fix

### LINKED ISSUE #3819: Make sure docs build ok
Something's fishy on develop right now.

MCGallaspy:
Fixes #3819. Docs were not building on develop, for a number of reasons, including:
* An unmet dependency (`django-extensions`). But it didn't really need that module anyway, so I removed the offending lines.
* Some urls in the screenshot directives were no longer valid.

Also some random fixups:
* Use an in-memory database for screenshots; I dunno why we weren't already. :(
* Better exception handling in the management command, by which I mean more specific.
* Building the docs fails if it can't make all the screenshots; before it would silently continue and produce broken image links.
* Add a warning to the screenshot directive usage notes. (Read it!)
* Change some screenshot directives to be more robust (see warning in the usage notes).

A little bit more to come, don't merge yet! I have to run to the shuttle.


MCGallaspy:
There's a problem with screenshots: the app is sometimes just really slow, and so the screenshots management command suffers from unpredictable timeouts. Page loads and user-interaction with `:navigation-steps:` is fraught with peril. This is compounded by the fact that with the screenshot directive there's really no way to specify a user interaction followed by a wait; that's on me, lesson learned. I'm seeing a lot of parallels here with the problems we've had with browser tests, since both need to use the site automatically. Also, the screenshots management command has some warts.

Fortunately, the screenshot Sphinx directive (the plug-in to the docs) can be de-coupled pretty easily from the actual implementation of the screenshots management command, so making the toolchain nicer can be done piecemeal. In an ideal world:

* We write methods for interacting with the site using selenium that can be used by both the browser tests and the screenshot directive.
* We polish the screenshots tool chain. (A subject on which I have some opinions, but in broad strokes by first making the management command nicer, then by making use of safe pre-defined actions in the directive instead of relying on more general but error-prone specifications.)

((This also dovetails nicely into my desire to revisit the whole test suite.))

MCGallaspy:
The upshot to this PR is that it builds on my machine, but not on field for reasons TBD.

MCGallaspy:
The reason turned out to be the window size. (And maybe to a lesser extent, poor performance due to lack of memory.)

MCGallaspy:
This is ready to be merged in, pending approval.

sphinx-docs/usermanual/userman_coach.rst
```diff
@@ -28,13 +28,6 @@ Overview
 
 The overview will display the activity of the group: the total hours logged, the average progress in exercises, the total exercises attempted, and the total time spent on content.
 
-.. screenshot::
```
MCGallaspy: Actually, these might also have been failing because of the window size. Let me try it again before I nix them.

MCGallaspy: These really do make it barf. I'll have to revisit this.

kalite/distributed/management/commands/screenshots.py
```diff
@@ -211,7 +214,7 @@ def __init__(self, *args, **kwargs):
 
         self.loginfo("==> Setting-up browser ...")
         super(Screenshot, self).setUp()
-        self.browser.set_window_size(1024, 768)
+        self.browser.set_window_size(1920, 1080)
```
aronasorman: Can you add a note why we increased the resolution to HD?

kalite/distributed/management/commands/screenshots.py
```diff
@@ -256,14 +257,21 @@ def snap(self, slug, focus, note):
...
+                    self.browser.execute_script('$("%s").css("%s", "%s");' % (selector, key, value))
+                if note:
+                    self.browser.execute_script("$('%s').qtip({content:{text:\"%s\"},show:{ready:true,delay:0,effect:false}})" % (selector, note))
+            except WebDriverException as e:
+                log.error("Error taking screenshot:")
+                log.error(str(e))
+                log.error("Screenshot info: {0}".format((focus, note)))
+                log.error("Current url: {0}".format(self.browser.current_url))
+                import sys
+                sys.exit(1)
```
aronasorman: Why exit this way rather than just reraising the error?

MCGallaspy: Because I want the process to exit with a non-zero code, so that the parent process (building the docs) recognizes something amiss. Simply raising an exception doesn't appear to give a nonzero exit code.

kalite/settings/__init__.py
```diff
@@ -148,11 +145,12 @@ def package_selected(package_name):
     # use another sqlite3 database for the screenshots
     DATABASES = {
         SCREENSHOTS_ROUTER: {
-            "ENGINE": getattr(local_settings, "DATABASE_TYPE", SQLITE3_ENGINE),
-            "NAME": SCREENSHOTS_DATABASE_PATH,
-            "OPTIONS": {
-                "timeout": 60,
-            },
-        }
+            "ENGINE": SQLITE3_ENGINE,
```
aronasorman: Rather than another `if` statement in our settings file, perhaps we should move to @benjaoming's `--settings` arg method instead.

sphinx-docs/screenshot.py
```diff
@@ -64,6 +64,8 @@ def process_screenshots(app, env):
         command += ["--lang", language]
     subprocess = Popen(command)
     subprocess.wait()
+    if subprocess.returncode:
```
aronasorman: clever!

kalite/distributed/management/commands/screenshots.py
aronasorman: Oh, that's strange. Maybe something up the chain is catching that error.

kalite/settings/__init__.py
MCGallaspy: I don't think it's usable until #3809 is merged in, but we'll see what @benjaoming has to say about it.

aronasorman:
The only issue here is another conditional setting. We can refactor that out once #3809 is approved. We can merge this in the meantime.

kalite/distributed/management/commands/screenshots.py
benjaoming: yeah sounds weird, exceptions always exits with non-zero.

kalite/settings/__init__.py
benjaoming: Yeah that sounds about right.

So both for docs building and for the central server, we should be using custom settings modules rather than conditions. I will try to work something out..

benjaoming: Btw, great to merge this, for me this workflow makes more sense... it's a "TODO" but not really a blocking issue for merging, much better to see things in action :)

PR #3827, opened 2015-06-12
Add --port option and fix issues with proxying django options

benjaoming:
Replaces #3776


MCGallaspy:
This one's a little fishy. What's the plan for replacing this if/when docopt merges your PR? We'll just have to add it to our TO-DO list? Should we remove `docopt` from our requirements in the meantime?

benjaoming:
@MCGallaspy I'm still importing everything used in `docopt()` from `docopt` (the module).

benjaoming:
But the plan would be to use docopt() as is and then use the exception thrown from docopt to collect the left-over parts of sys.argv and pass to underlying django commands called... i think it's acceptable
 and won't take up more than a couple of LOC.

MCGallaspy:
@benjaoming 
All the test failures here should be fixed in #3867.

Issue #3704, opened 2015-05-22
Cannot connect to server for downloading video files from current development installs


EdDixon:
While downloading videos from the current development installations we are not able to connect with the server in order to do so? The screen shot provided is from a MAC current in development installer installation but with no modification to the local_settings.py which was double checked to ensure there was no redirection to the video resources
![2015-05-22_15-17-46](https://cloud.githubusercontent.com/assets/4349806/7780563/92601d16-009d-11e5-8801-2ee741bf62e3.png)
.

EdDixon:
Just a note here this test was performed immediately after installation restarting the server seems to have resolved this problem. I am rebooting now to see if the problem returns or remains resolved.

EdDixon:
A reboot did not resolve the issue and to the user it appears as if the server is down or the url is not resolving to the appropriate
![2015-05-24_05-31-58](https://cloud.githubusercontent.com/assets/4349806/7787709/79070d9e-01d7-11e5-8295-8740c044b659.png)
 server?

MCGallaspy:
Does this issue only occur on mac? Have you been able to reproduce it @ed? Were you logged in as a coach or an admin?

MCGallaspy:
Unable to reproduce, @cpauya can you try this on mac?

cpauya:
I was able to reproduce this a few days ago @MCGallaspy - I will try again with the latest build from dungeon server: http://dungeon.learningequality.org:8085/browse/KL-KALITE014/latest and post a screenshot here.

cpauya:
Yes @MCGallaspy, video downloads are failing on the latest Mac installer too (after successfully doing a One-click Registration).  I am logged-in as admin.

Here's the browser screenshot on Firefox v38.0.5:
![screenshot 2015-06-03 14 34 38](https://cloud.githubusercontent.com/assets/175580/7954074/b25cfd72-09fe-11e5-99f6-31b10f99ae8c.png)

After the failed attempt to download a video, I ran `kalite status` on the terminal and got an "Unclean shutdown (7)" message.

```bash
cyril-lappy6:~ cyril$ kalite status
Unclean shutdown (7)
```

There are also no logs because it crashed on the pyrun2.7 binary, meaning the ka-lite server has stopped.  Here's the first few lines of the crash report:

```
Process:               pyrun2.7 [17404]
Path:                  /Applications/KA-Lite Monitor.app/Contents/Resources/pyrun-2.7/bin/pyrun2.7
[...]

PR #3830, opened 2015-06-12
Spawn processes by default for OS X whenever we call call_command_async

aronasorman:
Workaround for #3704. Running videodownload on OS X results in a strange quitting behaviour by the python process if run as a thread. If run as a process, it should work fine, albeit it might take longer. Good enough for now.

### DIFF
--- a/python-packages/fle_utils/django_utils/command.py
+++ b/python-packages/fle_utils/django_utils/command.py
@@ -131,7 +131,15 @@ def call_command_async(cmd, *args, **kwargs):
     that stringify in a way that commands can parse
     (which will work for str, bool, int, etc).
     """
+    #  Workaround for #3704. It seems like running cron for downloading videos
+    # as a different thread on the same process in OS X results in a normal exit.
+    # That's it, the kalite executable just quits, no segfaults or anything.
+    # The workaround for now is to spawn subprocesses by default on OS X,
+    # and have everyone else spawn threads.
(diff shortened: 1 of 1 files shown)


aronasorman:
No syntax errors yet. Will merge this now. @cpauya please review once you have a chance to do so.

cpauya:
I tested this by building a local `.dmg` but it still crashing the other `pyrun` process.

python-packages/fle_utils/django_utils/command.py
```diff
@@ -131,7 +131,15 @@ def call_command_async(cmd, *args, **kwargs):
...
     (which will work for str, bool, int, etc).
     """
+    #  Workaround for #3704. It seems like running cron for downloading videos
+    # as a different thread on the same process in OS X results in a normal exit.
+    # That's it, the kalite executable just quits, no segfaults or anything.
+    # The workaround for now is to spawn subprocesses by default on OS X,
+    # and have everyone else spawn threads.
+    is_osx = sys.platform == 'darwin'
+    in_proc = kwargs.pop('in_proc', not is_osx)
     in_proc = kwargs.pop('in_proc', True)
```
cpauya: I commented this line-out and the other process didn't crash.  The `/progress/` API call is still working but with 0 bytes being downloaded, it's only a 1.3MB video.

Testing it further...

PR #3864, opened 2015-06-13
Fixes #3861: Videos not rendering properly when subtitles are present.

### LINKED ISSUE #3861: Videos not rendering properly when subtitles are present
Expected: Subtitles render on videos A-OK.

Observed: Videos with subtitles aren't rendering at all.

Steps to replicate: Download videos, then upgrade EN language pack, then go to a video page which should have subtitles.

image

With JS error message `Uncaught TypeError: Cannot read property 'apply' of undefined`, which comes from the Django JS reverse stuff, for the line `window.Urls[url_name].apply(window.Urls,arguments);`.

The problem is that we're referencing `{{ url }}` in the [video player handlebars template](https://github.com/learningequality/ka-lite/blob/develop/kalite/distributed/hbtemplates/video/video-player.handlebars#L5) for the subtitles, but @rtibbles recently also added a Handlebars helper for url reversal, with the same name, which is now getting called instead. Not sure the best way to resolve this, but we probably either need to rename the helper method, or modify the name of the subtitle URL attribute getting stamped onto video data.

jamalex:


kalite/distributed/hbtemplates/video/video-player.handlebars
```diff
@@ -2,7 +2,7 @@
     <video class="video-js vjs-default-skin" id="{{ random_id }}" controls preload="auto" width="100%" height="600" {{#if content_urls.thumbnail }}poster="{{ content_urls.thumbnail }}" {{/if}}data-setup="{}">
         <source src="{{ content_urls.stream }}" type="{{ content_urls.stream_type }}" />
         {{#each subtitle_urls }}
-        <track kind="captions" src="{{ url }}" srclang="{{ code }}" label="{{ name }}" {{#ifcond code "==" ../selected_language }}{{#ifcond ../code "!=" "en" }}default="True"{{/ifcond}}{{/ifcond}}/>
+        <track kind="captions" src="{{ this.url }}" srclang="{{ code }}" label="{{ name }}" {{#ifcond code "==" ../selected_language }}{{#ifcond ../code "!=" "en" }}default="True"{{/ifcond}}{{/ifcond}}/>
```
rtibbles: Looks like the fix I would make.

rtibbles:
Test failures consistent with current develop.

PR #3876, opened 2015-06-15
Remove _built.touch logic, don't use fle_utils.testing as an app and …

### LINKED ISSUE #3875: testing application problems
So two issues:

 1. We should put a models.py in `fle_utils.testing` if it's supposed to be used -- or just remove the thing
 1. Make a decision about "_built.touch" because I don't see where this thing is even created, so we can get rid of `kalite.settings.BUILT` all-together and clean up the logic of that.

Currently, stuff is breaking because of having enabled `fle_utils.testing` which was a consequence of when I tried cleaning up settings modules of all applications.

benjaoming:
…remove unused "runcode" management script

Fixes #3875


aronasorman:
`runcode` is used by the central server for the integration tests.

benjaoming:
@aronasorman as was the case before, there's still some possible local settings to fix for the central server after I cleaned up settings.. so one of them would be to write a settings module with this content:

    INSTALLED_APPS += ('fle_utils.testing',)

PR #3878, opened 2015-06-16
Some logging is better than none

MCGallaspy:
Here's some logging. It's opt-in, since it doesn't use any rotation. A simpler solution than #3800 for now. In general it's possible to use fancy logfiles with rotation using the `logging` module, but since we import that from settings I'm not touching it right now.

### DIFF
--- a/kalite/django_cherrypy_wsgiserver/cherrypyserver.py
+++ b/kalite/django_cherrypy_wsgiserver/cherrypyserver.py
@@ -142,7 +142,12 @@ def run_cherrypy_server(host="127.0.0.1", port=None, threads=None, daemonize=Fal
         # stop_server(pidfile)
 
         from django.utils.daemonize import become_daemon
-        become_daemon()
+        kalite_home = os.environ.get("KALITE_HOME", None)
+        logfile = os.path.join(kalite_home, "kalite.log") if (kalite_home and os.environ.get("NAIVE_LOGGING", False)) else None
+        if logfile:
+            become_daemon(out_log=logfile, err_log=logfile)
+        else:
+            become_daemon()
 
         with open(pidfile, 'w') as f:
             f.write("%d\n" % os.getpid())
--- a/sphinx-docs/developer_docs/logging.rst
+++ b/sphinx-docs/developer_docs/logging.rst
@@ -6,6 +6,7 @@ If you wish to view output from the server, you have a few options:
 *  Start the server using the command `kalite manage kaserve --production`.
    This will start the server using CherryPy and a single thread, with output going to stdout and stderr.
 *  If you wish to capture the output from `kalite start`, then you need to do two thing:
+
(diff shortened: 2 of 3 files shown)


kalite/django_cherrypy_wsgiserver/cherrypyserver.py
```diff
@@ -142,7 +142,12 @@ def run_cherrypy_server(host="127.0.0.1", port=None, threads=None, daemonize=Fal
         # stop_server(pidfile)
 
         from django.utils.daemonize import become_daemon
-        become_daemon()
+        kalite_home = os.environ.get("KALITE_HOME", None)
+        logfile = os.path.join(kalite_home, "kalite.log") if (kalite_home and os.environ.get("NAIVE_LOGGING", False)) else None
```
aronasorman: Should we default this to `True` instead? If we can have the logger log on both a file and stdout, that would be great.

MCGallaspy: Unfortunately this implementation of `become_daemon` can only log to one file at once. I would also argue that if you want output to stdout, then you should not use `become_daemon` at all.

aronasorman:
Can you add a documentation entry for this?

aronasorman:
Once that's done feel free to merge

benjaoming:
This doesn't deprecate kaserve, but instead adds a mention of it in the docs. Why is the logging option not on `kalite start` ?

Also see [my comment of June 9th](https://github.com/learningequality/ka-lite/pull/3800#issuecomment-110352006)

MCGallaspy:
> Why is the logging option not on kalite start ?

You mean as another option? That's possible.

> This doesn't deprecate kaserve, but instead adds a mention of it in the docs.

It's in the "developer" section. Plus, I think it's preferable to run `kaserve --production`; you get the output and you can do whatever you'd like with it (including redirection). The option to get output from `kalite start` is there for convenience (by setting some environment variables), though I suppose it could be more convenient if that's what you mean.

sphinx-docs/developer_docs/logging.rst
```diff
@@ -0,0 +1,13 @@
+Logging
+=======
+
+If you wish to view output from the server, you have a few options:
+
+*  Start the server using the command ``kalite manage kaserve --production``.
+   This will start the server using CherryPy and a single thread, with output going to stdout and stderr.
```
benjaoming: What I would suggest would rather be `kalite start --foreground` or `kalite manage runserver` and then not mention `kalite manage kaserve`.

benjaoming:
@MCGallaspy would you mind if I take this PR from here? I really wanna get rid of the need to mention `kaserve` so we can just go for `kalite start --foreground` and `kalite manage runserver` for all other needs.

MCGallaspy:
Feel free.

PR #3888, opened 2015-06-17
asking re-enter password when create superuser

### LINKED ISSUE #3833: No need (or use) in asking for email address when creating superuser, and password should be hidden
Expected: The form that shows up when the server is first run without running setup asks for the critical information for setting up the server, and handles it appropriately.

Observed:

- The user is asked for an email address, which won't get used anywhere, and might be problematic for some users who are totally offline. We should just put a dummy email address into the user creation, and not show this to the user at all.
- The password field is shown in plaintext, rather than as ******. To fix this, set `<input type="password" ...>`. There should also be a second "confirm password" box, to avoid problems from typos, since there's no easy way to reset the password later.

image

66eli77:
fix #3833 by replacing email input with re-enter password on superuser create modal, also dot the input of password


MCGallaspy:
@66eli77 you may wish to fork from `develop` and apply your commit https://github.com/66eli77/ka-lite/commit/cb036a577c346619084708dbfcb02930b5785cbb instead of working from the `supermodaltest` branch, because it appears you are reverting some changes.

PR #3890, opened 2015-06-17
restore the bundled docopt.docopt function.

aronasorman:
The docopt function deleted here doesn't process positional arguments
correctly. This breaks a lot of management commands. We restore the old
one for now 'til we get this resolved.

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -75,6 +75,7 @@ else:
 import httplib
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


aronasorman:
Will merge this once tests are passing, as I need this working to make progress with the OS X installer.

benjaoming:
I recommend filing an issue about the nature of the problem, then I can fix it.

PR #3893, opened 2015-06-17
asking re-enter password when create superuser

### LINKED ISSUE #3833: No need (or use) in asking for email address when creating superuser, and password should be hidden

66eli77:
fix #3833 by replacing email input with re-enter password on superuser create modal, also dot the input of password


kalite/distributed/features/superuser_create.feature
```diff
@@ -2,58 +2,47 @@ Feature: Create superuser from the in browser modal
 
     Scenario: Superuser already exists
         Given I am on the homepage
+        And there is superuser
```
MCGallaspy: This step no longer exists, I removed it since a superuser always exists. (It's created in base_environment, from your last PR.)

kalite/distributed/features/superuser_create.feature
```diff
@@ -2,58 +2,47 @@ Feature: Create superuser from the in browser modal
...
         And I enter password correctly
-        And I enter email correctly
+        And I re-enter password correctly
         When I click the create button
-        Then the modal will dismiss
-        Given I am on the homepage
-        Then there should be no modal displayed
\ No newline at end of file
+        Then a superuser is created
+        And the modal will dismiss
```
MCGallaspy: I would say it's problematic to be reverting the changes to this and other scenarios in this file. I would delete `Scenario: Create superuser with email address longer than 40` as you have done, use the following steps for `Scenario: Create superuser with correct username and password and re-enter password`
```gherkin
    Scenario: Create superuser with correct username and password and re-enter password
        Given superuser is deleted
        And I am on the homepage
        And I enter username correctly
        And I enter password correctly
        And I re-enter password correctly
        When I click the create button
        Then the modal will dismiss
        Given I am on the homepage
        Then there should be no modal displayed 
```

and leave the other scenarios unchanged.

66eli77: you mean I should put "Given superuser is deleted" at the beginning of each Scenario except the first one?

MCGallaspy: Yeah, you should use that "given" for any scenario that requires a superuser does not exist, since a superuser is now created if it doesn't exist before each feature.

66eli77:
@MCGallaspy   the test pass on my local laptop but failed on Travis, I got this error message 
""This code isn't under transaction management")
      TransactionManagementError: This code isn't under transaction management"
does this means I cannot delete superuser within the test on Travis?

kalite/distributed/features/superuser_create.feature
```diff
@@ -5,55 +5,51 @@ Feature: Create superuser from the in browser modal
         Then there should be no modal displayed
 
     Scenario: Superuser does not exist yet
-        Given superuser is deleted
-        And I am on the homepage
-        Then I should see a modal
+        Given I am on the homepage
+        And superuser is deleted
```
MCGallaspy: I would reverse the order here: 
```gherkin
Given superuser is deleted
And I am on the homepage
```

Transaction errors seem to occur if you attempt to access the database (as with an API endpoint) while a write is still happening; I think that's the case here. Reversing the order should ensure the write finishes before any API endpoints are called.

kalite/distributed/templates/admin/superuser_form.html
```diff
@@ -8,18 +8,27 @@
...
             <div class="field">
+            <!--[if IE]>
+                <label for="id_superusername" style="float:left">{% trans 'Username' %}</label>
+            <![endif]-->
                 {{ form.superusername.errors }}
                 {{ form.superusername }}
             </div>
             </br>
             <div class="field">
+            <!--[if IE]>
```
MCGallaspy: Why only in IE?

66eli77: all other browsers support placeholder in input tag, only IE deserves the ugly labels.

MCGallaspy:
Screenshots in IE?

66eli77:
I was hoping @EdDixon can test this on IE once merged. I don't have IE setup on my machine

MCGallaspy:
I'm going to take a peke right now, we should definitely check it out __before__ it's merged. :)

MCGallaspy:
Here's IE 11:
ss

It appears that the label tags are being ignored, and also the placeholder seems to work ok. Is this targeting a specific version if IE?

66eli77:
any version lower than Internet Explorer 10 should display the label, and on IE8, placeholder is not supported, not sure about IE9

PR #3894, opened 2015-06-17
Django options were not parsed when they were already matched by doco…

### LINKED ISSUE #3892: Problems with arguments in kalitectl.py
See #3890. In the version of `kalitectl.py` now extant on `develop`, one needs to pass in --options to the kalitectl.py before the command, otherwise it's proxied to django management commands, e.g.:
```sh
kalite --port=1234 start --a-django-option
```
instead of
```sh
kalite start --port=1234 --a-django-option
```

In the version reverted by #3890 that is resolved, but then extra arguments are discarded, so the following two commands (for example) are equivalent:
```sh
kalite test my_test_label
kalite test
```
And this is true for all management commands that take extra arguments.

benjaoming:
…pt parser, only when they were NOT! #3892 

Fixes #3892

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -654,11 +654,19 @@ def docopt(doc, argv=None, help=True, version=None, options_first=False):
     # if matched and left == []:  # better error message if left?
     if collected:  # better error message if left?
         result = Dict((a.name, a.value) for a in (pattern.flat() + collected))
-        result['DJANGO_OPTIONS'] = sys.argv[len(collected) + 1:]
+        collected_django_options = result.get('DJANGO_OPTIONS', [])
+        result['DJANGO_OPTIONS'] = (
+            result.get('DJANGO_OPTIONS', []) +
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Looks like it pukes on this `bin/kalite start --traceback -v 2`, can you handle single-`-` options?

MCGallaspy:
Otherwise looks ok, very... __functional__.

benjaoming:
Yeah it doesn't like the space between `-v` and `2`. `-v2` works.

aronasorman:
Note: tests are failing because of that: we have `kalite manage test --traceback -v 2` in there.

benjaoming:
Ah yes, makes sense, I already changed them once to --version=2 but them
changed them back to -v 2 thinking that I had finally made the perfect
options parser :D

aronasorman:
Superseded by #3904.

PR #3897, opened 2015-06-17
Handle path names in cross-platform way

### LINKED ISSUE #3831: Subtitles not displaying
I'm having problems figuring out why no subs are showing, even though I have them downloaded, and can see them in the kalite\static\srt folder. 

**Branch**: today's develop

**Expected behavior**: Videos that have subs should show the subtitle or caption (not really sure which one, probably caption since that's the type of track in the player's code) button in the player control bar.
 
**Actual behavior**: No sub/caption button is visible in the player control bar:

sub-button-display-none

Since both buttons have `< style="display: none;" >` , I'm gonna venture a guess that the subtitle URLs for the  `<track >`  are not generated properly in the kalite/topic_tools/_ _init_ _.py file.

**Steps to reproduce**:

1. Set the DEBUG=False in the local settings
2. Choose the video that has subs from the sidebar (I checked the [APRIL 2015 MASTER](https://docs.google.com/spreadsheets/d/1k5xh2UXV3EchRHnYzeP6-YGKrmba22vsltGSuU9bL88/edit#gid=0) file, maybe there's another version?)

I tested with this file which has both English transcript & Spanish subtitle:
**Comparing numbers through 10**
comparing-numbers-through-10-example
https://www.khanacademy.org/math/early-math/cc-early-math-counting-topic/cc-early-math-comparing-numbers/v/comparing-numbers-through-10-example
mp4 & srt file name: **tJrSILRXOUc**

MCGallaspy:
Fixes #3831.

### DIFF
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -282,8 +282,8 @@ def get_content_cache(force=False, annotate=False, language=None):
         if os.path.exists(i18n.get_srt_path()):
             for (dirpath, dirnames, filenames) in os.walk(i18n.get_srt_path()):
                 # Only both looking at files that are inside a 'subtitles' directory
-                if dirpath.split("/")[-1] == "subtitles":
-                    lc = dirpath.split("/")[-2]
+                if os.path.basename(dirpath) == "subtitles":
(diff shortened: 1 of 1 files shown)


rtibbles:
Does the right fix. Tests pass. Merging.

PR #3879, opened 2015-06-16
Coach login redirects to coach reports

MCGallaspy:
Fixes #3857 and adds a test for it.

PR #3899, opened 2015-06-18
Coach login redirects to coach reports

### LINKED ISSUE #3857: Logging in as a teacher redirects to splash screen instead of coach reports
teacher login

Also, if you navigate to the "learn" page as a guest, and then try to login as a teacher from the learn page, it stays on the learn page instead of doing any kind of redirect.

MCGallaspy:
Fixes #3857 and adds a test for it. Supersedes #3879, since I rewrote history. Warning: looking at that PR now may cause headaches.


rtibbles:
Looks right to me.

PR #3900, opened 2015-06-18
Fixes #3885. Don't reference object directly in Backbone Model defaults, as object will be shared across instances.

### LINKED ISSUE #3885: ResponseLog does not get cleared when new AttemptLog is created.
branch: develop
Expected Behaviour: ResponseLog should only reflect events in this attempt.
Current Behaviour: ResponseLog contains events from all previous attemptlogs in this session.

rtibbles:
Not sure how long this has been extant, but it's going to give us some awfully crappy data.

Fixes #3885


rtibbles:
Fortunately, only extant on develop, not master.

PR #3904, opened 2015-06-18
Parse positional arguments correctly

### LINKED ISSUE #3892: Problems with arguments in kalitectl.py

aronasorman:
Adds a commit on top of #3894. In addition to that PR, we update the test calls to conform to the new parser.

Fixes #3892.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,11 @@ install:
 -  pip install coveralls
 script:
 # Test that we can run from source directory
-- bin/kalite start --traceback -v 2
+- bin/kalite start --traceback -v2
 - bin/kalite status
-- bin/kalite stop --traceback -v 2
+- bin/kalite stop --traceback -v2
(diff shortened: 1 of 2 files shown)


aronasorman:
Tests pass. Let's merge this.

benjaoming:
I would've done exactly the same, we can introduce 100% correct support for the short-hand django notations later.

PR #3905, opened 2015-06-18
Safely wait for page load in tests

MCGallaspy:
Just waits for a page to load before continuing. Should address the test failures in #3809.

### DIFF
--- a/kalite/testing/behave_helpers.py
+++ b/kalite/testing/behave_helpers.py
@@ -16,6 +16,10 @@ For finding and interacting with elements safely:
 
 For navigating the site:
 * build_url
+The following functions will wait for a page to load before continuing:
+* go_to_homepage
+* go_to_coachreports
+You can build your own such function using the "private" function `_go_to_url`.
 
 For logging in and out:
 * login_as_coach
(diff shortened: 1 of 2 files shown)


MCGallaspy:
@benjaoming

kalite/testing/behave_helpers.py
```diff
@@ -239,9 +243,33 @@ def build_url(context, url):
...
+def go_to_coachreports(context):
+    """
+    Goes to coach reports. Raises a TimeoutException if page does not load.
+    """
+    url = reverse("coach_reports")
+    _go_to_url(context, build_url(context, url))
 
+def _go_to_url(context, url, wait_time=MAX_PAGE_LOAD_TIME):
+    """
+    Tries to browse to a page. Waits for the page to load, since WebDriver.get does not.
```
rtibbles: Weird, I was under the impression that WebDriver.get does wait for page load?

http://selenium-python.readthedocs.org/en/latest/navigating.html

MCGallaspy: Hmm, seems you are right.

MCGallaspy: I guess my assumption about the test failures in #3809 was wrong, unless there's a race condition because stuff is happening after onload but before the rest of the JS stuff is ready.

rtibbles: Possible - but that's why we have all the explicit waits before interacting or checking any HTML elements?

MCGallaspy: What I think is happening is this. First of all, scenarios within a feature share one WebDriver instance. One of the failing steps is this:
```
When I click on the Show Tabular Report button     # kalite/coachreports/features/steps/coachreports.py:131 2.298s
    Then I should see the tabular report               # kalite/coachreports/features/steps/coachreports.py:204
Then I should see the tabular report               # kalite/coachreports/features/steps/coachreports.py:204 5.510s
```

And my suspicion is that the wait to find the element is not failing because all this is taking place on the _same_ page, so the element _exists_ but hasn't been updated/refreshed according to new conditions stipulated in the "given" steps. The other failures are similar -- the element being searched for _could_ exist, but may not reflect new conditions.  I assumed that `WebDriver.get` was the culprit, but if it does in fact wait for a page load then perhaps not. Unless it doesn't load the page again if the url is the same?

rtibbles: It's possible it won't, no.

I don't understand why these are failing only on @benjaoming''s PR though? Could he pull in latest develop and see if it is still extant?

MCGallaspy: @rtibbles I did that and pushed to my repo, here is the travis build for it: https://travis-ci.org/MCGallaspy/ka-lite/builds/67443425

MCGallaspy: Same failures. Now let me try it after merging this branch in.

MCGallaspy: @rtibbles the build: https://travis-ci.org/MCGallaspy/ka-lite/builds/67447341

MCGallaspy:
@rtibbles I think you're right, these waits are irrelevant. See the last build.

benjaoming:
@MCGallaspy sorry didnt see this before now, I'm wondering which build you were referring to when closing?

PR #3906, opened 2015-06-18
Video download and topic tree fixes

### LINKED ISSUE #2976: Video downloading invalidates content and topic tree caches at the end of every download
Branch: develop
Expected Behavior: hold off on invalidating content and topic_tree cache until all downloads have finished
Current Behavior: invalidates the content and topic_tree caches after every video completes
Steps to reproduce: install several language packs, download more than one video. Watch your Debug output go crazy as it continually invalidates all the languages
Screenshot(s):

image

### LINKED ISSUE #3743: Uncaught error on Learn page

### LINKED ISSUE #3835: Some topic tree API calls take a long time
Branch: develop
OS: Ubuntu

Expected: sidebar navigation should now be fast, given the fancy refactoring @rtibbles did.

Observed: Clicking around the top-level topics, often it would take much longer than usual (8+ seconds) to load the contents of the topic into the second level of the sidebar.

image

Note: The video download process was going on in parallel, which could be affecting this.

rtibbles:
Fixes #2976, and attempts to resolve #3835, and perhaps closes #3743 permanently.

Summary of changes:
* Only assign topic tree to global variable after the flat topic tree has been generated.
* Removes signals that trigger cache invalidation after modification of VideoFile model.

### DIFF
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -66,7 +66,7 @@ def get_topic_tree(force=False, annotate=False, channel=None, language=None, par
     if TOPICS.get(channel) is None:
         TOPICS[channel] = {}
     if annotate or TOPICS.get(channel, {}).get(language) is None:
-        TOPICS[channel][language] = softload_json(TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
+        topics = softload_json(TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
 
         # Just loaded from disk, so have to restamp.
         annotate = True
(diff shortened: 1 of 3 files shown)


MCGallaspy:
So when is video availability now updated?

rtibbles:
When the management commands have finished running (so, after all batched downloads or deletions have occurred).

MCGallaspy:
I see, so the key errors were because the topic tree cache variable was in an intermediary state of some sort?

MCGallaspy:
After starting the server the first time, the first call to `/api/topic_tree` still took ~1min. Subsequent calls were much faster. This is pretty much expected behavior. Also note that if initiating a large video download job, the results won't be available on the learn tab until the whole batch is finished. Looks ok!

Issue #3601, opened 2015-05-04
"Already stopped. Status was: Failed to start (check logs)"


jamalex:
On: demo server (master branch).

At one point, the server stopped responding, and when I went in to try to restart it, I got the following:

```
$ sudo bin/kalite stop
Already stopped. Status was: Failed to start (check logs)
```

Where are these logs of which it speaks? Also, I couldn't find a clean way to get the server restarted after that.

EdDixon:
if you issue a second sudo bin/kalite stop stop it should stop clean I think allowing a restart. I have seen that behavior a few times in the past with the older bat scripts as well. In a real pinch, sudo killall python will stop it also but if you have any other python apps running you may want to seriously consider what effects that will have.

benjaoming:
@jamalex you mean restarting by doing `kalite restart` ?

I think that if you can't do `kalite stop` in a clean and predictable way, it can fail as long as doing `kalite start` will still start it in case it had an "Already stopped" status.

But I would think that `kalite restart` shouldn't fail in case the server wasn't running... because that's how service commands work on Linux so I would think it's expected behaviour.

That message about logs doesn't make any sense if there aren't any logs, which I guess there aren't by default? So that should go.... But we really need to clean up the way things are logged so it's easy to determine log output and where it goes...

EdDixon:
Yea I always kind of felt the second stop I issued was just confirmation that the first one worked. CherryPy should be logging to /var/logs right?  As an httpd service I would think they would want that?

EdDixon:
Was just looking over their docs and CherryPy does support logging http://docs.cherrypy.org/en/3.3.0/refman/_cplogging.html#module-cherrypy._cplogging if configured also found this example: http://stackoverflow.com/questions/4056958/cherrypy-logging-how-do-i-configure-and-use-the-global-and-application-level-lo If we are not taking advantage of logging we might want to consider it as it?

jamalex:
It seems to be getting into a zombie state, where python2 (with `kalitectl.py start`) is still running in the background (though not serving requests), but `kalite status` shows "Failed to start (check logs) (6)". When I try to either start or restart it, it says "DatabaseError: database is locked". Trying to do anything with the DB via the `sqlite3` REPL directly also throws a "locked" error. So, it seems the background zombie process is locking up the DB and preventing a restart.

benjaoming:
@jamalex would you by any chance have found a recipe for recreating the issue?

[...]

PR #3917, opened 2015-06-23
Run both cherrypy and chronograph in same process #3550 #3601 #3827, …

benjaoming:
…add --foreground option and respect --port option and env KALITE_LISTEN_PORT

This PR centralizes the whole daemon behavior to one single call in kalitectl and removes lots of chained calls.

It also adds a `--foreground` option to run kalite in the foreground instead of going to daemon mode.

Tasks:

 * [x] Add `--foreground` option to `kalite start`.
 * [ ] Default logging to ~/.kalite/kalite.log and have "Already stopped (check logs)" message output path to log.
 * [x] Put deprecation warnings in kaserve etc.

Solves: #3550 #3601 #3827


benjaoming:
This includes #3809

MCGallaspy:
A few small comments, and let me test it out, but otherwise good to go. Nice!

MCGallaspy:
Tried running "python.exe kalite start" in Windows 8, everything worked okay, but the process never relinquished control back to the terminal...

MCGallaspy:
I guess I shoulda just used `kalite.bat`. Works ok, but has the opposite problem of not really accepting input (with e.g. breaks "kalite manage setup"). But given that start, stop, and status all require no input, I think that's okay for now.

Seems great. A++++ would merge.

benjaoming:
@MCGallaspy the input problem on Windows 8 was also an issue before, right? Would you be able to do a basic test of `raw_input` and see where the chain falls off? Would it be that `subprocess.Popen` can't handle it?

MCGallaspy:
Will this 
>Default logging to ~/.kalite/kalite.log and have "Already stopped (check logs)" message output path to log.

Appear here or in a separate PR?

PR #3918, opened 2015-06-23
Skip to links

radinamatic:
This is still a work-in-progress, so I'm submitting a PR mostly for feedback & help. 

The best implementation of "skip to" links should be useful for both screen-reader & sighted keyboard navigation users. This solution is inspired by the one on the http://webaim.org/

Two things remain to be solved/added here:  

1. After skipping to a particular point, there should be a way to grab the focus of the element. One possible solution would be something like described here (still working on how to implement it):
https://www.bignerdranch.com/blog/web-accessibility-skip-navigation-links/  

2. For some reason the two links inside the `{% if is_logged_in %}` condition are not showing now, even though they did display correctly (once the user logged in) before I put them inside the `<div id="skipto">`. I'm not really sure why... :(


MCGallaspy:
We haven't forgotten about this :)

radinamatic:
Thanks, @MCGallaspy! 
Yes, this still needs work, so 0.15 is a good milestone...

rtibbles:
For cleaner commit histories, doing rebase instead of merge can help!

kalite/distributed/templates/distributed/base.html
```diff
@@ -185,6 +185,21 @@
             <div class="navbar kalite-navbar header-footer-bg" role="banner">
               <div class="container">
                   <div class="row">
+
```
rtibbles: This needs to go into the relevant handlebars template, I believe, not here!

rtibbles:
Stale contrib and merge conflict.

rtibbles:
@radinamatic What's the status on this?

radinamatic:
Mea culpa, this got tota11y forgotten... :(

One of the reasons is that my plan was to implement this together with site-wide visual-focus-outline strategy, and for that I'd like to confer with @jtamiace. How soon should it be done to be viable for 0.15?

rtibbles:
Code freeze is end of next week.

radinamatic:
Then we better bump it to 0.16, since I doubt we'll have it by then (I'm due to travel next week, fever permitting).

rtibbles:
What's the prospectus for this one?

rtibbles:
Bumpity bump bump?!

rtibbles:
A bump bump?

benjaoming:
Closing this as @radinamatic has indicated there won't be time to finalize it immediately.

PR #3922, opened 2015-06-23
Fixup auto-generated username

### LINKED ISSUE #3530: 0.13.x kalite commands do not resolve paths with specific user names
Many of the related kalite command scripts such as kalite start rely on env paths and system paths, then building complete paths out from those. The current scripts are not allowing the use of certain systems usernames on Windows which are required by the scripts to create the internal paths needed. Currently the restriction is that "User Name must contain only letters, digits, and underscores, and start with a letter." So a user name with first and last will not resolve etc. @MCGallaspy and @benjaoming had previously resolved this issue by using proper escape characters in usability tests months ago but those may not have been merged into 0.13.x This can be replicated by running with a non compliant user name and then running the bat files directly as shown in the screen capture.
screenshot from 2015-04-16 16 34 21

MCGallaspy:
Fixes #3530. Tested it out by passing in a bad usernames using debugger. Hopefully this is uncontroversial!


aronasorman:
Merge conflict.

PR #3925, opened 2015-06-24
Docs setup customizing

benjaoming:
Sorry about the settings cleanup that went in this PR, but I was kind of looking at settings while writing the docs for them and felt inclined to get these two silly context processors out :)

Fix for #3925


benjaoming:
Btw. is INLINE_HELP a mature feature yet?

benjaoming:
I've been working on an idea to allow for a conventional settings module to be located in KALITE_HOME, `~/.kalite/custom_settings.py`. This would make instructions for users simpler, make it possible to always put an example file, and UIs would have a place to dump their settings without requiring `kalite --settings=some_ui.settings_module` which would break off from command line (e.g. you UI customizations to KA Lite would not appear when running `kalite` from the command line).

rtibbles:
@benjaoming INLINE_HELP is merged to keep the code in develop, but it is not currently active. The framework is all there, I believe, just needs the 'scripts' for different features.

I believe @tk21 may be generously volunteering some time alongside her full time gig to help us get those implemented over the summer.

tk21:
@rtibbles Yes, I'm still working on writing the 'scripts' for the different pages -- this is just a matter of appending to the "narratives.yaml" file. There is a bit of error-checking I need to implement, though. Such as only triggering a click on an element after checking that it has been loaded, or something along the lines of that. This affects the tutorials/scripts for which an action is triggered before showing the next tutorial modal.

benjaoming:
Overruled by #3938.

Issue #3921, opened 2015-06-23
Feature Request - User Accessibility Settings


radinamatic:
After we discussed the general color & contrast issues for better accessibility during our recent meeting, @jtamiace and I came with the idea to implement **User Accessibility Settings**. We thought of starting with options that should help primarily the users with low vision, and therefore nicely complement the efforts on video/captions improvements front (for hearing disabilities), and keyboard navigation improvement front for blind users. 

### Implementation examples  

**FluidProject - Learner Options** http://fluidproject.org/  
Check the **color & contrast** options in the **show display preferences** tab in the upper right corner. This solution is based on the whole **Infusion framework** and probably not the option we should try to implement right now, but serves as the nice example.  

![fluid](https://cloud.githubusercontent.com/assets/1457929/8317912/19c75842-1a01-11e5-8caa-73c9980a6cef.png)

**Alzheimer's Society - User Settings** http://www.alzheimers.org.uk/site/scripts/user_settings.php
Simple **High Contrast & Background** options that the users can choose from.  

![alzheimersociety](https://cloud.githubusercontent.com/assets/1457929/8317919/22c96098-1a01-11e5-8850-506cab0109c2.png)

-----------------------------------
I will be joining the Design team weekly meetings and @jtamiace mentioned that one of their interns may be available to work together on implementing this feature... :)

Anybody with other ideas, opinions and suggestions, feel free to chime in!

rtibbles:
Not going to happen within the scope of KA Lite.

@radinamatic please open an issue for this on the Kolibri repo.

PR #3929, opened 2015-06-24
Enable tab key access for User tab/dropdown - fixes #3928

### LINKED ISSUE #3928: User tab (dropdown menu) inaccessible by keyboard
**Branch**: latest develop  

**Expected Behavior**: Users should be able to access the content of the user tab (Points & Progress) by navigating with TAB key. 

**Current Behavior**: When navigating with TAB key, focus jumps from **Learn** tab to content further down, effectively impeding the user to access the elements inside the dropdown menu.

**Steps to reproduce**:

1. Log in as Learner.
2. Start navigating with TAB key.

radinamatic:
There is still room for improvement, but this fixes the immediate issue of accessing the dropdown contents.  
 
Tested with NVDA, but I have to retest with JAWS in another moment as my copy is currently misbehaving... :(


radinamatic:
On another matter, does anybody know why is GH pulling this `.gitignore` file into my PR? Shouldn't the "ignore" be, I don't know... ignored by default?

MCGallaspy:
This one is also still on the radar. :)

radinamatic:
This could technically be merged as it fixes the issue of access to options (I don't know if the build fail is a stopper).

On a second thought, though, I think I'm going to investigate a bit more the `role="presentation"` applied to elements of the menu, as it might need some revision. I'm reinstalling JAWS tomorrow, so I'll do some more testing...

rtibbles:
.gitignore is not ignored by default. You would have to add .gitignore to your .gitignore file in order for git to ignore it!

radinamatic:
@rtibbles Added as you suggested, but it seems mighty unintuitive to me...

Latest commit cleaned a few `role="presentation"` attributes that make sense only on `li class="divider"` elements. 

The only way `Points` element gets read as an element of the menu is if I put it inside `<a>`, otherwise as `<span>` it is ignored by my NVDA. Empty link is not a good solution though, and I'll look for an alternate.

rtibbles:
It's tricky though, because the 'points' is not a navigation item, so it's probably not as helpful to have it appear as a menu item in that way. 

Almost feels like the current presentation has the semantics wrong for the 'Points' item, as it is displayed as a menu item, but has no clickability.

Two possible solutions - we change it to not be a menu item, or we collapse it with the 'My Progress' indicator to make it a clickable link that is semantically related.

radinamatic:
"Dealing with tricky" seems to be the leitmotif of this line of work... :P 

I don't think it matters which solution we adopt as long as screen readers produce meaningful output for the user. Something like:

**Menu User Options for (`name-of-the-user`)**
- Points won (`number-of-points`)
- My Progress Report
- Accessibility Options _(feature proposed #3921)_
- Documentation _(structure for menuitem present)_
- Logout   

Which is how NVDA reads the menu with this commit (minus the two option/features not yet implemented). I don't have it available now, but I remember at some previous point JAWS read this menu (opened by mouse click) as:

**Menu Points**
- 3045   
- My Progress Report
- Logout

I agree that after a while users might get used and understand what is inside this menu, but I'd prefer if we could arrive to something like the first option. Pro-efficient screen-reader users tab-run through the available links and options anyway, so I'd rather give first-timers and the inexperienced users meaningful menu name and clear menu items.

rtibbles:
Plans for merge conflict resolution here?

radinamatic:
I'm working on `navigation.handlebars` for other issues, so I'll probably close this and open another one.

radinamatic:
Replaced by #4240. Closing.

Issue #3914, opened 2015-06-22
Delete skipped tests


aronasorman:
There's now a big mismatch between what tests are run on travis and what we run when we do a `kalite manage test` locally. Tests will be failing locally while they pass on Travis. Since we now depend on travis to run our tests rather than run them locally before pushing, this has led to these tests to rot away, and now they're failing.

Resolution: revisit all `@skipIf`ed tests, and delete them. Or put them in another file. But then that just shuffles things around. So I'd rather we delete them.

aronasorman:
Handled somewhat in #3932. Open another issue when needed.

PR #3932, opened 2015-06-24
Migrate to CircleCI

aronasorman:
Tests should now be passing in CircleCI. Changes:

- delete tests that are skipped in Travis. Addresses #3914.
- Rebased from #3817 to get some behave tests fixed.
- Tweaked circle.yml to get it up-to-date with travis.yml and the latest develop.


aronasorman:
Supersedes #3817.

aronasorman:
Rebased to develop.

aronasorman:
I'm getting behave errors again after rebasing to develop.

rtibbles:
Guessing these are related to the errors @benjaoming was getting on his pr that was merged earlier today?

benjaoming:
Yes that sounds plausible! I was looking previously at slowness of certain api_view functions that only happen with `runserver` and DEBUG=True but not with `kalite start`... if there's any kind of hint in that.

aronasorman:
After a bit more testing, I noticed that running our behave tests in different ways (i.e. just distributed tests vs. coachreports vs. all the tests) results in different tests failing each time.

I feel right now that our behave tests need more isolation with each other, if we wanna move forward with just making a more robust test suite in general. Shifting to another CI service won't change that.

However, I still see an advantage to switching to CircleCI, mainly because:
1. We can SSH in to the containers.
2. We have 4 containers available PER USER (so we can either make a build run faster by splitting it up, or run more concurrent builds)
3. We can pay to add more containers. Travis doesn't allow you to pay if you're open source.

Right now the plan is to split our tests 2 ways -- run behave tests in one container, run the rest in another. That would mean each build occupies 2 containers, and we can run 2 builds concurrently.

aronasorman:
By per user I mean if you register on CircleCI. Organizations also have 4 containers.

benjaoming:
And Circle builds faster than Travis?

aronasorman:
Without the split it does, but then that's because we've dumped 2.6 at this point. With the split between behave tests and other tests I'm expecting it to be faster, since we're running tests in parallel.

benjaoming:
If it's already evidently faster, I would recommend to remove Travis so we can start benefiting from this new fast CI :)

aronasorman:
Here's the latest build: https://circleci.com/gh/aronasorman/ka-lite/63

CircleCI has a degraded performance right now, which might explain why a majority of the behave tests are suddenly failing.

aronasorman:
So I feel that there are more pressing issues for me to solve now. There are still some test failures, but hey at least tests run a bit faster now :)

I'll leave this up to you guys to merge. Once it's done we can then turn off travis and fully migrate to CircleCI.

MCGallaspy:
I'm ready to merge this, like I mentioned in the other threads. We're pretty confident the features work as intended. That said, merging this will introduce test failures into develop. My suggestion is that we live with this for now (and maybe disable them so we don't wash out other potential legitimate failures) with the promise to revisit them after 0.14 release. And just to be explicit, I'm the one promising to revisit them.

benjaoming:
Agreed, and Travis is already failing on develop: https://travis-ci.org/learningequality/ka-lite/branches

Issue #3898, opened 2015-06-17
Assessment items as application data


benjaoming:
Currently the settings path for assessment items is hard-coded to be in the content root, which is a user data directory.

However, since this should be a shared piece of application data AND a download-able piece of user data, we need a condition that checks if the installation already provides bundled-in application data in a system-wide directory, `kalite.ROOT_DATA_PATH`.

Relevant code from `kalite.settings.base`:

```
KHAN_CONTENT_PATH = os.path.join(CONTENT_ROOT, "khan")
ASSESSMENT_ITEM_DATABASE_PATH = os.path.join(KHAN_CONTENT_PATH, 'assessmentitems.sqlite')
ASSESSMENT_ITEM_VERSION_PATH = os.path.join(KHAN_CONTENT_PATH, 'assessmentitems.version')
ASSESSMENT_ITEM_JSON_PATH = os.path.join(USER_DATA_ROOT, "data", "khan", "assessmentitems.json")
```

My proposal would be something like:

```python
_system_wide_assessment_items = os.path.join(ROOT_DATA_PATH, 'assessment_items')

if os.path.exists(_system_wide_assessment_items):
    ASSESSMENT_ITEM_ROOT = _system_wide_assessment_items
else:
    ASSESSMENT_ITEM_ROOT = KHAN_CONTENT_PATH

ASSESSMENT_ITEM_DATABASE_PATH = os.path.join(ASSESSMENT_ITEM_ROOT, 'assessmentitems.sqlite')
ASSESSMENT_ITEM_VERSION_PATH = os.path.join(ASSESSMENT_ITEM_ROOT, 'assessmentitems.version')
ASSESSMENT_ITEM_JSON_PATH = os.path.join(ASSESSMENT_ITEM_ROOT, "assessmentitems.json")
```

I'll have a look at this and figure out a way that does not affect old locations of assessment items.

benjaoming:
@jamalex would you care to share some insights on this? I think it's very closely related to our discussion on how to distribute for .deb vs. NOT distributing updates that have assessment items and still being able to run on a multi-user system :)

MCGallaspy:
Oh, looks like you have an open PR for this.

MCGallaspy:
Is this addressed in #3809?

benjaoming:
@MCGallaspy no, this is a PR on top of #3809 in order to keep it shorter.. but the other PR is a dependency :)

MCGallaspy:
[...]

PR #3934, opened 2015-06-24
Assessment item path fix

benjaoming:
PR for #3898

Is not entirely done necessarily, still needs some testing.. this is more to show the general idea.

Will re-assign PR as soon as I'm confident that it needs testing and reviewing by others.

Sorry about the PEP8 stuff in setup.py, there was just a lot of mis-aligned indentation, so I autopep8'ed it :)


kalite/distributed/management/commands/setup.py
```diff
@@ -50,18 +59,41 @@ def raw_input_password():
...
+        return
+    print("Cleaning *pyc files (if writable) from: {0}".format(path))
+    for root, __dirs, files in os.walk(path):
+        pyc_files = filter(
+            lambda filename: filename.endswith(".pyc"), files)
+        py_files = set(
+            filter(lambda filename: filename.endswith(".py"), files))
+        excess_pyc_files = filter(
+            lambda pyc_filename: pyc_filename[:-1] not in py_files, pyc_files)
+        for excess_pyc_file in excess_pyc_files:
```
aronasorman: Based on this line, we don't remove pyc files that have preexisting py files. Is that correct? If so, the docstring needs to reflect that.

kalite/distributed/management/commands/setup.py
```diff
@@ -50,18 +59,41 @@ def raw_input_password():
...
+
+
 def validate_username(username):
     return bool(username and (not re.match(r'^[^a-zA-Z]', username) and not re.match(r'^.*[^a-zA-Z0-9_]+.*$', username)))
 
+
 def get_clean_default_username():
     return (getpass.getuser() or "").replace("-", "_")
 
+
```
aronasorman: You linter!

benjaoming: sorry, there were so many formatting errors in the file, so I just autopep8'ethed it :)

Issue #3939, opened 2015-06-26
Running `kalite manage setup` with arguments yields different values.


cpauya:
**Branch** develop
**Target** 0.14.x
**Current Behavior** When running `kalite manage setup --username user --password pass --noinput` - the value of the `username` actually is set to the "cleaned" username of the logged-in user of the computer and the `password` to `None`.
**Expected Behavior** Running the above command should assign the values to the corresponding arguments.

**Further Testing** 

1. Try running `kalite manage setup --noinput --username user --password pass` and the script will be set to interactive-mode.  Meaning it did not follow the `--noinput` directive, the `username` is `None`, and only the password argument has correct value of `pass`.
1. To preserve the `--username` and `--password` values in non-interactive mode:  Run it in this format: `kalite manage setup --host '' --description '' --username user --password pass --noinput`.

**Related Issue/s**
These are the related / affected issue/s based on the above observations: 
* learningequality/installers#63 - Can not Log Into KA-Lite post Install MAC

benjaoming:
@cpauya can you try the same but putting `--option=value` ?

cpauya:
That works @benjaoming - I'm just amazed how this has caused an issue on the installer/s.

But reading about it on [Python's optparse module re *option argument*](https://docs.python.org/2/library/optparse.html#terminology) - I now understand why using your suggested format of `--option=value` works.

Thanks much for the speedy response. :)

benjaoming:
It's an unintended side-effect of a missing feature in docopt. I tried writing a patch for it, but it became a mess.

I will try to detect if there is an illegal pattern used and fail telling users to use --option=value

cpauya:
That'll surely be helpful in case someone forgets the format for option arguments.  :+1:

PR #3940, opened 2015-06-26
Fail when illegal option pattern is used #3939

benjaoming:
The regex isn't super perfect, but this is not mission critical.. I don't see it matching and failing with acceptable patterns.

#3939

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -76,6 +76,13 @@ import httplib
 import re
 import subprocess
 
+# We do not understand --option value, only --option=value.
+# Match all patterns of "--option value" and fail if they exist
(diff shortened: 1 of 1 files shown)


kalitectl.py
```diff
@@ -76,6 +76,13 @@
 import re
 import subprocess
 
+# We do not understand --option value, only --option=value.
+# Match all patterns of "--option value" and fail if they exist
+__validate_cmd_options = re.compile(r"--[^\s-]+\s+[^-\s][^-\s]+")
```
aronasorman: This is an opaque regex right here. Can we get away with normal substring matching?

benjaoming: I would not think so. Got suggestions?

PR #3946, opened 2015-06-27
Fix for extant test failures on develop. Add options for skipping bdd to TestRunner.

rtibbles:
It is likely that the login test failure is as a result of the incorrect facility being selected during login (hence giving a 'bad username' rather than 'bad password' error).

In addition, the proposed --bdd-only flag does not actually function, and is meant to come from a bdd_only flag. This option has been added to allow only running the bdd tests.


rtibbles:
And... doesn't work. Never mind!

PR #3947, opened 2015-06-27
Fixes "Computing" domain description.

### LINKED ISSUE #3834: Less informative description for "Computing" topic needed
If you don't have anything good to say, don't say anything at all. We should edit topics.json to get rid of the "s" (it came from KA).

image

rtibbles:
Fixes #3834


rtibbles:
image

PR #3954, opened 2015-06-29
for review

foolchan2556:
Hi Aron,

I just want to create a basic url in a basic app, only to find it return 404. Do you have any idea about what's wrong? Thanks!


aronasorman:
You forgot to add an entry to the distributed/urls.py file: https://github.com/aronasorman/ka-lite/blob/dca213ae4c08ea106e2d3133df04b7224a9d40b9/kalite/distributed/urls.py

PR #3955, opened 2015-06-29
Simplify logging and docs, and command line instructions

benjaoming:
Final amendment to #3601

@MCGallaspy I replaced NAIVE_LOGGING with a default log file that's just truncated every time you run `kalite start`. I think it matches the needs better on a production system.

I don't mind that it's there by default, because you'd have to have an incredible uptime and usage to make a significant impact on file size. Everything else will memory leak and die before this is an issue, and restarting kalite always truncates the log file.

In case there's a failure, normal users can refer to the log. I could add more interesting features that kept latest lines from a previous log, but on the other hand, I feel like we need to get a proper logging mechanism and not spend more time on this :)

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -353,7 +353,7 @@ class ManageThread(Thread):
         utility.execute()
 
 
-def manage(command, args=[], as_daemon=False, as_thread=False):
+def manage(command, args=[], as_thread=False):
     """
     Run a django command on the kalite project
 
@@ -363,9 +363,7 @@ def manage(command, args=[], as_daemon=False, as_thread=False):
     :param as_thread: Runs command in thread and returns immediately
     """
     
-    assert not (as_daemon and as_thread), "You cannot have a daemon and a thread"
-    
-    if not as_daemon and not as_thread:
+    if not as_thread:
         if PROFILE:
             profile_memory()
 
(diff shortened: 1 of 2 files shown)


MCGallaspy:
The changes look fine. I agree a proper logging mechanism is needed. I still think it should be disabled by default, unless we can show that the rate at which the logfile grows is super slow. I'm going to time it for a few minutes and see what happens.

kalitectl.py
```diff
@@ -480,12 +454,13 @@ def start(debug=False, daemonize=True, args=[], skip_job_scheduler=False, port=N
...
-            print("With logging...")
-            become_daemon(out_log=logfile, err_log=logfile)
-        else:
-            become_daemon()
+        kwargs = {}
+        # Truncate the file
+        open(SERVER_LOG, "w").truncate()
+        print("Going to daemon mode, logging to {0}".format(SERVER_LOG))
+        kwargs['out_log'] = SERVER_LOG
+        kwargs['err_log'] = SERVER_LOG
```
MCGallaspy: Seems to grow slow enough (about 50 bytes/second), but I'm seeing this error repeatedly in the logs:
```py
  File "C:\Python27\lib\logging\__init__.py", line 880, in emit
    stream.write(fs % msg)
ValueError: I/O operation on closed file
Logged from file _cplogging.py, line 280
```

MCGallaspy: Error is probably unrelated to the changes here, though.

benjaoming: Sounds like an interesting error, could you post a separate issue?

It *could* be because that the stream isn't threadsafe... logging writes to stdout and then that's actually redirected to our file stream. It might not happen with DEBUG=False, which is the default... which were you testing with?

benjaoming: And yes, agreed, this would have happened both with the old and new approach anyways... I'm thinking that making failure output available is extremely crucial, so we really have to make it default. We need an option for the future to select a different file and switch it off.

PR #3961, opened 2015-06-30
Implementing the tota11y accessibility visualization toolkit

radinamatic:
http://khan.github.io/tota11y/

Handy tool for checking the most common a11y pitfalls. So far I haven't seen any reason why would this overload the browser with violations for Khan & Perseus, but atm I'm having some issues with assessments on my end so if anybody can test the branch on their end with working exercises.


radinamatic:
Another tota11y (I'm gonna start using this everywhere ;) ) unintuitive thing about GH: why the heck is it pulling commits from another branch???

kalite/distributed/templates/distributed/base.html
```diff
@@ -224,6 +224,9 @@
         </div> {# End wrapper #}
         {# Footer #}
         {% include "distributed/partials/_footer.html" %}
+
+        <script type="text/javascript" src="{% static 'js/tota11y.min.js' %}"></script>
```
aronasorman: Can you have this loaded when a setting is `True`? Something like `IS_TOTA11Y`.

aronasorman:
@radinamatic you probably branched off from that branch, not develop :) `git rebase -i` to the rescue!

kalite/distributed/templates/distributed/base.html
rtibbles: I would rather just have this automatically come on when DEBUG is True, @aronasorman. That way we can be poking devs in the eye when they make bad accessibility choices!

rtibbles: @radinamatic See here for an example of how to only load this script when DEBUG is true: https://github.com/radinamatic/ka-lite/blob/tota11y/kalite/distributed/templates/distributed/base.html#L69

@aronasorman As I start to play around with the browserify refactor, I will look at how to exclude stuff like this from the built Javascript.

radinamatic: Ok, I get the idea... :) 
Since we do not have an `else` option here, this should be enough, right?

```
{% if settings.DEBUG %}
<script type="text/javascript" src="{% static 'js/tota11y.min.js' %}"></script>
{% endif %}
```

rtibbles: Yep, no alternative required.

radinamatic: Will do, captain!

radinamatic:
Shoot, why do I never think of these possibilities... :(
So if I do the `git rebase -i` from this sub-branch `tota11y` and eliminate the commit which pertains to the the parent branch `assesmentzip_to_docs`, will the commit still be available for the parent branch?

aronasorman:
@radinamatic yup! `git rebase -i` only affects the current branch.

radinamatic:
Ok, understood, child-branches have no influence on parents, nothing like real life... :)

Having a total git-meltdown these days!

rtibbles:
Only the sins of the parent shall be borne by the sins of the children. So git branches are like Klingons.

radinamatic:
Nothing like a good Trek reference to make coding more colorful!

radinamatic:
Seems I finally made some sense of it... (Couns. Troy would be proud ;)

If the rest of the command crew has no other objections (Circle CI included), this can actually be merged!

radinamatic:
I'm working of this branch, and then cherry-picking commits to other branches with their own PRs. I will do a final rebase before this is to be merged.

aronasorman:
Merge conflict :)

aronasorman:
@radinamatic I think a developer docs entry describing the motivation for using this would be helpful. I'd like to see a short motivation section, a couple of links that will help us get up to speed with this tool, and some tips, tricks and recommendations regarding tota11y and maybe a11y in general.

radinamatic:
Makes sense @aronasorman! I will start working on the dev docs entry right away.

Docs have been changing so much lately, it's hard to keep up. Is this a good place to add the a11y entry: https://learningequality.org/docs/developer_docs/contributing.html

If not, where?

aronasorman:
Yeah, the developer docs section would be good. Maybe make another page. "How can I contribute to... Accessibility."

radinamatic:
Added the dev doc page as suggested and rebased! @aronasorman

radinamatic:
@aronasorman Can you give me a hint as why this branch has conflicts?

aronasorman:
My guess is it's with `base.html`, but i'm not too sure as the diff is too small for that. But it should be easy to see why on the command line!

radinamatic:
I don't know if it was the `base.html`, but I just made a fresh PR ( #4209) with just that and the .js library, and it seems fine. Well, CircleCI is complaining, but what else is new.

I'll cherry-pick the commit with the developer docs entry, and if everything goes well, we can close this and you can merge the new one.

@aronasorman

radinamatic:
Since we have #4209 now (with no merge conflicts, apparently), can we close this?
@aronasorman

PR #3965, opened 2015-07-01
0.14.x into develop

MCGallaspy:
It is time.


benjaoming:
Agreed, especially the 30,000 removed LOC makes it unbearable to read diffs currently.

MCGallaspy:
It was suggested to me that our version numbers (due to internal use) _need_ to be of the form "N.N.N". This precludes PEP440 compliant version strings like "0.14a3", and of course strings like "0.14.alpha3" are _not_ PEP440 compliant. I think the commit you're commenting on here is outdated! In a later commit (and this is reflected currently on develop) I explained what the intersection of PEP440 and our requirements are:
```py
# Must be PEP 440 compliant: https://www.python.org/dev/peps/pep-0440/
```

benjaoming:
Yeah, just to sum up: Because of the way that A) the central server works (doesn't understand queries involving "0.14a3") and B) that debian packages don't order 0.14a3 ahead of 0.14.alpha2, I stuck to this naming convention in later version changes. We can stop it when we release the final 0.14.

PR #3966, opened 2015-07-02
Update installation docs, Raspberry Pi finalization + settings module / config package cleanup

benjaoming:
Clean up settings regarding config packages, removing the demo one. Add deprecation warnings. Move screenshot stuff to kalite.project.settings.screenshots ( @MCGallaspy )

* [x] Clean up config packages and add deprecation warnings
* [x] Add `kalite.project.settings.raspberry_pi`
* [x] Add `kalite.project.settings.demo`
* [x] Add `kalite.project.settings.screenshots`
* [x] Update Raspberry Pi config script
* [x] Update docs


benjaoming:
Please do not merge yet :)

kalite/project/settings/screenshots.py
```diff
@@ -0,0 +1,27 @@
+"""
+This module is for running a screenshot server (internal FLE purpose mainly).
+To use it, run kalite with --settings=kalite.project.settings.screenshots or
+set then environment variable DJANGO_SETTINGS_MODULE to
+'kalite.project.settings.screenshots'.
+"""
```
MCGallaspy: This will probably affect building the docs, since the screenshot directive attempts to run a management command without specifying a settings module. Is `--settings` honored by all management commands? I'll try this out.

benjaoming: Yes, --settings is picked up by the basic django management execution stuff.. it overwrites DJANGO_SETTINGS_MODULE

benjaoming: And sorry about changing the interface, but I'm sure it's easy to fix :)

MCGallaspy: > but I'm sure it's easy to fix :)

Sometimes a dangerous assumption!

kalite/settings/__init__.py
```diff
@@ -65,16 +69,16 @@ def package_selected(package_name):
...
+        RemovedInKALite_v015_Warning
+    )
     TURN_OFF_MOTIVATIONAL_FEATURES = True
     RESTRICTED_TEACHER_PERMISSIONS = True
     FIXED_BLOCK_EXERCISES = 5
     QUIZ_REPEATS = 3
 UNIT_POINTS = 2000
 
-# for extracting assessment item resources
-ASSESSMENT_ITEMS_ZIP_URL = "https://learningequality.org/downloads/ka-lite/%s/content/assessment.zip" % version.SHORTVERSION
```
benjaoming: This is moved to a location further up the file so can be overwritten by other settings modules.

scripts/raspberry_pi_setup.sh
```diff
@@ -1,18 +1,32 @@
 #!/bin/bash
 
-SCRIPT_DIR=`dirname "${BASH_SOURCE[0]}"`
-if [ -e "$SCRIPT_DIR/python.sh" ]; then
-    KALITE_DIR=$SCRIPT_DIR/../kalite
-else
-    KALITE_DIR=$SCRIPT_DIR/kalite
```
benjaoming: This block was no longer necessary because the script is not supposed to be run in a git checkout.

scripts/raspberry_pi_setup.sh
```diff
@@ -1,18 +1,32 @@
...
+    echo "Changing user config to use raspberry pi settings: $USER_CONFIG"
+    echo "from kalite.project.settings.raspberry_pi import *" > $USER_CONFIG
 fi
 
-#we_are_rpi=`"$current_dir/get_setting.sh" package_selected\(\"RPi\"\)`
-#if [ $we_are_rpi != "True" ]; then
-#    echo "Error: we aren't configured as a Raspberry Pi, cannot continue"
-#    read WAITING
-#    exit
-#fi
```
benjaoming: outdated, look at the above code -- it forces the user's settings to use the raspberry_pi settings module instead.

MCGallaspy:
@benjaoming this one is on my to-do list, but first I want to make sure it doesn't break building the docs.

benjaoming:
Cool, thanks!! I've added the settings option now. So it should build, but I still don't know quite how to run `make html`, it just dies with "No module named selenium.webdriver.common.keys"

benjaoming:
Functional on RTDs, also screenshots:

http://ka-lite.readthedocs.org/en/raspberry_pi_finalizations/usermanual/userman_coach.html

Recommending merging this, and I will change the setup of the RTD account to run off the learningequality github account instead.

benjaoming:
@jamalex are you okay about publishing on RTD as a supplement of current docs? It's not a consequence of merging, just an option.

jamalex:
RTD is cool, but I don't see big advantages to hosting the user docs both on RTD and on our own server. One advantage of running on our server is that I imagine we get better log data.

I could imagine a nice split would be to have our developer docs on RTD, and user docs on our own site. But not sure whether there's a big advantage to having that split either.

benjaoming:
RTD supports google analytics, so if you send me an ID for a tracker, I can add that.

benjaoming:
It's about off-loading the task of building docs. RTD implements github hooks, can track several different branches and versions. It's an advantage not having to deal with these things.

jamalex:
Oh, so is it running our custom Python code/directives and taking the screenshots itself now? In that case, might be handy -- my understanding had been that wouldn't be possible. What approach can we take for generating the PDFs, in that case? And can it give them to us in an HTML format that's good for embedding in the app? (Because, if we're going to get the advantage of "not having to deal with those things", then we should stop building docs ourselves on our own servers, and just use RTD for building).

benjaoming:
RTDs also generates PDFs. It's not something that I've looked into, but here it is:

http://readthedocs.org/projects/ka-lite/downloads/pdf/raspberry_pi_finalizations/

There's also an HTML and EPub for download.

You can see the results of changing the window size for the screenshots here:

http://ka-lite.readthedocs.org/en/raspberry_pi_finalizations/

But yes, Mike's screenshot.py is implemented and used in the docs, no going back from that. But instead of having a complicated build server that grabs new screenshots on every build, I just made a quick-fix so that the command can be run locally.

It also sets window heights correctly btw, which is really cool.

MCGallaspy:
I don't think this makes building the docs significantly easier... someone still has to run the command to generate screenshots and commit them. The option to skip creating screenshots is certainly helpful in the case that you wish to quickly create the docs to see how your changes look locally. I think in general having even an "optional" build process violates the principle of keeping the docs in one location. I agree that the current build process for the docs is frustrating, but I think getting wrapped up in another build process right now could lead to unnecessary heartache. "Better the devil you know..."

jamalex:
> "Better the devil you know..."

:+1:

benjaoming:
This is ready to merge. Just keep in mind that the fields server needs to be explicitly building screenshots after this:

    SPHINX_SS_USE_PVD=true make SPHINXOPTS="-D screenshots_create=1" html

benjaoming:
you might wanna do this to ensure that future git pull's dont create conflicts:

    SPHINX_SS_USE_PVD=true make SPHINXOPTS="-D screenshots_create=1" html && rm -rf .screenshots_dump && git checkout .screenshots_dump

docs/conf.py
```diff
@@ -111,23 +111,28 @@
...
+    import sphinx_rtd_theme
+    # html_theme = 'sphinx_rtd_theme'
+    html_theme = 'kalite-theme'
+    html_theme_path = ['.', sphinx_rtd_theme.get_html_theme_path()]
 
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
-html_theme_options = {
+# html_style = "kalite.css"
```
MCGallaspy: Intentional? If we don't wish to use these styles, we should remove the file.

docs/installguide/install_all.rst
```diff
@@ -0,0 +1,307 @@
+Windows Installation
+====================
+
+1. Download the KA Lite `Windows <https://learningequality.org/r/windows-installer-latest>`_ installer.
+2. After downloading the .exe file, double click it. A window will appear and guide you through the process of installing KA Lite on the server.
+
+Mac Installation
+================
+
+1. Download the KA Lite `OSX installer <https://disney.com>`_.
```
MCGallaspy: @aronasorman what shall we put here?

aronasorman: I'll upload something to the downloads server and send the ink.

docs/installguide/install_all.rst
```diff
@@ -0,0 +1,307 @@
...
+
+To install KA Lite from a file, go to our `PyPi page <https://pypi.python.org/pypi/ka-lite-static>`_ and download a .tar.gz or .zip. You can then install the file from a system by unpacking it and running *python setup.py install* or *pip install ka-lite-static-X.Y.Z.tar.gz*.
+
+Alternatively, install with *pip* using the instructions below:
+
+.. note:: Type the following commands in a terminal.
+
+#. Install pip (Ubuntu: ``sudo apt-get install python-pip``)
+#. (Recommended; essential on slower platforms like Raspberry Pi) Install M2Crypto (``sudo apt-get install python-m2crypto``).
+#. Run ``sudo pip install ka-lite-static`` (bundled dependencies) or ``sudo pip install ka-lite`` (normal pip behaviour, dependencies installed to host system).
```
MCGallaspy: So `ka-lite-static` installs as one package with deps bundled in, whereas `ka-lite` will actually install the deps, right? Does `ka-lite-static` have frozen versions of the deps?

docs/installguide/install_all.rst
```diff
@@ -0,0 +1,307 @@
...
+Configuration after installing / updating
+_________________________________________
+
+Every time you install or update kalite, you should (re)run ``kalite manage setup``
+to setup the database and download assessment items (video descriptions,
+exercises etc.).
+
+
+Uninstalling
+============
```
MCGallaspy: This section should also include instructions for Windows & Mac.

docs/installguide/install_all.rst
```diff
@@ -0,0 +1,307 @@
...
+
+You can remove KA Lite (when installed from pip or source distribution) with
+`pip uninstall ka-lite` or `pip uninstall ka-lite-static` (static version).
+
+For Ubuntu, use ``apt-get remove ka-lite``.
+
+
+Removing user data
```
MCGallaspy: Also OS specific. `~` at least is not conventional on Windows, but the Python docs [describe how it is resolved](https://docs.python.org/2/library/os.path.html#os.path.expanduser).

docs/installguide/install_all.rst
```diff
@@ -0,0 +1,307 @@
...
+In our tests, we found that the WiPi adaptor supported a higher number tablet connections.
+
+
+.. note:: The Raspberry Pi may crash if the USB adaptor is inserted or removed while the computer is switched on.
+
+    * Make sure to shutdown and remove the power from the Raspberry Pi.
+    * Afterwards, insert the wireless USB adaptor.
+    * Lastly, switch the Raspberry Pi on.
+
+#. Make sure the Raspberry Pi operating system is up-to-date.
```
MCGallaspy: Formatting!

MCGallaspy: Looks ok, actually.

docs/screenshot.py
```diff
@@ -10,19 +9,11 @@
 from docutils.parsers.rst import Directive
 import docutils.parsers.rst.directives as directives
 from exceptions import NotImplementedError
-from selenium.webdriver.common.keys import Keys
+print sys.path
```
MCGallaspy: why?

docs/screenshot.py
```diff
@@ -253,7 +258,10 @@ def run(self):
             self.url = self.options['url']
 
         if 'navigation-steps' in self.options:
-            self.filename = uuid.uuid4().__str__()
+            from hashlib import md5
```
MCGallaspy: Good.

docs/usermanual/userman_admin.rst
```diff
@@ -490,20 +494,26 @@ These are resized videos. All in all, this will take around 23 GB of space.
...
+
 If you skipped the step for adding assessment items during install, run the following command (beware it can take a while, the download size is over 500 MB):
 
 ``kalite manage unpack_assessment_zip https://learningequality.org/downloads/ka-lite/0.14/content/assessment.zip -f``
 
 As an alternative, you can download the zip file from https://learningequality.org/downloads/ka-lite/0.14/content/assessment.zip and run the command on your local copy. This is useful if you're deploying KA Lite on multiple computers.
 
+
 Language Packs
 ---------------------------
```
MCGallaspy: Might as well fix this one, too.

kalite/settings/__init__.py
```diff
@@ -32,22 +34,24 @@ def package_selected(package_name):
...
-
-# config_package (None|RPi) alters some defaults e.g. different defaults for Raspberry Pi(RPi)
-# autodetect if this is a Raspberry Pi-type device, and auto-set the config_package
-#  to override the auto-detection, set CONFIG_PACKAGE=None in the local_settings
 ########################
 
-CONFIG_PACKAGE = getattr(local_settings, "CONFIG_PACKAGE", "RPi" if (platform.uname()[0] == "Linux" and platform.uname()[4] == "armv6l") else [])
-
+# A deprecated setting that shouldn't be used
+CONFIG_PACKAGE = CONFIG_PACKAGE = getattr(local_settings, "CONFIG_PACKAGE", [])
```
MCGallaspy: What's the intention here?

kalite/settings/__init__.py
```diff
@@ -32,22 +34,24 @@ def package_selected(package_name):
...
 if isinstance(CONFIG_PACKAGE, basestring):
     CONFIG_PACKAGE = [CONFIG_PACKAGE]
 CONFIG_PACKAGE = [cp.lower() for cp in CONFIG_PACKAGE]
 
 
+if CONFIG_PACKAGE:
+    warnings.warn(
+        "CONFIG_PACKAGE is outdated, see docs (TODO MCGallaspy, URL ?)",
```
MCGallaspy: Why are you tagging me? I don't know what this setting is supposed to do, either... :(

@jamalex @rtibbles any insight?

MCGallaspy: Or rather, I don't know what the docs are supposed to explain here.

kalite/settings/__init__.py
```diff
@@ -82,21 +86,6 @@ def package_selected(package_name):
         KEY_PREFIX += "|restricted"  # this option changes templates
     DISABLE_SELF_ADMIN = True  # hard-code facility app setting.
 
-if package_selected("Demo"):
```
MCGallaspy: These settings now live in a different settings module, yes?

kalite/settings/__init__.py
rtibbles: I feel like the warning below is sufficient, no? If they had previously used a CONFIG_PACKAGE, then they should now use a specific settings module from kalite.project.settings.

scripts/raspberry_pi_setup.sh
```diff
@@ -1,18 +1,32 @@
 #!/bin/bash
```
MCGallaspy: It would be nice to have someone test this out on an RPi. @66eli77 ?

sphinx-docs/.screenshot_dump/.gitignore
```diff
@@ -1,3 +0,0 @@
-# Screenshots are dumped here by the build process
-*
```
MCGallaspy: Why'd ya remove this? git won't include empty directories, so this file ensures the directory exists but nothing in it is committed.

MCGallaspy: Ah, I see, you're now committing images.

MCGallaspy:
Okay, this doesn't seem too controversial! Let me update the build process on field to incorporate this. I'm against using RTD for reasons mentioned above (proliferation ruins authoritativeness), so please disable that.

MCGallaspy:
Also, why rename `sphinx-docs` to `docs`?

MCGallaspy:
I've made my suggested changes and will open a new PR superseding this one.

docs/installguide/install_all.rst
benjaoming: Yes, it does. It uses requirements.txt

docs/installguide/install_all.rst
benjaoming: I fixed up lots of stuff... and lots of stuff I didn't fix. But it looks like I wrote everything... but that's the sphinx-docs -> docs rename.

docs/screenshot.py
benjaoming: debug err!

kalite/settings/__init__.py
benjaoming: The URL for the docs -- sorry about tagging you, it was under the assumption that you were picking up after.. I was finding it difficult to link to a chapter in the docs because I didn't know the final location of them.

kalite/settings/__init__.py
benjaoming: yes, kalite.project.settings.demo

sphinx-docs/.screenshot_dump/.gitignore
benjaoming: yes, but thanks for the method, I used to wonder how to force git to include empty dirs

PR #3972, opened 2015-07-02
@echo off to make output redirection nicer

MCGallaspy:
In the future, in windows, we may wish to e.g. get the version number of a previous install with the command "kalite.bat --version 1> version.txt". Without `@echo off`, this will also print the contents of kalite.bat.

### DIFF
--- a/bin/windows/kalite.bat
+++ b/bin/windows/kalite.bat
@@ -1,2 +1,3 @@
+@echo off
 set "BIN_DIR=%~dp0"
 start /b python.exe "%BIN_DIR%..\kalite" %*


benjaoming:
FYI this was already in the bat template used in setup.py:

```
BAT_TEMPLATE = \
    r"""@echo off
set mypath=%~dp0
set pyscript="%mypath%{FNAME}"
set /p line1=<%pyscript%
if "%line1:~0,2%" == "#!" (goto :goodstart)
echo First line of %pyscript% does not start with "#!"
exit /b 1
:goodstart
set py_exe=%line1:~2%
call %py_exe% %pyscript% %*
"""
```

PR #3975, opened 2015-07-02
Unbundle kaa

benjaoming:
Causes Ubuntu Software Centre to not like the .deb because our bundled kaa contains .so files.


benjaoming:
Circle CI timeout, not going to rerun as it passed most.

benjaoming:
Codebase just got 1.8 MB smaller!

Issue #4004, opened 2015-07-03
`kalite start` should handle issues with acquiring port more gracefully


jamalex:
Running: http://ppa.launchpad.net/benjaoming/ka-lite/ubuntu/pool/main/k/ka-lite-source/ka-lite_0.14.alpha2-0ubuntu1_all.deb on Ubuntu (but assuming this is more general)

I had KA Lite running elsewhere, then installed, and went to the command line to try the `kalite start` command:

    jamalex@ace:~$ kalite start
    Running 'kalite start' as daemon (system service)

    Stand by while the server loads its data...

    INFO:kalite:Invalidating the web cache.
    INFO:kalite:Running videoscan.
    INFO:kalite:Preloading exercise data for language en.
    INFO:kalite:Preloading content data for language en.
    INFO:kalite:Preloading topic tree data for language en.
    INFO:kalite:Preloading exercise data for language es.
    INFO:kalite:Preloading content data for language es.
    INFO:kalite:Preloading topic tree data for language es.
    INFO:kalite:Copying static media...
    WARNING:py.warnings:/usr/share/kalite/python-packages/tastypie/resources.py:328: UserWarning: 'override_urls' is a deprecated method & will be removed by v1.0.0. Please rename your method to ``prepend_urls``.
      warnings.warn("'override_urls' is a deprecated method & will be removed by v1.0.0. Please rename your method to ``prepend_urls``.")

    To access KA Lite from another connected computer, try the following address(es):
        http://192.168.2.123:8008/
        http://10.8.0.1:8008/
    To access KA Lite from this machine, try the following address:
        http://127.0.0.1:8008/
    Going to daemon mode, logging to /home/jamalex/.kalite/server.log
    Exception in thread cronserver_blocking (most likely raised during interpreter shutdown):

Looking at the log file shows it didn't gracefully handle the fact that the desired port was taken, to inform the user of that fact:

    jamalex@ace:~$ cat /home/jamalex/.kalite/server.log
    [02/Jul/2015:17:27:33] ENGINE Listening for SIGHUP.
    INFO:cherrypy.error:[02/Jul/2015:17:27:33] ENGINE Listening for SIGHUP.
    [02/Jul/2015:17:27:33] ENGINE Listening for SIGTERM.
    INFO:cherrypy.error:[02/Jul/2015:17:27:33] ENGINE Listening for SIGTERM.
    [02/Jul/2015:17:27:33] ENGINE Listening for SIGUSR1.
[...]

PR #4014, opened 2015-07-03
Add port check, improve failure messages, default use of --traceback …

benjaoming:
…option, cronserver_blocking thread dies gracefully

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -59,6 +59,7 @@ from __future__ import print_function
 # DO NOT IMPORT BEFORE THIS LIKE
 import sys
 import os
+import socket
 
 # KALITE_DIR set, so probably called from bin/kalite
 if 'KALITE_DIR' in os.environ:
@@ -343,15 +344,11 @@ def get_pid():
 
 class ManageThread(Thread):
 
-    """
-    Runs a command in the background
-    """
-    daemon = True
-
     def __init__(self, command, *args, **kwargs):
         self.command = command
         self.args = kwargs.pop('args', [])
-        return super(ManageThread, self).__init__(*args, **kwargs)
+        super(ManageThread, self).__init__(*args, **kwargs)
+        self.daemon = False  # Main process does NOT exit until thread dies
 
     def run(self):
         utility = ManagementUtility([os.path.basename(sys.argv[0]), self.command] + self.args)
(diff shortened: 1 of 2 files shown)


kalitectl.py
```diff
@@ -471,7 +471,15 @@ def start(debug=False, daemonize=True, args=[], skip_job_scheduler=False, port=N
...
-    
+
+    # Start the job scheduler (not Celery yet...)
+    if not skip_job_scheduler:
+        manage(
+            'cronserver_blocking',
+            args=[],
+            as_thread=True
+        )
+
```
benjaoming: Was moved because of some vague idea that starting this thread should happen after become_daemon.. not sure if it really changes anything.

jamalex:
I've been getting the following a lot. Would this change that?

    Going to daemon mode, logging to /home/jamalex/.kalite/server.log
    Exception in thread cronserver_blocking (most likely raised during interpreter shutdown):
      File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
      File "/usr/share/kalite/kalitectl.py", line 361, in run
      File "/usr/share/kalite/python-packages/django/core/management/__init__.py", line 392, in execute
      File "/usr/share/kalite/python-packages/django/core/management/base.py", line 227, in run_from_argv

benjaoming:
@jamalex 
My theory is that the threads were killed with the main process before. This makes sense as you see the exception says `(...) Exception in thread cronserver_blocking (most likely raised during interpreter shutdown):`.

jamalex:
I've been testing a bit, and this does seem to help avoid "cronserver_blocking" errors and "not responding, killing with force" messages.

However, it seems it may have introduced a new problem, with `kalite restart`. Probably it was already a race condition, but the order of operations meant the port was generally freed by the time it tried to claim it again. Now, I pretty reliably get:

    (ka-lite)jamalex@hummingbird:~/fle/ka-lite$ bin/kalite restart
    kalite stopped
    Running 'kalite start' as daemon (system service)

    Stand by while the server loads its data...

    Port 8008 is occupied. Please close the process that is using it.(ka-lite)jamalex@hummingbird:~/fle/ka-lite$

benjaoming:
This is "funny" :) Seems to happen after I added a speedier shutdown in  09c986b

Nice testing btw! And thanks for fixing.

benjaoming:
Test failing because of usual timeout, merged this, fixes #4004

PR #4016, opened 2015-07-03
DB Backup for KA-Lite

SG345:
Initial Commit for Backup Application. Closed the previous PRs because of merge conflicts.


backup/__init__.py
```diff
@@ -0,0 +1,15 @@
+import dbbackup
+
+from django.core.management import call_command
+
+from fle_utils.chronograph.models import Job
+
+def setup_backup():
+	try:
```
aronasorman: Please format your code to use 4 spaces for indentation, as per the python style guide: https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

backup/tests.py
```diff
@@ -0,0 +1,16 @@
+"""
```
aronasorman: You can delete this file.

backup/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+		command = call_command('dbbackup')
+		schedule_job(command, frequency="DAILY")
+	except:
```
aronasorman: This will catch **all** errors, including Normal system exit (`sys.exit`) and system interrupts (`Ctrl+C`) :) Please make this catch a more specific error:

`except KeyError` should do it.

aronasorman:
Also, please make more descriptive commit messages. It's hard to discern what each one does when all of them have the same message.

benjaoming:
Hi there!

I would like to see this merged as a non-app. I disagree that it's a django application. It's a management command perhaps or something else. But you should try finding a more natural location for the code.

We are working to structure the application much better, so everything looks like a big mess of applications. We have to admit that. But I take it that we don't want to add to the confusion :)

Ben

benjaoming:
And great that you got it working! Kudos!

aronasorman:
@benjaoming the eventual plan for the backup subcomponent is to have a frontend for it as well, which will come in later. Making it as an app will prepare for the time when it's more feature complete.

benjaoming:
@aronasorman but we have 26 folders in kalite and this is not an app at present... so I kind of vote that we see it as a feature-complete automatic backup scheduler.

If UI is added later for restoring etc. we can also add that in `control_panel`.

aronasorman:
Hmm OK. In its current state I agree that it doesn't deserve to be an app when it's just a small snippet for now. This also won't be merged right now, but simply to do some early code review.

I don't have a good idea yet on how KA Lite will be structured moving forward (is that pending a refactor?), but this code should be self-contained enough to make it easy to move between folders. So perhaps once the new application layout has been committed to develop we can migrate to that here in this PR as well.

benjaoming:
Well, I'm all in favor of being open to new features and apps and not blocking things for the sake of a yet-to-be-planned refactor. I don't exactly know what application layout and refactoring is supposed to happen.

It's just damage control that we're strict on not creating new apps. Once they have models and table prefixes and migrations, it becomes a mess moving them.

aronasorman:
Sure I hear you. This feature will be implemented by 0.15 :+1:

backup/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    backup_filepath = os.path.join(backup_dirpath, restoref) #create file path
+    call_command('dbrestore', filepath=backup_filepath, database='default') #perform restore
+
+def rotate_backups():
+    now = time.time()
+    for fp in os.listdir(backup_dirpath):
+        if os.stat(os.path.join(backup_dirpath, fp)).st_mtime < now - 7 * 86400:
+            os.remove(os.path.join(backup_dirpath, fp))
+
+setup_backup()
```
aronasorman: Remember, this module can be imported multiple times, and this line will be executed multiple times too. Can you look into how we can guarantee that we schedule a backup only once?

backup/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+def restore_list():
+    file_list = os.listdir(backup_dirpath) #Retrieve the filenames present in the backup_dirpath
+    filelist = []
+    filenumber = 0
+
+    for eachfile in file_list:
+        filelist.append(eachfile) #Store files one by one
+        print filelist.index(eachfile), eachfile #show file number: index
+        filenumber += 1
```
aronasorman: This can be written in a more pythonic way, using `enumerate`:

```python
for i, f in enumerate(file_list):
    print i, f
```

aronasorman: Also, I prefer that we remove all input and output out of this `__init__.py` file and into the `restorebackup` management command. Instead `restore_list` should just take in the name of the backup and then restore that backup. The logic for printing out the list of backups and then asking for user input should be confined to the `restorebackup` management command.

backup/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+    print "Please enter file-number to restore:\n"
+
+    backup_option = int(raw_input())
+    restoref = filelist[backup_option] #file to be restored based on x
```
aronasorman: Will raise an error if the user inputs a number greater than the backup options. Need to catch that case!

backup/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    backup_filepath = os.path.join(backup_dirpath, restoref) #create file path
+    call_command('dbrestore', filepath=backup_filepath, database='default') #perform restore
+
+def rotate_backups():
+    now = time.time()
+    for fp in os.listdir(backup_dirpath):
+        if os.stat(os.path.join(backup_dirpath, fp)).st_mtime < now - 7 * 86400:
```
aronasorman: Please put these numerical constants (7 and 86400) into global constants, so other readers won't have to guess what they mean.

backup/__init__.py
```diff
@@ -0,0 +1,53 @@
+import dbbackup
```
benjaoming: Never put anything in __init__.py files unless it's an import that makes something available to users importing your module. In case of django apps, just don't put anything there.

http://axialcorps.com/2013/08/29/5-simple-rules-for-building-great-python-packages/

backup/__init__.py
SG345: I think we could add a conditional test within setup_backup() that checks if there's a recent backup file made in less than 24 hours.  We still might need setup_backup() inside __init__ as we need to ensure that it runs the first time kalite is launched. Or if there's an alternative way, we could do that too.

kalite/distributed/backup.py
```diff
@@ -0,0 +1,34 @@
...
+import datetime
+import os
+import sys
+import time
+from os.path import expanduser
+from django.core.management import call_command
+
+from fle_utils.chronograph.models import Job
+
+backup_dirpath = expanduser("~")
```
aronasorman: If these are constants, please use all-caps.

kalite/distributed/backup.py
```diff
@@ -0,0 +1,34 @@
...
+backup_dirpath = os.path.join(backup_dirpath, 'ka-lite-backups')
+
+def file_rename():
+    filename ='default.backup'
+    time = datetime.datetime.now().ctime()
+    path = os.path.join(backup_dirpath, filename)
+    target = os.path.join(backup_dirpath, time + '.backup')
+
+    if path.exists():
```
aronasorman: Oh, this works? I thought it should be `os.path.exists(path)`.

kalite/distributed/backup.py
```diff
@@ -0,0 +1,34 @@
...
+    try:
+        command = call_command('dbbackup')
+        force_job(command, frequency="DAILY")
+    except KeyError:
+        pass
+
+def rotate_backups():
+    now = time.time()
+    for fp in os.listdir(backup_dirpath):
+        if os.stat(os.path.join(backup_dirpath, fp)).st_mtime < now - 7 * 86400:
```
aronasorman: Can you make this if condition clearer by moving these inner calls to `os.path` and the constants into their own variables?

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,25 @@
...
+class Command(NoArgsCommand):
+    help = "Restore backup to a previous state"
+    def handle_noargs(self, **options):
+        file_list = os.listdir(backup_dirpath) #Retrieve the filenames present in the backup_dirpath
+        filelist = []
+        filenumber = 0
+
+        for i, f in enumerate(file_list):
+            print i, f
+            filenumber += 1
```
aronasorman: Rather than manually counting the files here by incrementing, just use `filenumber = len(file_list)` :)

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,25 @@
...
+
+        print "Please enter file-number to restore:\n"
+
+        backup_option = int(raw_input())
+        if backup_option>=0 and backup_option <=filenumber:
```
aronasorman: As per `the zen of python`, prefer catching the index error exception vs. checking that the index is correct. So something like:

```python
try:
   call_command(...)
except IndexError: # Not sure what the specific error is
   logging.error("Number option out of bounds.")
   raise
```

aronasorman:
Some cleanup needed, but otherwise this looks functional.

kalite/distributed/backup.py
benjaoming:
``` 
>>> os.path.join("a", "b").exists()
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'exists'
```

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,42 @@
+from os.path import join, expanduser
+import os
+from django.core.management.base import BaseCommand
+from django.core.management import call_command
+from optparse import make_option
+BACKUP_DIRPATH = expanduser("~")
+BACKUP_DIRPATH = os.path.join(BACKUP_DIRPATH, 'ka-lite-backups')
```
aronasorman: 2 spaces between imports and vars.

Also, please put this option in `kalite/settings.base.py`, and combine these into one file:
```python
BACKUP_DIRPATH = os.path.join(expanduser(~), 'ka-lite-backups')
```

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,42 @@
...
+
+backup_filepath=""
```
aronasorman: Remove

aronasorman: To clarify -- we want to avoid global variables. Only use them if they're a constant.

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,42 @@
...
+    option_list = BaseCommand.option_list + (
+        make_option(
+            '--file',
+            dest='backup_filepath',
+            default="",
+            help="File path"),
+        )
+    
+    def handle(self, *args, **options):
+            if(options['backup_filepath']):
```
aronasorman: Assign `backup_filepath` somewhere here.

kalite/settings/base.py
```diff
@@ -322,6 +323,10 @@
 
 ROOT_URLCONF = "kalite.distributed.urls"
 
+backup_dirpath = expanduser("~")
+backup_dirpath = os.path.join(backup_dirpath, 'ka-lite-backups')
+DBBACKUP_BACKUP_DIRECTORY = backup_dirpath
```
aronasorman: Oh, you keep redefining BACKUP_DIRPATH. Why not just use the one in settings? You can then use the value using:

```python

settings.BACKUP_DIRPATH
```

aronasorman:
Made an initial review. Comment here once you've addressed them!

SG345:
Comments have been addressed. Please review!

kalite/distributed/backup.py
```diff
@@ -0,0 +1,36 @@
+import dbbackup
+import datetime
+import os
+import sys
+import time
+from os.path import expanduser
```
benjaoming: you don't have to import these, you can actually just `os.path.join` etc after importing os.

kalite/settings/base.py
```diff
@@ -322,6 +323,9 @@
 
 ROOT_URLCONF = "kalite.distributed.urls"
 
+BACKUP_DIRPATH = os.path.join(expanduser("~"), 'ka-lite-backups')
```
benjaoming: I don't think it should be placed there by default, rather in the configured home directory, stored in `_data_path`.

kalite/distributed/backup.py
SG345: @benjaoming We'll be needing only os.path.expanduser and os.path.join. Should we still import "os" completely or should I replace import.os with two imports: expanduser/path.join?

SG345: @benjaoming : Yes, you're right. Multiple os methods are being used, so importing "os" once is sufficient. Thanks! I'll fix that.

benjaoming: Nah, you should be fine with just `import os`, AFAIK it's the common solution :)

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,41 @@
...
+                if(os.path.isfile(backup_filepath)):
```
aronasorman: Does this work? `backup_filepath` doesn't seem to be assigned anywhere within this file.

kalite/settings/base.py
aronasorman: I like the current behaviour better. Some users might mindlessly delete the .kalite directory. Having a visible folder for backups makes it clear that 1. backups are being done and 2. it's stored in here.

benjaoming: true, I concede... it might not make sense to encourage backing up in the source dir of some of the files that are backed up.

In the end, we should encourage/require people to manually set the backup path.

kalite/distributed/backup.py
```diff
@@ -0,0 +1,34 @@
...
+    defaultfilename ='default.backup'
+    time = datetime.datetime.now().ctime()
+    time = time + '.backup'
+    defaultpath = os.path.join(settings.BACKUP_DIRPATH, defaultfilename)
+    target = os.path.join(settings.BACKUP_DIRPATH, time)
+    if(os.path.exists(defaultpath)):
+        os.rename(defaultpath, target)   
+
+def setup_backup():
+    command = call_command('kalitebackup')
```
aronasorman: I don't think this will work :) Python does eager evaluation, meaning `call_command('kalitebackup')` already starts executing by this line. By the time you get to the `force_job` line, the backup already happened.

Instead, what `force_job` is expecting is simply the name of the command, and then it'll use `call_command` automatically when it's time to run it. Call it like this:

`force_job("kalitebackup", frequency="DAILY")`

kalite/distributed/management/commands/restorebackup.py
```diff
@@ -0,0 +1,44 @@
...
+                else:
+                    print 'File not found'
+            else:
+                file_list = os.listdir(settings.BACKUP_DIRPATH) #Retrieve the filenames present in the BACKUP_DIRPATH
+                filelist = []
+                filenumber = 0
+                for i, f in enumerate(file_list):
+                    print i, f.replace(".backup", "")
+                    filelist.append(f)
+                if(len(filelist) > 0):
```
aronasorman: First off:

Unlike the other *uncool* languages, python does not need parens in its if statements, unless you really need it to layout the order of evaluation. So this statement becomes:

`if len(filelist) > 0:`

Python also treats any value that's not an empty list, empty dict, empty set, 0 or `False` as a `True` value. So then your if statement can be restated to:

`if len(filelist):`

However, since a nonempty list is treated as `True`, you can reduce this to:

`if filelist`:

Much simpler and pythonic! Can you go over your if statements in this PR and see if you can simplify them?

Issue #4024, opened 2015-07-04
The --git-migrate setup option is not working


jamalex:
Installed: ka-lite-source - 0.14.alpha2-0ubuntu1

Tried to follow the [release note instructions](http://field.learningequality.org/develop/docs/installguide/release_notes.html#id1) for importing stuff from an old Git-based installation. It didn't seem to do anything (it just created a new database from scratch as usual, and didn't give any indication of importing anything from the old install).

    jamalex@hummingbird:~$ kalite manage setup --git-migrate /home/jamalex/fle/ka-lite
                                      
      _   __  ___    _     _ _        
     | | / / / _ \  | |   (_) |       
     | |/ / / /_\ \ | |    _| |_ ___  
     |    \ |  _  | | |   | | __/ _ \ 
     | |\  \| | | | | |___| | ||  __/ 
     \_| \_/\_| |_/ \_____/_|\__\___| 
                                      
    http://kalite.learningequality.org
                                      
             version 0.14.alpha2
                                      
    WARNING:root:It's recommended that you install Python version 2.7.9. Your version is: sys.version_info(major=2, minor=7, micro=6, releaselevel='final', serial=0)

    --------------------------------------------------------------------------------
    This script will configure the database and prepare it for use.
    --------------------------------------------------------------------------------
    Press [enter] to continue...
    Please choose a username and password for the admin account on this device.
      You must remember this login information, as you will need
      to enter it to administer this installation of KA Lite.
    Username (leave blank to use 'jamalex'): 
    ...

(Removing UX note from here, as that was meant for #4025)

jamalex:
Note: I'm in the process of rewriting `gitmigrate` -- so nobody take this one just now, please.

jamalex:
Fixed in #4026.

benjaoming:
@jamalex I moved the issue of `--option value` to #4039, thinking that this was indeed the problem you were experiencing as I could reliably see that options with a dash '-' were not blocked.

benjaoming:
Suggesting to close as a blocker for 0.14?

PR #4026, opened 2015-07-05
Updated gitmigrate command to a) work, b) be friendlier. Also, fixed docopt option parsing.

### LINKED ISSUE #4025: The gitmigrate management command isn't working
After failing with #4024, I tried running `gitmigrate` directly, but it failed as well (see below). I'm already working on trying some things for a fix.

    jamalex@hummingbird:~$ kalite manage gitmigrate
    CommandError: Must specify a path to migrate the Git installed version of KA Lite from

    jamalex@hummingbird:~$ kalite manage gitmigrate /home/jamalex/fle/ka-lite
    -------------------------------------------------------------------
    Attempting to copy Database file from previous installation
    -------------------------------------------------------------------

    IOError: [Errno 2] No such file or directory: '/usr/share/kalite/kalite/database/data.sqlite'

Also, on a UX note, it's hard to know what arguments to use, since it doesn't tell you about the `args[0]` option in the help, and the `--path` option is well-hidden:

    jamalex@hummingbird:~$ kalite manage help gitmigrate
    Usage: kalite manage gitmigrate [options] 

    Options:
      -v VERBOSITY, --verbosity=VERBOSITY
                            Verbosity level; 0=minimal output, 1=normal output,
                            2=verbose output, 3=very verbose output
      --settings=SETTINGS   The Python path to a settings module, e.g.
                            "myproject.settings.main". If this isn't provided, the
                            DJANGO_SETTINGS_MODULE environment variable will be
                            used.
      --pythonpath=PYTHONPATH
                            A directory to add to the Python path, e.g.
                            "/home/djangoprojects/myproject".
      --traceback           Print traceback on exception
      --auto-pdb            Break into PDB on an uncaught exception
      --path=PATH           Git Install Path
      --version             show program's version number and exit
      -h, --help            show this help message and exit

jamalex:
The command was trying to put stuff in `KALITE_DIR` instead of `DEFAULT_DATABASE_PATH` and `CONTENT_ROOT`, which prevented it from working at all.

The improvements here also present more options to the user, allowing them to select an alternate location to load videos from, see how big the files there are, and choose whether to copy or move. More granular progress is then displayed during the transfer.

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -663,12 +663,17 @@ def docopt(doc, argv=None, help=True, version=None, options_first=False):  # @Re
     
     # if matched and left == []:  # better error message if left?
     if collected:  # better error message if left?
+        
         result = Dict((a.name, a.value) for a in (pattern.flat() + collected))
-        collected_django_options = len(result.get('DJANGO_OPTIONS', []))
-        result['DJANGO_OPTIONS'] = (
-            result.get('DJANGO_OPTIONS', []) +
-            sys.argv[len(collected) + (collected_django_options or 1):]
-        )
+        
+        # count the number of arguments that were not Django options, and then properly read in Django
+        # options from sys.argv (we can't get them from "DJANGO_OPTIONS", since that doesn't catch 
+        # long-form options like "--git-migrate")
+        kalite_command_arg_count = len(collected)
+        if "DJANGO_OPTIONS" in result:
+            kalite_command_arg_count -= 1
+        result['DJANGO_OPTIONS'] = sys.argv[kalite_command_arg_count+1:]
+        
         # If any of the collected arguments are also in the DJANGO_OPTIONS,
         # then exit because we don't want users to have put options for kalite
         # at the end of the command
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -669,9 +669,7 @@ def docopt(doc, argv=None, help=True, version=None, options_first=False):  # @Re
         # count the number of arguments that were not Django options, and then properly read in Django
         # options from sys.argv (we can't get them from "DJANGO_OPTIONS", since that doesn't catch 
         # long-form options like "--git-migrate")
-        kalite_command_arg_count = len(collected)
-        if "DJANGO_OPTIONS" in result:
-            kalite_command_arg_count -= 1
+        kalite_command_arg_count = len([x for x in collected if x.name!="DJANGO_OPTIONS"])
         result['DJANGO_OPTIONS'] = sys.argv[kalite_command_arg_count+1:]
         
         # If any of the collected arguments are also in the DJANGO_OPTIONS,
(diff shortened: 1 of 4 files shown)


kalitectl.py
```diff
@@ -663,12 +663,17 @@ def docopt(doc, argv=None, help=True, version=None, options_first=False):  # @Re
...
-        )
+        
+        # count the number of arguments that were not Django options, and then properly read in Django
+        # options from sys.argv (we can't get them from "DJANGO_OPTIONS", since that doesn't catch 
+        # long-form options like "--git-migrate")
+        kalite_command_arg_count = len(collected)
+        if "DJANGO_OPTIONS" in result:
+            kalite_command_arg_count -= 1
+        result['DJANGO_OPTIONS'] = sys.argv[kalite_command_arg_count+1:]
+
```
jamalex: @benjaoming The old code here didn't work when mixing argument types for Django options (long-form-args like "--git-migrate" were not caught in "collected", but left in "__left", and the indexing counts were off).

Let me know if you see any issues with the way I've structured it now. It just ignores any stuff matched by `DJANGO_OPTIONS` and then pulls all the Django options in from `sys.argv`.

jamalex:
Also added in a fix for https://github.com/learningequality/ka-lite/issues/4024 (which includes a general fix to docopt parsing of Django options).

jamalex:
If anyone knows whether it's possible to ensure the server isn't running (or to stop it) from within the management command, that might make this process safer. In fact, it might be good for us to be doing that with the `setup` command in general?

jamalex:
Some failing tests; I need to fix the docopts stuff further still; doing so now.

jamalex:
Remaining errors look to be unrelated (timeouts in a couple behave tests), so this should be good to go (following review).

benjaoming:
@jamalex the fix provided fixes a one-off error, but you will have other problems if you add more "--long-option value" arguments. Intentionally, you should have triggered this (see beginning of kalitectl.py)

```python
# We do not understand --option value, only --option=value.
# Match all patterns of "--option value" and fail if they exist
__validate_cmd_options = re.compile(r"--[^\s-]+\s+[^-\s][^-\s]+")
if __validate_cmd_options.search(" ".join(sys.argv[1:])):
    sys.stderr.write("Please only use --option=value patterns. The option parser gets confused if you do otherwise.\n")
    sys.exit(1)
```

benjaoming:
I will see if I can find a solution for this and get back.

benjaoming:
It was the dash in "--git-migrate" that made it to not fail :/

benjaoming:
Opened up #4038 instead

jamalex:
> It was the dash in "--git-migrate" that made it to not fail :/

Ah, I tried renaming the argument to `gitmigrate` and I see the parse error now.

In the release notes, we currently don't tell people to use `=`. I'll fix that too.
http://field.learningequality.org/develop/docs/installguide/release_notes.html#id1

jamalex:
Doing the full docopt reversion in this PR, and fixing the docs here.

rtibbles:
As I am technically assigned to this, it looks reasonable to me. If @benjaoming thinks the option parsing is right, then I would say we're good to merge.

jamalex:
Option parsing has been reverted, so it's not included. Merging.

Issue #4003, opened 2015-07-03
[Errno 13] Permission denied: '/usr/lib/python2.7/dist-packages/kalite/i18n/static/js/i18n/en.js'


radinamatic:
**Branch**: 0.14 on Ubuntu

**Current Behavior**: Error while installing trying to update English language pack:

`Error during update: Error: [Errno 13] Permission denied: '/usr/lib/python2.7/dist-packages/kalite/i18n/static/js/i18n/en.js'`

![update languages ka lite](https://cloud.githubusercontent.com/assets/1457929/8490587/b88f3c34-212b-11e5-9bd2-2c2b5bf734ec.png)

**Steps to reproduce**:

1. Login as Admin
2. Try to update English language pack

radinamatic:
The same error appeared when I tried to install the Spanish language pack. Seems as if the packs are downloaded, though...

benjaoming:
Grrr, that used to work on the .deb

radinamatic:
Sorry, @benjaoming  I have a new one for you after the latest update to alpha2:

`[Errno 13] Permission denied: '/usr/lib/python2.7/dist-packages/kalite/i18n/data/dubbed_video_mappings.json'`  


![update languages ka lite2](https://cloud.githubusercontent.com/assets/1457929/8502255/e04c5838-21ad-11e5-918a-d8c071ef3f8b.png)

Same as yesterday, the update itself seems to finish fine (see the 'Up to Date' status), it just throws these errors at the end.  

Do you think it's the same problem, or should I open a separate issue?

benjaoming:
this is definitely an error, I will look at it, no need to test it further before there's a patch... thanks!

benjaoming:
There are two different issues at play here, but they both revolve around that the language packs are  mis-designed, something that was missed initially when moving the language data to user directories.. we only created ~/.kalite/locale

We need two new user-writable locations:

1. A general user-writable static_dir that will be picked up by `collectstatic`. This will address the write-error on `/usr/lib/python2.7/dist-packages/kalite/i18n/static/js/i18n/en.js`.
1. We have start copying out the json sources that are modified when new language packs are added and read them from a user-location. This will address the latter write error on dubbed_video_mappings.json

This will also prevent future updates from over-writing user-downloaded language packs.

benjaoming:
@radinamatic this is fixed now and released in alpha7

[...]

PR #4028, opened 2015-07-05
0.14.alpha3

benjaoming:

 * [x] ka-lite on pypi
 * [x] ka-lite-static on pypi
 * [x] ka-lite on Launchpad PPA
 * [ ] OSX installer
 * [ ] Windows installer

**WARNING** Merging this, merges and closes all PRs contained in this PR, and they may have a separate review process and amendments following. So don't merge this PR!! :)

## Known issues

 * languagepacks do not fully work in distributed environments #4003
 * syncing to central server is broken #4032


kalite/version.py
```diff
@@ -7,7 +7,7 @@
 # Must also be of the form N.N.N for internal use, where N is a non-negative integer
 MAJOR_VERSION = "0"
 MINOR_VERSION = "14"
-PATCH_VERSION = "0"
+PATCH_VERSION = "alpha3"
```
MCGallaspy: This will get automatically "cleaned up" by setuptools, which I discovered when running the command `setup.py sdist --static`. "0.14.alpha3" gets turned into "0.14a3" (or something like that). Basically you can get into trouble if you expect the version to look like the former but it looks like the latter.

jamalex: This will also likely break syncing and other version-parsing dependent features.

PR #4036, opened 2015-07-06
No failure on "--long-option value" when dash in long option

benjaoming:
command line failure on --option=value pairs **without** = (e.g. "--long-option value") should understand --long-option containing a dash

#4026


benjaoming:
The whole point of this was to ship a short-term solution that would be acceptable for 0.14 -- but short-term solutions have a way of proving why they can be considered short-term..... within a short term!!!!!

benjaoming:
Merging, usual timeout problems in tests.

PR #4038, opened 2015-07-06
Fix management command gitmigrate and setup --git-migrate=/old_install

benjaoming:
#4024
#4026

@jamalex I reverted the part with the docopt stuff, I don't know exactly what is ups and downs here, but I suspect that the old method was working, not saying that the change doesn't work just that I don't understand the implications atm!! Will open up a new issue addressing this.


jamalex:
Looks like you reverted part of the docopt changes, 1393b2, but not 2bdc9f8, so this isn't going to work.

benjaoming:
@jamalex thanks, well timed!! was about to release alpha3 when I noticed :/

jamalex:
Made the full reversions and some docs changes in the original PR; closing this one.

Issue #3995, opened 2015-07-02
"Create Super User"???


aleeshanderson:
Branch: 0.14
Date: 7/2/15
Build Affected: OSX
Expected behavior: not quite sure
Actual behavior: user is prompted to "Create Super User" The pop-up does not allow you to access the site or interact with it in anyway.

Steps to reproduce: Open KA Lite in Browser(chrome) after a fresh install

![screen shot 2015-07-02 at 3 59 06 pm](https://cloud.githubusercontent.com/assets/12928107/8489713/968416f8-20d3-11e5-9702-a065d0cdc8b0.png)
![screen shot 2015-07-02 at 3 59 21 pm](https://cloud.githubusercontent.com/assets/12928107/8489717/9d3264c8-20d3-11e5-80e3-e1e5a81d6f6c.png)

aleeshanderson:
* I closed KA-Lite and re opened it again, and it seems to be working properly the second time.

66eli77:
@aleeshanderson  great finding, so this issue happens on macbook too. confirmed that it's a cross platform issue

radinamatic:
Confirmed on Ubuntu too, correct modal appears after re-loading the page.

aronasorman:
Quite urgent. We don't want the user to be blocked from interacting with the site.

66eli77:
@aronasorman I'm working on it

66eli77:
here is how to reliably reproduce this issue:
1. start kalite server
2. on Chrome, open Settings, click "show advanced settings..."
3. under "Privacy", click "Clear browsing data..."
4. make sure "Cookies and other site and plug-in data" is checked
5. click "Clear browsing data..."
6. refresh kalite homepage, you will see this issue.

Issue #3973, opened 2015-07-02
KA-Lite superuser access creation is being denied to the installers 


EdDixon:
In all tests on every desktop OS supported (All  WIndows version, MAC, and Linux) after installation the user is taken to a "modal" super user creation page which is intended to be a fall back from the proper landing page which would be our typical login page. Having researched this extensively with the team I can really offer few clues. Screen Shot of the modal superuser page from the most recent WIndows 8 test but again, all desktop OS's affected by this same issue which is why I am issuing it against the ka-lite repo rather than installers it is common to all. Pinging @rtibbles @benjaoming HELP!
![2015-07-02_13-34-23](https://cloud.githubusercontent.com/assets/4349806/8486762/62e4e192-20c4-11e5-9773-db25cc744695.png)

benjaoming:
I would like to draw attention to the fact that the landing page causes not only one view function to be called, but also several API views.

This in case, causes a number of modules to be initialized.

Cherrypy uses threads.

See where I'm going? :)

rtibbles:
The issue as described does not suffer this problem, @benjaoming. @EdDixon can you clarify what the issue is here?

Either:
a) We should be creating the super user during setup rather than relying on the modal.
b) The issue we actually discussed in HipChat, whereby the first time the modal opens, and the content does not load, which *would* be explained by @benjaoming's explanation.

benjaoming:
a) doesn't work, we have to provide a browser-based superuser creation. This is already integrated for Raspberry Pi and Debian UIs where `kalite manage setup` is not meant to be mandatory for the future.

benjaoming:
@rtibbles what is the issue described? That KA Lite prompts for a superuser when there isn't one? Because that's expected behavior.

I thought we were continuing on the talks from Hipchat about the modal itself being broken?

rtibbles:
That's what I thought too, @benjaoming which is why I was seeking clarification, because both the title and the screenshot do not indicate this.

jamalex:
I agree that this is standard/expected behavior now, in general, so I think this may be more of a UX issue than anything else.

Perhaps we could add some "welcome" text, along the lines of "Thanks for installing KA Lite! Your first step now is to create an administrator account -- you'll use this later for creating coach accounts or administering the server."

One question here, though: does the Windows installer still ask the user to provide a username and password during the wizard process? Because if it's doing that, and then it still pops up this modal, that sounds like a bug.

rkurata:
[...]

PR #4041, opened 2015-07-06
fix superuser modal 403, which is caused by CSRF related problem

66eli77:
address the issues stated in these #3995   #3973


aronasorman:
@66eli77 can you open another PR with these commits on the 0.14.x branch?

rtibbles:
Looks like this is now live in 0.14.x

PR #4046, opened 2015-07-06
Return code 0 on "kalite stop" when already stopped

benjaoming:
It is problematic that "kalite stop" exits with a non-zero code just because it was already stopped. That should be okay, otherwise other scripts that call "kalite stop" will think it failed to stop!


benjaoming:
Marking as release blocker because installing ka-lite as a .deb will fail if the init script fails to stop kalite... which it will if it's already stopped!

PR #4048, opened 2015-07-06
Use SHORTVERSION for display in docs, instead of PATCH_VERSION

### LINKED ISSUE #4043: Version number in docs is confusing to users
(From call with Team4Tech today)

Users were confused about why it said "KA Lite 2" in the docs, and weren't sure if they were on the right page:

image

It looks like, on the latest develop build of the docs, it's got: "KA Lite dev11 documentation".

Ideally, this should read "KA Lite 0.14 documentation" (or possibly the full "0.14.0", though I think we decided to have the docs just versioned by minor release).

MCGallaspy:
Should fully fix #4043.


benjaoming:
sphinx, why you so weird!!

PR #4051, opened 2015-07-07
warn users on debian if they are running with a non-default user

benjaoming:
Example:

```
✗ bin/kalite asdasd
You are not running kalite as the default user ka-lite. This is recommended unless you want to re-create the database and start storing new videos/exercises etc. for a different user account

To run the command as the default user, run this instead:

    sudo su ka-lite -s /bin/sh -c bin/kalite asdasd

Do you wish to continue? [y/N] 
```

Pressing enter exits, entering y/Y continues. Having something in .kalite makes it skip.


MCGallaspy:
This one looks pretty critical. Can you add milestones to your PRs when you open them?

benjaoming:
Sorry, that was a miss. Is definitely intended for 0.14. It was also included in alpha7.

benjaoming:
Going to make an amendment after a change of /etc/ka-lite layout....

benjaoming:
So this is ready to be merged, any critical views?

jamalex:
I think this might interact with some of the stuff @rtibbles was working on, so perhaps he could take a look here as well?

benjaoming:
So I think this one is definitely a prerequisite for our default debconf flow, @jamalex could you have the final check and merge it? I've tested this already in alpha7 and find that it's "low-risk".

bin/kalite
```diff
@@ -78,6 +85,28 @@ if not os.path.exists(kalite_path):
 # Converts to normal string for Windows users, where it gets turned into a unicode string and errors out in subprocess
 env['KALITE_DIR'] = str(kalite_path)
 
+# TODO(benjaoming): Delegate responsibility to the .deb package
+if os.path.isfile("/etc/ka-lite/username"):
```
benjaoming: This is tested on Windows: It just returns False.

MCGallaspy: Still, you should probably use `os.sep` in path names, and use [os.name](https://docs.python.org/2/library/os.html#os.name) to make OS-conditional behavior.

MCGallaspy: Or [sys.platform](https://docs.python.org/2/library/sys.html#sys.platform).

benjaoming: I don't know if there's any point in assembling a path that's invalid on windows, anyways... I mean: `os.join(os.sep, "etc", ...)` is a bit odd. I think verifying that this worked was all we needed :)

MCGallaspy: My point is that if this is supposed to be skipped on windows, isn't it more explicit to condition on `os.name` or `sys.platform`? If nothing else, I think it's easy and worthwhile to note the intent in a comment (`# The intent is that this fails on Windows`) . After that I'm happy to merge this.

benjaoming:
This has been part of 5 alpha releases now. Fixed @MCGallaspy 's comments.

Merging now.

PR #4057, opened 2015-07-07
0.14.alpha7

benjaoming:

 * [x] ka-lite on pypi
 * [x] ka-lite-static on pypi
 * [x] ka-lite on Launchpad PPA
 * [ ] OSX installer
 * [ ] Windows installer

**WARNING** Merging this, merges and closes all PRs contained in this PR, and they may have a separate review process and amendments following. So don't merge this PR!! :)

## Known issues

 * None ATM!


kalite/version.py
```diff
@@ -7,7 +7,7 @@
 # Must also be of the form N.N.N for internal use, where N is a non-negative integer
 MAJOR_VERSION = "0"
 MINOR_VERSION = "14"
-PATCH_VERSION = "0"
+PATCH_VERSION = "alpha7"
```
MCGallaspy: Don't use version numbers like this. distutils will normalize the version number to something else (with e.g. `python setup.py sdist --static`) which causes problems when other components get the version number from `version.py` and expect something else. Please just user numbers, as the comment above suggests.

benjaoming: I've explained why I'm doing this a lot of times, I hope that's clear to all that it's a temporary thing.

Issue #4058, opened 2015-07-07
languagepackdownload always uses the default parameters


aronasorman:
branch: 0.14.x

I'm trying to download pt-BR through the commandline using this command:

```
kalite manage languagepackdownload -l 'pt-BR' --commandline
```

However, the `lang_code` parameter always uses `settings.LANGUAGE_CODE`. Something's wrong with our parameter parsing.

benjaoming:
I think the problem is that it isn't complaining about your "illegal" space after `-l`.

benjaoming:
I think the problem is that it isn't complaining about your "illegal" space after `-l`.

MCGallaspy:
"Fixes" in #4061... You should use the following command instead:
```
kalite manage languagepackdownload -l'pt-BR' --commandline
```

aronasorman:
Fixed in the current 0.14.

Issue #4039, opened 2015-07-06
Get docopt to understand "kalite manage django_command --option value" with spaces instead of '='


benjaoming:
Problem:

    ✗ bin/kalite manage command --option value
    Please only use --option=value patterns. The option parser gets confused if you do otherwise.

Expected:

Command runs.

Instead of messing around with a parser that detects "--option value" and then fails and even fails to fail (see #4036), we should have a parser that understands "--option value" and not only "--option=value".

jamalex:
I started describing some of these in a new issue, because I couldn't find this one. Putting the notes relevant to this one here instead (also, renamed this one to include keywords I was looking for when searching for the issue):

- Long-form options need to be separate from their values by `=`, not a space (e.g. `--path=/a/b/c` not `--path /a/b/c`)
- Short-form options cannot have spaces between the option letter and the argument (e.g. `-v3` not `-v 3`)

MCGallaspy:
Look, it seems like the original motivation for dropping in our own modified version of `docopt` comes from #3827. Hindsight is 20/20, but I think it's safe to say this has crept beyond the initial motivation of being a quick fix and instead is becoming one more weird thing to maintain. Before we go down this rabbit hole any further, I suggest we revert to using "plain" docopt and live with whatever constraints it imposes.

benjaoming:
@MCGallaspy agreed that it sucks to do get our hands dirty with command line parsing.. that's what docopt was for.. the goal is to make this layer as thin as possible. But I don't have any idea about make the layer thinner without modifying the command line API, which I think should be a no-go.. which makes the docopt constraints a bit hard to live with, unless there is a way that we've overlooked.

MCGallaspy:
> agreed that it sucks to do get our hands dirty with command line parsing

I would say that it more than sucks. Modifying 3rd party libraries is classically forbidden. Now we're maintaining that code in addition to our own. We've already spent a lot of time debugging unrelated issues when command-line argument parsing changed because of our modifications. It's impossible to say with certainty how much more dev time we will spend on that and the fallout from it, so I say we do indeed live with modifying the `kalite` command line API, hard though it may be. It will be a pain now to revert it (but hopefully not too much), but I think it will be substantially more painful down the line not to.

MCGallaspy:
To be clear, I'm suggesting we revert it in 0.15.

benjaoming:
[...]

PR #4061, opened 2015-07-07
Also fail on broken -x 123, should be -x123 for now

benjaoming:
Fix for #4058

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -81,9 +81,9 @@ import cherrypy
 
 # We do not understand --option value, only --option=value.
 # Match all patterns of "--option value" and fail if they exist
-__validate_cmd_options = re.compile(r"--[^\s]+\s+(?:(?!--|-[\w]))")
+__validate_cmd_options = re.compile(r"--?[^\s]+\s+(?:(?!--|-[\w]))")
 if __validate_cmd_options.search(" ".join(sys.argv[1:])):
-    sys.stderr.write("Please only use --option=value patterns. The option parser gets confused if you do otherwise.\n")
+    sys.stderr.write("Please only use --option=value or -x123 patterns. No spaces allowed between option and value. The option parser gets confused if you do otherwise.\n\nWill be fixed for next version 0.15")
     sys.exit(1)
 
 from threading import Thread


MCGallaspy:
I want to settle this option parsing stuff once and for all, so I made a big ol' regex that should match any valid long or short form argument. Here it is:
```re
(-[0-9a-zA-Z]\s?(?P<sh_value>\w+|'[^']*')|--[\w-]+[\s=]{1}(?P<lg_value>\w+|'[^']*'))
```
And at [pythex.org, with examples](http://pythex.org/?regex=%5E(-%5B0-9a-zA-Z%5D%5Cs%3F(%3FP%3Csh_value%3E%5Cw%2B%7C%27%5B%5E%27%5D*%27)%7C--%5B%5Cw-%5D%2B%5B%5Cs%3D%5D%7B1%7D(%3FP%3Clg_value%3E%5Cw%2B%7C%27%5B%5E%27%5D*%27))&test_string=--blah-di-da%3D1correct%0A--blah%202correct%0A--blah-di-da%203correct%0A--blah%3D4correct%0A--another-long-thing%20%275correct%20quotes%20with%20spaces%27%0A--an--even-longer--long-thing%3D%276correct%20quotes%20with%20spaces%20and%20-s%27%0A-x%20123%0A-x123%0A-l%27pt-Br%27%0A-l%20%27pt-br%27%0A-b%3Ddoesntwork&ignorecase=0&multiline=1&dotall=0&verbose=0). Disregard the initial `^` there, that's so it works in multiline mode.

What this matches in plain english:

First, it either tries to match short-form or long-form arguments.

A short-form argument is matched if there is exactly 1 `-` followed by 1 alphanumeric character (`\w`). Then there is exactly one optional whitespace (in particular, = is not a valid character). Then it can match a value, as described below.

A long-form argument is matched if there are exactly two dashes (`--`) followed by 1 or more alphanumeric characters, a dash (`-`), or an underscore (`_`). This is followed by exactly 1 of a whitespace character or `=`, and then a value as described below.

Values are matched the same way for long- or short-form arguments. They can either be 1 or more alphanumeric or `_` (`\w+`) or _anything_ inside of two `'` (aside from `'` itself), including nothing. The `'` are also matched, so if you don't want 'em then you gotta strip 'em. This allows us to pass whitespace characters within arguments. This also shouldn't break any existing uses, as it continues to match them.

Do with this as you will!

MCGallaspy:
Okay, after further discussion it's not a simple matter of dropping in a such a regex... Seems like the best course of action is to merge this in for now. However, we also decided it was better to live with the "lesser evil" of having options out of order than having to maintain a separate version of docopt. It's too much to revert it for 0.14, but let's put a moratorium on further changes and revert it for 0.15.

benjaoming:
@MCGallaspy #4039

benjaoming:
I've opened an upstream PR for docopt to fix some of this.

PR #4067, opened 2015-07-08
Fix for #3990 issue - For Mac OS, Video Downloading.

### LINKED ISSUE #3990: For Mac OS, Video Downloading Issue
After selected videos and clicking on download, it will take really long to process, even if the notification bar shows: "Launched video download process successfully." , it is said to take longer than 10 mins.

djallado:
@aronasorman. This fixes #3990.

I create a warning message to be shown in download video, prompting the user to wait for the video download to start.

### DIFF
--- a/kalite/updates/static/js/updates/update_videos.js
+++ b/kalite/updates/static/js/updates/update_videos.js
@@ -167,6 +167,7 @@ $(function() {
         numVideos = youtube_ids.length;
 
         // Do the request
+        show_message("warning", "Wait for the video download to start.");
         doRequest(window.Urls.start_video_download(), {youtube_ids: youtube_ids})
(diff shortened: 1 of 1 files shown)


kalite/updates/static/js/updates/update_videos.js
```diff
@@ -167,6 +167,7 @@ $(function() {
         numVideos = youtube_ids.length;
 
         // Do the request
+        show_message("warning", "Wait for the video download to start.");
```
aronasorman: Can you have this shown only when the user is using OS X? Also, please add a `gettext` call.

aronasorman:
With the port to using a virtualenv for the mac, this fix might not be needed anymore.

PR #4068, opened 2015-07-08
explicitly recognize the --settings option, and set the

aronasorman:
DJANGO_SETTINGS_MODULE with it.

This will allow 'kalite start' to work with a custom settings module
specified with --settings=<module>.

@benjaoming @jamalex please take a look.

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -24,6 +24,8 @@ Options:
   --debug               Output debug messages (for development)
   --port=<arg>          Use a non-default port on which to start the HTTP server
                         or to query an existing server (stop/status)
+  --settings=<arg>      Specify Django's settings module. Must follow python's
+                        import syntax.
   --skip-job-scheduler  KA Lite runs a so-called "cronograph", it's own built-in
                         automatic job scheduler required for downloading videos
                         and sync'ing with online sources. If you don't need this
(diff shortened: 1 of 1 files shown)


aronasorman:
Tested with `kalite start --settings=kalite.project.settings.dev` and `kalite shell --settings=kalite.project.settings.dev`. Both now have 'dev' settings turned on.

kalitectl.py
```diff
@@ -699,7 +701,10 @@ def docopt(doc, argv=None, help=True, version=None, options_first=False):  # @Re
     # Since positional arguments should always come first, we can safely
     # replace " " with "=" to make options "--xy z" same as "--xy=z".
     arguments = docopt(__doc__, version=str(VERSION), options_first=False)
-    
+
+    settings_module = arguments.pop('--settings', None)
+    os.environ['DJANGO_SETTINGS_MODULE'] = settings_module
```
jamalex: This probably requires that `--settings=X` be specified prior to any actual DJANGO_OPTIONS, due to the ordering constraint currently in place for `kalite` command vs Django options, right? Probably not a big deal, and I imagine it will tell the user if there's an ordering issue, but might be worth testing a few different cases of ordering (especially any we have in the docs) to make sure it handles things nicely.

jamalex:
Very cool! If you try a couple tests based on what I wrote above, and it all looks ok, merge away!

benjaoming:
sent you a PR @aronasorman otherwise I think it's cool to extract the --settings option and make it universal to all commands, that's exactly what was missing

aronasorman:
Merged your PR @benjaoming!

rtibbles:
Assuming this is good to merge.

benjaoming:
:+1:

PR #4071, opened 2015-07-09
Tweaks needed in order to work with central server

jamalex:
This doesn't really affect anything on the distributed server (other than adding `fle_utils.testing` to `INSTALLED_APPS` to support the `runcode` command), but is needed to get the submodule stuff on ka-lite-central working.

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -340,6 +340,7 @@ INSTALLED_APPS = [
     'fle_utils.config',
     'fle_utils.backbone',
     'fle_utils.chronograph',
+    'fle_utils.testing', # needed to get the "runcode" command, which we sometimes tell users to run
     'kalite.django_cherrypy_wsgiserver',
     'kalite.coachreports',
     'kalite.distributed',
--- a/python-packages/fle_utils/django_utils/command.py
+++ b/python-packages/fle_utils/django_utils/command.py
@@ -173,7 +173,14 @@ def call_outside_command_with_output(command, *args, **kwargs):
     cmd = (kalite_bin, "manage", command)
     for arg in args:
         cmd += (arg,)
-    for key, val in kwargs.items():
+
+    kwargs_keys = kwargs.keys()
+    
+    # Ensure --settings occurs first, as otherwise docopt parsing barfs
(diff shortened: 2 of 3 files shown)


jamalex:
@aronasorman This is very low-risk, but needs to get into 0.14, so that it will still support the `runcode` command, which I sometimes need to direct users to run in order to fix something.

kalite/settings/base.py
```diff
@@ -340,6 +340,7 @@
     'fle_utils.config',
     'fle_utils.backbone',
     'fle_utils.chronograph',
+    'fle_utils.testing', # needed to get the "runcode" command, which we sometimes tell users to run
```
jamalex: This is the critical piece needed for 0.14

python-packages/fle_utils/django_utils/command.py
```diff
@@ -169,16 +169,27 @@ def call_outside_command_with_output(command, *args, **kwargs):
...
         key = key.replace(u"_",u"-")
         prefix = u"--" if command != "runcherrypyserver" else u""  # hack, but ... whatever!
         if isinstance(val, bool):
             cmd += (u"%s%s" % (prefix, key),)
         else:
-            cmd += (u"%s%s=%s" % (prefix, key, unicode(val)),)
+            # TODO(jamalex): remove this replacement, after #4066 is fixed:
+            # https://github.com/learningequality/ka-lite/issues/4066
+            cleaned_val = unicode(val).replace(" ", "")
+            cmd += (u"%s%s=%s" % (prefix, key, cleaned_val),)
```
jamalex: This stuff is only used via the central server (the only place that calls `call_outside_command_with_output`).

python-packages/securesync/management/commands/register.py
```diff
@@ -50,7 +50,7 @@ def handle(self, *args, **options):
         assert client.test_connection() == "success", "Unable to connect to the central server!"
 
         # ensure a zone was specified
-        assert "zone" in options, "You must specify a zone."
+        assert options.get("zone", None), "You must specify a zone."
```
jamalex: In practice, this only ever gets called by the ecosystem tests.

Issue #4074, opened 2015-07-09
cannot download second video on RPi


66eli77:
platform: PRi 512RAM Raspbian
kalite: ka-lite_0.14.alpha7-0ubuntu1_all.deb
steps to reproduce:

1. login as admin and go to video download page and download one video
2. once the first video download finished, select another video and hit download video button
3. you will never see the downloading progress bar again and have no way to know if the video is downloaded even after long waiting
4. if you refresh the page, you will find out the second video was not downloaded, and you can repeat the problem again following steps 1, 2, 3.

66eli77:
@rtibbles the problem seems that after the first video download, kalite will re-Preloading exercise data/content data/topic tree for language en. This process takes a long time(15+ mins). If I wait this process to be finished and then choose the second video to download, then everything works fine.

benjaoming:
@rtibbles assuming we could mark content and topic tree data *changed*, would there be a way to load these data structures quicker into memory when initializing the server? It's not just that RPi is slow to start, but all other platforms could be said to have the same symptoms more or less. A slow or busy computer could produce the above error just as well as the RPi can?

rtibbles:
@benjaoming More intelligently invalidating the caches will be the ultimate way to fix this, but I don't want to churn the code that much just before release. I have made a PR that just invalidates the disk based caches for now, meaning that the caches will get rebuilt at next reboot.

rtibbles:
Fixed in #4077

Issue #3782, opened 2015-06-03
RPi warning message - can it be removed?


benjaoming:
The video download page warns users about only downloading one video at a time. But it doesn't say that this is already what happens. So it kind of makes you look for the button that goes "Fetch videos one by one" or even worse would make users tick one video, download it, and then tick the next.

![one_at_a_time](https://cloud.githubusercontent.com/assets/374612/7958984/0912efe6-09f5-11e5-8cef-e5f8d22476ad.png)

My suggestion: Remove the warning, write something in the Docs on bulk downloading with btsync because that's where the problem occurs.

jamalex:
> or even worse would make users tick one video, download it, and then tick the next

I think this is actually what it is recommending, as we've had problems on low-powered devices when many video files are queued at once. One at a time is probably a bit conservative, but more than, say, 100 at a time may cause problems. More testing is needed, and improvements to the way the queuing works are also needed.

benjaoming:
Let's get it sorted out like this: We have a `raspberry_pi` settings module and an automated script for setting up Raspberry Pis (that doesn't look nice in plural), and so we should have a setting that forces linear downloading.

rtibbles:
Downloading *is* already linear.

The issue here is that the current downloading code uses the database to track downloading progress. Trying to create the entries for too many videos at once in an SD card backed SQLite database causes problems, because there is too much write activity at once.

It is not about downloading, so much as queuing for download.

rtibbles:
(But I think the user is best spared that level of semantics)

benjaoming:
Thanks for explaining, @rtibbles 

Maybe we can fix this for 0.15 by having a better queue mechanism and thus make the message redundant.

rtibbles:
I think that is definitely the best way. @jamalex and I had discussed only
holding the queue in memory as an alternative.

Another possibility, which could tie into making the cache invalidation
more efficient, is to complete @MCGallaspy's database backed content
caches, and we could then track downloading there, and then trigger cache
updates there too.

rtibbles:
This issue is no longer extant since #4630

PR #4077, opened 2015-07-10
Don't invalidate caches except at boot when using disk based caches

rtibbles:
To improve performance on the Raspberry Pi, don't invalidate the cache at download finish, just remove the disk based cache to force a refresh of the content cache at startup.

### DIFF
--- a/kalite/caching/__init__.py
+++ b/kalite/caching/__init__.py
@@ -69,6 +69,13 @@ def invalidate_all_caches():
     """
     invalidate_inmemory_caches()
     if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP:
+        
+        # The underlying assumption here is that if generating in memory caches is too onerous a task to conduct
+        # at every system start up, then it is too onerous a task to conduct during server operation.
+        # We defer the regeneration of these caches to next system startup, by deleting the existing disk based
+        # copies of these caches.
+        # This will prompt the caches to be recreated at next system start up, and the disk based copies to be rewritten.
+        
         for filename in glob.glob(os.path.join(settings.CHANNEL_DATA_PATH, "*.cache")):
             os.remove(filename)
     else:
--- a/kalite/caching/__init__.py
+++ b/kalite/caching/__init__.py
@@ -24,6 +24,9 @@ from django.test.client import Client
 from fle_utils.internet.webcache import *
 from kalite import i18n, topic_tools
 
+import os
+import glob
+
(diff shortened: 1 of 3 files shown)


rtibbles:
Note: Fixes #4074

kalite/updates/views.py
```diff
@@ -27,7 +27,9 @@ def update_context(request):
 @render_to("updates/update_videos.html")
 def update_videos(request, max_to_show=4):
     context = update_context(request)
-    messages.warning(request, _('For low-powered devices like the Raspberry Pi, please download videos one at a time.'))
+    messages.warning(request, _('For low-powered devices like the Raspberry Pi, please download less than 25 videos at a time.'))
```
benjaoming: Wondering a bit why this is included, but I previously had this issue opened because the message is very odd anyways, because it's difficult to understand exactly what it means: #3782

rtibbles: It means.. don't download more than 25 videos at once?

benjaoming:
Great!! Let's aim to include this in Alpha8 and have it for a spin.

kalite/updates/views.py
benjaoming: How is it related to this PR and have you read the issue I've opened about the message? It's more about UI.

kalite/caching/__init__.py
```diff
@@ -65,7 +68,11 @@ def invalidate_all_caches():
     call in here.
     """
     invalidate_inmemory_caches()
-    initialize_content_caches(force=True)
+    if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP:
```
MCGallaspy: I don't really understand _why_ this is necessary, can you help me make the connection in plain English? If we `DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP`, then we should remove all these `*.cache` files, but why?

On the other hand, if that setting is false then we should `initialize_content_caches`, which creates in-memory (as opposed to disk-based) caches?

rtibbles: The underlying assumption here is that if generating in memory caches is too onerous a task to conduct at every system start up, then it is too onerous a task to conduct during server operation.

So we defer the regeneration of these caches to next system startup, by deleting the existing disk based copies of these caches.

This will prompt the caches to be recreated at next system start up, and the disk based copies to be rewritten.

MCGallaspy: Okay, can you explain as much in a comment there? It is perhaps due to my unfamiliarity with this code, but I did not make the connection between deleting these files and not reloading the cache.

PR #4085, opened 2015-07-13
master into 0.13.x

MCGallaspy:
Discovered some changes to master that don't exist on 0.13.x.


MCGallaspy:
In particular, docs built from the 0.13.x branch have a strange display name, "KA Lite 2 documentation" (from the PATCH_VERSION), when the SHORTVERSION should be preferred.

benjaoming:
Merging this, the docs version issue is known and the fix worked in other branches.

IMO, we could also move the docs to `docs/` on both branches.

MCGallaspy:
I see no reason to move `sphinx-docs` to `docs`. It's done on later branches, that's fine, but why worry about it on earlier branches if it's not broken?

benjaoming:
It would be to serve 0.13 docs on the same server that serves 0.14 docs.. at least on RTD this would be the case, don't know how the fields server works.

PR #4094, opened 2015-07-14
Fixed i18 wrapping in content rec cards.

radinamatic:
Found these non-wrapped strings while fixing cards for screen readers.


rtibbles:
Unproblematic. Merging.

Issue #4054, opened 2015-07-07
Use app-based settings


benjaoming:
We need to complete the last part of #3757 - details are in this Issue about how to work with project settings vs. app settings and re-introduce the possibility of more convenient "derived" settings where sensible.

This is somewhat necessary to finally get rid of local_settings when we consider a setting like CONTENT_ROOT that could potentially be a very efficient way of changing lots of other settings. Currently, they would all have to be specified manually in `~/.kalite/settings.py` which is ugly.

Issue #4040, opened 2015-07-06
Settings derived from CONTENT_ROOT are inconsistent with assessment zip layout and settings module


benjaoming:
Previously, one would set CONTENT_ROOT and other settings would be derived from it.

We need good copy-paste friendly code whereby the user can have the option to customize settings that previously were affected by CONTENT_ROOT.

This does not affect users still running on local_settings.py

Current docs only have this section:

> CONTENT_ROOT = “<path to desired content folder>” (default=ka-lite/content) This is the path that KA Lite will use to look for KA Lite video files to play. Change the path to another local directory to get video files from that directory. NB! Directory has to be writable for the user running the server in order to download videos.

MCGallaspy:
The only two derived settings I could track down (in `kalite.settings.base`) are:
```python
KHAN_CONTENT_PATH = os.path.join(CONTENT_ROOT, "khan")
ASSESSMENT_ITEM_ROOT = os.path.join(CONTENT_ROOT, 'assessment')
```

Can we put these settings in an installed app? So that if a user creates a custom settings module, instead of setting `CONTENT_ROOT`, `KHAN_CONTENT_PATH`, AND `ASSESSMENT_ITEM_ROOT` separately, they can still just set `CONTENT_ROOT`? Or is this an anti-pattern?

benjaoming:
@MCGallaspy yes, can't think of a better proposal... we've discussed something about assessment items having issues because 3 of the settings diverge from the layout of the current assessment items zip.

To summarize: The settings say that assessment_items.sqlite, assessment_items.versions, and assessment_items.json are located differently from where they are unzipped.

To fix this, we should probably move these specific files so they respect the settings.

Anyways, the point of this is also, that while fixing that issue, and because these settings are all derived from ASSESSMENT_ITEM_ROOT which is derived from CONTENT_ROOT, we should fix all of this together.

I don't mind assigning myself to this, because I'm packing up the .deb with assessment items bundled and I'm relying on this functionality to be fixed, anyways.

MCGallaspy:
Okay, but I'm not sure if you're saying "yes" to this:

> Can we put these settings in an installed app?

or this:

> Or is this an anti-pattern?

If the former, then the documentation doesn't require much updating. If the latter, then we'll need to mention all three settings (though I think it's a bit overkill to have these all be separate settings).

benjaoming:
[...]

PR #4095, opened 2015-07-14
Refactor topic_tools to use app-based settings and remove unused ASSE…

benjaoming:
We found a couple of early blockers before 0.14 is released. This was supposed to be schedule for 0.15, but it's hard to fix without solving parts of #4054.

 * [x] Paths for assessment items are inconsistent with zip layout #4040
 * [x] Delete old assessment items
 * [x] Settings are derived from CONTENT_ROOT but will be inconsistent if CONTENT_ROOT is changed later - move to app based settings layout for those settings #4040
 * [x] Ensure that users are informed about the non-derived assessment items database setting


docs/usermanual/userman_admin.rst
```diff
@@ -662,8 +662,10 @@ Most common settings
 ^^^^^^^^^^^^^^^^^^^^
 
 * DEBUG = <True or False> (default = False) -- Enables debug mode. In case you run into technical issues, enable this setting before troubleshooting / reporting.
-* CONTENT_ROOT = "<path to desired content folder>" (default=ka-lite/content)
+* CONTENT_ROOT = "<path to desired content folder>" (default=~/.kalite/content)
   This is the path that KA Lite will use to look for KA Lite video files to play.  Change the path to another local directory to get video files from that directory. NB! Directory has to be writable for the user running the server in order to download videos.
+* ASSESSMENT_ITEMS_DATABASE_PATH = "<path to desired content folder>" (default=~/.kalite/content/assessmentitems.sqlite)
```
jamalex: "<path to desired content folder>"? Probably better to have this be a locally derived setting, since otherwise things will break. e.g.: 
`* ASSESSMENT_ITEMS_DATABASE_PATH = os.path.join(CONTENT_ROOT, "assessment/khan/assessmentitems.sqlite")`

And what about the .version file? How is it managed?

kalite/settings/base.py
```diff
@@ -274,7 +255,14 @@
 STATIC_ROOT = getattr(local_settings, "STATIC_ROOT", STATIC_ROOT)
 MEDIA_URL = getattr(local_settings, "MEDIA_URL", "/media/")
 STATIC_URL = getattr(local_settings, "STATIC_URL", "/static/")
+
+
 DEFAULT_DATABASE_PATH = getattr(local_settings, "DATABASE_PATH", DEFAULT_DATABASE_PATH)
+
+# This database is located in the content root because then it can be copied
+# together with the other media files located there.
+ASSESSMENT_ITEMS_DATABASE_PATH = os.path.join(CONTENT_ROOT, 'assessmentitems.sqlite')
```
jamalex: I don't understand how this is overridable in project settings, if it's defined here and then used immediately below.

docs/usermanual/userman_admin.rst
benjaoming: The problem is that `ASSESSMENT_ITEMS_DATABASE_PATH` is used in `settings.DATABASES` so it can't be changed by apps unless we start figuring out how that pattern would look.. and I'm not optimistic that it's actually possible. So I left it as a top-level setting.

The other settings for assessment items are non-configurable derivatives of CONTENT_ROOT.

kalite/contentload/management/commands/generate_assessment_zips.py
```diff
@@ -8,17 +8,18 @@
 from multiprocessing.dummy import Pool as ThreadPool
 from threading import Lock
 
-from django.conf import settings
+from django.conf import settings as django_settings
```
MCGallaspy: When should one use `from django.conf import settings` vs `from kalite import settings`? What is the recommended way for devs to access the user's settings module?

kalite/settings/base.py
benjaoming: True that, needs another revisit

kalite/contentload/management/commands/generate_assessment_zips.py
benjaoming: The recommended way is django.conf.settings because it respects DJANGO_SETTINGS_MODULE and it's lazy

docs/usermanual/userman_admin.rst
MCGallaspy: I suggest we remove ASSESSMENT_ITEMS_DATABASE_PATH as a setting, and simply hard-code the path in the DATABASES setting. I don't see a scenario when this should really be user-configurable.

benjaoming:
Going to work on @jamalex 's comment about settings.DATABASES being initialized from a default ASSESSMENT_ITEM_DATABASE_PATH

benjaoming:
The easy solution seems to be that we should not let the ASSESSMENT_ITEM_DATABASE_PATH be configurable for now.

It's an odd case: We want CONTENT_ROOT to be configurable for serving media, yet we need a database connection established on the basis of where CONTENT_ROOT points. I would suggest to leave it out of the docs.

In a longer perspective, things like settings.DATABASES should always be configured through ENVIRONMENT, it's a suitable "high-level" approach for this specific setting and that would also follow from normal conventions.

We *could* mention settings.DATABASES in the docs instead, but it's kind of a heavy chapter and people who want to mess with it are better off reading the source code and the django documentation.

benjaoming:
Another issue regarding this is about the system-wide assessment items.... they might mean that the database is in a different location, forcing this logic to be pushed back the settings.base :(

MCGallaspy:
Okay, so do assessment items count as user data or application data? In one sense they are application data since they are associated with a version, but in another sense they are user data since they _can_ be downloaded after installation and their location on disk is configurable. (Though this is not the case in Windows! where assessment items _are_ bundled with the installer.)

What are the likely cases where the end-user will install ka-lite and download the assessment items separately?

benjaoming:
@MCGallaspy 
Yes, to summarize our current situation: We offer an assessment-free version (understandably since its 500 MB) where users download the items into their USER_DATA_ROOT. But we also offer bundles where they will be available from a system-wide location.

The most likely cases are:

1. For devs, we constantly change ka-lite and the assessment items, so we need an easy way around it
1. Testing: We shouldn't put 500 MB of system data on the test server
1. Users with online access: They will find it easier to install a light-weight KA Lite and then download the assessment items if they need them - some might just want videos?
1. THE FUTURE! Assessment items might be available from separate sources.

docs/usermanual/userman_admin.rst
```diff
@@ -645,14 +645,23 @@ You can also run the ``kalite`` with a completely different Python settings modu
...
-By default, ``~/.kalite/settings.py`` will load ``kalite.project.settings.base`` which are the basic settings. But you can also load Raspberry Pi settings by changing the file to read something like:
+By default, ``~/.kalite/settings.py`` will load ``kalite.project.settings.base`` which are the basic settings. But you can also load Raspberry Pi settings by changing the file to read something like::
 
-  .. ::
-    from kalite.project.settings.raspberry_pi import *
-    # Put your settings here, e.g.
-    # MY_SETTING_VAR = 123
+  from kalite.project.settings.raspberry_pi import *
+  # Put your settings here, e.g.
+  # MY_SETTING_VAR = 123
```
aronasorman: Maybe remove the `#`? Might confuse people and put them in there.

benjaoming: But MY_SETTING_VAR is not a valid setting :) I mean, if the user doesn't know what `#` means in Python, then they are lucky to even be able to open the file in a text editor :)

docs/usermanual/userman_admin.rst
```diff
@@ -661,37 +670,45 @@ Available settings
 Most common settings
 ^^^^^^^^^^^^^^^^^^^^
 
-* DEBUG = <True or False> (default = False) -- Enables debug mode. In case you run into technical issues, enable this setting before troubleshooting / reporting.
-* CONTENT_ROOT = "<path to desired content folder>" (default=ka-lite/content)
-  This is the path that KA Lite will use to look for KA Lite video files to play.  Change the path to another local directory to get video files from that directory. NB! Directory has to be writable for the user running the server in order to download videos.
-* TIME_ZONE = <desired time zone>  (default = "America/Los_Angeles")
+* ``DEBUG = <True or False> (default = False)``
```
aronasorman: Does this setting even have a substantial effect anymore? It doesn't for me. I have to manually specify dev settings using `--settings`. Can we remove it?

docs/usermanual/userman_admin.rst
```diff
@@ -661,37 +670,45 @@ Available settings
...
+  Enables debug mode. In case you run into technical issues, enable this setting before troubleshooting / reporting.
+* ``CONTENT_ROOT = "<path to desired content folder>" (default=~/.kalite/content)``
```
aronasorman: Rather than `~`, can we explicitly label that as `<user's home directory>/`? Normal users might not know what `~` means.

docs/usermanual/userman_admin.rst
benjaoming: It isn't very important, no. It doesn't change much to our LOGGING, but it would enable static serving when using kalite manage runserver.

But putting it under "Most common settings" is maybe a tad too much. Could you address this in #4106 ?

docs/usermanual/userman_admin.rst
```diff
@@ -645,14 +645,23 @@ You can also run the ``kalite`` with a completely different Python settings modu
...
 
+.. NOTE::
+    When changing ``CONTENT_ROOT``, you should also change your ``DATABASES`` **if you
+    have downloaded your own assessment items** and you want to keep the
+    read-only assessment_items database (~50 MB) together
+    with your other media contents (for portability). Example::
+      
+      from kalite.project.settings.base import *
+      CONTENT_ROOT = '/example'
+      DATABASES['assessment_items']['NAME'] = os.path.join(CONTENT_ROOT, 'assessmentitems.sqlite')
```
MCGallaspy: This message has too much technical information, and the _if_ makes it seem optional. Plus I assume most users have no opinion about portability. I prefer this:
```rst
.. warning::
    When changing ``CONTENT_ROOT``, you **must** also change the ``DATABASES['assessment_items']['NAME']`` setting to ``os.path.join(CONTENT_ROOT, 'assessmentitems.sqlite')`` on a following line. See this example::

        CONTENT_ROOT = '/example'
        DATABASES['assessment_items']['NAME'] = os.path.join(CONTENT_ROOT, 'assessmentitems.sqlite')
```

docs/usermanual/userman_admin.rst
MCGallaspy: I would also put the warning here.

docs/usermanual/userman_admin.rst
benjaoming: But if they change it on a system that is using a system-wide location for assessment items, that won't work :) Because then assessmentitems.sqlite **won't** be in CONTENT_ROOT :)

benjaoming: It's also conveniently located in a "note"

MCGallaspy: > But if they change it on a system that is using a system-wide location for assessment items...

Then we should spell out all the possible situations are and the implications of changing `CONTENT_ROOT`. How does a user configure a system-wide assessment items location?

p.s. `.. note::` and `.. warning::` are similar but just have different styles applied.

benjaoming: In case you have system-wide assessment items, you don't have to worry about the assessment items database... its already located outside of CONTENT_ROOT and that cannot be reconfigured.

benjaoming: You mean it's better to post it as a warning? Agreeing that it should be moved, too.

kalite/updates/management/commands/videodownload.py
```diff
@@ -150,7 +149,6 @@ def handle(self, *args, **options):
 
                 # Initiate the download process
                 try:
-                    ensure_dir(settings.CONTENT_ROOT)
```
benjaoming: The directory is already ensured in settings.

kalite/topic_tools/__init__.py
```diff
@@ -33,24 +35,19 @@
...
-from kalite.topic_tools import models as main_models
+from . import models as main_models
+from . import settings
 
-TOPICS_FILEPATHS = {
-    settings.CHANNEL: os.path.join(settings.CHANNEL_DATA_PATH, "topics.json")
-}
-EXERCISES_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "exercises.json")
-ASSESSMENT_ITEMS_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "assessmentitems.json")
-CONTENT_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "contents.json")
```
benjaoming: All of these have been moved to topic_tools.settings

kalite/settings/base.py
```diff
@@ -285,7 +284,7 @@
     },
     "assessment_items": {
         "ENGINE": "django.db.backends.sqlite3",
-        "NAME": KHAN_ASSESSMENT_ITEM_DATABASE_PATH,
+        "NAME": __assessment_items_database_path,
```
benjaoming: It's an internal and non-configurable variable, rather than something we expose *as if* it was configurable.

kalite/django_cherrypy_wsgiserver/cherrypyserver.py
```diff
@@ -30,6 +30,14 @@ def start(self):
...
         # Serve the content files
         if getattr(settings, "CONTENT_ROOT", None):
+            # Assessment items
+            from kalite.contentload.settings import KHAN_ASSESSMENT_ITEM_ROOT
+            static_handler = cherrypy.tools.staticdir.handler(
+                section="/",
+                dir="",
+                root=os.path.abspath(KHAN_ASSESSMENT_ITEM_ROOT)
+            )
+            cherrypy.tree.mount(static_handler, settings.CONTENT_URL + "khan/")
```
benjaoming: KHAN_ASSESSMENT_ITEM_ROOT can be changed, so we have to serve it as it was configured rather than expecting it to always be nested in CONTENT_ROOT

benjaoming:
This PR is very critical, as there are no one else who have expressed a desire for a *review*, I'm going to merge it so we can see if it works.

PR #4097, opened 2015-07-14
Remove old raspberry pi script

benjaoming:
It's okay to remove the old script not that the ka-lite-raspberry-pi deb takes care of it, right?


benjaoming:
I vote for removing this so it doesn't lead to any futher misunderstandings

benjaoming:
PR also includes docs improvements

benjaoming:
Not seeing any responses to this after 12 days of lingering. So merging since it doesn't involve changing any active code-base but merely removing a script that everybody aims to never use again :)

benjaoming:
Also, the documentation included in the PR should be made visible.

PR #4099, opened 2015-07-14
Add versioning to database; migrate if version is not current

### LINKED ISSUE #4069: More robust detection of whether or not to run `setup` during server start (even for upgrades)
For context, see: https://github.com/learningequality/installers/issues/86#issuecomment-119763059

Currently, we auto-run `setup` within the `initialize_kalite` command, before starting the server, when the following conditions are met:

```python
    # Ensure that the database has been synced and a Device has been created
    try:
        assert Settings.get("private_key") and Device.objects.count()
    except (DatabaseError, AssertionError):
        # Otherwise, run the setup command
        self.stdout.write("Setting up KA Lite; this may take a few minutes; please wait!\n")
```

However, this doesn't get triggered during an upgrade, but it needs to be, to avoid problems and confusion. One way we could detect whether or not to auto-run setup would be:
- Have a value called `initialized_version` set in the `Settings` model.
- Check this value in `initialize_kalite`, and run `setup` if it isn't defined or if it's older than the current version.
- Upon completion of the `setup` command, update the value of `initialized_version` if needed.

This would be pretty slick and efficient, unless I'm overlooking something critical. Any thoughts?

MCGallaspy:
Fixes #4069 and learningequality/installers#96.

Edit: fixes learningequality/installers#96. Gotta trigger those robots.


benjaoming:
Just making sure... Will it pass this use case?

1. User installs ka-lite and immediately runs `kalite start`. This way, a non-interactive setup is run.
1. There are not assessment items. So user runs `kalite manage setup` and is prompted to download the assessment items.

kalite/distributed/management/commands/initialize_kalite.py
```diff
@@ -22,17 +22,20 @@ class Command(BaseCommand):
...
-        
-        # Ensure that the database has been synced and a Device has been created
+        """Run the setup command, if necessary.
+            It's necessary if the Settings model doesn't have a "database_version" or if that version doesn't match
+            kalite.version.VERSION, indicating the source has been changed. Then setup is run to create/migrate the db.
+        """
+
         try:
-            assert Settings.get("private_key") and Device.objects.count()
+            from kalite.version import VERSION
```
benjaoming: We *could* decide that no minor releases can include database changes and use SHORTVERSION ?

MCGallaspy:
You're right, setting the database_version should be handled in `setup`. In both your cases, the database will be created for the first time. The use case I wished to address is when a user upgrades from 0.13 to 0.14 without migrating the database (as happens in Windows).

kalite/distributed/management/commands/initialize_kalite.py
MCGallaspy: Is it harmful to use the full version?

benjaoming: it will prompt database migrations on every minor upgrade, it's not harmful, no, and maybe we will want this feature in alpha releases anyways

benjaoming:
This is looking good. I've merged it into alpha8.

MCGallaspy:
Can we merge this in to 0.14.x? Or will it be merged from the alpha branches?

benjaoming:
No, nothing is supposed to be merged via the alpha PRs... they just show what I included in the Alpha.

MCGallaspy:
Right, then shall I resolve the conflict and merge this?

benjaoming:
:+1:

benjaoming:
Since it was part of alpha8, I have had it tested already a fair bit, no problems occurred.

kalite/distributed/management/commands/setup.py
```diff
@@ -365,6 +365,15 @@ def handle(self, *args, **options):
...
         # call_command("clean_pyc", interactive=False, verbosity=options.get("verbosity"), path=os.path.join(settings.PROJECT_PATH, ".."))
 
+        # Migrate the database
+        call_command(
+            "syncdb", interactive=False, verbosity=options.get("verbosity"))
+        call_command("migrate", merge=True, verbosity=options.get("verbosity"))
+        # Create *.json and friends database
+        call_command("syncdb", interactive=False, verbosity=options.get(
+            "verbosity"), database="assessment_items")
+        Settings.set("database_version", VERSION)
```
MCGallaspy: @benjaoming merge conflict resolved.

benjaoming: yes, this looks like the right choice!

PR #4114, opened 2015-07-17
Settings object has no attribute "DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP"

amodia:
@rtibbles & @benjaoming  this fixes #4113 

I used the patch suggested by @cpauya [here](https://github.com/learningequality/ka-lite/issues/4113#issuecomment-122160760)

PR #4115, opened 2015-07-17
Hot fix for failing video downloads on OSX. 

### LINKED ISSUE #3704: Cannot connect to server for downloading video files from current development installs

### LINKED ISSUE #3990: For Mac OS, Video Downloading Issue

cpauya:
Hi @aronasorman, this PR does the following:
* fixes #3704 
* fixes #4114 
* fixes #3990 

As per learningequality/installers#100 - PyRun doesn't include the `multiprocessing` module by default.  So we check if `call_command()` is being run inside PyRun and call another method that uses the `subprocess` module.  

If not running on Pyrun, we resume with current functionality that calls the `CommandProcess` class which uses the `multiprocessing` module.

**To test this with Pyrun**

1. Checkout the [installers repo](https://github.com/learningequality/installers) on another directory, then run `setup.sh` of the OSX installers repo.
2. Press Ctrl+C or cancel when it's already downloading the "ka-lite" repo, we only need the downloaded PyRun package.
3. Checkout this PR's branch.
4. Run `"<installers-working-dir>/installers/osx/temp/pyrun-2.7/bin/pip" install .` to install the `ka-lite-static` module.
5. Run `"<working-dir>/installers/osx/temp/pyrun-2.7/bin/pip" freeze` - to see double-check that `ka-lite-static` is installed.
6. Run `launchctl setenv KALITE_PYTHON "<bin/pyrun full path>"`.  Example: `launchctl setenv KALITE_PYTHON "/Users/cyril/w/fle/ka-lite/installers/osx/temp/pyrun-2.7/bin/pyrun"`
7. Restart the terminal.
8. Check that we have a `KALITE_PYTHON` env var with `env | grep KA`.
9. You can now run `bin/kalite manage setup` command and test the device registration, video downloads, language downloads, and etc to see if they work.

To test with the OSX default Python, just unset the `KALITE_PYTHON` with `launchctl unsetenv KALITE_PYTHON` and restart the server.

**Tips**
1. Make an alias of the 'pyrun' and 'pip' to ease typing.
2. Set DEBUG=True in your `local_settings.py` for easier tracing.

**Other fix**
This also sets the default value for the `DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = False` at `kalite/settings/__init__.py` to fix crashes when `settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP` is used in other Python codes as per #4114.  I had to do this to continue testing the fixes above.


kalite/settings/__init__.py
```diff
@@ -46,6 +46,8 @@ def package_selected(package_name):
         RemovedInKALite_v015_Warning
     )
 
+DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = getattr(local_settings, "DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP", False)
```
benjaoming: Okay, if you want to retrieve it from local_settings, I can see the point... but are you still using local_settings for OSX?

cpauya:
Hi @benjaoming - yes we are still using `local_settings.py` on OSX.  I tried using `bin/kalite start --settings kalite.my_settings` but couldn't remember why I can't make it work.

But we are leaving `local_settings.py` as soon as 0.14.x is released - had to deal with more important tasks. :)

I saw your merged PR already and will use that.  Thanks.

python-packages/fle_utils/django_utils/command.py
```diff
@@ -51,36 +50,6 @@ def call_command_with_output(cmd, *args, **kwargs):
         sys.exit   = backups[2]
 
 
-def call_command_subprocess(cmd, *args, **kwargs):
```
benjaoming: This is safe, the function was re-declared further down in the original, so this code block was always dead.

python-packages/fle_utils/django_utils/command.py
```diff
@@ -138,11 +109,27 @@ def call_command_async(cmd, *args, **kwargs):
     # and have everyone else spawn threads.
     is_osx = sys.platform == 'darwin'
     in_proc = kwargs.pop('in_proc', not is_osx)
-    #in_proc = kwargs.pop('in_proc', True)S
     
     if in_proc:
         return call_command_threaded(cmd, *args, **kwargs)
     else:
+        # MUST: Check if running on PyRun to prevent crashing the server since it doesn't have
```
benjaoming: This will only affect pyrun, and the OSX team are making pretty solid assertions that this works :)

benjaoming:
I think it's about time we merge this so we don't have stuff critical lying around unmerged just before launch!

In a more general setting: Trying to import a module or an object from a module and catching `ImportError` is normally a better pattern for these cases.

But this `python-packages/fle_utils/django_utils/command.py` is terrible and should go, I opened #4139 for this.

cpauya:
Thanks @benjaoming - looking forward for the refactor at #4139 later. :)

PR #4119, opened 2015-07-17
Replace the 'close' X link with an accessible button in user alert messages

### LINKED ISSUE #4117: User alert messages use link with an empty 'href' as a close button - A11y issue
**PR is coming!**  

**Branch**: develop

**Current Behavior**: The use of links with an empty `href` should be avoided at all cost as they represent a problem for users with screen readers. The `X` at the right border of user alert messages, which is used to close them, is constructed with a link instead of the more appropriate `button` element. 

a-close

radinamatic:
Fixes  #4117.

button-close

It might have some merge conflicts, so will have to work on that...


radinamatic:
This is still WIP, and it must have been a bad rebase that emptied this PR... :)
I'll try to dig the commit out of the limbo and push it again.

PR #4122, opened 2015-07-20
Write cache data in user space #4109

### LINKED ISSUE #4109: empty topic tree when DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = True
steps to reproduce:
1. delete the topics.json_en.cache file in your data/khan folder
2. set DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = True in local_settings.py or your way of assigning tags
3. start kalite server
4. go to learn page and you will see an empty sidebar

benjaoming:
Resolves #4109.

### DIFF
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -79,24 +79,20 @@ def get_topic_tree(force=False, annotate=False, channel=None, language=None, par
     if TOPICS.get(channel) is None:
         TOPICS[channel] = {}
 
-    if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
-        cached_topics = softload_json(
-            cache_file_path("topic_{0}_{1}.json".format(channel, language)),
-            logger=logging.debug,
-            raises=False
-        )
+    if annotate or TOPICS.get(channel, {}).get(language) is None:
+        if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
+            cached_topics = softload_json(
+                cache_file_path("topic_{0}_{1}.json".format(channel, language)),
+                logger=logging.debug,
+                raises=False
+            )
         if cached_topics:
             TOPICS[channel][language] = cached_topics
-            if parent:
-                return filter(lambda x: x.get("parent") == parent, cached_topics)
-            else:
-                return cached_topics
-    
-    if annotate or TOPICS.get(channel, {}).get(language) is None:
-        topics = softload_json(settings.TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
-
-        # Just loaded from disk, so have to restamp.
-        annotate = True
+            annotate = False
+        else:
+            topics = softload_json(settings.TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
+            # Just loaded from disk, so have to restamp.
+            annotate = True
 
     if annotate:
         flat_topic_tree = []
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -80,6 +80,7 @@ def get_topic_tree(force=False, annotate=False, channel=None, language=None, par
         TOPICS[channel] = {}
 
     if annotate or TOPICS.get(channel, {}).get(language) is None:
+        cached_topics = None
         if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
             cached_topics = softload_json(
                 cache_file_path("topic_{0}_{1}.json".format(channel, language)),
--- a/kalite/topic_tools/__init__.py
+++ b/kalite/topic_tools/__init__.py
@@ -46,11 +46,22 @@ if not os.path.exists(django_settings.CHANNEL_DATA_PATH):
     logging.warning("Channel {channel} does not exist.".format(channel=django_settings.CHANNEL))
 
 
+def cache_file_path(basename):
+    """Consistently return path for a cache filename. This path has to be
+    writable for the user running kalite."""
+    assert "/" not in basename, "Please use a valid filename"
+    cache_dir = os.path.join(django_settings.USER_DATA_ROOT, 'cache')
+    if not os.path.exists(cache_dir):
+        os.makedirs(cache_dir)
+    return os.path.join(cache_dir, basename)
+
+
 # Globals that can be filled
 TOPICS = None
 CACHE_VARS.append("TOPICS")
+CNT=0
 def get_topic_tree(force=False, annotate=False, channel=None, language=None, parent=None):
-
+    
     # Hardcode the Brazilian Portuguese mapping that only the central server knows about
     # TODO(jamalex): BURN IT ALL DOWN!
     if language == "pt-BR":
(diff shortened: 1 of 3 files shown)


benjaoming:
Now it resolves #4109 ! The topics cache was written wrongly, works after renaming `topics` to `cached_topics`. Also moved the block earlier so default topics JSON isn't *always* loaded.

kalite/topic_tools/__init__.py
```diff
@@ -46,11 +46,22 @@
     logging.warning("Channel {channel} does not exist.".format(channel=django_settings.CHANNEL))
 
 
+def cache_file_path(basename):
+    """Consistently return path for a cache filename. This path has to be
+    writable for the user running kalite."""
+    assert "/" not in basename, "Please use a valid filename"
```
MCGallaspy: What is the intent? That `basename` is not a directory? (Like `mydir/`?)

benjaoming: Yeah, mainly that the channel name was being used for one of the file names and I didn't know where the channel names were coming from... very much not important assertion.

kalite/topic_tools/__init__.py
```diff
@@ -168,12 +187,18 @@ def get_exercise_cache(force=False, language=None):
...
+                raises=False
+            )
             if exercises:
                 EXERCISES[language] = exercises
                 return EXERCISES[language]
         EXERCISES[language] = softload_json(settings.EXERCISES_FILEPATH, logger=logging.debug, raises=False)
-        if language == "en":  # English-language exercises live in application space, translations in user space
+        
+        # English-language exercises live in application space, translations in user space
+        if language == "en":
```
MCGallaspy: Purely a matter of style, but I like comments that live on the same line as the thing they are annotating, as it makes them harder to separate/get out of sync. Imagine this
```python
# talking about foo
do_foo()
```

then later something gets added:
```py
# talking about foo
do_bar()
do_foo()
```

Then `do_foo` is decided to be obsolete and replaced with `do_baz`, so:
```python
# talk about foo
do_bar()
do_baz()
```

benjaoming: Yes it just made the line exceed 80 chars, guessing it was a fly-by subliminal action :)

kalite/topic_tools/__init__.py
```diff
@@ -67,6 +78,20 @@ def get_topic_tree(force=False, annotate=False, channel=None, language=None, par
         TOPICS = {}
     if TOPICS.get(channel) is None:
         TOPICS[channel] = {}
+
+    if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
```
MCGallaspy: I'm confused, why move this block of code around?

benjaoming: Because otherwise we call `softload_json(...)` (loading the original file), then discards all that in favor of a cached version.. that didn't seem to make sense.

MCGallaspy: Yeah, I see what you mean. But now you will call `softload_json` every time `force` is `False` (if this setting is enabled). Seems like the intended behavior of the original version is something like this:

1. If `TOPICS[channel][language]` doesn't exist, then load the _un_-annotated version from disk, then annotate it.
2. If you must annotate the `TOPICS[channel][language]`, under certain conditions (`settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force`) first check if the annotated version is cached, and load the cache if it is.
3. In all other cases, do something else.

In order to preserve this behavior but cut out the unnecessary `softload_json` that you identified, I think it should be like this:

```python
def get_topic_tree(force=False, annotate=False, channel=None, language=None, parent=None):
    
    # Hardcode the Brazilian Portuguese mapping that only the central server knows about
    if language == "pt-BR":
        language = "pt"

    if not channel:
        channel = django_settings.CHANNEL

    if not language:
        language = django_settings.LANGUAGE_CODE

    global TOPICS
    if not TOPICS:
        TOPICS = {}
    if TOPICS.get(channel) is None:
        TOPICS[channel] = {}
    
    if annotate or TOPICS.get(channel, {}).get(language) is None:
        if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
            cached_topics = softload_json(
                cache_file_path("topic_{0}_{1}.json".format(channel, language)),
                logger=logging.debug,
                raises=False
            )
        if cached_topics:
            TOPICS[channel][language] = cached_topics
            annotate = False
            # Get rid of this and let it drop down to similar statement below
            #if parent:
            #    return filter(lambda x: x.get("parent") == parent, cached_topics)
            #else:
            #    return cached_topics
        else:
            topics = softload_json(settings.TOPICS_FILEPATHS.get(channel), logger=logging.debug, raises=False)
            # Just loaded from disk, so have to restamp.
            annotate = True

    if annotate:
        flat_topic_tree = []

        # Loop through all the nodes in the topic tree
        # and cross reference with the content_cache to check availability.
        content_cache = get_content_cache(language=language)
        exercise_cache = get_exercise_cache(language=language)
```

benjaoming: This PR was about fixing stuff that was already broken in some code that I didn't write, I don't feel much like cleaning up the code.. it's a bit like combing the desert, if you know what I mean... I can copy paste your above suggestion, but I also think we should move on :) thanks for the suggestion.

benjaoming: I don't know what `force` was for, but it implies forcing it to load... although you're right it didn't really..

MCGallaspy: I know what you mean, but the tagged issue was about writing caches in user data. Moving around that block of code had nothing to do with that, and it looks like it changes the behavior in the way that I noted. If you don't wanna clean up code, then I say resist the temptation to make unrelated changes!

kalite/topic_tools/__init__.py
```diff
@@ -266,16 +287,22 @@ def get_content_cache(force=False, annotate=False, language=None):
 
     if CONTENT is None:
         CONTENT = {}
+
+    if settings.DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP and not force:
```
MCGallaspy: Same situation as in https://github.com/learningequality/ka-lite/pull/4122#discussion_r35056371.

MCGallaspy: The smallest change would be to restore this block to its original location. Or wait for me to open a PR targeting your branch here. :)

kalite/topic_tools/__init__.py
benjaoming: >  Moving around that block of code had nothing to do with that,

Moving the code block was about only hitting `softload_json` once. I wasn't just cleaning up code, I was fixing it by moving it. Since the original intention was to avoid stuff that made the RPi way too slow, I think parsing the topic tree json once less is okay to push for in the scope of the PR.

I see that it might have delivered results that didn't respect the `annotated` kwarg, thanks for fixing that.. I don't know if it would have mattered since these functions are really bloated kwarg-wise :)

rtibbles:
Yep, this seems to preserve the intended functionality, while allowing flexible cache path writing, and preventing an unnecessary load of the JSON files.

PR #4124, opened 2015-07-20
Fix --foreground process leak

benjaoming:
Instead of leaving an eternally blocking thread behind because cherrypy calls its join(), we use setDaemon and wait for it after asking it to shutdown

### DIFF
--- a/bin/kalite
+++ b/bin/kalite
@@ -109,5 +109,9 @@ p = subprocess.Popen(
     env=env,
 )
 
-p.communicate()
+try:
+    p.communicate()
+except KeyboardInterrupt:
+    # We are not going to show this un-interesting trackeback!
+    pass
 sys.exit(p.returncode)
--- a/bin/kalite
+++ b/bin/kalite
@@ -112,6 +112,6 @@ p = subprocess.Popen(
 try:
     p.communicate()
 except KeyboardInterrupt:
-    # We are not going to show this un-interesting trackeback!
+    # We are not going to show this un-interesting traceback!
     pass
 sys.exit(p.returncode)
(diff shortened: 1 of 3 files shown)


bin/kalite
```diff
@@ -109,5 +109,9 @@ p = subprocess.Popen(
     env=env,
 )
 
-p.communicate()
+try:
+    p.communicate()
+except KeyboardInterrupt:
```
aronasorman: :+1:

bin/kalite
```diff
@@ -109,5 +109,9 @@ p = subprocess.Popen(
...
+    # We are not going to show this un-interesting trackeback!
```
aronasorman: s/trackeback/traceback/

python-packages/fle_utils/chronograph/management/commands/cronserver_blocking.py
```diff
@@ -37,7 +38,8 @@ def handle(self, *args, **options):
                 logging.debug("No jobs due to run.")
             
             # Sleep a little bit at a time to discover if we have to shutdown
-            for __ in range(6):
-                time.sleep(sleep_time // 6)
+            for __ in range(60):
```
aronasorman: Why was this upped?

benjaoming: It's supposed to sleep very little otherwise it won't see the shutdown flag. I mistook the default setting to be 60, it was actually set to 600 in our complex setting stack :)

bin/kalite
benjaoming: Oh that was there on purpose... Just testing your review eyes aren't sleepy :) :)

Issue #4125, opened 2015-07-20
Default cache settings are weird


benjaoming:
This block of code in `kalite.settings.base` is very hard to figure out:

```python
if CACHE_TIME != 0:  # None can mean infinite caching to some functions
    # When we change versions, cache changes, too
    KEY_PREFIX = ".".join(version.VERSION)

    # File-based cache
    install_location_hash = hashlib.sha1(".".join(version.VERSION)).hexdigest()
    username = getpass.getuser() or "unknown_user"
    cache_dir_name = "kalite_web_cache_%s" % (username)
    CACHE_LOCATION = os.path.realpath(getattr(local_settings, "CACHE_LOCATION", os.path.join(tempfile.gettempdir(), cache_dir_name, install_location_hash))) + "/"
    CACHES["file_based_cache"] = {
        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
        'LOCATION': CACHE_LOCATION,  # this is kind of OS-specific, so dangerous.
        'TIMEOUT': CACHE_TIME,  # should be consistent
        'OPTIONS': {
            'MAX_ENTRIES': getattr(local_settings, "CACHE_MAX_ENTRIES", 5*2000)  # 2000 entries=~10,000 files
        },
    }

    # Memory-based cache
    CACHES["mem_cache"] = {
        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
        'LOCATION': 'unique-snowflake',
        'TIMEOUT': CACHE_TIME, # should be consistent
        'OPTIONS': {
            'MAX_ENTRIES': getattr(local_settings, "CACHE_MAX_ENTRIES", 5*2000)  # 2000 entries=~10,000 files
        },
    }

    # The chosen cache
    CACHE_NAME = getattr(local_settings, "CACHE_NAME", "file_based_cache")

```

What does it do?

1. Creates a file-based cache with a 5 year time out.
1. Stores the cache in a temporary location!? (So much for the 5 years)
1. Initializes settings for `"mem_cache"` which are NEVER used anywhere.
1. Then finally, it sets `CACHE_NAME`, used in `fle_utils.internet.webcache` (yes, we have a package called `internet`). So what's CACHES['default'] supposed to be for!?
1. At every start of KA Lite `kalite start` the cache is explicitly deleted anyways? Why do we bother to have a cache? Why do we bother with the tempdir? What's with the 5 years?

[...]

PR #4128, opened 2015-07-20
Cache settings #4125

benjaoming:
Builds upon #4122

Fixes one main issue:

* File based caching is not persistent because it's stored in a tempdir!

New feature:

* The dev settings now include memory based caching by default
* Docs mention need to clear webcache at one of the settings where it's necessary
* Settings for caching much less messy
* More efficient topic_tree view now that it doesn't have to write the topic tree to a useless file-based cache

#4125


benjaoming:
@MCGallaspy assigning you because it's an easy task IMO... just check that the code doesn't have any obvious problems... it's been tested in 3 alphas now without issues.

rtibbles:
Seems OK to me, and has been tested.

PR #4130, opened 2015-07-20
0.15 testing fixup

MCGallaspy:
This is WIP. Create expected level of isolation for behave tests. Basically, flush the database inbetween each scenario. Consequently, some tags which were applied to features need to be applied to each scenario.

Working now on individual test failures.


kalite/coachreports/features/environment.py
```diff
@@ -1,4 +1,4 @@
 """
 environment.py specific to the this app
 """
-from kalite.testing.base_environment import before_all, after_all, before_feature, after_feature
+from kalite.testing.base_environment import *
```
benjaoming: This is usually not cool, any reason you want to import * ?

benjaoming: test

MCGallaspy: laziness?

MCGallaspy: But I will change it, since you are of course right

benjaoming: laziness can be a valid argument :)

Laziness is a good reason for PyCharm or LiClipse btw :)

kalitectl.py
```diff
@@ -400,7 +400,7 @@ def start_watchify():
     sys.stdout.write('Starting watchify')
 
     watchify_process = subprocess.Popen(
-        ['node compile_javascript.js --debug --watch --staticfiles'],
+        args='node compile_javascript.js --debug --watch --staticfiles',
```
MCGallaspy: [This is recommended.](https://docs.python.org/2/library/subprocess.html#subprocess.Popen) Old version fails on Windows.

benjaoming: But Popen doesn't have a kwarg `args=...` ?

MCGallaspy: True, but it still works. e.g.
```python
>>> def f(a, b):
...     return a - b
...
>>> f(3,1)
2
>>> f(b=3,a=1)
-2
>>>
```

benjaoming: But is it necessary for it to work on Windows? Or was that the part where you changed the list to a string?

MCGallaspy: list to string. But it's still perfectly valid python. It seems [any argument can be referenced by name](https://docs.python.org/2/tutorial/controlflow.html#keyword-arguments), whether the argument is required or not (require==does not have a default value).

kalite/distributed/static/js/distributed/bundle_modules/homepage.js
```diff
@@ -14,4 +14,12 @@ $(function() {
         }
     }
     window.statusModel.listenTo(window.statusModel, "change", content_rec_load);
+
+    window.statusModel.listenTo(window.statusModel, "change:has_superuser", function () {
+        if (!window.statusModel.get("has_superuser")) { // "has_superuser" is true if an admin user exists.
+            window.super_user_create_modal = new user.SuperUserCreateModalView();
+        } else if ( typeof window.super_user_create_modal !== "undefined" ) {
+            window.super_user_create_modal.remove();
+        }
```
MCGallaspy: @rtibbles is this kosher?

rtibbles: As Christmas... I mean, yeah, looks fine.

kalite/coachreports/features/environment.py
rtibbles: I think in this instance it is actually a good idea.

kalite/coachreports/features/steps/coachreports.py
```diff
@@ -81,12 +81,40 @@ def impl(context):
 
 @given(u"I am on the tabular report")
 def impl(context):
+    create_some_learner_data()
```
rtibbles: Not sure I am keen on this. Data should be ensured by other 'given' steps.

kalite/coachreports/features/steps/coachreports.py
```diff
@@ -241,4 +269,4 @@ def impl(context):
 
 @when(u"I click on the completed colored cell")
 def impl(context):
-    click_and_wait_for_id_to_appear(context, find_css_with_wait(context, "td.complete[value=subtraction_1]"), "details-panel-view")
+    click_and_wait_for_id_to_appear(context, find_css_with_wait(context, "td.complete"), "details-panel-view")
```
rtibbles: This specificity should be coming from a 'given' statement.

rtibbles:
Merging. I will continue the test cleanup from develop.

Torch passed!

kalite/coachreports/features/environment.py
benjaoming: @rtibbles `import *` is never a good idea when you are importing from a module that doesn't define `__all__`

rtibbles: In that case, I would say the problem here is that the module being imported from is missing an ```__all__```.

PR #4136, opened 2015-07-21
New diagnose command diagnoses system (not so robust)

benjaoming:
Example:

```
✗ bin/kalite diagnose

KA Lite diagnostics

Calculating diagnostics...

KA LITE VERSION:      0.14.0
PYTHON:               2.7.6 (default, Mar 22 2014, 22:59:56) 
                      [GCC 4.8.2]
PLATFORM:             Linux
                      hostname
                      3.16.0-38-generic
                      #52~14.04.1-Ubuntu SMP Fri May 8 09:43:57 UTC 2015
                      x86_64
SERVER ADDRESS:       http://192.168.40.148:8008/
SERVER ADDRESS:       http://127.0.0.1:8008/
SERVER STATUS:        OK, running
INSTALLED IN:         code/ka-lite/kalite
CONTENT ROOT:         code/ka-lite/content
CONTENT SIZE:         3.0 MB
USER DATABASE:        code/ka-lite/kalite/database/data.sqlite
ASSESSMENT DATABASE:  code/ka-lite/content/assessmentitems.sqlite
DEVICE NAME:          hostname
DEVICE UUID:          60d28ba6c1d9593b8f9c6b61d1ae7935
DEVICE REGISTERED:    True
DEVICE SYNCED:        10 times
ZONE:                 Zone for public key MIIBCgKCAQEA1o91D7UGvm8wfovNq92h6Uzxm9C5QDXRl43268
```

### DIFF
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -197,7 +197,7 @@ def update_default_args(defaults, updates):
 
 
 def get_size(start_path):
-    """Utility function"""
+    """Utility function, returns the size (bytes) of a folder"""
     total_size = 0
     for dirpath, __, filenames in os.walk(start_path):
         for f in filenames:
@@ -607,11 +607,15 @@ status.codes = {
 
 def diagnose():
     """
-    This command diagnoses a running instance of KA Lite
+    This command diagnoses an installation of KA Lite
     
     It has to be able to work with instances of KA Lite that users do not
     actually own, however it's assumed that the path and the 'kalite' commands
     are configured and work.
+    
+    The function is currently non-robust, meaning that not all aspects of
+    diagnose data collection is guaranteed to succeed, thus the command could
+    potentially fail :(
     
     Example: KALITE_HOME=/home/otheruser/.kalite kalite diagnose --port=7007
     """
@@ -620,6 +624,8 @@ def diagnose():
     print("KA Lite diagnostics")
     print("")
     
+    # Tell users we are calculating, because checking the size of the
+    # content directory is slow. Flush immediately after.
     print("Calculating diagnostics...")
     sys.stdout.flush()
     print("")
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -62,6 +62,7 @@ from __future__ import print_function
 # Add distributed python-packages subfolder to current path
 # DO NOT IMPORT BEFORE THIS LIKE
 import os
+import platform
 import socket
 import sys
 import time
@@ -638,7 +639,7 @@ def diagnose():
     
     diag("KA Lite version", kalite.__version__)
     diag("python", sys.version)
-    diag("platform", "\n".join(os.uname()))
+    diag("platform", platform.platform())
     
     try:
         __, __, port = get_pid()
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -667,11 +667,14 @@ def diagnose():
         try:
             from securesync.models import Device
             device = Device.get_own_device()
+            sync_sessions = device.client_sessions.all()
+            zone = device.get_zone()
             diag("device name", str(device.name))
-            diag("device UUID", str(device.get_uuid()))
+            diag("device ID", str(device.id))
             diag("device registered", str(device.is_registered()))
-            diag("device synced", str(device.client_sessions.all().count()) + " times")
-            diag("zone", device.get_zone())
+            diag("synced", str(sync_sessions.latest('timestamp').timestamp if sync_sessions.exists() else "Never"))
+            diag("sync result", ("OK" if sync_sessions.latest('timestamp').errors == 0 else "Error") if sync_sessions.exists() else "-")
+            diag("zone ID", str(zone.id) if zone else "Unset")
         except:
             diag("Device failure", traceback.format_exc())
     
--- a/kalitectl.py
+++ b/kalitectl.py
@@ -13,6 +13,7 @@ Usage:
   kalite shell [options] [DJANGO_OPTIONS ...]
   kalite test [options] [DJANGO_OPTIONS ...]
   kalite manage [options] COMMAND [DJANGO_OPTIONS ...]
+  kalite diagnose [options]
   kalite -h | --help
   kalite --version
 
@@ -42,6 +43,7 @@ Examples:
   kalite stop           Stop KA Lite
   kalite shell          Display a Django shell
   kalite manage help    Show the Django management usage dialogue
+  kalite diagnose       Show system information for debugging
 
   kalite start --foreground   Run kalite in the foreground and do not go to
                               daemon mode.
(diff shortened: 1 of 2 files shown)


benjaoming:
Marking this a release blocker because I want the ka-lite-gtk UI to depend on it, hope that's okay.

benjaoming:
On a further note: Since 0.15 should move kalitectl, I would like to keep the function here for now. Later, if needed, we could create a utility function that outputs raw data for other views, like a web page, to utilize.

jamalex:
> DEVICE SYNCED:        10 times

This won't be so meaningful, since we limit the number of locally stored syncsessions to 10. Perhaps "LAST SYNCED" and "LAST SYNC RESULTS" (success, failed, etc). Perhaps it could list "RECORDS IN PURGATORY" or something, too.

> ZONE:                 Zone for public key MIIBCgKCAQEA1o91D7UGvm8wfovNq92h6Uzxm9C5QDXRl43268

Looks like it's using the name (which in some cases could be ambiguous); Zone ID would be better (`zone.id`).

kalitectl.py
```diff
@@ -592,6 +605,85 @@ def status():
 }
 
 
+def diagnose():
+    """
```
jamalex: Ooh, any way we could default to `traceback = True` for this command, so that in the case even this command fails, the thing they copy will provide useful feedback to us?

kalitectl.py
```diff
@@ -592,6 +605,85 @@ def status():
...
+        diag("installed in", os.path.dirname(kalite.__file__))
+        diag("content root", settings.CONTENT_ROOT)
+        diag("content size", filesizeformat(get_size(settings.CONTENT_ROOT)))
+        diag("user database", settings.DATABASES['default']['NAME'])
+        diag("assessment database", settings.DATABASES['assessment_items']['NAME'])
+        try:
+            from securesync.models import Device
+            device = Device.get_own_device()
+            diag("device name", str(device.name))
+            diag("device UUID", str(device.get_uuid()))
```
jamalex: This should just be `device.id`, rather than recomputing the id from first principles.

jamalex:
A couple notes, then looks good to merge as a first pass. Thanks!

benjaoming:
> > DEVICE SYNCED: 10 times
> 
> This won't be so meaningful, since we limit the number of locally stored syncsessions to 10.

In that case, we should remove it here?

```
➜  ka-lite git:(info-command) ✗ ack-grep --html num_times_synced kalite
kalite/control_panel/templates/control_panel/partials/_device_table.html
35:                                {% if device.num_times_synced != None %}<a href="{% url 'device_management' zone_id=zone_id device_id=id %}">{% endif %}
38:                                {% if device.num_times_synced != None %}</a>{% endif %}
41:                                {% if device.num_times_synced != None %}
42:                                    {{ device.num_times_synced }}
```

kalitectl.py
benjaoming: I guess worst case is they get a failure with traceback on the command line when running it.

benjaoming:
```
✗ bin/kalite diagnose

KA Lite diagnostics

Calculating diagnostics...

KA LITE VERSION:      0.14.0
PYTHON:               2.7.6 (default, Mar 22 2014, 22:59:56) 
                      [GCC 4.8.2]
PLATFORM:             Linux
                      hostname
                      3.16.0-38-generic
                      #52~14.04.1-Ubuntu SMP Fri May 8 09:43:57 UTC 2015
                      x86_64
SERVER ADDRESS:       http://192.168.40.148:8008/
SERVER ADDRESS:       http://127.0.0.1:8008/
SERVER STATUS:        OK, running
INSTALLED IN:         /media/archive/code/ka-lite/kalite
CONTENT ROOT:         /media/archive/code/ka-lite/content
CONTENT SIZE:         3.0 MB
USER DATABASE:        /media/archive/code/ka-lite/kalite/database/data.sqlite
ASSESSMENT DATABASE:  /media/archive/code/ka-lite/content/assessmentitems.sqlite
DEVICE NAME:          hostname
DEVICE ID:            60d28ba6c1d9593b8f9c6b61d1ae7935
DEVICE REGISTERED:    True
SYNCED:               2015-07-21 02:19:05.435716
SYNC RESULT:          OK
ZONE ID:              756ae3bfefe4439fa6602a983607f40a
```

kalitectl.py
jamalex: I guess we're not running any Django management commands, so `--traceback` won't be helpful anyway. If it fails for some reason, does it display a full traceback (rather than swallowing it)?

jamalex:
> In that case, we should remove it here?

Most of those are just checking whether it was synced *at all*, which is still useful. The "# Times Synced" column isn't so useful, but let's leave it for now -- we need to rethink those pages in general at some point, as the UX is bad.

jamalex:
> USER DATABASE:        code/ka-lite/kalite/database/data.sqlite

Are these absolute/relative paths? Why no leading `/`?

jamalex:
Beauty! Looks good to merge, if it's been testing in `IS_SOURCE = False` mode as well?

benjaoming:
No, but I think it might be relevant to look at it in relation to calls from another user, something like

    KALITE_HOME=/home/otheruser/ kalite diagnose

benjaoming:
Yes, that works, too!

```
➜  ~  KALITE_HOME=/var/ka-lite/.kalite kalite diagnose

KA Lite diagnostics

Calculating diagnostics...

KA LITE VERSION:      0.14.0
PYTHON:               2.7.6 (default, Mar 22 2014, 22:59:56) 
                      [GCC 4.8.2]
PLATFORM:             Linux-3.16.0-38-generic-x86_64-with-Ubuntu-14.04-trusty
SERVER ADDRESS:       http://127.0.0.1:8008/
SERVER STATUS:        OK, running
INSTALLED IN:         /home/benjamin/.virtualenvs/kalite_pip/local/lib/python2.7/site-packages/ka_lite-0.14.0-py2.7.egg/kalite
CONTENT ROOT:         /var/ka-lite/.kalite/content
CONTENT SIZE:         0 bytes
USER DATABASE:        /var/ka-lite/.kalite/database/data.sqlite
ASSESSMENT DATABASE:  /var/ka-lite/.kalite/content/assessmentitems.sqlite
DEVICE NAME:          hostname
DEVICE ID:            5eaa4d250e4654938b5986f1840ed48e
DEVICE REGISTERED:    False
SYNCED:               Never
SYNC RESULT:          -
ZONE ID:              Unset
```

benjaoming:
Merging this because it's harmless and we can do improvements in new issues.

kalitectl.py
benjaoming: Wondering what happened to my comment. Anyways, the diagnose command does not swollow tracebacks from failures in its own sub-calls... if it fails, it will fail loudly... and that traceback is pretty much more important than other parts of the diagnostics :)

PR #4142, opened 2015-07-22
Replace the 'close' <a> element with an accessible button in user alert messages

### LINKED ISSUE #4117: User alert messages use link with an empty 'href' as a close button - A11y issue

radinamatic:
Fixes #4117, instead of #4119 that suffered a bad rebase.

button-close

### DIFF
--- a/kalite/distributed/static/js/distributed/utils/messages.js
+++ b/kalite/distributed/static/js/distributed/utils/messages.js
@@ -34,7 +34,7 @@ function show_message(msg_class, msg_text, msg_id) {
         return $("#message_container");
     }
 
-    var x_button = '<a class="close" data-dismiss="alert" href="#">&times;</a>';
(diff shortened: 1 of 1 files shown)


radinamatic:
I don't know if it's just me, or CircleCI is messing with everybody...
Anyway, let me know if there is something I should correct on my end.

benjaoming:
Usually we have timeouts. It may be worth verifying if the change has an effect on these behave tests and if the behave tests that fail are really testing something essential.

radinamatic:
All I can see are coachreports and login modal issues, nothing supposedly touched by this PR.

PR #4149, opened 2015-07-24
0.15 - Content Rating

MCGallaspy:
WIP. Added minimal front-end stuff to let @jtamiace prototype.


kalite/distributed/static/js/distributed/topics/hbtemplates/content-area.handlebars
```diff
@@ -1,3 +1,3 @@
 {{!-- <h3>{{ title }}</h3> --}}
-
-<div class="content"></div>
\ No newline at end of file
+<div class="content"></div>
+<div id="rating-container-wrapper">rating container yeah!</div>
```
MCGallaspy: @rtibbles or anyone, better place to put this?

rtibbles: Seems fine to me.

rtibbles:
Looks good as a starting point.

MCGallaspy:
Waiting for a new spec. Made another branch off this branch in the meantime: https://github.com/MCGallaspy/ka-lite/commits/content-rating-wip

MCGallaspy:
@rtibbles the front end code here should be essentially done. Can you review?

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,251 @@
...
+            Prepare self and subviews.
+        */
+        this.model = options.model || function() {
+            var model = new Backbone.Model();
+            model.url = "/api/not/gonna/work";
+            return model;
+        }();
+        _.bindAll(this, "delete_callback", "submit_rating", "edit_callback", "render1st", "render2nd", "render3rd", "renderTextArea", "renderAll");
+        this.model.fetch().done(this.renderAll).fail(this.render1st);
+        window.rating_view = this;
```
MCGallaspy: oops. Will remove.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,251 @@
...
+
+    edit_callback: function() {
+        this.text_view.toggle_disabled(false);
+        this.$(".rating-submit").show();
+        this.$(".rating-edit").hide();
+    },
+
+    delete_callback: function() {
+        // delete the model, then start afresh!
+        // To-do: delete model
```
MCGallaspy: Will remove these comments as well.

MCGallaspy:
@aronasorman let's review this together on Friday, August 14, 2015.

kalite/distributed/static/js/distributed/rating/models.js
```diff
@@ -0,0 +1,18 @@
+var Backbone = require("../base/backbone");
+
+var RatingModel = Backbone.Model.extend({
+    urlRoot: "/api/content_rating"
+});
+
+var ContentRatingCollection = Backbone.Collection.extend({
+    parse: function(response) {
```
rtibbles: Is this necessary?

c.f. in Backbone-tastypie:
```
	Backbone.Collection.prototype.parse = function( data ) {
		if ( data && data.meta ) {
			this.meta = data.meta;
		}

		return data && data.objects || data;
	};
```

Should already be happening.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,263 @@
...
+            el: this.$("#star-container-3"),
+            model: this.model,
+            rating_attr: "rating3"
+        });
+        var self = this;
+        this.model.once("change:rating3", function() {
+            self.star_view_3.remove();
+            self.renderTextArea();
+        })
+    },
```
rtibbles: These three methods seem rather WET.

rtibbles: Might be a case for subviews that govern the individual ratings, rather than all being handled by the top level view?

kalite/distributed/static/js/distributed/rating/models.js
MCGallaspy: Ah, I guess not. I assumed the `meta` object returned along with objects would throw off Backbone.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,263 @@
...
+        this.star_view_3 = this.add_subview(StarView, {
+            el: this.$("#star-container-3"),
+            model: this.model,
+            rating_attr: "rating3"
+        });
+        this.text_view = this.add_subview(TextView, {
+            el: this.$("#text-container"),
+            model: this.model,
+            rating_attr: "text"
+        });
```
rtibbles: Again, this seems repetitious of methods above.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,263 @@
...
+        this.submit_rating();
+    },
+
+    submit_rating: function() {
+        this.text_view.toggle_disabled(true);
+        this.$(".rating-submit").hide();
+        this.$(".rating-edit").show();
+        this.model.save();
```
rtibbles: Do we want to do anything depending on success or failure of the save?

kalite/distributed/static/js/distributed/topics/hbtemplates/content-area.handlebars
rtibbles: Might want to delete the text inside it now though?

MCGallaspy: Yes, I've mentally filed that away under the "Making beautiful templates" task.

requirements.txt
```diff
@@ -8,3 +8,4 @@ ply==3.4
 slimit==0.8.1
 six==1.9.0
 django-appconf==1.0.1
+python-dateutil==2.4.2
```
rtibbles: Where is this used?

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: Certainly adding a subview & adding the attribute change callback could be spun-off.

requirements.txt
MCGallaspy: [See here, for example](https://github.com/learningequality/ka-lite/blob/develop/kalite/control_panel/views.py#L4). It was a missing dependency for me -- could be my distribution of python? But it doesn't seem to be a standard library module.

rtibbles: Weird.

MCGallaspy:
For reviewers, you'll have to create and run migrations for this to work. I'm waiting until it's finalized to commit any.

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: The only thing I can think of is to give the user some feedback e.g. a message with `Your rating was not submitted! :(`

MCGallaspy:
Will work on templates/display after this is merged. Ready now @aronasorman.

kalite/control_panel/api_resources.py
```diff
@@ -358,3 +358,38 @@ def alter_list_data_to_serialize(self, request, to_be_serialized):
...
+        authorization = ObjectAdminAuthorization()
+        excludes = ['signed_version', 'counter', 'signature', 'id', 'resource_uri']
+        serializer = CSVSerializer()
+
+    def obj_get_list(self, bundle, **kwargs):
+        self._facility_users = self._get_facility_users(bundle)
+        content_ratings = ContentRating.objects.filter(user__id__in=self._facility_users.keys())
+        return super(ContentRatingExportResource, self).authorized_read_list(content_ratings, bundle)
+
+    def alter_list_data_to_serialize(self, request, to_be_serialized):
```
aronasorman: Can you add a docstring briefly describing what you're doing here?

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,238 @@
...
+        this.model = options.model || function() {
+            var model = new Backbone.Model();
+            model.url = "/api/not/gonna/work";
+            return model;
+        }();
+        _.bindAll(this, "delete_callback", "submit_rating", "edit_callback", "renderQuality", "renderDifficulty", "renderTextArea", "renderAll", "render");
+    },
+
+    render: function() {
+        if( this.model.isNew() || this.model.get("quality") === 0) {
```
aronasorman: Can you put a comment explaining this if statement.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,238 @@
...
+            Called when the view's model is not fetched successfully -- particularly when it doesn't yet exist.
+        */
+        this.$el.html(this.template());
+        this.$(".rating-submit").hide();
+        this.$(".rating-edit").hide();
+        this.$(".rating-skip").hide();
+        this.$(".rating-delete").hide();
+        this.add_subview_and_once(StarView, "star_view_quality",
+            {el: this.$("#star-container-quality"), model: this.model, rating_attr: "quality"},
+            "change:quality", this.renderDifficulty);
```
aronasorman: Rather than always explictly calling `renderDifficulty`, just have it raise its own event and have another function attach the next function.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,238 @@
...
+
+    edit_callback: function() {
+        this.text_view.toggle_disabled(false);
+        this.$(".rating-submit").show();
+        this.$(".rating-edit").hide();
+    },
+
+    delete_callback: function() {
+        var self = this;
+        this.$el.html("Deleting your review...");
```
aronasorman: Never goes away from our testing

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -0,0 +1,238 @@
...
+        this.render();
+    },
+
+    render: function() {
+        this.$el.html(this.template(this.model.attributes));
+        return this;
+    },
+
+    /* Debounce'd -- waits until the input stops arriving (500 ms) to update the model */
+    text_changed: _.debounce(function() {
```
aronasorman: Why not update every keypress? Tablets can take it :D

aronasorman:
There were no checks for hiding the content rating view when the user gets logged out. Can you add that?

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -34,10 +53,45 @@ var ContentAreaView = BaseView.extend({
...
+                    "user": window.statusModel.get("user_uri"),
+                    "content_kind": this.model.get("kind"),
+                    "content_id": this.model.get("id")
+                })
+            });
+        }
+        if( this.rating_view.model.get("content_id") !== this.model.get("id") ) {
+            var self = this;
+            this.content_rating_collection.fetch().done(function(){
+                if(self.content_rating_collection.models.length == 1) {
```
aronasorman: `===`

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -34,10 +53,45 @@ var ContentAreaView = BaseView.extend({
...
+                    self.rating_view.model = self.content_rating_collection.pop();
+                    self.rating_view.render();
+                } else {
```
aronasorman: This else statement handles the case of having more than 1 content rating model. It will never happen in the backend, but you never know. Can you print out an error saying "Something's wrong, you shouldn't have more than 1 content rating model!"

kalite/distributed/static/js/distributed/utils/api.js
```diff
@@ -76,7 +76,9 @@ function handleFailedAPI(resp, error_prefix) {
             try {
                 messages = $.parseJSON(resp.responseText || "{}").messages || $.parseJSON(resp.responseText || "{}");
             } catch (e) {
-                var error_msg = sprintf("%s<br/>%s<br/>%s", resp.status, resp.responseText, resp);
+                // Replacing resp.responseText with "blah blah blah" is just a workaround... this should be fixed.
+                // See https://github.com/learningequality/ka-lite/issues/4203
+                var error_msg = sprintf("%s<br/>%s<br/>%s", resp.status, "blah blah blah", resp);
```
aronasorman: Mike: "Take out"

kalite/testing/base_environment.py
```diff
@@ -56,6 +69,11 @@ def before_scenario(context, scenario):
 def after_scenario(context, scenario):
     if context.logged_in:
         logout(context)
+
+    if "download_csv" in context.tags:
+        import shutil
```
aronasorman: Move to top

kalite/testing/base_environment.py
```diff
@@ -79,3 +97,15 @@ def database_teardown(context):
     """
     for alias in connections:
         call_command("flush", database=alias, interactive=False)
+
+def do_fake_registration():
+    """
+    Register the device, in case some feature being tested depends on it. Will be undone by the database teardown.
+    """
+    # Create a Zone and DeviceZone to fool the Device into thinking it's registered
+    from securesync.models import Zone, Device, DeviceZone
```
aronasorman: Also move to top

kalite/testing/base_environment.py
```diff
@@ -30,7 +30,20 @@ def after_feature(context, feature):
...
     database_setup(context)
 
-    browser = context.browser = webdriver.Firefox()
+    if "registered_device" in context.tags:
+        do_fake_registration()
+
+    profile = webdriver.FirefoxProfile()
+    if "download_csv" in context.tags:
+        # Let csv files be downloaded automatically. Can be accessed using context.download_dir
+        import tempfile
```
aronasorman: Move to top

kalite/testing/behave_helpers.py
```diff
@@ -232,14 +233,28 @@ def _visibility():
         return None
 
 
-def build_url(context, url):
-    return urljoin(context.config.server_url, url)
+def build_url(context, url, params={}):
```
aronasorman: add docstring

kalite/testing/behave_helpers.py
```diff
@@ -232,14 +233,28 @@ def _visibility():
...
     url = reverse("homepage")
     context.browser.get(build_url(context, url))
 
 
+def go_to_content_item(context):
+    """
+    Go to the page for a content item (video, exercise, etc). This method does not care which one, it's to ensure
+    that an instantiated ContentAreaView exists on the page. Not the content item may not be flagged as "available".
+    """
+    content = get_content_cache().itervalues().next()
```
aronasorman: Use `get_content_data`

kalite/distributed/static/js/distributed/utils/api.js
MCGallaspy: Removing the "blah blah blah" but on second thought I stand by the discussion in #4203... printing the `responseText` is potentially very ugly.

kalite/testing/behave_helpers.py
MCGallaspy: Won't work, as we discussed.

MCGallaspy:
> There were no checks for hiding the content rating view when the user gets logged out. Can you add that?

I was wrong, learners are redirected to the homepage after logging out. But I added logic to remove the view anyway in case a user logs out.

Otherwise, everything is addressed @aronasorman.

kalite/main/content_rating_models.py
```diff
@@ -0,0 +1,18 @@
+from django.db import models
+
+from kalite.facility.models import FacilityUser
+
+from securesync.models import DeferredCountSyncedModel
+
+class ContentRating(DeferredCountSyncedModel):
```
jamalex: Need to add `minversion = "0.15.0"` in here (top-level within the class definition), ala https://github.com/MCGallaspy/ka-lite/blob/beautify-content-rating/kalite/main/models.py#L491

kalite/main/content_rating_models.py
```diff
@@ -0,0 +1,18 @@
...
+from securesync.models import DeferredCountSyncedModel
+
+class ContentRating(DeferredCountSyncedModel):
+    class Meta:
+        unique_together = ("content_kind", "content_id", "user")
+
+    # Maintain info on content type -- could be video, exercise, etc, but we should be able to uniquely id it
+    content_kind = models.CharField(max_length=100, db_index=True, blank=False)
+    content_id = models.CharField(max_length=100, db_index=True, blank=False)
+
```
jamalex: Need to add "content_source" in here as well, so we'll later be able to distinguish channels more easily. See https://github.com/MCGallaspy/ka-lite/blob/beautify-content-rating/kalite/main/models.py#L531

kalite/main/content_rating_models.py
```diff
@@ -0,0 +1,18 @@
...
+    user = models.ForeignKey(FacilityUser, blank=False)
+    quality = models.IntegerField(blank=False, default=0)
+    difficulty = models.IntegerField(blank=False, default=0)
+    text = models.TextField(blank=True, null=True)
```
jamalex: Usually don't put `null=True` on `CharField` or `TextField`.

MCGallaspy: Okay. For my benefit, any explanation?

benjaoming: `null=True` is fine on CharFields, because lots of databases actually support a null-value and will throw an error if you populate a row without inserting to a non-null CharField.

It's another case for TextField, AFAIK it's not nullable in any db engines.

jamalex: https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.null

> Avoid using null on string-based fields such as CharField and TextField because empty string values will always be stored as empty strings, not as NULL. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL.

benjaoming: @jamalex cool, thanks for that!! I was thinking about how the database works, didn't know Django handles this for CharFields

PR #4151, opened 2015-07-24
Changed the VideoJS version number in the package.json to 4.11

radinamatic:
`4.12` dropped the support for **srt** files, in favor of implementing **vtt**. `4.11` version is the most recent (Jan. 2015) that supports **srt**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "scriptjs": "^2.5.7",
     "soundmanager2": "^2.97.20150601",
     "underscore": "~1.8.3",
-    "video.js": "^4.12.10",
+    "video.js": "^4.11.4",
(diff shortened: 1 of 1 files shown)


benjaoming:
Why did they drop support for srt?

radinamatic:
VTT is becoming a standard for timed tracks:
https://github.com/videojs/video.js/issues/1956#issuecomment-85617607

benjaoming:
Then opening an issue about migrating to VTT would make sense :+1:

radinamatic:
Yes, we mentioned that with @rtibbles. It will have to be done eventually, especially considering the future upgrade to `videojs` version 5, which is even better, a11y wise.

PR #4168, opened 2015-07-27
make sure the language-specific i18n js catalog file is loaded before en.js.

aronasorman:
Or else, we get english catalog files, i.e. it won't get translated.

Also added a few more gettext calls in the coachreports JS.

Assigned you @rtibbles since I moved en.js out of the compress block. Aside from a slightly longer asset loading process, are there any other implications?


aronasorman:
Merging this now, as the changes seem harmless*.

* As uttered by every zombie virus scientist ever.

PR #4169, opened 2015-07-27
Better version info

benjaoming:
Also adds a change so docs reflect the version precisely

Example:

```
bin/kalite shell
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
(InteractiveConsole)
>>> from kalite import version
>>> version.VERSION
'0.14.dev20150727211121'
```


kalite/version.py
```diff
@@ -5,10 +5,15 @@
...
-PATCH_VERSION = "0"
-VERSION = "%s.%s.%s" % (MAJOR_VERSION, MINOR_VERSION, PATCH_VERSION)
+
+from django.utils.version import get_version
+
+VERSION_TUPLE = (0, 14, 0, "alpha", 0)
+
+MAJOR_VERSION, MINOR_VERSION, PATCH_VERSION = map(str, VERSION_TUPLE[:3])
+
+VERSION = get_version(VERSION_TUPLE)
```
MCGallaspy: What's the output of `get_version`? Does it comply with the warning given above?

> Must also be of the form N.N.N for internal use, where N is a non-negative integer

MCGallaspy: Here's the code, for convenience: https://github.com/learningequality/ka-lite/blob/develop/python-packages/django/utils/version.py

MCGallaspy: So it will comply with the warning if it's a `"final"` release. Perhaps @jamalex can comment on whether that's acceptable? Unfortunately I can't supply an example of what a non-compliant version number will break.

benjaoming: Regarding using letters in the version string, it does what we've been doing with all the alphas and will make all dev work be based on a "123.123.dev123" string... so that part will be well-tested though not so crucial for  releases (unless we do 123.123.post123).

MCGallaspy: I'd like someone else to review this. I added the warning, but it was because @jamalex advised me to. I think he's in a better position to say whether this is ok or not.

benjaoming: It really isn't doing something that we weren't doing already. I.e. releasing with `VERSION_TUPLE = (0, 14, 0, "final", 0)` is equivalent of releasing with `VERSION="0.14"` and so basically it's safe enough as long as we bump the version before releasing.

benjaoming:
```
>>> get_version((0, 14, 0, "final" ,0))
'0.14'
```

MCGallaspy: All I'm saying is that I'm not in a position to evaluate whether this is ok if, e.g. `VERSION_TUPLE = (0, 14, 0, "alpha", 42)`. I get that this is already the case on the alpha branches. In my mind that might be an issue, just because I was told the version string had to have a certain format, though I don't know why.

Also `get_version((0, 14, 0, "final" ,0))` does in fact return something different from before. `0.14` vs `0.14.0`. I would just like someone else to clarify what is required internally for our version numbers.

jamalex: @benjaoming @mcgallaspy I spent some time digging into this on the airplane, and I'm glad I did!

My concern was that the function `fle_utils.general.version_diff` (which gets used for minversion comparison for critical securesync functions, in `fle_utils.django_utils.serializers.versioned_json.Serializer`) barfs if there are non-integer components passed in as part of the version string. However, it is (meant to be; see below) using `kalite.version.VERSION`, which only includes the first three components, so only the first three components need to be integers. Hence, `(0, 14, 0, "alpha", 42)` would be OK, but `(0, 14, "alpha", 42)` or `(0, 14, "0alpha")` etc would NOT be.

However, in looking into this, I discovered that as of 0.14, the proper `VERSION` is no longer being used by securesync, with [this commit](https://github.com/learningequality/ka-lite/commit/66f996e1d3157201dcdf2b8984be51eb13f77bf1#diff-2fe3caeae433a31561ccb9c8334aaa92R15) in April (as part of the settings refactor, I believe; @benjaoming), and now `securesync.VERSION` is always hard-coded to "1.0". Since "1.0" is bigger than all our `minversion` values, we would never exclude any models/fields when syncing with older versions, leading to bad sync errors.

I put a fix in at https://github.com/learningequality/ka-lite/pull/4205

MCGallaspy: @jamalex note also that `get_version(VERSION_TUPLE)` does not necessarily return three dot-separated integers even if the first three items of `VERSION_TUPLE` are integers.

jamalex: Good to note. And yeah, things like '0.14.dev20150727211121' will break syncing. Perhaps we should update the `version_diff` code to handle such things in a sensible way? Or switch to a system that ensures the first three components are integral?

benjaoming:
okay, moving to 0.15

benjaoming:
I've been finding it tedious that every time I do dev/alpha/beta releases, I have to manually bump versions... if syncing is broken by this, then syncing is going to be broken whether the version info is appended manually or by `get_version`. Making a "0.14" release without making note of its alpha state is a no-go because PyPi, PPAs etc. will never allow you to overwrite releases with identical versions, and pip and apt won't upgrade them automatically. Maintaining a sleek release process is the basis of distributing nice tests.

That's why I wanted this to be implemented ASAP.... but I guess we can wait until 0.15, I'll keep bumping versions manually.

Issue #4176, opened 2015-07-28
Page not found when clicking on Sync to central server.


djallado:
**Branch:** 0.14.x

**Expected Behavior:** Will not error when clicking on `device management page` button.

**Current Behavior:** Page not found error when clicking `device management page` button.

**Steps to reproduce:** 

1. Login as `admin`.
2. Click on `Register device` link in `Available Actions` column.
3. Click on `device management page` in info message.
4. Click on `Sync to central server!`.
5. Click on `device management page` in info message.

Error in the screenshot below:
![after-fix-the-4174](https://cloud.githubusercontent.com/assets/8664071/8925270/fcd19556-3538-11e5-838a-f063d55dbbc1.png)

rtibbles:
This is a duplicate of #4101.

Issue #4101, opened 2015-07-15
404 from 'device management page' link while syncing to central server 


radinamatic:
**Branch**: develop  (continued from #4100) 

**Current Behavior**: While on Device management page (/management/zone/...), pressing the `Sync to central server!` button opens the  `Successfully launched data syncing job. After syncing completes, visit the device management page to view results.` info message.  

![404-device0](https://cloud.githubusercontent.com/assets/1457929/8688615/9442716c-2a9e-11e5-8555-88f4ea2b0740.png)  

A part from being redundant/self-referencing (user is already on the **Device management**), the link inside info message leads to 404 page:

![404-device](https://cloud.githubusercontent.com/assets/1457929/8688644/f3b749d8-2a9e-11e5-8343-63242eb9626e.png)  

**Expected Behavior**:  I'm not sure what is supposed to happen here, maybe simply:  

 `Successfully launched data syncing job. See **Sync Sessions** below for details.` 

**Steps to reproduce**:

1. Log in as Admin.
2. Go to Manage > Facilities tab.  
3. Select the device.
4. On **Device management** page press the `Sync to central server!` button.
5. Press the ` device management page` link inside the info message.

aronasorman:
Working for me on the latest 0.14.x.

radinamatic:
I'm really sorry to be the party-breaker, but no... :(

**Branch**: Windows installer @MCGallaspy complied yesterday (so technically 0.14.x) on Windows XP.

**Steps to reproduce**:  

1. Login as Admin
2. Go to `Manage > Facilities`
3. Press the `Sync Now!` button, and you will get the `Successfully launched data syncing job. After syncing completes, visit the device management page to view results.` message. 

![sync1](https://cloud.githubusercontent.com/assets/1457929/9141278/bc248c78-3d37-11e5-8045-3897beeafb66.png)  

4. Press the `device management page` link in the above alert message, to get to the sync results:

![sync2](https://cloud.githubusercontent.com/assets/1457929/9141338/3a6e75da-3d38-11e5-8e2c-daca03c02156.png)

Up to here, all is correct!  

5. But if you press the `Sync to central server button` again (from the above step):

![sync3](https://cloud.githubusercontent.com/assets/1457929/9141385/89f46d94-3d38-11e5-8783-a3e91e8d00ee.png)

6. Once you get the same alert message as before and you press the link `device management page` again, you'll get:

[...]

PR #4175, opened 2015-07-28
Fix no module named yaml

arceduardvincent:
Hi @aronasorman and @djallado this fixed #4174 

I add the PyYAML 3.11 in production requirements to fix the no module named yaml.
After I installed the PyYAML 3.11 in my virtualenv. The problem was fixed see the image below.
screenshot 2015-07-28 14 43 25

PR #4177, opened 2015-07-28
Fixed the issue 4176 Page not found when clicking on device management page.

djallado:
Hi! @aronasorman. This fixes the issue #4176.

After the fixed by this PR #4175 by @arceduardvincent , another issue is #4176
This will fix the issue of page not found error when clicking on `device management page`.


rtibbles:
Fixes #4101

PR #4042, opened 2015-07-06
Inline

tk21:
Start inline help by pressing "SHOW ME HOW!" button located in top right corner of pages with narratives. Pressing button will trigger lightboxed tooltip modals to show up, with the text being pulled from narrative.yaml.

screen shot 2015-07-06 at 12 32 46 am
screen shot 2015-07-06 at 12 32 48 am
screen shot 2015-07-06 at 12 32 49 am

Still need to work on triggering actions between steps.

PR #4187, opened 2015-07-30
Inline help feature

MCGallaspy:
Replaces #4042, resolving the merge conflicts.


MCGallaspy:
oops. weird commit msg, because I carelessly used `git commit -F`. Don't pull, going to rewrite history to change the message.

PR #4188, opened 2015-07-30
Inline help feature

MCGallaspy:
Replaces #4042, resolving the merge conflicts.


data/narratives.yml
```diff
@@ -1,23 +1,57 @@
...
+  - 'li.video.active':
+    - step: 1
+    - text: 'Selecting this "Videos" tab will lead you to the place where you can download new videos for all topics!'
+  - '#content_tree':
+    - step: 2
+    - text: 'Downloadable content will be organized in this topic tree.'
+  - 'span.dynatree-node.dynatree-folder.dynatree-has-children.unstarted.dynatree-exp-c.dynatree-ico-cf':
+    - step: 3
+    - text: 'Simply toggle the "+" button to view see more subtopics under a topic....'
+    - before-showing: 'click'
```
MCGallaspy: This doesn't conform to the [specification](https://docs.google.com/document/d/11wY1EViZrCKXZkus-7Pf4RMCbEYuaTNSCRa5YCglMMU/edit).

kalite/inline/static/js/inline/views.js
```diff
@@ -24,7 +24,8 @@ var ButtonView = BaseView.extend({
                 var narr = self.model.attributes;
 
                 //translate narrative into build options for introjs
-                var options = self.parseNarrative(narr);
+                var before_showing = [];
+                var options = self.parseNarrative(narr, before_showing);
```
MCGallaspy: Instead of passing in an object as a parameter and modifying it, I think it's preferable to include it in the returned object. Something like
```js
var options = self.parseNarrative(narr);
var before_showing = options.before_showing;
```

kalite/inline/static/js/inline/views.js
```diff
@@ -66,18 +99,25 @@ var ButtonView = BaseView.extend({
...
+                else if (key === "step") {
+                    newkey = "step";
                 }
                 else if (key === "before-showing") {
-                    //TO DO: user actions to taken before showing modal
+                    //add the before-showing attribute to a separate
+                    //"before-showing" array to be used in intro.onchange()
+                    console.log("inside before-showing building");
+                    var before_showing_obj = {};
+                    before_showing_obj["element"] = selectorKey;
```
MCGallaspy: Again, this and the logic that depends on it don't conform to [the spec](https://docs.google.com/document/d/11wY1EViZrCKXZkus-7Pf4RMCbEYuaTNSCRa5YCglMMU/edit).

MCGallaspy:
@aronasorman can you review and merge this after the 0.15 branch split?

MCGallaspy:
TODO: Add some documentation about this feature.

docs/usermanual/userman_coach.rst
```diff
@@ -56,3 +56,10 @@ Detailed View
 1. Click on the colored cells.
 2. A panel should open up below the cell to reveal the number of questions attempted and the actions made on each exercise. 
 3. Click on the same cell to close the panel. 
+
+Inline Help
+-----------
+
+KA Lite has an inline help system on some pages.
+On parts of the site where it is available, you can find a "Show me how!" button.
```
MCGallaspy: Need to elaborate?

docs/usermanual/userman_coach.rst
```diff
@@ -56,3 +56,10 @@ Detailed View
...
+If you click this button, you can view a guided tour of the currently shown interface.
```
MCGallaspy: Need to elaborate?

MCGallaspy:
Ahhh, test failures!

aronasorman:
Is this super critical for 0.15 @MCGallaspy? If not, let's punt this to 0.16.

MCGallaspy:
No, but like we discussed, I'd like to merge this in to develop directly after we branch off 0.15.x. Will look at the test failures, too, but given they once passed and only started failing after I changed the documentation, I feel somewhat confident about it.

aronasorman:
Sounds good. I moved it to 0.16 then.

MCGallaspy:
Something is amiss!

MCGallaspy:
Can we now merge this in to develop?

MCGallaspy:
@aronasorman I had assigned this to you in the distant past.

MCGallaspy:
Notice me!
image

aronasorman:
image

I'll take a look later today before I get home.

data/narratives.yml
```diff
@@ -1,23 +1,58 @@
 management/zone/.*:
-  - 'a#manage.admin-only':
+  - li.manage-tab.active:
     - step: 1
-    - text: 'Welcome! This is the landing page for admins. If at any point you would like to navigate back to this page, click on this tab!'
-  - 'li.facility':
+    - text: Welcome! This is the landing page for admins. If at any point you would like to navigate back to this page, click on this tab!
```
aronasorman: Have you looked into how we can get this translated?

kalite/inline/features/steps/inline-help.py
```diff
@@ -29,15 +33,18 @@ def step_impl(context):
...
+        actual_num_text = number_el.text
+        return int(actual_num_text) == expected_num if actual_num_text else False
+
+    WebDriverWait(context.browser, 30, ignored_exceptions=NoSuchElementException).until(condition)
 
 
 @then("an element is highlighted")
 def step_impl(context):
     elem = find_css_class_with_wait(context, HIGHLIGHTED_CLASS)
-    assert elem is not None, "Couldn't find a highlighted element on the page."
```
aronasorman: why were the asserts for this deleted?

aronasorman:
In the spirit of nitpicking, I get a slightly misaligned number in the red close button:
screen shot 2015-11-05 at 08 45 29

aronasorman:
My other main concern is internationalization. If we're not ready to have non-english text here, then let's have a conditional that will hide the "Show me how!" button when our language != "en".

data/narratives.yml
MCGallaspy: Yeah, we can wrap it with gettext here: https://github.com/MCGallaspy/ka-lite/blob/inline/kalite/inline/static/js/inline/views.js#L109

kalite/inline/features/steps/inline-help.py
MCGallaspy: `find_css_class_with_wait` will raise a TimeoutError already. The assertion is redundant.

MCGallaspy:
We can wrap the text with gettext on the front end easily enough, but how can we include the messages into our strings to be translated? Will we have to add them manually?

aronasorman:
We're gonna have to make our own way of parsing yml files, much like what we did for handlebars templates until recently (when `makemessages` supported them recently). You can pass that off to me, but I might not get to that until the code freeze. So I recommend adding the conditional for now.

MCGallaspy:
> We're gonna have to make our own way of parsing yml files
I don't think that's so hard -- we just have to grab the `text` key. Is this handled by the central server in [the update_pot command](https://github.com/fle-internal/ka-lite-central/blob/master/centralserver/i18n/management/commands/update_pot.py)? If not, can you point me in the right direction?

aronasorman:
You'll want to edit KA Lite's `makemessages` command: https://github.com/learningequality/ka-lite/blob/master/python-packages/django/core/management/commands/makemessages.py#L156-L170

Add another conditional. If we're processing the `djangojs` domain and looking at a yml file, then do something in there that will output a file in a programming language that's parseable by `xgettext`. Here's a list of supported languages:

```
C, C++, ObjectiveC, PO, Shell, Python, Lisp, EmacsLisp, librep, Scheme, Smalltalk, Java, JavaProperties, C#, awk, YCP, Tcl, Perl, PHP, GCC-source, NXStringTable, RST, Glade, Lua, JavaScript, Vala, GSettings, Desktop.
```

Django 1.5 before I edited the file, had to convert JS files into C files, and then make xgettext scan them.

aronasorman:
Here's the exact change I made to make things simpler: https://github.com/learningequality/ka-lite/commit/88b63f2ad168f94f57dd8139de20a5013ea5bcd0

MCGallaspy:
I'm opposed to editing Django files, this prevents easy upgrading. I'd like to revert makemessages to django 1.5, then write a script (like another mgmt command) that modifies the *output* of vanilla makemessages. Seem reasonable?

MCGallaspy:
If you agree to that, I'd like to first open separate PR to achieve equivalent results to develop, then rebase this PR onto that and introduce any further changes I need to make specific to inline help.

aronasorman:
That's a good idea, actually. The main blocker if you're reverting is reintegrating handlebars parsing. This was hacky before, but perhaps now we can just delegate to nodejs for these kinds of things: https://www.npmjs.com/package/handlebars-xgettext

But we are going lose the current translations for the handlebars strings if we don't do this correctly. Rather than revert now for 0.16, I'm thinking we should keep the current makemessages for now, then add the new po files from the script you're going to create. We can then remove the handlebars from djangojs.po in the next version.

What do you think?

MCGallaspy:
Okay, I'll create a script that runs makemessages and then handles adding the inline help messages to the po file.

Issue #4172, opened 2015-07-28
Successfully add coach but it returns an error.


djallado:
**Branch:** 0.14.x

**Expected Behavior:** Successfully  add coach without error.  

**Current Behavior:** Successfully add coach but it returns an error.

**Steps to reproduce:**

1. Log in as `admin`.
2. Click on `Default Facility`.
3. Click on `Add a new coach`.
4. Put coach info in every textboxes and click on `Create user` button.

Error in the screenshot below:
![add-coach](https://cloud.githubusercontent.com/assets/8664071/8923467/b833cc1c-3523-11e5-9786-b54b2f5495fe.png)

PR #4192, opened 2015-07-31
Fix the issue 4172 - Successfully add coach but it returns an error.

djallado:
Hi! @aronasorman. This fixes the issue #4172.

I just removed the `getattr` attribute of `get_zone()` of `facility/views.py`

### DIFF
--- a/kalite/facility/views.py
+++ b/kalite/facility/views.py
@@ -123,7 +123,7 @@ def _facility_user(request, facility, title, is_teacher=False, new_user=False, u
                 if request.next:
                     return HttpResponseRedirect(next)
                 else:
-                    zone_id = facility.get_zone().id
+                    zone_id = getattr(facility.get_zone(), "id", None)
                     return HttpResponseRedirect(reverse("facility_management", kwargs={"zone_id": zone_id, "facility_id": facility.id}))
 
             # New student signed up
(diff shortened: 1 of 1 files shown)


kalite/facility/views.py
```diff
@@ -123,7 +123,7 @@ def _facility_user(request, facility, title, is_teacher=False, new_user=False, u
                 if request.next:
                     return HttpResponseRedirect(next)
                 else:
-                    zone_id = facility.get_zone().getattr(id, None)
+                    zone_id = facility.get_zone().id
```
rtibbles: Can you change this to this:
```
getattr(facility.get_zone(), "id", None)
```

This should guarantee a result even if get_zone returns None.

djallado:
@rtibbles. 

I tried to used this code: `getattr(facility.get_zone(), "id", None)` but it raises an error:
screen shot 2015-08-04 at 6 19 18 pm

cpauya:
@djallado You're ka-lite server may still be using the old code.  That's the manifestation of a method call for `getattr` - which I'm not sure is valid.  You must use the function call for `getattr` like @rtibbles suggested.

Another point is that the code `zone_id = facility.get_zone().getattr(id, None)` is using the [built-in Python function `id`](https://docs.python.org/2/library/functions.html#id), so that merits change.

One last note, there's a `next` variable declared at line 91.  Can you re-use that for the codes that use `if request.next...` on lines 117 and 123?  Is that safe @rtibbles - unless there's a compelling reason for using `request.next` there?

PR #4194, opened 2015-07-31
Create js bundle for register_public_key_client.html

### LINKED ISSUE #4156: Remove inline js in securesync module
In `develop`:

See here: https://github.com/learningequality/ka-lite/blob/develop/python-packages/securesync/templates/securesync/register_public_key_client.html#L22

Causes an error since dependent js files are missing. I guess the recommended thing to do here is create a bundle which includes both the inline script and its dependencies, right @rtibbles?

MCGallaspy:
The js module has the same name. Lives in the `distributed` app. Fixes #4156.

### DIFF
--- a/kalite/distributed/static/js/distributed/bundle_modules/register_public_key_client.js
+++ b/kalite/distributed/static/js/distributed/bundle_modules/register_public_key_client.js
@@ -1,0 +1,29 @@
+var $ = require("../base/jQuery");
+var messages = require("../utils/messages");
+var api = require("../utils/api");
+
+function reload() {
+    window.location.reload();
+}
+
+function auto_register() {
+    // Disable and show message
+    $("#one-click-register").toggleClass("disabled", true);
+    messages.show_message("info", "{% trans 'Contacting central server to register; page will reload upon success.' %}");
+
+    // window.auto_registration_url is defined by a template context variable.
+    api.doRequest(window.auto_registration_url, null, {dataType: "jsonp"})
+        .success(function() {
+            messages.show_message("success", "Auto-registered.");
+            window.location.reload();
+        })
+        .fail(function(a, b) {
+            // Re-enable
+            $("#one-click-register").toggleClass("disabled", false);
+        });
+}
+
+$(function() {
+    $(".refresh-link").click(reload);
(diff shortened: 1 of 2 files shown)


jamalex:
Question for @rtibbles: in cases like this, would it not be nicer to have
the js module in the same app as the template etc that uses it? Otherwise,
it kind of feels like we're spreading stuff around in weird ways...

rtibbles:
I think in general, yes - @jamalex. I think the only motivation for separating them at this point is to get a minimal fix that works into develop, and then we can open an issue for migrating securesync back to its final resting place?

MCGallaspy:
If it's worth doing, it's probably worth doing now. I'll add some commits to this PR to move securesync into kalite.

rtibbles:
Woohoo!

jamalex:
Cool! And then the js bundle can live directly inside the kalite.securesync app?

Out of curiosity, was there a technical, or just aesthetic, reason that it couldn't have been put inside securesync while in python-packages?

rtibbles:
I didn't build the build script to look in there, either I had to change the build script, or this get changed.

I prefer this way.

jamalex:
Oh, hmm... doesn't it look inside any app in `INSTALLED_APPS`? (like templates, models, etc?) There's example code for auto-finding the paths for these apps. If it did that, we could use the idiom inside pluggable apps as well.

rtibbles:
As implemented, it does not.

The front end build process can definitely be improved and made more general, but this is how it is for now. Happy to look at a redesign for 0.16.

jamalex:
Sounds good! https://github.com/learningequality/ka-lite/issues/4195

kalite/distributed/static/js/distributed/bundle_modules/register_public_key_client.js
```diff
@@ -0,0 +1,29 @@
+var $ = require("../base/jQuery");
```
rtibbles: Not necessary to change, but worth noting that it is possible to use non-relative imports here as well (the build script collects all django app static/js/<appname> paths and makes them available).
```
var $ = require("base/jQuery");
```

PR #4196, opened 2015-08-01
Time picker for Coach Reports

rtibbles:
Fixes https://github.com/fle-internal/ka-lite-central/issues/269

Summary of changes:
* Implements a time picker on the coach report landing page
* Allows for the setting of arbitrary date ranges for summary view and tabular report.
* Adds a dynamic setting to allow different default time windows.

@jamalex


jamalex:
This looks good to me. Let's get it merged in and get some testing on it.

PR #4197, opened 2015-08-01
Add more explicit wait in test

MCGallaspy:
Does what it says. Also add something to .gitignore that I've been meaning to do for a while.

### DIFF
--- a/kalite/distributed/static/js/distributed/contentrec/hbtemplates/content-nextsteps-lesson.handlebars
+++ b/kalite/distributed/static/js/distributed/contentrec/hbtemplates/content-nextsteps-lesson.handlebars
@@ -21,7 +21,7 @@
     </a>
     <!--The URL without the specific video, so that we see the path
     along the nav and show them the entire series-->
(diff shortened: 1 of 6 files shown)


kalite/distributed/static/js/distributed/contentrec/hbtemplates/content-nextsteps-lesson.handlebars
```diff
@@ -1,6 +1,6 @@
 
 <div class="row" class="content-nextsteps-lesson">
-	<a href="/learn/{{ path }}">
+	<a href="/learn/{{ path }}" id="content-nextsteps-lesson-link">
```
rtibbles: This is a repeated element in the page, we shouldn't use an ```id``` attribute here - can you change it to a class?

MCGallaspy:
@rtibbles added another commit. See commit message for explanation.

rtibbles:
Looks good, let's see what the Circle master says!

rtibbles:
No test failures I would not expect. Merging.

PR #4217, opened 2015-08-08
  Better version info

benjaoming:
Reopening #4169 for 0.15.


benjaoming:
@jamalex perhaps since you have the overview of the syncing process, you can open up related issues regarding fixing syncing? The issues have persisted ever since I started releasing stuff with "alpha" in the version string, and possibly also in previous versions if it was the case they also had alphas and betas.

jamalex:
`get_version` is quite variable in what it returns (it follows "rules", but not simple ones):

`get_version((0, 15, 0, "alpha", 0))` -> '0.15.dev20150805164928' (if run from within a git repo)
`get_version((0, 15, 0, "alpha", 0))` -> '0.15' (if **not** run from within a git repo)

`get_version((0, 15, 1, "alpha", 0))` -> '0.15.1.dev20150805164928' (if run from within a git repo)
`get_version((0, 15, 1, "alpha", 0))` -> '0.15.1' (if **not** run from within a git repo)

`get_version((0, 15, 0, "alpha", 1))` -> '0.15a1'

`get_version((0, 15, 0, "beta", 0))` -> '0.15b0'

`get_version((0, 15, 1, "beta", 0))` -> '0.15.1b0'

(sometimes 2 components, sometimes 3, sometimes 4; also, it varies depending on context -- inside a git repo or not -- which will lead to strange situations with "non-source" installed code reporting itself as a final release)

Main problem: We don't have any good way to compare these version numbers (which is what is needed for syncing). The two (current) built-in methods for version comparison don't handle them properly:

`StrictVersion('0.15.dev20150805164928')` -> `ValueError: invalid version number '0.15.dev20150805164928'`

`LooseVersion('0.15.dev20150805164928') < LooseVersion('0.15.4')` -> `False`
`LooseVersion('0.15.4b4') < LooseVersion('0.15b1')` -> `True`

If we didn't use the `.dev234234234` part (i.e. we didn't have "alpha 0" releases), then StrictVersion would work fine, and perform accurate comparisons. But that defeats @benjaoming's goal of not needing to bump version numbers.

Any suggestions? I don't think it would be wise to merge this yet as-is.

benjaoming:
The good news, @jamalex is that this is based on a tuple that's easier to compare than importing StrictVersion and LooseVersion.

I.e. `(0, 15, 0, "alpha", 0)` < (0, 15, 0, "alpha", 1) == True` etc.

If we can change version comparisons to this behaviour, that'd be sufficient.

In case we need to store versions as a string via `get_version` and parse them back into a tuple, it would mean to write an inverse of `get_version`... but that shouldn't be too hard.

jamalex:
Version comparison happens over the wire (comparing local version with the version of the other device we're communicating with), and since we need to be backwards compatible with older versions (which use the "N.N.N" format), we can't actually switch up the version comparison approach much.

The simplest (and I'm now thinking best) solution would be for us not to try to use the same version string for both securesync version comparison and installer versioning. Could we add a "FULLVERSION" variable or something for your purposes, and leave "VERSION" as it was (major.minor.patch)?

Even then, I think having a version format (for installers) that's more readable/consistent will make it simpler for users to understand what version they're dealing with. I don't like the way `get_version` strips off components of the version that are 0, etc. Is that format somehow required for Launchpad etc to handle things correctly? Otherwise, it might be nicer to use something like "0.15.0" instead of "0.15", "0.15.0-alpha1" instead of "0.15a1", "0.15.0-dev20150805164928" instead of "0.15.dev20150805164928", etc.

Also, any suggestions for the issue of the "dev20150805164928" component only being possible to calculate when inside a git repo? Should we add a step to the build process that writes the git HEAD to a file that can then be read?

benjaoming:
I didn't mean to replace VERSION, that's not in this PR... it's just VERSION_TUPLE being introduced with a preference of making internal comparisons of versions align with that tuple rather than relying on string interpretation.

We've already made ourselves guilty of making pre-releases with uncontrolled versions strings, so making it an explicit behaviour would be true to the need of making lots of pre-releases (the more, the merrier testing-wise).

> I don't like the way get_version strips off components of the version that are 0, etc. Is that format somehow required for Launchpad etc to handle things correctly?

I think it's Launchpad + [Pep 440](https://www.python.org/dev/peps/pep-0440/). I see that former versions have done 0.13.0, but then again, that leaves out doing a 0.13.0.1. The point is that you can make another .point release beyond 0.13.0 -- not saying that it's super essential, because it's also allowed to do 0.13.post1 or 0.13.1.post1.. but yes those standards tend to dis-favour the redundant zero.

> Even then, I think having a version format (for installers) that's more readable/consistent will make it simpler for users to understand what version they're dealing with.

That's the aim of aligning with other popular version formats.. that we just don't bother with inventing the wheel but accept that other people had the discussion before us and take their word :) I think you might be making lots of sense, but at the same time I don't know exactly if it's beyond our scope to actually get too deep into this... my goal was really honestly to just implement a known method and align the rest of our logic.. and so if there's no actual blockers, I think we should do it.

> Otherwise, it might be nicer to use something like "0.15.0" instead of "0.15", "0.15.0-alpha1" instead of "0.15a1", "0.15.0-dev20150805164928" instead of "0.15.dev20150805164928", etc.

I did 12 alpha releases. They *need* to be versioned for said reasons. The point is to get a rolling, and fully automated release process independent of the need for odd pre-releases.

>  Also, any suggestions for the issue of the "dev20150805164928" component only being possible to calculate when inside a git repo?

It's date based. It's not for public releases, but for nightly builds.. we simply constrain ourselves to the date stamp with no mention of the git commit hash. Did you mean something else?

> Should we add a step to the build process that writes the git HEAD to a file that can then be read?

Hmmm, no, I think that releasing something and making an amend is possible. The main point is that it's a non-announced release for testing purposes, and we don't intend to make several of them or tag them. But we could in theory use git tags for important betas that we ask a relatively broader public to use.

jamalex:
> I didn't mean to replace VERSION, that's not in this PR

Hmm, but it does seem to be in this PR? It deletes `VERSION = "%s.%s.%s" % (MAJOR_VERSION, MINOR_VERSION, PATCH_VERSION)` and adds `VERSION = get_version(VERSION_TUPLE)`.

> We've already made ourselves guilty of making pre-releases with uncontrolled versions strings, so making it an explicit behaviour would be true to the need of making lots of pre-releases (the more, the merrier testing-wise).

Yes, definitely agree with having lots of prereleases for testing. Just want to make sure it's done in a clean/predictable way.

> that leaves out doing a 0.13.0.1

We're following http://semver.org/, which means if we make a change to 0.13.0, we release it as a patch release, 0.13.1 -- so, I don't think we would need that extra digit (it would just be confusing -- when would we use that vs bumping the patch version?)

> accept that other people had the discussion before us and take their word

Yes, that's why we settled on http://semver.org/, as it is well-defined, interpretable, and very popular, and has been serving us well. There are great Python libraries (e.g. [this](https://github.com/k-bx/python-semver)) for dealing with these versions, which handle alpha/beta/rc/dev components etc.

> I did 12 alpha releases. They need to be versioned for said reasons. The point is to get a rolling, and fully automated release process independent of the need for odd pre-releases.

Yes, I agree (and don't think I said anything that contradicts that). I was just suggesting that we not drop `0` components -- basically, that we follow SemVer closely (and now, I'm thinking, that we just go full-hog with SemVer).

> It's date based. It's not for public releases, but for nightly builds.. we simply constrain ourselves to the date stamp with no mention of the git commit hash. Did you mean something else?

[`get_version` calls `get_git_changeset`, which gets the timestamp of the last git commit](https://github.com/django/django/blob/1.5.12/django/utils/version.py#L43). If it's run from a non-git directory, it returns `None`, and `get_version` then just skips adding the `.devXX` suffix. This means that our prerelease installer builds will report themselves (internally) as final versions (as there's no git dir). Try it! :)

That said, I'm all for the idea of having auto-generated `devXX` releases, I'm just saying that calculating `VERSION` on the fly within the distributed codebase using `get_version` won't work. I would propose that we write our own `get_version` function that uses the [python-semver](https://github.com/k-bx/python-semver) lib, and calculates its own `devXX` suffices in a better way.

> > Should we add a step to the build process that writes the git HEAD to a file that can then be read?

> Hmmm, no, I think that releasing something and making an amend is possible. The main point is that it's a non-announced release for testing purposes, and we don't intend to make several of them or tag them. But we could in theory use git tags for important betas that we ask a relatively broader public to use.

I think I probably didn't explain well enough what I meant. I was trying to address the problem of `get_version` not returning `devXX` suffices outside of a git repo. I was saying we could write a special file as part of the build process (at the point the codebase is pulled out of its git context) so that a predictable `devXX` suffix could be calculated at runtime.

benjaoming:
> get_version calls get_git_changeset, which gets the timestamp of the last git commit. If it's run from a non-git directory, it returns None, and get_version then just skips adding the .devXX suffix. This means that our prerelease installer builds will report themselves (internally) as final versions (as there's no git dir). Try it! :)

Oh! Trying it now, doesn't look pretty.... pretty useless if anything.. unless of course having the internal string rep. be "0.15" for somehow emulating the final release.

Okay, let's just ditch it then. But I really liked the tuples though, as they made it much easier and without any need for parsing when doing version comparisons.

I'm with the proposal to add a semver dependency... it's like a 3.3 kB file. But I suggest sending it through the dev room and then opening up a separate issue about adding it.

jamalex:
Cool, thanks for opening that issue! I think that should work well.

> Oh! Trying it now, doesn't look pretty.... pretty useless if anything.. unless of course having the internal string rep. be "0.15" for somehow emulating the final release.

I think having the internal version number include the pre-release info would be ideal, since it will help with bug reporting during prerelease stages (especially now, with the output of `kalite diagnose`!)

> But I really liked the tuples though, as they made it much easier and without any need for parsing when doing version comparisons.

Tuples are nice in that they're natively comparable. Can you clarify the use case, though? The only cases in which I know we're doing version comparison is over the wire (via the API for syncing), and so there would still be serialization/deserialization of the tuple in any case. And, since we need to be backwards compatible with previous versions that send their version as a `.`-delimited semver string, we would now just need to support both formats.

benjaoming:
Nah, I can't actually think of something because kalite is just one big codebase so everything is always the same version and doesn't have to detect it :) But if we want to decouple stuff, it's nice for another component to be able to do stuff like...

    from kalite import VERSION_TUPLE
    if VERSION_TUPLE > (0,14):
        print("Now we know that we can call this new command")

...because our API is somewhat changing. For instance if say we choose to run the central server as a truly decoupled application.

jamalex:
Ah, makes sense, yes! So, down the road, once we get to the point of a big happy plugin family!

PR #4222, opened 2015-08-10
Remove out-dated setup scripts (no mentions found in docs or codebase)

benjaoming:
I don't think we should continue to encourage these installation procedures anymore as they are sub-optimal and untested... worst case broken.

On Linux: They won't generate start up scripts.

On Windows and Mac: Not sure if they work anymore.

@jamalex sorry about un-assigning you, seems you're busy, assigning @aronasorman for this and inviting others for inputs.

## Fallback

If someone wants to run KA Lite from a Git checkout, *we still support this* but not with system start up scripts, user interfaces etc.

I would suggest that people having installed directly from git just run `bin/kalite manage setup`. The server will work, their data will be intact.

## Risks

There are none IMO. If people already used a setup script, it will have installed the system startup service already. Having already introduced the kalite command, this system startup service should still work after updating.

The new kalite command automatically updates the database when needed.


benjaoming:
@EdDixon you know anybody on the setup scripts?

@j-schwartz this is the removal I was talking about in learningequality/installers#113. Is there any specifically binding reason or would you be able to either use `kalite manage setup` or install from the .deb version?

jamalex:
I know the SMILE folks were using the setup shell scripts, and previously got thrown off by `setup_linux.sh` being renamed to `setup_unix.sh`. But a lot of things have changed, including not downloading via Git, and hence people will need to adjust their procedures anyway, so now's a good time to remove those scripts. Let's just make sure this is explicitly noted in the release notes.

benjaoming:
Good idea -- talkning about release notes @aronasorman can you make sure this file's contents are moved and the file deleted?

https://github.com/learningequality/ka-lite/blob/master/DEPRECATIONS_0.13.rst

PR #4225, opened 2015-08-11
0.14b2

benjaoming:
**WARNING!** Merging this, merges and closes all PRs contained in this PR, and they may have a separate review process and amendments following. So don't merge this PR!! :)

 * [x] [ka-lite on pypi](https://pypi.python.org/pypi/ka-lite)
 * [x] [ka-lite-static on pypi](https://pypi.python.org/pypi/ka-lite-static)
 * [x] [ka-lite on Launchpad PPA](https://launchpad.net/~learningequality/+archive/ubuntu/ka-lite)
 * [ ] OSX installer
 * [x] Windows installer

## Known issues

* [ ] #4226 Some translated videos do not match their title/description
* [ ] #4232 Doesn't work with DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP (Raspberry Pi)

Debian/Ubuntu:

* [ ] learningequality/installers#132 - Raspberry Pi: Bad ownership set for KALITE_HOME, server not starting
* [x] learningequality/installers#130 - Raspberry Pi: The nginx.conf file does not seem to be reliably installed on 15.04 and reinstalled after removal of ka-lite-raspberry-pi
* [x] learningequality/installers#116 - `cannot create /etc/ka-lite/username: Directory nonexistent` -- only happens on some Debian/Ubuntu (unknown which)
* [x] learningequality/installers#124 - Ubuntu 15.04 complains that /etc/init.d/ka-lite is not executable

## Included PRs

 * #4222

## Installer links

  * [Windows](http://field.learningequality.org/KALiteSetup-0.14.0.exe)
  * [Mac]()


benjaoming:
@MCGallaspy @cpauya please feel welcome to add links for new installer builds

benjaoming:
@MCGallaspy the link that @aronasorman posted for the Windows installer, `http://field.learningequality.org/KALiteSetup-0.14.0.exe` -- how is this built? Automatically on every commit pushed to 0.14.x?

aronasorman:
@benjaoming I think it's built when we ask Mike to build one, and he just
maintains the same link.

MCGallaspy:
> I think it's built when we ask Mike to build one, and he just
maintains the same link.

Yes. Would be nice to return to the automated build process -- something about unpacking the assessment zip was failing with Wine on OSX, but works ok in Windows, if I recall correctly.

benjaoming:
Could the Windows and OSX installers build from PyPi sources or are they already doing this? It would be nice if somehow the pre-release process was more sleek.

Many people can have access to releasing on PyPi, it's a matter of having an account there and optionally (but ideally) a PGP key. I've documented this in the "Release Workflow" document.

aronasorman:
@benjaoming I don't think they use the PyPi sources. Since the dungeon build server is meant to create a new build for each new merge into our develop branch I don't think using PyPi would be a good idea for dev builds.

However, the dungeon build server also has the concept of "promoting" a dev build into a release. So if we're ready for one we can theoretically promote those installers, as well as push the current commit they're based on into PyPi, such that all the installation methods are in sync.

benjaoming:
I find that we can save some efforts keeping testers and other stake-holders up to date with the state of our software by creating pre-releases and maintaining the structure that you see above... if you track back through previous betas and alphas, you'll see that the table was set but there wasn't really any guests :)

Releases them on PyPi as a premise is IMO a possible first step.. but really only relevant for pre-releases, we shouldn't put nightly builds there. So yeah, maybe it's better to have the build server create them and then promote them to releases.... BUT that leaves out a huge gap: I've had to pre-merge lots of PRs for the pre-releases to even be released. Practice, I think, shows us that it's possible to test PRs that're not yet merged in a pre-release and then continue working on them after. Either that, or we have to be more quick and liberal about merging PRs. Anyways, that's why I like pre-releasing this way by creating a special branch, bumping the version, and then throwing it out once the next pre-release is done.

benjaoming:
Oh and btw... this is done and beta 3 is out :) #4235

PR #4238, opened 2015-08-14
Log data migration

rtibbles:
Summary of changes:
* Data migration for latest_activity_timestamp to put in values where available
* Increases size of 'id' fields on log models to 200.

### DIFF
--- a/kalite/main/migrations/0039_log_timestamps.py
+++ b/kalite/main/migrations/0039_log_timestamps.py
@@ -1,0 +1,246 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """
+        Create latest_activity_timestamp data when possible by inspecting the model.
+        For ExerciseLog inspect relevant AttemptLogs to infer most recent activity.
+        If not available, and for other Logs, fallback to completion_timestamp if available.
+        """
+        for e in orm.ExerciseLog.objects.filter(latest_activity_timestamp=None):
+
+            alogs = orm.AttemptLog.objects.filter(user=e.user, exercise_id=e.exercise_id, context_type__in=["playlist", "exercise"])
+
+            if alogs:
+                e.latest_activity_timestamp = sorted([a.timestamp for a in alogs])[-1]
+                e.zone_fallback = e.get_zone()
+                e.save()
+            else:
+                e.latest_activity_timestamp = e.completion_timestamp
+                e.save()
+
+        for v in orm.VideoLog.objects.filter(latest_activity_timestamp=None):
+            v.latest_activity_timestamp = v.completion_timestamp
+            v.save()
+
+        for c in orm.ContentLog.objects.filter(latest_activity_timestamp=None):
+            c.latest_activity_timestamp = c.completion_timestamp
+            c.save()
+
+
+    def backwards(self, orm):
+        raise RuntimeError("Cannot reverse this migration.")
+
+    models = {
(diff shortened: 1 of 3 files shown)


kalite/main/migrations/0039_log_timestamps.py
```diff
@@ -0,0 +1,246 @@
...
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """
+        Create latest_activity_timestamp data when possible by inspecting the model.
+        For ExerciseLog inspect relevant AttemptLogs to infer most recent activity.
+        If not available, and for other Logs, fallback to completion_timestamp if available.
+        """
+        for e in orm.ExerciseLog.objects.filter(latest_activity_timestamp=None):
```
jamalex: Would be good to add a surrounding `try...except` around everything here, from my experience with weird missing users in running similar code on the ExerciseLogs on globe (with `e.user` not pointing to a valid user).

kalite/main/migrations/0039_log_timestamps.py
```diff
@@ -0,0 +1,246 @@
...
+
+            alogs = orm.AttemptLog.objects.filter(user=e.user, exercise_id=e.exercise_id, context_type__in=["playlist", "exercise"])
+
+            if alogs:
+                e.latest_activity_timestamp = sorted([a.timestamp for a in alogs])[-1]
+                e.zone_fallback = e.get_zone()
+                e.save()
+            else:
```
jamalex: Would be more efficient to have this be an `elif e.completion_timestamp:`, so that it doesn't save models unless needed.

kalite/main/migrations/0039_log_timestamps.py
```diff
@@ -0,0 +1,246 @@
...
+                e.latest_activity_timestamp = e.completion_timestamp
+                e.save()
+
+        for v in orm.VideoLog.objects.filter(latest_activity_timestamp=None):
```
jamalex: For this one (and the ContentLog), the filter could be `.filter(latest_activity_timestamp=None).exclude(completion_timestamp=None)`, to avoid unnecessary saves.

kalite/main/migrations/0039_log_timestamps.py
```diff
@@ -0,0 +1,246 @@
...
+            v.latest_activity_timestamp = v.completion_timestamp
+            v.save()
+
+        for c in orm.ContentLog.objects.filter(latest_activity_timestamp=None):
+            c.latest_activity_timestamp = c.completion_timestamp
+            c.save()
+
+
+    def backwards(self, orm):
+        raise RuntimeError("Cannot reverse this migration.")
```
jamalex: Rather than raising an error, I'd say just pass. The `latest_activity_timestamp` will be deleted in the previous migration, which will effectively bring it back to the previous state anyway.

jamalex:
Looks good, after suggested tweaks.

kalite/main/migrations/0039_log_timestamps.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        for e in orm.ExerciseLog.objects.filter(latest_activity_timestamp=None):
+
+            try:
+
+                alogs = orm.AttemptLog.objects.filter(user=e.user, exercise_id=e.exercise_id, context_type__in=["playlist", "exercise"])
+
+                if alogs:
+                    e.latest_activity_timestamp = sorted([a.timestamp for a in alogs])[-1]
+                    e.zone_fallback = e.get_zone()
```
jamalex: Oh, sorry, missed this in the first pass. This shouldn't be getting set here, if we're not running the overridden save method (so the zone stuff won't be getting affected, but changing this here might then break signatures).

PR #4239, opened 2015-08-14
Fix them blasted tests

rtibbles:
Tests pass.

Ready to merge. Let's do this so we can rely on our Circle tests.


jamalex:
I very much want to merge this, but am also least qualified to review it. So I won't. But I am excited.

radinamatic:
Way to go...!!! :+1: 

I don't recall seeing a CircleCI green light since..., well ever. :grinning:

kalite/control_panel/tests/control_panel.py
```diff
@@ -551,60 +557,60 @@ def test_device_log_csv_endpoint(self):
         self.assertEqual(len(rows), 2, "API response incorrect")
 
 
-class CSVExportBrowserTests(CSVExportTestSetup, BrowserActionMixins, CreateAdminMixin, KALiteBrowserTestCase):
+# class CSVExportBrowserTests(CSVExportTestSetup, BrowserActionMixins, CreateAdminMixin, KALiteBrowserTestCase):
```
MCGallaspy: I'm assuming you're removing these tests because they were giving too many false positives? Ideally, if we consider what they're _intended_ to test useful, then we should write new tests. Otherwise if they're not even useful in principle, then I'd just remove them instead of commenting them out.

kalite/distributed/features/exercise.feature
```diff
@@ -3,11 +3,6 @@ Feature: Exercises on the Learn page
     Internally we distinguish between Khan exercises and perseus exercises,
     But the user doesn't and we shouldn't here.
 
-    Scenario: Exercise is not available
```
MCGallaspy: Can you check the step implementation? I think this "feature" was intended as a proof of concept, so if it's not performing any useful checks then we can just remove this feature file entirely.

rtibbles: It does not do anything. Can delete.

kalite/control_panel/tests/control_panel.py
rtibbles: What they are intended to test is useful, but they give false positives on CircleCI.

They are commented out here as a placeholder for writing new tests, given that the current cost of having a failing build for a very small number of tests that pass locally is high, as we cannot rely on the tests at all.

kalite/facility/templates/facility/facility_user.html
```diff
@@ -74,7 +27,12 @@ <h3 class="no-margin-top">{{ title }}</h3>
                     {{ field.label_tag }}
                     {{ field }}
                     {{ field.errors }}
-                    <span class="helptext">{{ field.help_text }}</span>
+                    {% if field.id_for_label == "id_group" and request.is_admin %}
+                        {# Allow admins to add a new group #}
+                        <span class="helptext">(<a href="{% url 'add_group' %}?facility={{facility.id}}&next={{ request.get_full_path|urlencode }}">{% trans "Add a new group" %}</a>)</span>
```
MCGallaspy: I dunno why, but github is highlighting the `&` in a very ominous red color

rtibbles: That is odd, but I would guess it's just its syntax highlighting parsing it as html.

kalite/testing/behave_helpers.py
```diff
@@ -69,6 +78,17 @@ def assert_no_element_by_css_selector(context, elem, wait_time=MAX_PAGE_LOAD_TIM
     with context._runner.test_case.assertRaises(NoSuchElementException):
         context.browser.find_element_by_css_selector(elem)
 
+def assert_no_element_by_xpath_selector(context, elem, wait_time=MAX_PAGE_LOAD_TIME):
```
MCGallaspy: Please add a docstring.

kalite/facility/templates/facility/facility_user.html
MCGallaspy: I'm worried it's some sort of weird unicode character.

rtibbles: Searching the page for '&' highlights all instances of that character on the page, so if it is a weird unicode character, it always was.

kalite/control_panel/tests/control_panel.py
MCGallaspy: I get what you're saying, my concern is that the tests _won't_ be re-implemented, and then we'll just have a bunch of useless old code sitting around. I suggest one of:
1. Re-implementing these tests immediately. The obvious concern is that we have other equally as important/urgent stuff to do. (Do you have the time and inclination to do this?)
2. We set a date when someone will begin re-implementing them, so it doesn't languish.
3. We remove the tests entirely and don't give it a second thought until we actually encounter a bug, then we re-implement them. Obviously this is a gamble.

My contention is that just commenting them out is essentially the same as choosing 3.

rtibbles: I am happy to take on 2. Creating an issue for these tests with a 0.16 timeline on them should be sufficient for that.

MCGallaspy:
I'm happy to resolve the conflict and merge this when you're done, FYI.

kalite/control_panel/tests/control_panel.py
rtibbles: Created here: https://github.com/learningequality/ka-lite/issues/4249

rtibbles:
This should now be ready to merge, assuming that tests pass.

rtibbles:
Tests pass. Merging.

kalite/control_panel/tests/control_panel.py
benjaoming: > I get what you're saying, my concern is that the tests won't be re-implemented

That was spot on :)

I'm gonna get a status update on this, if it still fails, I remove the stale code.

PR #4240, opened 2015-08-14
Fix several a11y issues on homepage, navigation and user menu

radinamatic:
Eliminates a11y validation errors reported by various automatic tools:

- [x] Headings order
- [x] Context/Roles/Scope of navigation items
- [x] Some SR improvements

Replaces #3929 that had old merge conflicts. As suggested by @rtibbles, I collapsed the `Points` item with `My progress`: it remains as non clickable `span`, but it gets announced by the screen reader (I inverted the order of  `" %(points)d points "` in the `/user/views.js` file, so the phrase would sound more natural when announced. 

user-menu

User menu can be opened by both `Enter` and `Space` keys, and is navigable by both `Tab` and arrow keys. One pending issue would be the visible focus outline for the user toggle, but I believe we should resolve it by setting visual focus for keyboard users globally.


radinamatic:
Not sure why is this failing tests, seems to have problem with coachreports (surprise, surprise).

rtibbles:
Merge conflicts abound once more!

kalite/distributed/static/js/distributed/user/hbtemplates/navigation.handlebars
```diff
@@ -17,21 +17,19 @@
...
-  <li class="dropdown " id="user-name">
-    <a href="#" class="dropdown-toggle" data-toggle="dropdown" id="user-name-a">
+  <li class="dropdown " id="user-name" role="menuitem">
+    <a href="#" class="dropdown-toggle" data-toggle="dropdown" id="user-name-a" tabindex="0">
         <span id="username">{{!-- Username injected here w/ js --}}</span>
       <span class="caret hidden-xs"></span>
     </a>
 
-    <ul class="dropdown-menu" data-no-collapse="true" role="menu">
+    <ul class="dropdown-menu" data-no-collapse="true" aria-label="User menu" role="menu">
```
rtibbles: Should this be translated?

kalite/distributed/templates/distributed/base.html
```diff
@@ -92,7 +92,7 @@
                   <div class="row">
                       <div class="navbar-header">
                           {# collapse entire menu #}
-                          <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-collapse" aria-hidden="true" id="navbar_toggle">
+                          <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-collapse" title="Open menu" id="navbar_toggle">
```
rtibbles: Should this be translated?

rtibbles:
Couple of translation tags missing, methinks.

PR #4245, opened 2015-08-16
Calls remove on parent view first to remove from DOM to prevent multiple reflows.

rtibbles:
General fix for a specific problem, witnessed when Hiding the Tabular Report View.

By removing the parent view first, we reflow only once.


rtibbles:
Superseded by #4253

PR #4246, opened 2015-08-17
Correct docs folder name, and fix for a broken link and couple of minor build errors

radinamatic:
Tried to build docs locally, but was getting `'sphinx-build' command was not found` errors with sphinx installed via pip (Ubuntu 15.04). Only after I checked the official docs and installed with `apt-get install python-sphinx` did I manage to build. Added the command to `README.md`, fixed a broken `:ref:` link and a couple of missing line breaks in the `install_all.rst` file.


radinamatic:
Reversed the `apt-get` install option as not necessary.

PR #4248, opened 2015-08-17
0.15 - beautify content rating

MCGallaspy:
A WIP, but opened so that it can be continually reviewed. Screenshots incoming.


MCGallaspy:
ss

Looks like this ^

MCGallaspy:
@jtamiace now it looks like this
ss

@radina do you find sufficient contrast between activated (FFFF00) & un-activated (FFFFCC) color for stars, along with the background color (FFFFFF)?

radinamatic:
@MCGallaspy If we were talking about text, the contrast would be an epic fail, but we are not. However I'd definitely go towards an orange hue (you double the contrast compared to yellow), unless you want to experiment with red or other hues.

Another suggestion would be to leave un-activated stars white:

stars-contrast

MCGallaspy:
@radinamatic @jtamiace 
A star can be activated or not (a class indicating that it's been previously selected as a rating or not) and also be `active` or not `active` (a pseudoselector indicating that the user is clicking the element).

ss


From left to right on the Quality row: 

* activated, not `active`
* activated, `active`
* not activated, not `active`
* not activated, `active`

MCGallaspy:
Okay, this is more or less complete. It would be good to get a second pair of more design-savvy eyes to clean it up. In order this shows what the user will see when they navigate to a content item without a rating (or a deleted rating):

ss1

First, they'll be prompted to rate quality...

ss2

Then difficulty...

ss3

Then optional comments...

ss4

Then they can review, revise, or delete the rating.

Uses the same button colors from https://github.com/learningequality/ka-lite/pull/4248#issuecomment-132312692.

jamalex:
Would be good to save progress after every rating event, rather than waiting until final submission, as I can imagine people tapping the star ratings and then not completing the process. Should be as simple as a `this.model.save()` inside the rating callbacks.

jamalex:
I wonder whether it might be good to indicate what the range of the rating scale is -- e.g. put "Low quality" at one end of the row, and "High quality" at the other end. Might help novice users understand what the scale means.

jamalex:
This light yellow is what the hover event looks like, I'm guessing? And the darker color (with stars to the left also filled in) is when it's been clicked?
image

MCGallaspy:
> This light yellow is what the hover event looks like, I'm guessing? And the darker color (with stars to the left also filled in) is when it's been clicked?

No, the background color doesn't change on hover right now. As described above (https://github.com/learningequality/ka-lite/pull/4248#issuecomment-132312692) the yellow background appears when a user is clicking that particular element, but before they release the mouse button. Darker colors indicate the star is "activated" (a user previously clicked that or a higher-value star and set the rating).

> I wonder whether it might be good to indicate what the range of the rating scale is -- e.g. put "Low quality" at one end of the row, and "High quality" at the other end. Might help novice users understand what the scale means.

Good call. Lemme add that in.

> Would be good to save progress after every rating event, rather than waiting until final submission, as I can imagine people tapping the star ratings and then not completing the process. Should be as simple as a this.model.save() inside the rating callbacks.

Okay.

MCGallaspy:
Regarding styling, I'm hoping to meet with @jtamiace to come up with something more final on Thursday, but I will gladly incorporate any suggestions in the meantime.

MCGallaspy:
ss

Added a fixed-height label element. The text is set when a user hovers over a star rating option, and unset when they leave an option.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -146,11 +146,24 @@ var StarView = BaseView.extend({
...
+        "mouseenter .star-rating-option": "mouse_enter_callback",
+        "mouseleave .star-rating-option": "mouse_leave_callback"
+    },
+
+    label_values: {
+        1: "Very Low",
+        2: "Low",
+        3: "Normal",
+        4: "High",
+        5: "Very High"
```
MCGallaspy: Star rating labels are set here. This could easily be configurable on a per-view basis.

jamalex: @mcgallaspy did you see the comments on models.py in the previous, merged
PR?

MCGallaspy:
@jtamiace @jamalex 

* Now the "submit", "edit", and "skip" buttons have been removed -- text is always editable, and updates on a (throttled) continual basis.
* Rating elements grow out as the user interacts with them.
* Labels appear on either side of star ratings, in order to give context.

ss2
ss3
ss4

To-do is to indicate that ratings remain editable even after they are set.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -160,13 +148,31 @@ var StarView = BaseView.extend({
...
+        var template_options = {};
+        _.extend(template_options, this.model.attributes, {title: this.title, lowest_value: this.label_values[1], highest_value: this.label_values[5]});
+        this.$el.html(this.template(template_options));
         this.rating_change();
     },
 
-    rate_value_callback: function(ev) {
-        var val = $(ev.target).attr("data-val");
+    // Throttled so it's triggered only once when several children are the targets of the "click" event
+    rate_value_callback: _.throttle(function(ev) {
```
jamalex: I'd suggest debounce instead of throttle (doesn't need to be saving constantly during the typing -- adds extra load -- just when they pause, as that's when they may then go on to close it). And could be 2 secs instead of 100ms.

jamalex: (this is intended for both places -- I know the one here doesn't involve typing)

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -44,41 +56,20 @@ module.exports = BaseView.extend({
...
 
         this.listenToOnce(this.model, "change:quality", function(){
-            self.star_view_quality.remove();
-            self.star_view_difficulty = self.add_subview(StarView, {el: self.$("#star-container-difficulty"), model: self.model, rating_attr: "difficulty"});
+            self.star_view_difficulty = self.add_subview(StarView, {title: "Difficulty", el: self.$("#star-container-difficulty"), model: self.model, rating_attr: "difficulty", label_values: this.difficulty_label_values});
         });
 
-        this.listenToOnce(this.model, "change:difficulty", function(){
-            self.star_view_difficulty.remove();
-            self.text_view = self.add_subview(TextView, {el: self.$("#text-container"), model: self.model, text_attr: "text"});
```
jamalex: Could you clarify why `text_attr` needs to be configurable?

jamalex:
Looking good!

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -27,7 +25,21 @@ module.exports = BaseView.extend({
             model.url = "/api/not/gonna/work";
             return model;
         }();
-        _.bindAll(this, "delete_callback", "submit_rating", "edit_callback", "renderAll", "renderSequence", "render");
+        _.bindAll(this, "delete_callback", "renderAll", "renderSequence", "render");
+        this.quality_label_values = {
+            1:  "Poor quality",
```
aronasorman: Wrap these labels in a `gettext` call.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -88,32 +79,15 @@ module.exports = BaseView.extend({
                          2) The view's model is not fetched, and the user finishes filling out the new form.
         */
         this.$el.html(this.template());
-        // Explicitly hide/display desired buttons, since we re-render the template
-        this.$(".rating-skip").hide();
-        this.$(".rating-edit").hide();
         var views_and_opts = [
-            ["star_view_quality", StarView, {el: this.$("#star-container-quality"), model: this.model, rating_attr: "quality"}],
-            ["star_view_difficulty", StarView, {el: this.$("#star-container-difficulty"), model: this.model, rating_attr: "difficulty"}],
+            ["star_view_quality", StarView, {title: "Quality", el: this.$("#star-container-quality"), model: this.model, rating_attr: "quality", label_values: this.quality_label_values}],
```
aronasorman: Is "Quality" shown to the user? If so, please wrap in a `gettext` call. Same for "Difficulty".

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: This one is actually throttled so that clicking on a child element within the `.star-rating-option` container doesn't trigger multiple events (`trailing: false` option prevents it from being called at the end of the throttle period as well). Note the callback from the events hash:
```js
        "click .star-rating-option": "rate_value_callback",
        "click .star-rating-option >": "rate_value_callback",
```

MCGallaspy: Basically the intent is that this callback is triggered _at most_ once per 100ms.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -203,19 +210,8 @@ var TextView = BaseView.extend({
         return this;
     },
 
-    text_changed: function() {
+    text_changed: _.throttle(function() {
```
MCGallaspy: @jamalex I prefer `throttle` to `debounce` here -- delaying execution until _after_ input stops (especially for a long amount of time) leaves a window where the view/model could be changed (as with a page close or navigating to a different content item) before the model is synced.

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: I see that `debounce` has the `immediate` option, seems to fit the bill. Will change that.

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: Aye aye, cap'n.

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: It doesn't for the current design.

rtibbles:
One other thing that occurs is to only show the content rating when the content itself is available.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -203,19 +210,8 @@ var TextView = BaseView.extend({
...
-            val = !this.is_disabled();
-        }
-        this.$(".rating-text-feedback").attr("disabled", val);
-        return this;
-    },
-
-    is_disabled: function() {
-        return this.$(".rating-text-feedback").attr("disabled") === "disabled";
-    }
+    text_changed: _.throttle(function() {
```
jamalex: This was the particular one where debounce might be better. Then, it'll wait until they stop typing for a bit to save, rather than pounding the server constantly. I'd recommend debounce with 2000ms.

jamalex: (other debounce is good too, just more important on this one)

MCGallaspy: @jamalex 
Ok, I'll change it, but just to reiterate [what I said above](https://github.com/learningequality/ka-lite/pull/4248#discussion_r37657574): if someone navigates away from a content item in the 2s window after they stop typing, then when the debounced `save` is called it will not sync whatever text they have entered -- instead of creating a new view every time the user navigates to a different item, only the model object is switched out, so the debounced view method uses the new model attributes. It might be possible to instantiate a new view and keep a reference to the old one around until it's fired off any debounced functions queued up, or even to make this a model method and keep old models around in a similar way. Simply decreasing the debounce delay still leaves the potential for a race condition, but decreases the chance that a human will notice it. What do you think?

Just to make sure I tried it out, and it's not particularly challenging to cause the sync to fail as I've described with a 2s delay. Unlikely for a person to do IMO, but something which could be quite unexpected in automated contexts.

rtibbles: What is probably better to do, is not to debounce the set, but just debounce the save.

Then when the 'close' method is called on the contentareaview, you can call a non-debounced save to ensure no user data is lost.

This is what happens in point updating also.

MCGallaspy: Clever clever

rtibbles: The only place we might lose data then is if someone closes their browser window, but apart from preventing browser window closures until we have no 'dirty' models, there's not much we can do about that.

MCGallaspy:
@rtibbles 

>One other thing that occurs is to only show the content rating when the content itself is available.

There seems to be some disparity in how different content types are handled, so that this would have to be handled differently for each of exercises, assessment items, videos, etc. Can we address that in a separate PR? Pending any other comments, I think this is done.

jtamiace:
@MCGallaspy , the design works and is good to merge

aronasorman:
Code looks good to me. Minor nits can be fixed as we go.

PR #4253, opened 2015-08-18
Fixes #4244: Safer BaseView.remove: remove el before reflow, no infinite recursion.

### LINKED ISSUE #4244: Hiding large tabular report takes... an amount of time that is indistinguishable, in practical terms, from forever
Loading a tabular coach report with quite a few cells, and then clicking the "Hide Tabular Report" button, hangs the page for me.

Looking into the stack trace, it looks as if it might be because we remove the view, which then recursively removes all subviews, which then removes their DOM elements one by one, which causes a cascading sequence of page reflows.

Possible solution: remove top-level element (via `Backbone.View.prototype.remove.call(this);`) before doing the recursive subview removal, in `BaseView.remove`.

Temporary solution: hide the "Hide Tabular Report" button?

jamalex:
Fixes #4244.

- Remove the top-level view itself (via Backbone.View.remove) to get it out of the DOM before removing subviews, to minimize reflow overhead.
- Make sure we never recurse infinitely in subview removal, in case there's weird circularity.

### DIFF
--- a/kalite/distributed/static/js/distributed/base/baseview.js
+++ b/kalite/distributed/static/js/distributed/base/baseview.js
@@ -73,6 +73,15 @@ module.exports = Backbone.View.extend({
     },
 
     remove: function() {
+
+        // make sure we never end up removing the same view twice, in case there's weird circularity
+        if (this._removed) return;
+        this._removed = true;
+
+        // remove this view using the default Backbone code, which removes the DOM element
+        Backbone.View.prototype.remove.call(this);
+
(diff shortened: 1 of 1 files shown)


jamalex:
This has been merged into central-develop and launched to the central server. Much speediness ensued: http://globe.learningequality.org:8008/coachreports/coach/zone/630bec0f378455d4a14768d2f15fffad/c2b99d05cc1052b0acd61139be66fd99/8d93d033f5655231b245f318a291685d/

rtibbles:
Looks good. Tests pass. Merging.

PR #4254, opened 2015-08-18
Fix container-inside-container overflow issue.

jamalex:
(with a .container inside a .container, these inner divs were overflowing their parents; not so apparent on the distributed server, though noticeable, but super bad in the context of the central server's styling, which motivated the fix)

Before:

image

After:

image


rtibbles:
Low risk. Merging.

PR #4257, opened 2015-08-18
Expire cookies whenever we close the browser.

aronasorman:
Since sessions are now stored in cookies, our default cookie expiration date may have some unintended consequences for computers with multiple KA Lite users. Admins who expect to get logged out on reboot, actually don't get logged out. 

This simple change ensures that the session cookie is cleared *when the browser is closed* ([not tab closed, browser closed](http://stackoverflow.com/questions/3976498/why-doesnt-session-expire-at-browser-close-true-log-the-user-out-when-the-bro)).

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -467,6 +467,9 @@ KEY_PREFIX = version.VERSION
 # Separate session caching from file caching.
 SESSION_ENGINE = getattr(
     local_settings, "SESSION_ENGINE", 'django.contrib.sessions.backends.signed_cookies' + (''))
+    
+# Expire session cookies whenever we close the browser.
+SESSION_EXPIRE_AT_BROWSER_CLOSE = True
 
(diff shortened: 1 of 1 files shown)


aronasorman:
@jamalex we were discussing in the dev meeting whether this is an appropriate solution to @EdDixon's concern. My understanding is that it is a matter of urgency --  whether we want to merge this in now, or to bundle it in for a patch release to avoid Yet Another Manual Build Of The Installers. The default had cookies persist for up to a year when I looked at their values.

There is also the question of any far reaching consequences of this -- what comes to my mind are whether we have tests that depend on the session being retained across browser reboots. That's not too critical for me and @EdDixon's concerns about security trump that.

jamalex:
Should be fine, yep. No strong opinions here. Would be nice to add an option that would let users change this if desired, but should be OK with this as the default.

jamalex:
I guess this can be overridden in custom settings. Would be good to add to the docs.

Issue #4258, opened 2015-08-18
Fixes needed on the "installation" page in the docs


jamalex:
Based on http://field.learningequality.org/docs/0.14/en/installguide/install_main.html (looks identical, in terms of the things I'm mentioning here, to the develop version).

- [x] PPA and pip instructions, and the "advanced topics" stuff, should move to a separate advanced page (and not show up on this main installers page), as discussed previously with @benjaoming 
- [x] OSX installer needs to point to something like https://learningequality.org/r/osx-installer-latest rather than disney.com
- [x] Primary Debian installation instructions should point to https://learningequality.org/r/deb/ka-lite-bundle/0.14
- [x] Raspberry Pi instructions need to be similarly updated (CC https://github.com/learningequality/ka-lite/issues/4082)
- [x] "KA Lite is currently not supported on Internet Explorer version 8 or lower. You must use IE9 or later." should be changed to mention that essentially any version of Chrome, Firefox, or Safari should work fine too. We've had people email asking why they can only use IE.

MCGallaspy:
> OSX installer needs to point to something like https://learningequality.org/r/osx-installer-latest rather than disney.com

Yes, what is the link for the latest OSX installer? Do you know @aronasorman?

It seems also the windows installer link learningequality.org/r/windows-installer-latest needs to be updated to point to the latest window installer. How can I upload the installer (which is currently on field) to the homepage server? The admin interface?

MCGallaspy:
> OSX installer needs to point to something like https://learningequality.org/r/osx-installer-latest rather than disney.com

It seems you were viewing an outdated version of the docs.

jamalex:
> Yes, what is the link for the latest OSX installer? Do you know @aronasorman?

We should have it point to https://learningequality.org/r/osx-installer-latest or https://learningequality.org/r/mac-installer-latest, which can then be configured to point to the latest version of the docs.

> It seems also the windows installer link learningequality.org/r/windows-installer-latest needs to be updated to point to the latest window installer. How can I upload the installer (which is currently on field) to the homepage server? The admin interface?

The files under /downloads/ are on the `pantry` server, which it looks like you should have SSH access to (according to the Ansible scripts). Windows installer should go in `/var/www/downloads/ka-lite/0.14/installers/windows`. We'll want to keep using a consistent naming scheme (that includes the LONG_VERSION).

(Long-term, the build box will be uploading this directly itself, rather than us uploading it, but for now you should be able to scp the file in manually)

[...]

PR #4260, opened 2015-08-18
0.14 docs fixup

MCGallaspy:
Addresses these two points from #4258:

* PPA and pip instructions, and the "advanced topics" stuff, should move to a separate advanced page (and not show up on this main installers page), as discussed previously with @benjaoming
* "KA Lite is currently not supported on Internet Explorer version 8 or lower. You must use IE9 or later." should be changed to mention that essentially any version of Chrome, Firefox, or Safari should work fine too. We've had people email asking why they can only use IE.


MCGallaspy:
Remember, once these are merged you'll have to run the playbook in order to build and deploy the updated docs.

benjaoming:
:+1:

In the longer run, we probably need to give a better overview such that installation methods that may be extremely convenient to some aren't hidden behind "Advanced installation". Also, I should make updates that reflect what we're learning with the Raspberry Pi package currently.

jamalex:
What's here looks good, and then there will be a few more things to do, e.g. the PPA instructions should also be moved to this section (can be done as part of updating the Deb links to use the /r/deb-latest-installer redirect) -- I would also recommend having the page be called "Advanced installation methods" rather than the more generic "Advanced topics".

MCGallaspy:
@jamalex okay, as soon as the .deb link exists I will update the docs.

benjaoming:
Merging this, so I can continue the work in a separate PR.

PR #4261, opened 2015-08-18
Cache invalidation

### LINKED ISSUE #4107: Intelligently invalidate content caches
At the moment we invalidate content caches all at once. Instead, we should use the information we have about what has changed to only invalidate the parts of the cache that have changed.

See #4074 for more details on this issue.

rtibbles:
Fixes #4107 

Summary of changes:
* Moves all topic tree and content to a database backed system, with separate databases per language and channel.
* Refactors annotation to be done per item, rather than all at once.
* Only reannotate videos that have been downloaded.

This is a WIP, but I wanted @aronasorman and @jamalex to take a look and comment on approach.

As it stands, Sidebar and Exercise/Video rendering are working fine with it, but I will have to make some further modifications in order to fully deprecate get_topic_tree and get_content/exercise_cache.

Things that still need to be done:
- [x] Remove all references to get_topic_tree, get_content_cache, get_exercise_cache.


aronasorman:
Thanks! This will make further optimizations much easier \o/

kalite/topic_tools/content_models.py
```diff
@@ -0,0 +1,222 @@
...
+    def __init__(self, *args, **kwargs):
+        kwargs = parse_model_data(kwargs)
+        super(Item, self).__init__(*args, **kwargs)
+
+
+def parse_model_data(item):
+    extra_fields = item.get("extra_fields", {})
+
+    if type(extra_fields) is not dict:
+        extra_fields = json.load(extra_fields)
```
aronasorman: Shouldn't this be `loads`, since it's a string?

rtibbles:
@aronasorman Punting this to 0.16 for the sake of code stability.

rtibbles:
@aronasorman Test failures here are cosmetic, but will be fixed. This is now profileable, so would appreciate you running this with profiling tools to check performance!

aronasorman:
What're the series of management commands I need to run to get this set up? `init_content_items` then `annotate_content_items`?

kalite/control_panel/api_resources.py
```diff
@@ -132,12 +133,15 @@ def _get_facility_users(self, bundle):
...
-        return response
+        try:
+            response = super(ParentFacilityUserResource, self).create_response(request, data, response_class=response_class, **response_kwargs)
+            # add a suggested download filename if we're replying with a CSV file
+            if response["Content-Type"].startswith("text/csv"):
+                params = ["%s-%s" % (k,str(v)[0:8]) for (k,v) in request.GET.items() if v and k not in ["format", "limit"]]
+                response["Content-Disposition"] = "filename=%s__%s__exported_at-%s.csv" % (request.path.strip("/").split("/")[-1], "__".join(params), datetime.now().strftime("%Y%m%d_%H%M%S"))
+            return response
+        except Exception as e:
+            import pdb; pdb.set_trace()
```
aronasorman: Might want to clean this up.

kalite/contentload/management/commands/init_content_items.py
```diff
@@ -0,0 +1,146 @@
+import os
```
aronasorman: Module docstrings

kalite/control_panel/api_resources.py
rtibbles: And this is one of many reasons I need people to review my code...

kalite/contentload/management/commands/init_content_items.py
rtibbles: :+1:

rtibbles:
@aronasorman Yep - ```init_content_items``` creates the initial databases and could be run remotely, and then we could distribute the database files, rather than the json files.

```annotate_content_items``` needs to happen locally, and can be run with a specific content id to only reannotate that particular file (horror of horrors, we can now download a video file, and just reannotate that changed video file rather than the whole topic tree!).

Also, at some point in the near future, we should probably also add a 'filename' field to the models, so that we are not reliant on id + format for identifying the file in question.

File: kalite/testing/testrunner.py [hunk 4]
```diff
@@ -150,6 +152,15 @@ def build_suite(self, test_labels, extra_tests, **kwargs):
         browser = webdriver.Firefox()
         print("Successfully setup Firefox {0}".format(browser.capabilities['version']))
         browser.quit()
+
+        if not database_exists():
+            call_command("init_content_items")
+            call_command("annotate_content_items")
+
+            print("Successfully setup content database")
```
aronasorman: `logging.info`

kalite/topic_tools/content_models.py
```diff
@@ -0,0 +1,389 @@
...
+    extra_fields = item.get("extra_fields", {})
+
+    if type(extra_fields) is not dict:
+        extra_fields = json.loads(extra_fields)
+
+    remove_keys = []
+    for key, value in item.iteritems():
+        if key not in Item._meta.fields:
+            extra_fields[key] = value
+            remove_keys.append(key)
```
aronasorman: Why not just remove the key directly in this line, rather than adding them to a list?

rtibbles: Can cause a Runtime error.

https://docs.python.org/2/library/stdtypes.html#dict.iteritems

aronasorman:
First pass. This will take a few more passes since it's so huge :)

Looks like there's a ton of code cleanup going on in here as well. Can you point out which files I should focus on? Also, can you add more detail as to what kind of refactorings and cleanup you did? My main concern is that some of the deleted files are test files (including Nalanda stuff, which is OK I guess).

rtibbles:
Yep - I really didn't want to make this so big, because the important changes are actually quite carefully encapsulated. But at the same time I didn't want to do a bunch of refactor on e.g. the playlist code because it is beyond deprecated.

Will mark the important sections and rationalize deletions.

kalite/caching/__init__.py
```diff
@@ -1,99 +0,0 @@
-"""
```
rtibbles: We now use neither the previous web page caching nor the in memory caches.

We should be using Django's in built caching mechanism for any caching in future.

kalite/coachreports/api_views.py
```diff
@@ -10,7 +10,7 @@
 from kalite.main.models import ExerciseLog, VideoLog, ContentLog, AttemptLog, UserLogSummary
 from kalite.facility.models import FacilityUser
 from kalite.shared.decorators.auth import require_admin
-from kalite.topic_tools import get_topic_leaves, get_exercise_cache, get_content_cache
+from kalite.topic_tools.content_models import get_topic_contents, get_topic_nodes
```
rtibbles: This file is a reasonable example of the slight switch this PR makes in terms of the API of interfacing with content caches.

kalite/coachreports/features/steps/coachreports.py
```diff
@@ -101,15 +100,15 @@ def create_some_learner_data():
         ("attempted", 50, 10),
         ("struggling", 30, 25),
     )
-    exercises = random.sample(get_exercise_cache().keys(), len(attempt_states))  # Important they are *distinct*
```
rtibbles: More efficient to randomly sample at the database level, so this convenience function handles that.

kalite/coachreports/models.py
```diff
@@ -36,20 +35,6 @@ def get_user_logs(cls, user, pl_video_ids=None, pl_exercise_ids=None):
 
         return (user_vid_logs, user_ex_logs)
 
-    @classmethod
```
rtibbles: Quizzes are awesome, but we should probably bring them back in another form, rather than attempting to raise the dead.

kalite/coachreports/models.py
```diff
@@ -110,9 +88,9 @@ def user_progress(cls, user_id):
             n_ex_started = len([ex for ex in pl_ex_logs if ex["attempts"] > 0])
             n_ex_incomplete = len([ex for ex in pl_ex_logs if (ex["attempts"] > 0 and not ex["complete"])])
             n_ex_struggling = len([ex for ex in pl_ex_logs if ex["struggling"]])
-            ex_pct_mastered = int(float(n_ex_mastered) / n_pl_exercises * 100)
-            ex_pct_incomplete = int(float(n_ex_incomplete) / n_pl_exercises * 100)
-            ex_pct_struggling = int(float(n_ex_struggling) / n_pl_exercises * 100)
+            ex_pct_mastered = int(float(n_ex_mastered) / (n_pl_exercises or 1) * 100)
```
rtibbles: Actually an extant bug - if a playlist doesn't have any exercises in it, this will cause a divide by 0.

kalite/coachreports/models.py
```diff
@@ -124,20 +102,6 @@ def user_progress(cls, user_id):
             else:
                 ex_status = "complete"
 
-            # Oh Quizzes, we hardly knew ye!
-            # TODO (rtibbles): Sort out the status of Quizzes, and either reinstate them or remove them.
```
rtibbles: True to my word.

kalite/contentload/management/commands/annotate_content_items.py
```diff
@@ -0,0 +1,51 @@
+import os
```
rtibbles: Management command - little more than a convenience wrapper around the annotate_content_models method.

kalite/contentload/management/commands/generate_assessment_zips.py
```diff
@@ -345,7 +347,7 @@ def _old_content_links_to_local_links(matchobj):
 def _get_content_by_readable_id(readable_id):
     global CONTENT_BY_READABLE_ID
     if not CONTENT_BY_READABLE_ID:
-        CONTENT_BY_READABLE_ID = dict([(c["readable_id"], c) for c in get_content_cache().values()])
+        CONTENT_BY_READABLE_ID = dict([(c.get("slug"), c) for c in get_content_items() if c.get("slug")])
```
rtibbles: This might be problematic, and should probably be reverted.

kalite/contentload/management/commands/init_content_items.py
```diff
@@ -0,0 +1,146 @@
...
+
+from django.core.management import call_command
```
rtibbles: Eventually, this management command can be deprecated, because we can just generate the databases directly from contentload.

At the moment it just transforms our current JSON files into database files.

Due to the default use of bulk inserts, this is a relatively speedy operation.

kalite/control_panel/api_resources.py
```diff
@@ -402,7 +407,7 @@ def alter_list_data_to_serialize(self, request, to_be_serialized):
             bundle.data.pop("user")
 
             content_id = bundle.data.pop("content_id", None)
-            content = get_content_data(request, content_id) or get_exercise_data(request, content_id)
+            content = get_content_item(language=request.language, content_id=content_id)
```
rtibbles: This is probably OK - but we have the slight issue of duplicate content_ids in the topic tree. (Note: this is an issue both with content items being repeated in the topic tree, and also with Topics and exercises sometimes having the same 'id').

We should switch to UUIDs.

kalite/control_panel/views.py
```diff
@@ -539,6 +537,7 @@ def _get_user_usage_data(users, groups=None, period_start=None, period_end=None,
 
     # Add group data.  Allow a fake group UNGROUPED
     for user in users:
+        user_data[user.pk]["pct_mastery"] = user_data[user.pk]["pct_mastery"]/(user_data[user.pk]["total_exercises"] or 1)
```
rtibbles: Previously % mastery was being calculated for mastery of *every single exercise*, seemed like a high bar.

kalite/distributed/static/js/distributed/content/views.js
```diff
@@ -6,6 +6,7 @@ var VideoModels = require("video/models");
 var $script = require("scriptjs");
 
 var ContentBaseView = require("./baseview");
+var ExerciseModels = require("exercises/models");
```
rtibbles: One consequence of this approach, is that Exercise content data is now accessible through the same API endpoint as other content. This means we can unify Exercises with all the other content type in the front end.

kalite/distributed/static/js/distributed/content/views.js
```diff
@@ -101,6 +109,18 @@ var ContentWrapperView = BaseView.extend({
...
                 });
                 break;
+
+            case "Exercise":
+                $script(window.sessionModel.get("STATIC_URL") + "js/distributed/bundles/bundle_exercise.js", function(){
+                    self.add_content_view(external("exercise").ExercisePracticeView);
+                });
+                break;
+
+            case "Quiz":
```
rtibbles: I probably could have deleted this, given my other changes.

kalite/distributed/static/js/distributed/exercises/models.js
```diff
@@ -13,7 +15,7 @@ var ExerciseParams = {
 };
 
 
-var ExerciseDataModel = Backbone.Model.extend({
+var ExerciseDataModel = ContentModels.ContentDataModel.extend({
```
rtibbles: Exercise data now comes from the backend with 'extra_fields' - so it needs the parsing magic of ContentDataModel to inflate this to full JSON.

kalite/distributed/static/js/distributed/search/views.js
```diff
@@ -94,89 +46,28 @@ var AutoCompleteView = BaseView.extend({
             minLength: 3,
             appendTo: ".navbar-collapse",
             html: true,  // extension allows html-based labels
-            source: this.data_source,
+            source: window.Urls.search_api(window.sessionModel.get("CHANNEL")),
```
rtibbles: Refactor to query an API endpoint, rather than load a huge pile of data into the frontend.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -717,48 +717,21 @@ var TopicContainerOuterView = BaseView.extend({
...
-        }
+
+        var view = new ContentViews.ContentWrapperView({
+            id: id,
+            kind: kind,
+            context_id: this.model.get("id"),
+            channel: window.channel_router.channel
+        });
+
+        this.content_view.show_view(view);
```
rtibbles: ContentAreaView should be almost completely useless by now. Can probably remove thanks to this.

kalite/distributed/templates/distributed/base.html
```diff
@@ -60,6 +60,7 @@
           GET_VIDEO_LOGS_URL          : "{% url 'api_dispatch_list' resource_name='videolog' %}",
           GET_EXERCISE_LOGS_URL       : "{% url 'api_dispatch_list' resource_name='exerciselog' %}",
           GET_CONTENT_LOGS_URL        : "{% url 'api_dispatch_list' resource_name='contentlog' %}",
+          GET_ATTEMPT_LOGS_URL        : "{% url 'api_dispatch_list' resource_name='attemptlog' %}",
```
rtibbles: This was previously hard coded in JS.

kalite/distributed/templates/distributed/search_page.html
```diff
@@ -30,9 +30,6 @@ <h2 class="errorlist">{{ query_error }}</h2>
                             <a href="{% url 'learn' %}{{ node.path }}" class="{{ node.kind }}-{% if not node.available %}un{% endif %}available">{% trans node.title %}</a>
                         </li>
                     {% endfor %}
-                    {% if hit_max|get_item:category %}
```
rtibbles: Querying by each type at a time is not very efficient, and also doesn't scale beyond Topic, Video, Exercise.

kalite/distributed/tests/browser_tests/__init__.py
```diff
@@ -4,6 +4,4 @@
 from .distributed import *
 from .knowledge_map import *
 from .language_packs import *
-from .nalanda import *
```
rtibbles: Have removed all features here, nothing to test.

kalite/distributed/tests/browser_tests/distributed.py
```diff
@@ -101,65 +100,6 @@ def test_register_login_exact(self):
         self.browser_logout_user()
 
 
-@unittest.skipIf(getattr(settings, 'HEADLESS', None), "Fails if settings.HEADLESS is set.")
```
rtibbles: Was being skipped anyway.

kalite/distributed/tests/browser_tests/nalanda.py
```diff
@@ -1,71 +0,0 @@
-"""
```
rtibbles: Relies on deleted playlist app.

kalite/distributed/views.py
```diff
@@ -24,8 +24,8 @@
 
 from fle_utils.internet.classes import JsonResponseMessageError
 from fle_utils.internet.functions import get_ip_addresses, set_query_params
-from kalite import topic_tools
 from kalite.shared.decorators.auth import require_admin
+from kalite.topic_tools.content_models import search_topic_nodes
```
rtibbles: Use the power of a database to do what it does well, searching.

kalite/distributed/views.py
```diff
@@ -148,23 +144,17 @@ def search(request):
     # Inputs
     page = int(request.GET.get('page', 1))
     query = request.GET.get('query')
-    category = request.GET.get('category')
-    max_results_per_category = request.GET.get('max_results', 25)
+    max_results = request.GET.get('max_results', 50)
```
rtibbles: Limit total results, rather than by category, do the factoring out post hoc.

kalite/distributed/views.py
```diff
@@ -172,29 +162,16 @@ def search(request):
...
-            if category and node_type != category:
-                # Skip categories that don't match (if specified)
-                continue
-
-            exact_match = filter(lambda node: node["kind"] in node_kinds[node_type] and node["title"].lower() == query, node_dict.values())[:1]
 
-            if exact_match:
-                # Redirect to an exact match
-                return HttpResponseRedirect(reverse('learn') + exact_match[0]['path'])
+        matches, exact, pages = search_topic_nodes(query=query, language=request.language, page=page, items_per_page=max_results)
```
rtibbles: Use paging methods on the database to determine pagination.

kalite/main/api_resources.py
```diff
@@ -91,81 +90,6 @@ class Meta:
         authorization = UserObjectsOnlyAuthorization()
 
 
-class Exercise():
```
rtibbles: Bye Bye, non-ORM TastyPie.

kalite/main/api_views.py
```diff
@@ -6,18 +6,63 @@
...
+    query = query.lower()
+    # search for topic, video or exercise with matching title
+
+    matches, exact, pages = search_topic_nodes(query=query, channel=channel, language=request.language, page=1, items_per_page=15)
+
+    return JsonResponse(matches)
+
+
+
+@api_handle_error_with_json
```
rtibbles: This view replaces 84 lines of Tastypie code.

kalite/playlist/models.py
```diff
@@ -1,206 +0,0 @@
...
-    class Meta:
-        ordering = ['playlist', 'sort_order']
-
-
-class PlaylistToGroupMapping(ExtendedModel):
-    playlist = models.CharField(db_index=True, max_length=15)
-    group = models.ForeignKey(FacilityGroup, related_name='playlists')
-
-
-class QuizLog(DeferredCountSyncedModel):
```
rtibbles: Only sync model in the app, and no registered uses as far as we can tell. Safe to delete.

kalite/topic_tools/__init__.py
```diff
@@ -1,618 +0,0 @@
-"""
-TODO: NOTHING SHOULD BE HERE! It's prohibiting the import of other topic_tools.xxx
```
rtibbles: Like @benjaoming said.

Also, not much of the code made it out of here alive.

kalite/topic_tools/annotate.py
```diff
@@ -0,0 +1,155 @@
+import os
```
rtibbles: This is where a lot of the surviving code from ```__init__.py``` made it to.

kalite/topic_tools/annotate.py
```diff
@@ -0,0 +1,155 @@
...
+
+def create_thumbnail_url(thumbnail):
+    if is_content_on_disk(thumbnail, "png"):
+        return django_settings.CONTENT_URL + thumbnail + ".png"
+    elif is_content_on_disk(thumbnail, "jpg"):
+        return django_settings.CONTENT_URL + thumbnail + ".jpg"
+    return None
+
+
+def update_content_availability(content_list, language="en"):
```
rtibbles: Essentially, this is a concatenation of the annotation code from get_exercise_cache and get_content_cache, properly ```if```ed.

kalite/topic_tools/base.py
```diff
@@ -0,0 +1,91 @@
+"""
```
rtibbles: This is where the remaining code made it to.

kalite/topic_tools/content_models.py
```diff
@@ -0,0 +1,389 @@
+import os
```
rtibbles: This is, in some ways, the most important, and also the least important part of the refactor.

All of the functions in here are designed to make the database implementation opaque to the rest of the app.

All functions should return a dict, or list of dicts, so that the rest of the app does not have to be concerned with the implementation details of the database backend.

If I have coded it correctly, then every function in here should be repurposable to use a completely different database manager (other than peewee) and the rest of the app could carry on without changes.

What this file needs is complete API documentation, to detail each of the methods, and what their output characteristics should be.

kalite/topic_tools/content_models.py
```diff
@@ -0,0 +1,389 @@
...
+    return item
+
+
+def database_exists(channel="khan", language="en", database_path=None):
+    path = database_path or CONTENT_DATABASE_PATH.format(channel=channel, language=language)
+
+    return os.path.exists(path)
+
+
+def set_database(function):
```
rtibbles: This allows us to hot swap databases, depending on channel and language.

kalite/updates/management/commands/videodownload.py
```diff
@@ -185,6 +186,8 @@ def youtube_dl_cb(stats, progress_callback, *args, **kwargs):
                     handled_youtube_ids.append(video.youtube_id)
                     self.stdout.write(_("Download is complete!") + "\n")
 
+                    annotate_content_models(ids=[video.youtube_id])
```
rtibbles: Update annotation for just the id that was updated! Simples!

rtibbles:
@aronasorman Tried to give some tips on review, and highlighted a big gap in documentation for myself.

circle.yml
```diff
@@ -4,6 +4,7 @@ machine:
 
 dependencies:
   cache_directories:
+    - "content"
```
aronasorman: Why do we need to cache the content directory? Is this where the sqlite files are written?

rtibbles: Yep, can be changed, ofc.

kalite/control_panel/api_resources.py
rtibbles: This function now does not return topics.

kalite/distributed/static/js/distributed/topics/views.js
rtibbles: Will leave for now as content rating relies on it. Once that has stabilised in 0.15, can work on extricating it.

kalite/coachreports/api_views.py
benjaoming: True that... I always use manual search tools like `ack-grep --python` to find these things... trying to use the most broad match and then review results manually. Not sure if it's the best possible way, but it usually works.

kalite/topic_tools/__init__.py
benjaoming: Huge step! Wonderful!! I'm sure from the experience of the problems with topic_tools.settings that this can be a benefit for the rest of ka-lite's life cycle, no matter which kinds of refactors we get into.

requirements.txt
```diff
@@ -7,6 +7,7 @@ ply==3.4
 slimit==0.8.1
 six==1.9.0
 django-appconf==1.0.1
+peewee==2.6.3
```
benjaoming: What's that? :) Is it necessary? Do we need it? :)

kalite/topic_tools/__init__.py
rtibbles: i18n is the only one left with code in its __init__.py, after I deleted the entire caching app. We should make sure to finish that in 0.16 too.

requirements.txt
rtibbles: This handles the database backend for the content caches. It is a light wrapper around SQLite, inspired by the Django ORM.

So far it seems pretty efficient and useable, and I've not had any of the issues that @MCGallaspy had with using the Django ORM for other content ORMing.

It also allows very easy hot swapping of database files during runtime, for, e.g., using different databases for different languages (and also channels).

That said, I'm not wedded to it, so if we can do this efficiently without adding a dependency, happy with that too.

benjaoming: Sounds like it has good benefits, and it looks pure-Python with some kind
of [Cython speed-ups](
https://github.com/coleifer/peewee/blob/master/setup.py#L10) that I don't
understand how works.. should probably test that it builds a cross-platform
result.

But if there aren't any critical optimization benefits, perhaps its better
to wait until Kolibri with introducing a whole new ORM?

I don't think KA Lite or Kolibri should ever have 2 ORMs.... would be a
nightmare to be honest... should we then define models for both ORMs? How
do we create a relation from one ORM to the other? How does query caching
work across ORMs? Who's responsible for migrations? etc.

kalite/caching/__init__.py
aronasorman: Do we only use the in-memory caches for i18n and topic tool-related functions?

rtibbles: I believe so. Searching for CACHE_VARS should show any instances.

gwong89:
running command 

`bin/kalite manage init_content_items --overwrite --no-bulk-create`

```python
  File "/Users/giowong/Documents/ka-lite/python-packages/django/core/management/base.py", line 224, in run_from_argv
  File "/Users/giowong/Documents/ka-lite/python-packages/django/core/management/base.py", line 263, in execute
  File "/Users/giowong/Documents/ka-lite/kalite/contentload/management/commands/init_content_items.py", line 156, in handle
    bulk_insert(items, database_path=database_path)
  File "/Users/giowong/Documents/ka-lite/kalite/topic_tools/content_models.py", line 103, in wrapper
    output = function(*args, **kwargs)
  File "/Users/giowong/Documents/ka-lite/kalite/topic_tools/content_models.py", line 353, in bulk_insert
    Item.insert_many(map(parse_model_data, items[idx:idx+500])).execute()
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 2980, in execute
    cursor = self._execute()
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 2470, in _execute
    return self.database.execute_sql(sql, params, self.require_commit)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3199, in execute_sql
    self.commit()
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3048, in __exit__
    reraise(new_type, new_type(*exc_args), traceback)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3191, in execute_sql
    cursor.execute(sql, params or ())
OperationalError: too many SQL variables
```

gwong89:
changing line 135 from
`bulk_create = not kwargs["no_bulk_create"]`

to

`bulk_create = False`

resulted in

```python
  File "/Users/giowong/Documents/ka-lite/python-packages/django/core/management/base.py", line 224, in run_from_argv
  File "/Users/giowong/Documents/ka-lite/python-packages/django/core/management/base.py", line 263, in execute
  File "/Users/giowong/Documents/ka-lite/kalite/contentload/management/commands/init_content_items.py", line 152, in handle
    create_table(database_path=database_path)
  File "/Users/giowong/Documents/ka-lite/kalite/topic_tools/content_models.py", line 103, in wrapper
    output = function(*args, **kwargs)
  File "/Users/giowong/Documents/ka-lite/kalite/topic_tools/content_models.py", line 388, in create_table
    db.create_tables([Item])
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3281, in create_tables
    create_model_tables(models, fail_silently=safe)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 4451, in create_model_tables
    m.create_table(**create_table_kwargs)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 4194, in create_table
    db.create_table(cls)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3278, in create_table
    return self.execute_sql(*qc.create_table(model_class, safe))
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3199, in execute_sql
    self.commit()
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3048, in __exit__
    reraise(new_type, new_type(*exc_args), traceback)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3191, in execute_sql
    cursor.execute(sql, params or ())
OperationalError: table "item" already exists
```

benjaoming:
Will this continue to rely on Peewee or what are the thoughts and plans atm?

gwong89:
turned this
```python
 if bulk_create and os.path.isfile(database_path):
   if kwargs["overwrite"]:
     os.remove(database_path)
   else:
     logging.info("Database already exists, use --overwrite to force overwrite")
     return None
``` 

into this

```python
  if kwargs["overwrite"]:
   os.remove(database_path)
  else:
   logging.info("Database already exists, use --overwrite to force overwrite")
   return None
```


error:
```python
  File "/Users/giowong/Documents/ka-lite/python-packages/django/core/management/base.py", line 224, in run_from_argv
  File "/Users/giowong/Documents/ka-lite/python-packages/django/core/management/base.py", line 263, in execute
  File "/Users/giowong/Documents/ka-lite/kalite/contentload/management/commands/init_content_items.py", line 152, in handle
    create_table(database_path=database_path)
  File "/Users/giowong/Documents/ka-lite/kalite/topic_tools/content_models.py", line 103, in wrapper
    output = function(*args, **kwargs)
  File "/Users/giowong/Documents/ka-lite/kalite/topic_tools/content_models.py", line 388, in create_table
    db.create_tables([Item])
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3281, in create_tables
    create_model_tables(models, fail_silently=safe)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 4451, in create_model_tables
    m.create_table(**create_table_kwargs)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 4194, in create_table
    db.create_table(cls)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3278, in create_table
    return self.execute_sql(*qc.create_table(model_class, safe))
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3199, in execute_sql
    self.commit()
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3048, in __exit__
    reraise(new_type, new_type(*exc_args), traceback)
  File "/usr/local/lib/python2.7/site-packages/peewee.py", line 3191, in execute_sql
    cursor.execute(sql, params or ())
OperationalError: table "item" already exists
Gios-MacBook-Pro:ka-lite giowong$ 
```

rtibbles:
@benjaoming Happy to migrate this to use the Django ORM, as long as we can cleanly make manual connections to different databases (rather than interfere with any of our existing Database connections, and have to manually route different models to different databases).

If this is as simple and straight forward as in Peewee, I am happy to use Django instead, and can migrate the code to do so.

aronasorman:
@rtibbles before I merge this in, can you update the getting started wiki page?

rtibbles:
Sure!

rtibbles:
@benjaoming I have looked at how Django manages database connections, and I can't seem to find a clean way to dynamically swap databases in a QuerySet with the ```using``` keyword, without those databases having not already been defined in the ```settings.py```. If there is a way to do it, I'd love to hear, but my first attempt have made it seem like the Django database API is not built to support it.

rtibbles:
@aronasorman Updated this, it didn't actually do the annotation at setup, so it does now (instead of just running videoscan). Have also updated the wiki.

One thing that we will want to consider is separating out the translation step from the annotation (they are still intermixed), as in future the database cache files should be translated when they are distributed, and hence should never have to be translated on the client machine.

benjaoming:
@rtibbles 
> @benjaoming I have looked at how Django manages database connections, and I can't seem to find a clean way to dynamically swap databases in a QuerySet with the using keyword

But the whole approach of needing to dynamically swap databases? What's the base of that? Could we not do it?

It sounds a bit like asking for lots of trouble IMI*

*) In My Intuition

rtibbles:
Dynamically swapping the databases is a nice clean solution for having different caches for every language. Meaning that instead of having to translate content metadata either on the fly, or in a costly upfront process on the client machine, we can include the database in the language pack instead.

This is important, because the content cache annotation is per language, and so if we are going to have persistent, low friction database annotation, then we need to be tracking all of the different languages.

The alternative would be to have a single database for every language, and pile each individual language specific field into the model, in a JSON blob. This could work, but would mean that doing things like searching would be much more inefficient, as we would no longer be able to query on titles, descriptions, etc, because we would have to search the JSON blob.

benjaoming:
That sounds reasonable for the database-oriented approach.

What about the gettext framework? It usually works by having loads of .mo files stacked... maybe for translations it's better to use this translation-intended approach?

Sorry about arriving late with inputs for this.

rtibbles:
The problem with using the mo files for content metadata (and particularly the assessment items) translation is that the mo files get huge in memory (~60Mb when loaded by Django), which means we can put large peak memory load on the server. By having all the translations done beforehand, and stored in a database, our hope is to reduce the overall memory load, and the peak memory load as well.

gwong89:
here is a gist with some benchmarks:


https://gist.github.com/gwong89/b05d4dc0b47af2438304

rtibbles:
Nice. As I feared, the annotation of all content at once causes memory
bloat, so it may be worth optimising the code to annotate content one at a
time (or in chunks to keep the size of uncommitted transactions down).

rtibbles:
@gwong89 Probably good to compare running the annotation command memory usage with what happens during downloading a single video, if that's possible. That should tell us something about the memory usage of annotating a single item at a time.

Then the next step would be measuring how much slower that is on the SanDisk or RPi. Given that we only need to run the overall content annotation rarely, this may be a reasonable trade off, even if it's slow.

rtibbles:
@benjaoming One more thing to allay fears. For these databases, we will not be running migrations, nor making references from these databases to the Django ORM. They are standalone, and will not interact with the Django Models. The design of ```content_models.py``` is to give a completely implementation agnostic API to the rest of the app, so it doesn't matter at all how the data is stored internally - it could still be our ugly old JSON blobs if we wanted! But we don't want that :)

aronasorman:
@gwong89 and I have tested it a bit in the process of benchmarking. I think this is good to go, and we can iron out issues as we move forward with 0.16.

PR #4262, opened 2015-08-18
Content rating model fixup

MCGallaspy:
See here: https://github.com/learningequality/ka-lite/pull/4149/files#r37342427

### DIFF
--- a/kalite/main/content_rating_models.py
+++ b/kalite/main/content_rating_models.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.db import models
 
 from kalite.facility.models import FacilityUser
@@ -5,14 +6,18 @@ from kalite.facility.models import FacilityUser
 from securesync.models import DeferredCountSyncedModel
 
 class ContentRating(DeferredCountSyncedModel):
+
+    minversion = "0.15.0"
+
     class Meta:
         unique_together = ("content_kind", "content_id", "user")
 
     # Maintain info on content type -- could be video, exercise, etc, but we should be able to uniquely id it
     content_kind = models.CharField(max_length=100, db_index=True, blank=False)
     content_id = models.CharField(max_length=100, db_index=True, blank=False)
+    content_source = models.CharField(max_length=100, db_index=True, default=settings.CHANNEL)
 
     user = models.ForeignKey(FacilityUser, blank=False)
     quality = models.IntegerField(blank=False, default=0)
     difficulty = models.IntegerField(blank=False, default=0)
-    text = models.TextField(blank=True, null=True)
+    text = models.TextField(blank=True)
--- a/kalite/main/content_rating_models.py
+++ b/kalite/main/content_rating_models.py
@@ -1,3 +1,5 @@
+import uuid
+
 from django.conf import settings
 from django.db import models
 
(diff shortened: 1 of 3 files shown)


kalite/main/content_rating_models.py
```diff
@@ -5,14 +6,18 @@
...
+
+    minversion = "0.15.0"
+
     class Meta:
         unique_together = ("content_kind", "content_id", "user")
 
     # Maintain info on content type -- could be video, exercise, etc, but we should be able to uniquely id it
     content_kind = models.CharField(max_length=100, db_index=True, blank=False)
     content_id = models.CharField(max_length=100, db_index=True, blank=False)
+    content_source = models.CharField(max_length=100, db_index=True, default=settings.CHANNEL)
```
benjaoming: Doesn't this change require a migration?

MCGallaspy: Yes, so does #4149. I didn't commit a migration file in anticipation of the fields changing. My thought is that we can wait for any new models to be introduced into 0.15, and then create migrations for whatever is necessary, in the interest of keeping the total number of migrations low and the history simple.

benjaoming: I would suggest this work flow: Create tonnes of new migrations! Then before the release, delete them and replace them with a single migration so we don't have multiple migrations in a release. I suspect KA Lite has shipped many times with multiple migrations for the same apps within one release.

jamalex:
Would be good to add a migration at this point. We can remove it and generate a new one after, before release, if anything else gets added here. Other than that, looks great.

jamalex:
Note https://github.com/learningequality/ka-lite/issues/4273 for this as well -- but can be done as a separate PR, after this is merged

MCGallaspy:
Now addresses one third of #4273.

kalite/main/content_rating_models.py
rtibbles: I approve of @benjaoming's suggested work flow.

benjaoming: Added the following to the Release Workflow:

> Squashing migrations
> With South: Delete all migrations that were added since the last release. Create one single migration for each app.

rtibbles: Can we combine schema and data migrations? (Admittedly data migrations are rare, but we do have one in 0.14).

benjaoming: They can actually be combined manually yes, just copy-paste the migration into where it fits. Admittedly, it may become a mess and perhaps worth putting separately in case it causes issues. AFAIK data and schemamigrations are made of the same and interpreted the same way by South.

kalite/main/content_rating_models.py
```diff
@@ -5,14 +8,28 @@
 from securesync.models import DeferredCountSyncedModel
 
 class ContentRating(DeferredCountSyncedModel):
+
+    minversion = "0.15.0"
+
     class Meta:
         unique_together = ("content_kind", "content_id", "user")
```
jamalex: Just to avoid weird possible edge-case collisions, probably best to update this to include `content_source` as well. May need to then regenerate the migration.

jamalex:
One small suggestion, and then looks dandy to me!

MCGallaspy:
I don't understand the cause and effect between that commit and this test failure, but I'll try to resolve it nonetheless.

jamalex:
Looks gUUID to me!

jamalex:
As it seems unlikely this broke coach reports, feel free to merge as deemed appropriate.

rtibbles:
I am rerunning the tests, as it appeared to be a timeout issue.

MCGallaspy:
That was it, however, I'd like to try to make the test timeout-proof if possible. Do you remember which one it was? I'll open a separate issue.

MCGallaspy:
@jamalex signed off, so merging.

Issue #3403, opened 2015-03-24
Dependency management: Unbundle python-packages


benjaoming:
# Overview

Currently, KA Lite ships with a directory `python-packages`, each of them have to be unbundled. That means, for some of them to be placed separately, and for others to be bundled with installers or as dynamic dependencies.

**Unbundling**: To remove a package from python-packages and include it as a dependency /or/ to completely remove the package as a dependency and write some new code.

**Scope**: If a package has been modified, we need to remove the causes of the modification from the kalite codebase. If the package has been updated, we should just pin it at its version from previously and schedule an upgrade for later in a separate issue (unless it's really trivial).
- [x] accenting.py #4706
- [x] announcements #4692
- [x] annoying #4690
- [x] async #4691
- [x] cherrypy #4693
- [x] collections_local_copy.py #4711
- [x] contextlib2.py #4712
- [x] dateutil #4264
- [x] debug_toolbar  #3691
- [ ] django
- [x] django_extensions - #3487
- [x] django_snippets #4714 #4822
- [x] docopt.py
- [ ] fle_utils
- [x] git #3784
- [x] gitdb #3785
- [x] httplib2 #4694
- [x] httreplay  #4695
- [x] ifcfg #4696
- [x] importlib #4697
- [x] iso8601 #4698
- [x] kaa
- [x] khanacademy #4699
- [x] khan_api_python #4700
- [x] memory_profiler.py #4701
- [x] mimeparse.py #4702
- [x] mock.py #3690
- [x] _multiprocessing.py #4703
- [x] oauth.py #4700
- [x] pbkdf2.py #4704
- [x] playground #3676
- [x] polib.py #4705
- [x] postmark #3677
- [x] pyasn1 #4707
- [x] pytz #3678
- [x] requests #5138 
- [x] rsa #4707
- [ ] securesync
- [x] six.py #3716
- [x] slugify #4708
- [x] smmap #4709
- [x] south
- [x] tastypie #3825
- [x] unittest2 #3675
- [x] werkzeug #3679
- [x] xlrd #4710
- [x] youtube_dl  #3696
# Offline Installers

It's important that an installer is portable, thus that it can easily be stored for offline usage.

For Mac / Windows, an approach could be to bundle a virtualenv. @ruimalheiro and @cpauya what are your opinions on this?

[...]

PR #4264, opened 2015-08-19
Remove dateutil, now it's in requirements.txt

benjaoming:
Seems a new dependency was snuck in through 763e153e3f0f7eed768c3eaa53bdb398958dcd15

#3403


benjaoming:
Preferably, this dependency should be completely removed, it's a bit overkill for these meagre purposes:

```
➜  ka-lite git:(kalite-command-proxy-status) ack-grep dateutil kalite
kalite/main/models.py
10:from dateutil import relativedelta

kalite/updates/api_views.py
4:import dateutil.parser
61:                start_time = make_naive(dateutil.parser.parse(request.GET["start_time"]), get_current_timezone())

kalite/basetests/tests.py
211:        "dateutil": "1.5",

kalite/control_panel/views.py
4:import dateutil
467:        period_end = dateutil.parser.parse(period_end)
468:        period_end = period_end + dateutil.relativedelta.relativedelta(days=+1, microseconds=-1)

kalite/facility/api_resources.py
3:from dateutil.tz import tzlocal
```

benjaoming:
I mean, 3700+ lines of code to parse a date is *too much* IMO.

aronasorman:
I see what you mean. Can we even get `datetime` to replace it? I know we can parse [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) dates there, and we can probably make our own parser using `datetime` as well as long as we know the format of the dates beforehand.

benjaoming:
I've opened up a separate issue on this, but for now, we can safely remove the package in python-packages as it's a dependency. It's still mentioned in LICENSE because we will continue to ship in the static build as long as its a dependency.

rtibbles:
I agree this is unnecessary - especially as we have seen that we cannot rely on server times, let alone their timezones for good timestamping - in fact, a lot of our code assumes as much.

benjaoming:
Merging this, sounds like there's plentiful of agreement to remove dateutil in #4266

PR #4265, opened 2015-08-19
Bumping version to 0.15 to distinguish current develop branch

benjaoming:
@rtibbles or @jamalex could you build a 0.15 assessment.zip?

Anything else that needs doing?


aronasorman:
You also need an entry in `version.yml`.

benjaoming:
@aronasorman so it's passing the tests, need anything else done?

Issue #4268, opened 2015-08-20
Slashes missing in Windows paths in docs


jamalex:
http://field.learningequality.org/docs/0.14/en/installguide/install_all.html#removing-user-data
has `C:UsersYourUsername.kalite`

PR #4269, opened 2015-08-20
Use double backticks for code sample; single backtick inappropriate

MCGallaspy:
Fixes https://github.com/learningequality/ka-lite/issues/4268.


jamalex:
Lovely! :cupid:

PR #4272, opened 2015-08-20
0.14.x merge and Zone space coach reports (currently live on central)

rtibbles:


rtibbles:
@aronasorman This is 0.14.x into develop, along with the zone spacing of the coach reports that is live on central now. Tests pass, should be ready to merge.

PR #4278, opened 2015-08-24
Much strings -- i18n performance fixes.

### LINKED ISSUE #4276: i18n js files are massive and slow to load
es.js is like... 40MB, is it? Takes several minutes to load the page on the Pi.

When pulling translations from KA, we should only put strings into the JS i18n files that will be used on the frontend. Most of them are only used on the backend.

This is a blocker for i18n on the Pi.

aronasorman:
Fixes #4276.

1. Restore the use of the `djangojs` domain. i18n js files are now less than 10KB.
2. Parse a lot of KA's JS strings, thanks to xgettext's builtin `JavaScript` parser.

### DIFF
--- a/kalite/i18n/__init__.py
+++ b/kalite/i18n/__init__.py
@@ -389,7 +389,7 @@ def update_jsi18n_file(code="en"):
     request.path = output_file
     request.session = {settings.LANGUAGE_COOKIE_NAME: code}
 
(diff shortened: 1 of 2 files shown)


aronasorman:
Very small, but possibly wide-ranging consequences in terms of i18n. We'll have to reupload the new strings since they now include a lot of KA's strings in there. We can probably get away with using "Translation Memory" in translating those strings though, since we already have KA's TM.

PR #4281, opened 2015-08-24
0.14 -- Detailed linux installation instructions & link cleanup

MCGallaspy:
Unsure if there's a corresponding issue for this. @jamalex this incorporates the new download redirect scheme. @EdDixon perhaps you can verify that the instructions here are accurate?


aronasorman:
Links work, text is sane. Merging.

MCGallaspy:
@benjaoming I believe you had wanted to see the url scheme we came up with.

MCGallaspy:
These can be easily changed at this point, but once they "go public" as it were, it might be more problematic.

benjaoming:
Alright, cool, thanks for that @MCGallaspy 

On a side-note, I would recommend that we fix the on screen information to make documentation redundant. I'm not even sure if the on-screen guide is so bad we need documentation for it.

MCGallaspy:
Yeah, I don't think it's bad or that it needs to be fixed, but I guess it can't hurt to have supplemental explanation? At any rate, I believe @jamalex had requested these changes last week, though please correct me if I misremember it.

PR #4284, opened 2015-08-25
Coach reports d3

sethachoi:
Used D3 to implement a radial graph for the hours (total vs spent on content) metric on coach reports. The given screenshot is... less than ideal because the total is like a billion times larger than the time spent on content, so it just looks like a blue circle. But it totally works I swear.
screen shot 2015-08-25 at 10 34 05 am


MCGallaspy:
Hey @sethachoi, you probably want to target the `develop` branch! `master` is currently version 0.13.

sethachoi:
o woops

PR #4285, opened 2015-08-25
Registration redirect does nothing when offline

### LINKED ISSUE #4282: Change require_registration decorator behavior
@jamalex
In branch `0.14.x` and `develop`, in `securesync.devices.decorators.require_registration`, the current behavior is to redirect to the `register_public_key_client` view if the device is not registered, regardless of whether it's online or not.

The desired behavior of the decorator is described by this table:

           | online | offline 
---------- | --------- | ------------
__registered__ |  do nothing   | do nothing
__not registered__ | redirect to registration page | do nothing

@jamalex does this affect any critical use cases? There _are_ workarounds -- register the devices before going offline, or register one device and then clone it to a number of installations not intended to ever be online, etc. Or in the simplest case where one simply doesn't wish to use the affected admin tabs at all (all the configuration is done through a settings module) then it has no effect.

MCGallaspy:
Fixes #4282. Replaces #4283. Implemented as a client test instead of a browser test (execution time of ~10s vs ~30s).

### DIFF
--- a/kalite/updates/tests/__init__.py
+++ b/kalite/updates/tests/__init__.py
@@ -4,3 +4,4 @@ from base import *
 from class_tests import *
 from command_tests import *
 from unicode_tests import *
+from regression_tests import *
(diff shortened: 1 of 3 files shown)


jamalex:
Looks great. Thanks for including a regression test!

PR #4293, opened 2015-08-26
Prevents django compressor happening dynamically per request

rtibbles:
Summary of changes:
* Set default to offline compression with Django Compressor.
* Run compression during setup.
* Turn offline compression off for dev.

### DIFF
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -455,6 +455,7 @@ class Command(BaseCommand):
         # Now deploy the static files
         logging.info("Copying static media...")
         call_command("collectstatic", interactive=False, verbosity=0)
+        call_command("compress", interactive=False, verbosity=0)
 
         # This is not possible in a distributed env
         if not settings.CENTRAL_SERVER:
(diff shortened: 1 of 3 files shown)


kalite/distributed/management/commands/setup.py
```diff
@@ -455,6 +455,7 @@ def handle(self, *args, **options):
         # Now deploy the static files
         logging.info("Copying static media...")
         call_command("collectstatic", interactive=False, verbosity=0)
+        call_command("compress", interactive=False, verbosity=0)
```
benjaoming: I think compressing relies on external libraries? Or not?

rtibbles: It has internally bundled libraries to do all the compressing by default, to use more efficient/'better' compilers you can set it to defer to external libraries.

rtibbles: c.f. http://django-compressor.readthedocs.org/en/latest/settings/#django.conf.settings.COMPRESS_CSS_FILTERS
http://django-compressor.readthedocs.org/en/latest/settings/#django.conf.settings.COMPRESS_JS_FILTERS

PR #4297, opened 2015-08-27
Update version.yml to 0.14.

aronasorman:


MCGallaspy:
What's the purpose of this file? Is it redundant along with version.py?

PR #4304, opened 2015-08-28
Convert README to RST (compatible with sphinx and PyPi), move badges

benjaoming:
Changes include:

 * Convert to RST with pandoc
 * Move badges to top

TODO:

 * [x] Include README contents in docs, eliminate redundant info, fix links
 * [x] Remove links to Github Wiki


benjaoming:
Okay not removing the Github Wiki link for now, too much stuff still needs to be migrated.

benjaoming:
@MCGallaspy assigning you ;)

I would think the only controversial part here is about including online badges on the documentation's index page, I think we could/should fix that later, but I'm not sure if it's really that severe. For now, the aim was more that we stop repeating information by actually including the README in the documentation as a kind of starting page. DRYish.

docs/index.rst
```diff
@@ -6,8 +6,12 @@
 KA Lite Documentation
 ===================================
 
-Welcome to the KA Lite Documentation page! Here, you will find all the information needed to set up the KA Lite software. Additionally, there's information on how to manage KA Lite and how to perform role-specific tasks. If you have any questions that are not addressed by the documentation and you have access to the Internet, please be sure to check our `FAQ`_ or `Forums`_!    
-
+Welcome to the KA Lite Documentation page! Here, you will find all the 
+information needed to set up the KA Lite software. Additionally, there's
```
MCGallaspy: My preference is now to split up lines by punctuation instead of by some character limit. Subsequent lines without intervening newlines are grouped together anyway, and it makes it semantically easier to read. In this case, I'd do this:

```rst
Welcome to the KA Lite Documentation page!
Here, you will find all the information needed to set up the KA Lite software.
Additionally, there's information on how to manage KA Lite and how to perform role-specific tasks.
If you have any questions that are not addressed by the documentation and you have access to the Internet, please be sure to check our `FAQ`_ or `Forums`_!    
```

Plus, it creates a bias towards shorter sentences when you're writing -- which is beneficial in my mind!

MCGallaspy:
As I've said before, I'm hesitant to have multiple versions of the docs floating around -- namely RTD vs our own server. After our many discussions I have no preference in principle for one over the other (both have different advantages and disadvantages), but I do insist that we pick one and stick to it. In practice we're already linking to our own docs server from the FLE home site, which gives us an incentive to stick with that.

docs/index.rst
benjaoming: I was editing in nano, and I couldn't read the line, so I just CTRL+J'ed it. Wrapping at 80 chars matters IMO.

It also helps you not creating too long paragraphs -- equally important as the sentence length ;)

benjaoming: Btw... this wasn't wrapped at 80 chars... I was lazy.

MCGallaspy:
Looks ok. Re: my above comment, we discussed in the dev room about using RTD and decided to do it.

PR #4305, opened 2015-08-28
DB Backup: Fix backup_filepath referencing issue and error out invalid filepath

SG345:
DB Backup currently has a functionality for restoring backup from a given file. This PR fixes the referencing issue in 'restorebackup' management command which occurs while restoring a backup from a given file. Also, prints an error message if the given filepath is invalid.

### DIFF
--- a/kalite/distributed/management/commands/restorebackup.py
+++ b/kalite/distributed/management/commands/restorebackup.py
@@ -18,9 +18,12 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
+        backup_filepath = options['backup_filepath']
         if(options['backup_filepath']):
             if(os.path.isfile(backup_filepath)):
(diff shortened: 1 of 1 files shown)


kalite/distributed/management/commands/restorebackup.py
```diff
@@ -18,9 +18,12 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args, **options):
+        backup_filepath = options['backup_filepath']
         if(options['backup_filepath']):
             if(os.path.isfile(backup_filepath)):
                 call_command('dbrestore', filepath=backup_filepath, database='default') #perform restore from the given filepath
+            else:
+                print 'The given file was not found.'
```
aronasorman: Maybe add a `sys.exit(1)` after this print statement, to make sure the command exits.

PR #4322, opened 2015-08-31
Docs for rtd

benjaoming:
Contains the following:

 * Remove selenium from python-packages, didn't work with latest Firefox anyways
 * Added updated screenshots to repo
 * Using RTD theme
 * Move the docs/ so it'll build on ReadTheDocs.org


benjaoming:
How it looks:

screenshot from 2015-08-31 23 24 24

benjaoming:
And for the part with the screen shots:

screenshot from 2015-08-31 23 25 53

benjaoming:
I guess we can live with the fact I didn't make Circle CI work :)

MCGallaspy:
I'll review this either today or tomorrow, I've seat a deadline to merge this for Sept 3 on the dev calendar.

benjaoming:
@aronasorman this is intended to be merged into 0.13.x so no need to push it for 0.16. It's very critical that it's prioritized, because it's intended to unify all the docs from 0.13 and up on RTD.

MCGallaspy:
When you say you "Added updated screenshots to repo" @benjaoming, where did you get them from? Did you build them by running sphinx-build?

circle.yml
```diff
@@ -0,0 +1,18 @@
+machine:
```
MCGallaspy: Why add this at all?

benjaoming:
> When you say you "Added updated screenshots to repo" @benjaoming, where did you get them from? Did you build them by running sphinx-build?

Yes! They were also generated using the updated screen resolution patch. Are they okay to your knowledge?

circle.yml
benjaoming: Oh this was an attempt to get the 0.13.x branch to build on Circle, AFAIK it was working on Travis before.

docs/kalite-theme/breadcrumbs.html
```diff
@@ -0,0 +1,28 @@
...
+          {% elif display_bitbucket %}
+            <a href="https://bitbucket.org/{{ bitbucket_user }}/{{ bitbucket_repo }}/src/{{ bitbucket_version}}{{ conf_py_path }}{{ pagename }}{{ source_suffix }}" class="fa fa-bitbucket"> Edit on Bitbucket</a>
+          {% elif show_source and source_url_prefix %}
+            <a href="{{ source_url_prefix }}{{ pagename }}{{ source_suffix }}">View page source</a>
+          {% elif show_source and has_source and sourcename %}
+            <a href="{{ pathto('_sources/' + sourcename, true)|e }}" rel="nofollow"> View page source</a>
+          {% endif %}
+        {% endif %}
+        {% if not display_github %}
+          <a href="http://github.com/learningequality/ka-lite/blob/{{ theme_git_branch }}/sphinx-docs/{{ pagename }}.rst" class="fa fa-github">
```
MCGallaspy: Why not remove this part, since this already incorporates a link to github above? Moreover, this is misconfigured because the variable `theme_git_branch` set in `theme.conf` below is inaccurate.

circle.yml
benjaoming: ...but we have switched off Travis.

MCGallaspy:
> Are they okay to your knowledge?

Checking.

docs/kalite-theme/breadcrumbs.html
benjaoming: This is intended for local builds, the above is for RTD. You are right that theme.conf needs fixing, thanks, doing that now!

circle.yml
MCGallaspy: I think if we begin to actively work on the app in this branch then we should turn on tests, but until then there's no need.

benjaoming: Oh this is just a first stab, we can always fix it later... I don't see a problem in merging in circle.yml, at least then we can start figuring out why it isn't running whenever we have the next PR.

MCGallaspy:
The main claim (building with `make html` without launching the screenshots directive) works, committed screenshots are not broken, should work fine on RTD. Now we need to update fle-home to point to RTD.

benjaoming:
@MCGallaspy thanks for merging and commenting!

I've now attempted the build on RTD, however the current global setting to install requirements using `requirements_sphinx.txt` is broken because I didn't think about adding this file, so I'm opening up another PR for this....

PR #4327, opened 2015-09-01
Delete the initdconfig management command.

### LINKED ISSUE #4137: Delete deprecated "initdconfig" management command
Deprecated by #3736

aronasorman:
Fixes #4137. These platform specific configuration files should be done within the installers.


benjaoming:
Agreed, there's also enough time in case the OSX installer still uses it, which I'm pretty sure has been disconfirmed previously. We've kept it and put a deprecation warning even.

PR #4328, opened 2015-09-01
Remove duplicate requirements

MCGallaspy:
These requirements are duplicated in `requirements_test.txt`, resulting in an error when you run `pip install -r requirements_dev.txt`.


rtibbles:
So, are we to assume that requirements for dev = requirements + requirements_test + requirements_dev ?

MCGallaspy:
That's the implication, isn't it? I think we have requirements_test.txt for the test server, since there may be some modules used in dev not used in testing.

PR #4332, opened 2015-09-01
Selectively hide hint indicator if not hints available.

### LINKED ISSUE #4001: For Mac OS, some Hint has no content but doesn't show system status
When there is no hint for some exercises, the users are not aware of that and will miss click and waste hints. Maybe better to show something like "No hint available" or disable the button.

rtibbles:
Fixes #4001

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -208,6 +208,15 @@ var ExerciseView = BaseView.extend({
                 answerType = $("span.sol").map(function(index, item){return $(item).attr("data-forms");}).get().join();
             }
 
+            var hints;
+            if (self.data_model.get_framework() == "khan-exercises") {
+                hints = data.hints;
+            } else if (self.data_model.get_framework() == "perseus") {
+                hints = Exercises.PerseusBridge.getNumHints() > 0;
+            }
+
+            self.trigger("hint_available", hints);
+
             var checkVal = /number|decimal|rational|proper|improper|mixed|radical|integer|cuberoot/gi;
 
             if (checkVal.test(answerType)){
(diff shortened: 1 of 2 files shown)


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -712,6 +721,8 @@ var ExercisePracticeView = ExerciseWrapperBaseView.extend({
         });
 
         this.listenTo(this.exercise_view, "check_answer", this.check_answer);
+
+        this.listenTo(this.exercise_view, "hint_available", this.toggle_hint_view);
```
MCGallaspy: `toggle_hint_view` takes an argument, but my understanding is that this form will not call toggle_hint_view with an argument. Is that correct? Can you use [_.partial](http://underscorejs.org/#partial) here?

MCGallaspy: Disregard, I see that it's handled by [Events.trigger](http://backbonejs.org/#Events-trigger).

MCGallaspy:
I am trying to locate an exercise without hints, in order to test this out -- quite tricky, any way to search through exercises/assessment items systematically?

rtibbles:
Would have to load up all the assessment items, parse them all, and then check to see if the hints property is an empty list or not.

I assume one must exist for the bug report to exist, but I could not find one either.

MCGallaspy:
> Would have to load up all the assessment items, parse them all, and then check to see if the hints property is an empty list or not.

My conclusion too. Wrote a quick script:
```python
for ai in AssessmentItem.objects.all():
    item_data = json.loads(ai.item_data)
    if not len(item_data.get("hints", [])):
        print ai.id

MCGallaspy:
Some of the assessment items for this content item have no hints, for example: `/learn/khan/computing/computer-science/algorithms/graph-representation/quiz-representing-graphs/`

Seems to work great. Merging.

PR #4333, opened 2015-09-01
Fix back button behaviour for sidebar.

### LINKED ISSUE #4310: Sidebar back button malfunction
Branch: 0.14.x
Browser: Chrome
OS: Mac
Reproduce: go to learn page to bring up the sidebar, then extend the siderbar to make the back button shown up. When there're more than two levels of back action, the first few clicks won't retrieve the sidebar until you reach the root level and the sidebar will suddenly display the root level.

rtibbles:
Fixes #4310


66eli77:
texted and fixed

PR #4344, opened 2015-09-02
0.15.x to develop

aronasorman:


aronasorman:
Look at all those check marks.

MCGallaspy:
The `requirements_dev.txt` fix on 0.15.x is necessary to build develop docs, so merging.

Issue #4336, opened 2015-09-02
Videos not displayed correctly


MCGallaspy:
On branch `0.15.x` on Windows 8, with videos in my ~/.kalite content directory from previous installation, when I log in as a learner and navigate to an available video I see this:

![ss2](https://cloud.githubusercontent.com/assets/8888020/9620440/efe5d490-50ce-11e5-8074-476b289e58cb.png)

Note how the pane containing the video is almost totally collapsed. Resizing the window results in the js error in the picture: `Uncaught TypeError: this.set_container_size is not a function` variously either from `bundle_videos.js` or `video/views.js` -- odd that both are included?

rtibbles:
Assume this is fixed as relevant PR has been merged.

PR #4345, opened 2015-09-02
Fix video player auto-resizing, and set default height, to avoid jump

jamalex:
Fixes https://github.com/learningequality/ka-lite/issues/4336

### DIFF
--- a/kalite/distributed/static/css/distributed/video-js-override.css
+++ b/kalite/distributed/static/css/distributed/video-js-override.css
@@ -1,5 +1,9 @@
 /* Responsive content display for tablets */
 /* Jessica TODO: take off !important tags */
+
+#video-player {
+    height: 500px;
+}
 div.video-player {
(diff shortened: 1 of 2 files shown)


aronasorman:
is there an issue for this?

jamalex:
Yes.

MCGallaspy:
Please be sure to set milestone and assignee @jamalex.

MCGallaspy:
Okay, this seems to work fine, except that now there is a lot of whitespace below the footer on _available_ videos _if_ you're not logged in as a learner. Doesn't occur if you're logged in as a learner or viewing an unavailable video. I want to check if it exists on `0.15.x`, and if it does I'll open a separate issue.

I'm disappointed by the test failure here, but it seems to be unrelated.

MCGallaspy:
The whitespace issue does indeed exist on `0.15.x`

PR #4347, opened 2015-09-02
Add a scenario option to generaterealdata mgmt command

MCGallaspy:
I wanted to test out coach reports with a more specific scenario.

* Adds a new kwarg to `generate_fake_facility_users` to make coaches. Since coach-facility group associations are not possible within the app, but permissible at a model level, I'm careful not to set the facility group for coaches.
* Adds a new option to `generaterealdata` (`--scenario-1`) as described in the help text. Behavior is exactly the same if this option is not passed.


rtibbles:
Looks fine, low risk. Merging.

PR #4359, opened 2015-09-03
Fixes to contentload and friends for loading latest content from KA for 0.15

jamalex:
Updated content itself coming in a separate PR.


jamalex:
Closing in favor of #4360, as the JSON files get collapsed down anyway, and review is easy.

PR #4370, opened 2015-09-03
Content rating - clean csv output

### LINKED ISSUE #4367: Content rating - extraneous columns in csv export output; filter empty rows
The columns look like this:
image

In particular, `deleted` and `resource_uri` columns could be removed, they're not helpful.

Rows are also created when a user visits a content item page for the first time, but if the row is empty (a value of 0 for both ratings), then perhaps we should filter it from the output -- it's indicative that the user did not interact with the rating widget at all, otherwise at least one rating would have a value between 1 and 5.

MCGallaspy:
Fixes #4367. See commit message:

```
Exportable content ratings are now filtered to be non-empty, where
  we take non-empty to mean that both the quantitative ratings are not 0.
  If they _are_ both 0, it indicates the user has not interacted with the
  rating widget at all.

Some irrelevant columns are also removed from the result.
```

 Now the output looks like this:

image

### DIFF
--- a/kalite/control_panel/api_resources.py
+++ b/kalite/control_panel/api_resources.py
@@ -370,8 +370,9 @@ class ContentRatingExportResource(ParentFacilityUserResource):
         queryset = ContentRating.objects.all()
         resource_name = 'content_rating_csv'
         authorization = ObjectAdminAuthorization()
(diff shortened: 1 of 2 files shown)


rtibbles:
Looks good to me.

PR #4371, opened 2015-09-03
Don't double import en.js.

aronasorman:
Or else the catalog files will reset to empty.


rtibbles:
Looks good, will merge when tests pass.

PR #4372, opened 2015-09-03
Wrap some JS strings in coachreports views.js.

aronasorman:
@rtibbles I'm not sure if these strings are displayed, but I wrapped them anyway. Can you confirm if they are?


rtibbles:
They are, good catch.

PR #4374, opened 2015-09-04
Content rating -- only show on available content

### LINKED ISSUE #4309: Content rating visible even when content is unavailable
Soon on `develop` (but for now on my branch `beautify-content-rating`) the rating widget is visible even when content is unavailable. @rtibbles [suggests that if the content is unavailable the widget should not appear.](https://github.com/learningequality/ka-lite/pull/4248#issuecomment-133592817)

MCGallaspy:
Fixes #4309, and also fixes two other bugs I discovered while working on that.
See the commit message, where I describe the changes:

```
Check if content is available, if it's not then don't show the rating.
  It took me a while to figure out that the "model" being passed around
  was actually the entry in the sidebar itself... :(

Improve clarity by trimming down `show_rating` function a bit
  and adding comments.

First error: Define a close function for RatingView and call it
  instead of BaseView.remove, which removes the container element from
  the DOM, thus foiling subsequent renders.

Second error: initialize rating_view's model property with
  blank "content_id", "content_kind" -- if that initalization
  logic is run it will cause the subsequent display logic to not
  execute, since it's conditional on "content_id".
```

### DIFF
--- a/kalite/distributed/static/js/distributed/rating/views.js
+++ b/kalite/distributed/static/js/distributed/rating/views.js
@@ -108,7 +108,7 @@ module.exports = BaseView.extend({
             },
             patch: true
         });
-    },
+    }
 });
 
 
--- a/kalite/distributed/static/js/distributed/rating/views.js
+++ b/kalite/distributed/static/js/distributed/rating/views.js
@@ -108,6 +108,12 @@ module.exports = BaseView.extend({
             },
             patch: true
         });
+    },
+
+    close: function() {
+        // Defined so that when BaseView.remove is called, we just clear the html instead
+        // of removing the view's element from the DOM.
+        this.$el.html("");
     }
 });
 
--- a/kalite/distributed/static/js/distributed/rating/views.js
+++ b/kalite/distributed/static/js/distributed/rating/views.js
@@ -109,12 +109,6 @@ module.exports = BaseView.extend({
             patch: true
         });
     },
-
-    close: function() {
-        // Defined so that when BaseView.remove is called, we just clear the html instead
-        // of removing the view's element from the DOM.
-        this.$el.html("");
-    }
 });
 
 
--- a/kalite/distributed/static/js/distributed/topics/views.js
+++ b/kalite/distributed/static/js/distributed/topics/views.js
@@ -67,10 +67,9 @@ var ContentAreaView = BaseView.extend({
     },
 
     remove_rating_view: function() {
-        // Remove the rating view if it exists. Uses rating_view.close, so that the container element is not removed
-        // from the DOM.
+        // Remove the rating view if it exists.
         if (typeof this.rating_view !== "undefined") {
-            this.rating_view.close();
+            this.rating_view.remove();
             delete this.rating_view;
         }
     },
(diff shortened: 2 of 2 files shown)


kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -108,6 +108,12 @@ module.exports = BaseView.extend({
             },
             patch: true
         });
+    },
+
+    close: function() {
+        // Defined so that when BaseView.remove is called, we just clear the html instead
```
rtibbles: Why is this desired behaviour?

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -56,24 +56,46 @@ var ContentAreaView = BaseView.extend({
         this.$(".content").html("").append(view.$el);
     },
 
+    should_show_rating: function() {
+        /*
+        This function determines whether a rating should be shown for the content item.
+        returns: true or false
+        */
+        var entry_available = (typeof this.model !== "undefined") && !!this.model.get("available");
+        var logged_in = typeof window.statusModel.get("user_id") !== "undefined";
```
rtibbles: Could also use the ```.has``` method here. http://backbonejs.org/#Model-has

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -56,24 +56,46 @@ var ContentAreaView = BaseView.extend({
...
+        return logged_in && entry_available;
+    },
+
+    remove_rating_view: function() {
+        // Remove the rating view if it exists. Uses rating_view.close, so that the container element is not removed
```
rtibbles: Still not sure I understand the rationale for this, given that we delete the reference to the view - do we now have a view without a reference lying around in the DOM?

kalite/distributed/static/js/distributed/rating/views.js
MCGallaspy: This view's `el` property is assigned to an element in the template of the parent view. When `remove` is called this view's `el` is removed from the DOM. What I have chosen to do instead is clear the html of the `el` so that it remains attached to the DOM. Then new instantiations of RatingViews can be attached to the same element

kalite/distributed/static/js/distributed/topics/views.js
MCGallaspy: Okay.

kalite/distributed/static/js/distributed/topics/views.js
MCGallaspy: When this is called, the intent is that we're totally done with the view and want to get rid of it. The reference is deleted so that the object can be garbage collected. I want to clear the html of the view, but I also want to keep the view's `el` (which belongs to the template of the parent view) attached to the DOM, so that if we make a new RatingView later we can attach it to the same element with no new logic.

rtibbles: It sounds like you want to keep the DOM element around, but don't care so much about the view.

I would recommend appending the view.el to the DOM element in question, therefore, rather than setting it as the 'el' property on instantiation. You will just end up with an extra nested div, which should make no difference.

kalite/distributed/static/js/distributed/rating/views.js
```diff
@@ -108,7 +108,7 @@ module.exports = BaseView.extend({
             },
             patch: true
         });
-    }
+    },
```
rtibbles: Trailing commas break Explorers of the Internet.

PR #4376, opened 2015-09-04
Coverage metrics for 0.15.x #3

benjaoming:
Okay, stuff's applying cleanly... you dig @aronasorman ? :)

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,2 +1,0 @@
-repo_token: YWMKkAVqIigWxX8XerfykVab17vEKmdXO
-
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,1 +1,2 @@
 repo_token: YfchfjB9wbHxDS7MDadVe09f2StV0pds1
+service_name: circleci
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,1 @@
+repo_token: YfchfjB9wbHxDS7MDadVe09f2StV0pds1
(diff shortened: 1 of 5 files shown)


benjaoming:
Btw. as I remember it, we need to push a change to the learningequality repo before we start seeing coverage metrics added to PRs.

benjaoming:
Builds so far:
https://coveralls.io/github/learningequality/ka-lite

.coveralls.yml
```diff
@@ -0,0 +1,2 @@
+repo_token: YfchfjB9wbHxDS7MDadVe09f2StV0pds1
```
aronasorman: This is ok to be public?

aronasorman:
We should discuss what we want to do about this new code coverage metric that we have. What's our timeline for bringing that to ~100%? And who would be responsible for that?

Also @benjaoming is there a way to view which lines aren't tested?

.coveralls.yml
benjaoming: yes, necessary

benjaoming:
> Also @benjaoming is there a way to view which lines aren't tested?

See the --omit option for coverage in circle.yml

aronasorman:
@benjaoming can you reopen another one for develop? I'd rather move all new development in there. There's a merge conflict anyway, so might as well :)

benjaoming:
I'll open up another one for 0.15.x

aronasorman:
For 0.15.x? You can also just reopen this.

But I disagree about opening for 0.15.x. 0.15 will go into bugfix mode soon, and I don't see us writing any substantial number of tests on that branch. Given that, I don't see any use for coverage metrics for 0.15.x anymore.

benjaoming:
Because look at the diffset :)

> Given that, I don't see any use for coverage metrics for 0.15.x anymore.

Well, the whole point of collecting coverage metrics is to see the changes! So we need to start collecting that data at some point.

PR #4381, opened 2015-09-04
Removes Django Compressor.

### LINKED ISSUE #4379: Remove django-compressor in 0.15.x
It's no longer used. See this patch in particular: https://github.com/learningequality/ka-lite/commit/9ba0f55c47108edad94f94f69aa45b8dc307f99d#commitcomment-13076497

rtibbles:
Fixes #4379


kalite/testing/testrunner.py
```diff
@@ -111,6 +115,15 @@ def __init__(self, *args, **kwargs):
         self._bdd_only = kwargs["bdd_only"]  # Extra options from our custom test management command are passed into
         self._no_bdd = kwargs['no_bdd']      # the constructor, but not the build_suite function where we need them.
 
+        # Django < 1.7 serves static files using the staticfiles app, not from static root.
+        # This causes django_js_reverse not to get served to the client, so we manually copy it into distributed.
+
+        call_command("collectstatic_js_reverse", interactive=False)
+
+        ensure_dir(os.path.join(os.path.dirname(os.path.dirname(__file__)), "distributed", "static", "django_js_reverse", "js"))
```
MCGallaspy: Should this be:
os.path.dirname(\_\_file\_\_)), "distributed", "static", __"js"__, "django_js_reverse", "js"
?

rtibbles: Nope.

MCGallaspy: Yeah, you're right. Much embarrassment.

MCGallaspy:
Maybe also increase the wait time [here](https://github.com/rtibbles/ka-lite/blob/ed86d12c8d17da233129857bd9e393da2123b393/kalite/distributed/features/steps/search.py#L26). I'm guessing the auto-complete logic is just taking longer given you know input each key one-by-one.

rtibbles:
The main issue with the search auto-complete (aside from the jQuery UI plugin not responding well to instantaneous filling) is that the JSON blob for search is 1.1Mb, which takes about 14 seconds to load when running the testing server locally (and hence probably takes even longer on Circle).

rtibbles:
This should be fixed in 0.16 when #4261 gets merged, as I have refactored the search there to use a query endpoint, rather than dumping the entire JSON blob into the client.

MCGallaspy:
I made my suggested changes in #4293.

PR #4384, opened 2015-09-07
Fix #4382 -- we are MIT licensed

### LINKED ISSUE #4382: License incorrectly marked in setup.py
Setup.py indicates that KA Lite is GPL, whereas it is actually MIT license:
https://github.com/learningequality/ka-lite/blob/0.15.x/setup.py#L355

benjaoming:
#4382

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -352,7 +352,7 @@ setup(
     author_email="info@learningequality.org",
     url="https://www.learningequality.org",
     description=DIST_DESCRIPTION,
-    license="GPLv3",
+    license="MIT",
     keywords=("khan academy", "offline", "education", "OER"),
     scripts=['bin/kalite'],
     packages=find_packages(exclude=["python-packages"]),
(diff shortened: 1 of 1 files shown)


aronasorman:
We probably need to include the license on top of each source file too at some point :)

jamalex:
Cool, thx -- I think it's best to have it MIT for now, but we could revisit earlier discussions we had (a couple years ago) about switching to a copyleft license like GPL, and examine how it might affect partnerships, strategy, or other practical issues.

jamalex:
I'm assuming the `OfflineGenerationError: You have offline compression enabled but key "9818b417fc796fd10733c626222e7217" is missing from offline manifest. You may need to run "python manage.py compress".` error is unrelated. :P Merging!

benjaoming:
@jamalex that's one of the many many test failures I'm trying to get rid of in #4385 and have finally succeeded in running tests locally.

PR #4385, opened 2015-09-07
Fix lots of unresolved imports / undefined stuff

benjaoming:
It worries me to see all the issues in 0.14.x :/ Luckily most issues are in tests.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -11,6 +11,7 @@ dependencies:
 
 test:
   override:
+    - kalite manage compress
     - kalite start --traceback -v2
     - kalite status
     - kalite stop --traceback -v2
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,6 @@ dependencies:
 
 test:
   override:
-    - kalite manage compress
     - kalite start --traceback -v2
     - kalite status
     - kalite stop --traceback -v2
--- a/kalite/caching/tests/cache_tests.py
+++ b/kalite/caching/tests/cache_tests.py
@@ -9,7 +9,7 @@ from django.conf import settings; logging = settings.LOG
 from django.test.client import Client
 from django.utils import unittest
 
-from .. import caching
+from ... import caching
 from kalite.testing.base import KALiteTestCase
 from kalite.topic_tools import get_content_cache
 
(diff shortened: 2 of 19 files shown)


benjaoming:
Notice! All the BDD tests are running now, except what I've disabled.

benjaoming:
Yay, we're now at this stage: Node 1 (non-BDD) times out.

benjaoming:
Good news: All the tests are now running on my local machine (non BDD) ! The BDD tests are already running on Circle.

MCGallaspy:
> It worries me to see all the issues in 0.14.x 

Why? Hadn't we decided to address those issues in (then) develop as a practical compromise, given that as you say the issues appear to be only test related? Trying to address them here is duplicating that effort from other branches in an incompatible way.

benjaoming:
The sentence was incomplete, it worries me that we have all these issues in the tests! It worries me that we released 0.14 without being able to run tests, locally nor in Travis/Circle.

The fixes I applied were very obvious stuff... broken imports etc...

MCGallaspy:
Yeah I get that it is worrisome, but you're rehashing the discussion we already had about this very topic, wherein we decided that it was better to release something that we had tested extensively by hand than to delay even further by working on protracted test issues.

And my point about incompatibility is not that your changes are non-obvious, but that you're taking a different approach than we already took in 0.15.x and develop to get the tests working. Like (this)[https://github.com/benjaoming/ka-lite/commit/182126a46659e167c9370c0c81fec8b05dc4f4bb#diff-daf6a8e2e2a3ca1e28ec7687d750febfR85]. Arguably if you're going to have an empty step you might as well delete it, otherwise it gives the wrong impression that something is being tested and passing when it's not.

My opinion is that we should focus on the more pressing 0.15.x and develop branches than worry about 0.14.x any more. Thus my question -- why worry about failing tests on 0.14.x? What value does it add? Or if you must worry about 0.14.x tests, why didn't you say anything back when we were discussing this before the 0.14.0 release? This is an honest question -- do you feel you were not included in that discussion?

benjaoming:
I asked a couple of times if anyone was gonna run the tests locally before releasing. I did say that, and I didn't follow up on it until now when I was doing a change for 0.14 and I started wondering about the test situation.

Then I found that the tests had lots of errors, and I'm suggesting that we don't do any patch releases without having functioning tests. I assumed that the test fixes from 0.15 would be too incompatible to backport.

Rather than complaining, I made a PR, I hope you can see the difference here ;)

Issue #4396, opened 2015-09-08
Duplicated exercise in topic tree


rtibbles:
For some reason, we have two almost completely identical exercises in the topic tree (I believe this is probably due to an editing error on KA's part, where the data has been inserted twice under two slightly different names).

This causes us to have non-unique paths.

Data below:
```
[{'all_assessment_items': ['{"sha": "d124afecdeb491da7ef297aafde30f824d142309", "live": true, "id": "x27af6ca3e3dd6586", "perseus_api_major_version": 3, "requires_screen_or_mouse": false}',
   '{"sha": "e71f206d4e3b0b4fc47d62a5b06244d4df34918b", "live": true, "id": "x503f312d62af87a3", "perseus_api_major_version": 3, "requires_screen_or_mouse": false}',
   '{"sha": "488f2b1dedb412b6a603b2434fcc4af9b4380481", "live": true, "id": "xa11ed08fb084c2fd", "perseus_api_major_version": 3, "requires_screen_or_mouse": false}',
   '{"sha": "e246cb8cc6e4b79a645278d0d9d0646b295b73a7", "live": true, "id": "x861e7f6afd36266d", "perseus_api_major_version": 3, "requires_screen_or_mouse": false}',
   '{"sha": "3bc3136183c1490e9b76a96121d2112dad36484a", "live": true, "id": "xf0f42f8dcea44fed", "perseus_api_major_version": 3, "requires_screen_or_mouse": false}'],
  'available': False,
  'basepoints': 10.0,
  'description': 'Passages that test your knowledge on fluids in motion',
  'display_name': 'A fluid flowing through a tube',
  'exercise_id': 'fluids-in-motion---passage-1',
  'id': 'fluids-in-motion---passage-1',
  'kind': 'Exercise',
  'name': 'fluids-in-motion---passage-1',
  'path': 'khan/test-prep/mcat/physical-sciences-practice/physical-sciences-practice-tut/fluids-in-motion-passage-1/',
  'prerequisites': [],
  'seconds_per_fast_problem': 4.0,
  'slug': 'fluids-in-motion-passage-1',
  'title': 'A fluid flowing through a tube',
  'uses_assessment_items': True},
 {'all_assessment_items': ['{"sha": "11e79a919a2dd45b896dcf94ac816bfc8ad1d242", "live": true, "id": "xf62865128a3d3ae7"}',
[...]

PR #4397, opened 2015-09-09
Don't assume that single kind topics won't have duplicate slugs.

rtibbles:
This will require a rerun of contentload, but not for assessment items.


rtibbles:
Fixes #4396

rtibbles:
Also removes items not in the topic tree from content caches.

Reran contentload too to update content caches.

```
INFO:kalite:Downloaded topic_tree data for
            1806 exercises
            7393 content files
```

kalite/contentload/management/commands/channels/khan.py
```diff
@@ -139,6 +139,12 @@ def build_full_cache(items, id_key="id"):
         except AttributeError:
             logging.error("Unable to serialize %r" % item)
 
+        item = whitewash_node_data(item)
+
+        if ids:
+            if item["id"] not in ids:
```
jamalex: Should this use `id_key` (or what's that for?) -- and is `ids` a blacklist? May be a better name for it?

kalite/contentload/management/commands/contentload.py
```diff
@@ -37,6 +37,20 @@ def scrub_topic_tree(node_cache, channel_data):
                     del node[att]
 
 
+def topic_tree_id_catalog(topic_tree):
+    # Takes the topic tree and returns a dictionary with the ids of all nodes in the tree
+    # This is for use in removing items from caches that are not in the topic tree
```
jamalex: Docstring?

kalite/contentload/management/commands/contentload.py
```diff
@@ -37,6 +37,20 @@ def scrub_topic_tree(node_cache, channel_data):
...
+    ids = {}
```
jamalex: Why not just a set...?

kalite/contentload/management/commands/channels/khan.py
rtibbles: I think id_key is actually redundant at this point, it was being used previously to double check things prior to whitewash_node_data being applied, but after whitewash_node_data, every node has 'id' as its id_key.

PR #4398, opened 2015-09-09
remove absolute with of the `th.headrow, th.headrow div`

### LINKED ISSUE #4340: Text display issues on tabular report
On `0.15.x`, this is really two separate issues.

First, long words in content titles have weird line breaks. See screenshot. Can we perhaps expand the column size to fit the longest word?

Secondly, the user's name (first + last name, not the username) overflows the row. See screenshot. Overflow is bad, scrolling seems less than ideal, and truncating names may be confusing. If the row must be fixed-height, then I suggest using smaller font size for too-large names. Otherwise I suggest expanding the height of the row.

ss3

arceduardvincent:
Hi @aronasorman and @MCGallaspy  - this fixes #4340

**Branch:** 0.15.x

**Expected Behaviour:** Able to remove the unwanted line breaks in the titles. 
screenshot 2015-09-09 14 57 23

**Current Behavior:**  Long words in content titles have weird line breaks.
screenshot 2015-09-09 14 56 41

I remove the declared absolute value of the `th.headrow` that causes to have line breaks in the titles.

### DIFF
--- a/kalite/coachreports/static/css/coachreports/tabular_view.css
+++ b/kalite/coachreports/static/css/coachreports/tabular_view.css
@@ -61,7 +61,6 @@ table.tab_report th.username {
 }
 
 th.headrow, th.headrow div {
-    width: 80px;
     table-layout: fixed;
     font-weight: bold;
     word-wrap: break-word;


rtibbles:
This is the wrong fix.

Intended behaviour is explicitly not to dynamically size the columns independently. Rather, the fix suggested by @MCGallaspy in #4340 is that we should change the overall size of columns depending on the longest word appearing.

PR #4403, opened 2015-09-09
 Fix non-deterministic behaviour for search autocomplete behave test

rtibbles:
Summary of changes:
* Wait for search autocomplete JSON blob to load before proceeding.
* Search for 'basic' rather than 'basic addition'.

Let's see what our circular overlord says!


MCGallaspy:
Re: the content rating availability issue, I wonder if it might be helpful to create a test-only topic tree with test-friendly paths? Something like:

```
/learn/khan-exercise
/learn/assessment-item/available
/learn/assessment-item/unavailable
/learn/video/available
...
```

MCGallaspy:
@rtibbles can't say why the content rating stuff is now failing, it passes locally. My hypothesis is that it's truly just taking a long time for the page to completely load. As for the other old-style test, again it seems to be a long page load.

rtibbles:
Yeah, this definitely worked for me too locally.

MCGallaspy:
Let's merge in #4412 and re-run the tests, in order to observe the screencast for failing tests. Will provide clues.

MCGallaspy:
I'll bet if we merge this https://github.com/learningequality/ka-lite/pull/4416, then everything will work beautifully.

PR #4409, opened 2015-09-10
Reduce number of ajax requests in content rating

### LINKED ISSUE #4407: On page load, content rating makes 4 POST requests, 2 of which return a 400 status.
Branch: 0.15.x

Expected Behaviour: We shouldn't have to make any saves until something actually changes in the content rating, and all those should be well formed requests.

Current Behaviour: Content Rating attempts four saves immediately on page load, even though I have not interacted with the content rating widget at all. 2 of these attempts at saves return a 400 error from the server.

image

MCGallaspy:
Fixes #4407. @rtibbles I wasn't able to replicate the exact circumstances you reported, but I did pare down the # of requests as much as possible. See commit messages for details. Also addresses another issue I noticed.

There is an extra PUT request after POST requests to create a new rating model. It seems to be done automatically -- can we live with that?

### DIFF
--- a/kalite/distributed/static/js/distributed/topics/views.js
+++ b/kalite/distributed/static/js/distributed/topics/views.js
@@ -84,27 +84,22 @@ var ContentAreaView = BaseView.extend({
 
         // Secondly, if the rating_view is previously deleted or never shown before at all, then define it.
         if( typeof this.rating_view === "undefined" ) {
-            this.rating_view = this.add_subview(RatingView, {
-                model: new RatingModel({
-                    "user": window.statusModel.get("user_uri"),
-                    "content_kind": "",
-                    "content_id": ""
-                })
-            });
+            this.rating_view = this.add_subview(RatingView, {});
             this.$("#rating-container-wrapper").append(this.rating_view.el);
         }
 
         // Finally, handle the actual display logic
-        if( this.rating_view.model.get("content_id") !== this.model.get("id") ) {
+        if( this.rating_view.model === null || this.rating_view.model.get("content_id") !== this.model.get("id") ) {
             var self = this;
             this.content_rating_collection.fetch().done(function(){
+                // Queue up a save on the model we're about to switch out, in case it hasn't been synced.
+                if (self.rating_view.model !== null && self.rating_view.model.hasChanged()) {
+                    self.rating_view.model.debounced_save();
+                }
                 if(self.content_rating_collection.models.length === 1) {
                     self.rating_view.model = self.content_rating_collection.pop();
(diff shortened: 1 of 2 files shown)


MCGallaspy:
It seems this PR suffers from the same issue in `0.15.x` that's being addressed in https://github.com/learningequality/ka-lite/pull/4403.

kalite/distributed/static/js/distributed/topics/views.js
```diff
@@ -84,27 +84,22 @@ var ContentAreaView = BaseView.extend({
...
-                    "content_id": ""
-                })
-            });
+            this.rating_view = this.add_subview(RatingView, {});
             this.$("#rating-container-wrapper").append(this.rating_view.el);
         }
 
         // Finally, handle the actual display logic
-        if( this.rating_view.model.get("content_id") !== this.model.get("id") ) {
+        if( this.rating_view.model === null || this.rating_view.model.get("content_id") !== this.model.get("id") ) {
```
rtibbles: If ```this.rating_view.model``` is ```undefined``` this will evaluate to ```false```.

rtibbles: (And hence throw an error as you try to ```get``` from an ```undefined```)

MCGallaspy:
@benjaoming never seen this before -- any idea why `kalite status` would return `Not Responding`?

benjaoming:
@MCGallaspy no, never seen it, and it only happened on one of the builder nodes... trying a rebuild...

rtibbles:
@MCGallaspy re: the garbage collection, I think due to the fact that there is an extant reference it won't be garbage collected.

Could you address the comment I made above?

MCGallaspy:
> Could you address the comment I made above?

Roger.

rtibbles:
For more info on garbage collection: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management

kalite/distributed/static/js/distributed/topics/views.js
MCGallaspy: The idea is that it's always either defined or set to null. From `rating\view.js`:
```js
module.exports = BaseView.extend({

    ...

    initialize: function(options) {
        ...
        this.model = options.model || null;
        ...
    },
```

I got tired of writing `if (typeof foo === "undefined")`, decided that `if( foo === null )` was cleaner. Do you have a preference for this project?

rtibbles: Fair enough, it's explicit, so that's fine.

PR #4404, opened 2015-09-10
Docs screenshots fixup

MCGallaspy:
Starts to fix #4296. Docs now build with screenshots when `screenshots_create = True` in `docs/conf.py`. But they need to be debugged -- they've gotten out of sync with the changes in the app. At the same time, we should add new screenshots to the docs where appropriate.

Find three commits:
* A squashed commit of many changes I made to get screenshots working. See the message for details.
* A commit with new screenshots -- as I said above, they need to be debugged.
* Increase timeout on a test that presumably started failing because I changed a browser_mixin method.

PR #4410, opened 2015-09-10
Consolidate fixes

MCGallaspy:
Consolidates #4403 and #4404, and I set an explicit page load timeout for old-style tests. Doesn't catastrophically fail locally. Let's see what the circle maters think.


MCGallaspy:
Closing. Didn't help.

PR #4411, opened 2015-09-10
expire sessions every 30 minutes.

### LINKED ISSUE #4399: 0.14.0 Cookie is not expiring when the the browser is closed on remote clients.
I have been experiencing this quite a bit while testing, it seems that from a remote machine after logging into a KA-Lite server instance as admin that if I then close the browser and reopen the browser to that same ip address, I am still logged in as before where the expected behavior is that the cookie would be deleted and I should have to select facility and provide username and password credentials to log back in.  This does not happen on the same machine that is hosting the KA-Lite server instance and seems to only happen on remote connections to that server. 

The use case I see this being an issue for is in a computer lab scenario with a dedicated server where the teachers and students are logging in using the same machines giving the students elevated privileges to the KA-Lite server instance.

aronasorman:
Fixes #4399. Expire cookies every 30 minutes instead of "browser close".

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -461,9 +461,10 @@ KEY_PREFIX = version.VERSION
 # Separate session caching from file caching.
 SESSION_ENGINE = getattr(
     local_settings, "SESSION_ENGINE", 'django.contrib.sessions.backends.signed_cookies' + (''))
-    
-# Expire session cookies whenever we close the browser.
-SESSION_EXPIRE_AT_BROWSER_CLOSE = True
+
+# Expire session cookies after 30 minutes, but extend sessions when there's activity from the user.
+SESSION_COOKIE_AGE = 60 * 30     # 30 minutes
+SESSION_SAVE_EVERY_REQUEST = True
 
 # Use our custom message storage to avoid adding duplicate messages
 MESSAGE_STORAGE = 'fle_utils.django_utils.classes.NoDuplicateMessagesSessionStorage'


kalite/settings/base.py
```diff
@@ -461,9 +461,10 @@
 # Separate session caching from file caching.
 SESSION_ENGINE = getattr(
     local_settings, "SESSION_ENGINE", 'django.contrib.sessions.backends.signed_cookies' + (''))
-    
-# Expire session cookies whenever we close the browser.
-SESSION_EXPIRE_AT_BROWSER_CLOSE = True
+
+# Expire session cookies after 30 minutes, but extend sessions when there's activity from the user.
+SESSION_COOKIE_AGE = 60 * 30     # 30 minutes
+SESSION_SAVE_EVERY_REQUEST = True
```
rtibbles: Just out of interest, does this have performance implications?

rtibbles:
Could you merge in latest 0.15.x to confirm tests passing?

kalite/settings/base.py
aronasorman: I have not measured, unfortunately. And the system @foolchan2556 made isn't ready yet. But hopefully we can get one by the end of 0.16.x. 

By my cursory testing though there's no noticeable performance change.

aronasorman:
@rtibbles I'm getting some sauce-related errors, can you take a quick peek and see what's causing them?

aronasorman:
Tests pass! (awyeah)

PR #4412, opened 2015-09-10
Use saucelabs remote webdriver

MCGallaspy:
Hopefully this configuration is right! Useless without adding secrets to the environment in Circle CI (which I did).

See: https://circleci.com/docs/browser-testing-with-sauce-labs

and: https://github.com/circleci/sauce-connect

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,6 +8,9 @@ dependencies:
     - pip install -e .
     - npm install
     - npm install -g jshint
+  post:
+    - wget https://saucelabs.com/downloads/sc-latest-linux.tar.gz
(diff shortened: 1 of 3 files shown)


MCGallaspy:
I recommend merging this and re-running failing tests, since saucelabs so nicely provides screencasts which could demonstrate the problem.

MCGallaspy:
Negatory, let me add a fallback for running the tests locally.

MCGallaspy:
I guess my `[ci skip]` didn't work for some reason. Anyway, this should be ready.

PR #4414, opened 2015-09-11
Coach report fixes

### LINKED ISSUE #4337: Coach reports date range should be inclusive
On `0.15.x`, created some student data, and then tried to view it on the coachreports as an admin. It seems that the date range is not inclusive in some cases, as student data from the same-day are not viewable:

ss1

I tired several different start & end dates with student data from 9/1. In all cases I expect data to be show:

      Begin date      | End date  | Data shown?
---------|-----------|---------------------
 9/1 | 9/1 | No
9/1 | 9/2 | Yes
8/31 | 9/1 | No
8/31 | 9/2 | Yes

In other words, date ranges should be inclusive with both the end and begin dates.

### LINKED ISSUE #4338: Coach reports exercise detail view inconsistently displayed
On `0.15.x`, when viewing the detail view for exercises from the coach reports, panels with only one exercise log have a buggy display:
ss4

Compare to:
ss_good_detail

In case of one attempt, I expect the "Attempt-Status-Time-Action" subtable to have the same alignment as with multiple attempts.

### LINKED ISSUE #4339: Coach reports video detail view does nothing on click
On `0.15.x`, I navigated to the video page as student to create an attempt record (%-completion seems to have no effect here). As an admin, clicking on the corresponding item in the coach reports does not expand a detail view, despite the panel being "clickable" -- the cursor changes and there is a hover effect. It _does_ close another open detail view for e.g. an exercise, but otherwise has no effect.

Opening the console reveals the following js error: `Uncaught TypeError: Cannot read property 'meta' of undefined` from `bundle_coach_report.js`.

Edit: Expected behavior is a detail view -- but if none exists, perhaps we should make the element "unclickable"?

Double edit: Perhaps we should also describe in the documentation what to expect from the detail view for various content types?

### LINKED ISSUE #4340: Text display issues on tabular report

### LINKED ISSUE #4341: Coach report date selector missing navigation icon
On `0.15.x`, on the coach reports, clicking on the date selectors reveals a calendar. Where I would expect an icon to navigate months I see nothing, however there is a clickable element there (my mouse is hovering over it in the screenshot) and it does work as expected. The icon is just missing:

ss

### LINKED ISSUE #4342: Coach report date refresh button -- can we remove it?
On `0.15.x`, when you change the date on the coach reports page, then the tabular report does not change -- you need to first click the refresh button next to the date range and then click the "show tabular report" button again. Can we eliminate the refresh button? I would expect the date to be updated automatically when I use the selector, without clicking the refresh button, and if the tabular report is expanded then I would expect it to be updated as well.

This is not a bug, hence the 0.16.0 milestone -- but if it's a safe change then we can consider incorporating it in `0.15.x`.

### LINKED ISSUE #4343: Update coach report documentation
We need to update the coach report documentation for 0.15. This means at least updating the screenshot as it does not include the date picker. That's blocked until we get that working again on `0.15.x`, so I'll make that a priority. We should also describe the use of the date picker, since the user flow with the refresh button may not be immediately obvious. It might be helpful to also describe what's expected in the detail view for various content types.

### LINKED ISSUE #4349: Coach reports summary main view issues
I'm collecting several different issues here. All of these are with data generated by running `generaterealdata --scenario-1` (see https://github.com/learningequality/ka-lite/pull/4347) on branch `0.15.x` (running from source in this case, not from this week's windows installer) and logged in as a coach user.

1. "Total Hours Logged" says is labeled "Hours spent on content vs total hours logged". Does "total hours logged" mean among all users for the Facility? The FacilityGroup? Among all facilities in the sharing network?
2. "Total Hours Logged" central value of "Total" is in hours (but its unlabeled), but the circle surrounding actually represents a ratio -- a bit misleading.
3. The central values in the main view are not vertically centered.
4. Some are just unadorned numbers, with no visualization as with "Total Hours Logged" -- is this intentional? See screenshot: 
ss1
5. Text overflows on mouseover of "Total Hours Logged": 
ss2
6. Selecting a new facility or group causes the main view to not be updated correctly -- on initially loading the page I see the screenshot from 4, but then choosing the "Ungrouped" group and then the original group again causes "Average progress in exercises" not to be updated: 
ss3
(Edit: same thing in 6 happens if I switch facilities)
7. Clicking "Show Tabular Report" and then changing the facility or group causes "Hide Tabular Report" to be shown even though the tabular report disappears. Clicking it seems to have the same effect as "Show Tabular Report", so it's apparently just the text that's not updated: 
ss3

### LINKED ISSUE #4351: Coach reports -- tabular report vertical scrolling
On `0.15.x` using `generaterealdata --scenario-1`: When there are many attempt logs, then the tabular report scrolls vertically. This creates very wide log detail views:

ss1

Additionally, the scroll bar at bottom of the tabular report pane is out of view when looking at the titles of the content items. I suspect this could make it not-so-obvious that it is indeed scrollable. I also wonder what effect what effect the touch-and-swipe style of scrolling will have (as with touchscreens) given that every element in the tabular report is clickable.

Finally, scrolling within the detail view of an attempt log blocks scrolling in the outer tabular report. Ideally the detail view would not be scrollable, and scrolling within it would cause the tabular report to scroll instead.

For future iterations, can we paginate the number of columns in the tabular report? Or perhaps make the tabular report searchable/filterable by content items as well? That could help to reduce the number of columns displayed at once.

### LINKED ISSUE #4366: student progress reports when no videos under the subject, video icon should display "0"
branch: 0.15.x
os: Mac
browser: Chrome
no number displayed when no videos under the subject
screen shot 2015-09-03 at 12 35 47 pm

rtibbles:
Fixes #4337, fixes #4338, fixes #4339, fixes #4340, fixes #4341, fixes #4342, fixes #4343, fixes #4349, fixes #4351, fixes #4366.

Summary of changes:
See commit messages!

@MCGallaspy these are mostly bugs filed by you, one by @66eli77 .


MCGallaspy:
Checking off as I go through. Incoming edits.
- [x] #4337
- [ ] #4338

MCGallaspy:
Re #4342 -- toggling the visibility of the datepicker refresh button changes the width of some elements, instead of appending the refresh button:

ss2


But otherwise the fix seems okay.

rtibbles:
Unfortunately appending causes some weird issues all of it's own.

Might be better just to enable/disable the button for a smooth experience.

rtibbles:
@MCGallaspy made the fix I suggested.

MCGallaspy:
Good, I'll just wait for the build to finish.

PR #4428, opened 2015-09-15
Try more robust sauce connect behavior

MCGallaspy:
I hope there are no syntax errors here.


MCGallaspy:
And of course there are.

MCGallaspy:
Don't pull this; I'm going to squash commits when I'm done.

rtibbles:
@MCGallaspy perhaps using systemd or upstart would be a better way forward?

rtibbles:
https://docs.saucelabs.com/reference/sauce-connect/#service-management

MCGallaspy:
> perhaps using systemd or upstart would be a better way forward?

I don't understand how to use those services in the context of a circle VM. As far as I know, the only possibility for configuration are using the configuration options in circle.yml (which has no built-in support for sc, upstart, or systemd that I can determine), running individual shell commands (which we have determined is insufficient), or scripting more sophisticated behavior as I've tried here.

At any rate, it appears that the sc process is being terminated when the command ends... so there may be no avoiding running it `background: true` as we've tried.

MCGallaspy:
Okay, I'm now of the opinion that trying to cleverly circumvent a sauce connection failure is probably more trouble than it's worth... instead let's redirect the `sc` executable output to a logfile so that when it fails, we can ssh in and take a peek. In the meantime, can we live with sometimes having to restart a build if sauce does not connect? And perhaps we can also alter the test behavior to gracefully continue when a sauce connections is not established?

MCGallaspy:
As for why `sc` does not connect sometimes, I think we can:

* Try to get output as I've suggested
* Contact saucelabs support

rtibbles:
Let's close this in favour of #4431

PR #4431, opened 2015-09-15
Graceful sauce failure

MCGallaspy:
Logs output and tries to gracefully degrade the tests if we can't connect to saucelabs. Supercedes #4428.


MCGallaspy:
@aronasorman could be an alternate reviewer.

Issue #4432, opened 2015-09-16
KA Lite splash image off-center


MCGallaspy:
On `0.15.x`. I pulled this from the docs, but I've also noticed when using the app, or viewing sl screencasts. The splash image overflows it's container.

![image](https://cloud.githubusercontent.com/assets/8888020/9893486/be9ca724-5bcd-11e5-9a35-f75233fccbf3.png)

I think it's a vector asset, right? Can we resize it dynamically?

66eli77:
if you set the width the same as its container, it should resize correctly

rtibbles:
Fixed in #4435

PR #4433, opened 2015-09-16
Docs screenshots fixup

MCGallaspy:
Replaces #4404. I created a weird history after accidentally merging develop instead of 0.15.x and continuing to work on top of it. This should incorporate all the changes.


MCGallaspy:
Two more things. Need to build the screenshots with:

* An internet connection, or the server configured to detect a local running instance of the central server. Otherwise some (registration) pages display incorrectly.
* A downloaded language pack, so that the delete button is visible for a screenshot. Or fiddle around with screenshot settings or setup process to make the server _think_ a language pack is downloaded. @aronasorman any advice on how to make a language pack appear to be downloaded without actually downloading it? Ideally a method which is future proof.

MCGallaspy:
Oh, and I will resolve #4432 and remove cruddy screenshots from the commit history before building and committing a good set.

aronasorman:
re: language pack -- you just need to create a json file in a folder named after the language code. If you wanna fake es:

```
locale/
  es/
     - es_metadata.json

> cat locale/es/
{}
```

That should be the minimal way to fool `kalite` that a language is installed.

MCGallaspy:
Are there any caveats? Does it need to be done before the server starts? Is there any guarantee it won't change in the future, quietly breaking the screenshot again?

aronasorman:
If you do it after server restart you need to force a cache invalidation. There should be functions for that. 

re: guarantee -- we're not planning any huge i18n refactoring soon, and that part of i18n doesn't change much. So it should be stable in the short term.

MCGallaspy:
Empty es_metadata.json as you've described doesn't work -- instead I get a bunch of errors when it fails to find some key from that file. I think for now I will simply actually download a language pack.

aronasorman:
Were you trying to view the `update language packs` page? It should only fail in that page. Otherwise you can fill it with these details:

```
{"num_dubbed_videos": 0, "code": "de", "native_name": "Deutsch", "subtitle_count": 1761, "ka_ntranslations": 184935, "language_pack_version": 17, "package_size": 8777710, "approved_translations": 185679, "percent_translated": 88.96293989411399, "zip_size": 3546260, "software_version": "0.15", "beta": false, "kalite_nphrases": 1435, "num_exercises": 0, "kalite_ntranslations": 744, "name": "German", "phrases": 208715, "ka_nphrases": 207280}
```

MCGallaspy:
It also failed on startup because it lacked the keys `name` and `code`.

MCGallaspy:
But yes, I was trying to "view" that page with a WebDriver, to take screenshots.

jamalex:
> remove cruddy screenshots from the commit history before building and committing a good set.

this sounds like the main remaining piece to me -- I skimmed the code, and nothing popped out (though I'm not really familiar with that code

:+1:

MCGallaspy:
Everything seems to work ok. Probably good to take another pass at it closer to release.

PR #4435, opened 2015-09-16
Style splash image responsively

### LINKED ISSUE #4432: KA Lite splash image off-center

MCGallaspy:
Fixes #4432. Here's how it looks in various viewport sizes:

__Normal__
ss2

__Smaller__
ss3

__Smallest__ (when the hamburger shows up)
ss4


66eli77:
look good

Issue #4438, opened 2015-09-16
Better way to specify CHANNEL setting


MCGallaspy:
This isn't a bug by a usability consideration. Since we're deprecating the user of local_settings.py, if one wishes to specify the CHANNEL setting, one has to also specify a number of derived settings (I'm looking at CHANNEL_DATA_PATH and CHANNEL_DATA).

I think the best solution is to remove derived settings into the relevant app within kalite, so that it's loaded _after_ the user's settings module. There may be other considerations in this specific case, depending on how the derived settings are used.

MCGallaspy:
@aronasorman incidentally, despite my best efforts I'm having some trouble importing a channel for testing. Could there be other settings I'm missing? What is the recommended way to import a channel?

rtibbles:
Fixed in #4440

PR #4440, opened 2015-09-17
Custom channel fixes

rtibbles:
Makes custom channels activatable in new settings paradigm.

Sets root node of custom channel import to "root".

### DIFF
--- a/kalite/topic_tools/settings.py
+++ b/kalite/topic_tools/settings.py
@@ -12,16 +12,22 @@ only on django.conf.settings
 import os
 from django.conf import settings
 
+CHANNEL = getattr(settings, "CHANNEL", "khan")
+
+CHANNEL_DATA_PATH = os.path.join(settings.CONTENT_DATA_PATH, CHANNEL)
+
+# Whether we wanna load the perseus assets. Set to False for testing for now.
+LOAD_KHAN_RESOURCES = getattr(settings, "LOAD_KHAN_RESOURCES", CHANNEL == "khan")
 
 DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = getattr(settings, "DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP", False)
 
 KHAN_EXERCISES_DIRPATH = os.path.join(settings.STATIC_ROOT, "js", "distributed", "perseus", "ke")
 
 TOPICS_FILEPATHS = {
-    settings.CHANNEL: os.path.join(settings.CHANNEL_DATA_PATH, "topics.json")
+    CHANNEL: os.path.join(CHANNEL_DATA_PATH, "topics.json")
 }
-EXERCISES_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "exercises.json")
-CONTENT_FILEPATH = os.path.join(settings.CHANNEL_DATA_PATH, "contents.json")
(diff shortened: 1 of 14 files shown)


rtibbles:
Fixes #4438 too...

kalite/topic_tools/settings.py
```diff
@@ -12,16 +12,22 @@
 import os
 from django.conf import settings
 
+CHANNEL = getattr(settings, "CHANNEL", "khan")
```
MCGallaspy: If the user doesn't specify CHANNEL in their settings module, could the following situation be problematic?

```python
from kalite.topic_tools import settings as tt_settings

settings.CHANNEL  # AttributeError
tt_settings.CHANNEL  # "khan"
```

MCGallaspy:
Looks good, just one question.

kalite/topic_tools/settings.py
rtibbles: Yes, but I explicitly changed all references to CHANNEL to only come from
topic tools settings.

MCGallaspy:
Still not working as I expected. Ran the same import command (`python bin\kalite manage contentload --import=D:\ka-lite\import_test --channel=test --data=D:\ka-lite\import_test`), no js console errors, but the sidebar is not responsive. I see the root node ("blah") but clicking it has no effect though the url is changed.

Navigating directly to the content item (`/learn/test/blah/astro/`) does indeed bring it up, but the sidebar is still wonky.

MCGallaspy:
That said, the PDF reader works without a hitch!

jamalex:
> That said, the PDF reader works without a hitch!

Note that there are two modes -- one in which we use pdf.js and one in which we launch an external viewer (for tablet clients, where pdf.js is too slow). We need to test that both modes work -- I'm worried that `settings.PDFJS = False` may no longer do anything, as I don't see a code path that references this, at first peek. We'll need this, for the Aflatoun deployment, as they're using tablets.

MCGallaspy:
@jamalex this is diverging from the PR, but is there any documentation about pdf content? Like, how it can be configured, and what the expected behavior is for either mode?

jamalex:
Not that I know of, no -- but the basic behavior is meant to be:

- `PDFJS = True` (default): Load the pdf.js library and use it to render the document inline within the page.
- `PDFJS = False`: Don't bother loading all the pdf.js libraries (they're big), and don't dare attempt to render in the browser -- just have an link to the PDF itself so that it will launch in an external viewer.

rtibbles:
```python bin\kalite manage contentload --import=D:\ka-lite\import_test\blah --channel=test --data=D:\ka-lite\import_test```

Is the correct command, the import path should always reference the root node, not the containing folder.

MCGallaspy:
@richard what's the difference between a root node and the base folder here? If I wanted everything in D:\ka-lite\import_test to become a top level topic/sidebar item, how could I do that?

rtibbles:
Then you would target import_test, but you have the channel metadata in
there, so I assume you do not want to do that.

MCGallaspy:
So is it forbidden to include the channel metadata there? Could it cause the sidebar problems I've been experiencing?

MCGallaspy:
It's probably worth testing out channel import more extensively if we wish to make it user-facing eventually, or have partners run the command. This PR looks good to me, though.

rtibbles:
I don't think we ever plan to make this user facing, as it should be
deprecated by content-curation.

PR #4441, opened 2015-09-17
Collectstatic

### LINKED ISSUE #4401: Remove collectstatic from startup
As per discussion on when to run `kalite manage collectstatic`, we have delegated it for `kalite manage setup` for now.

This should be a very low hanging fruit, does anyone have objections against removing it from `kaserve` where it's currently run?

        logging.info("Copying static media...")
        call_command("collectstatic", interactive=False, verbosity=0)

        call_command("collectstatic_js_reverse", interactive=False)

@rtibbles @jamalex @MCGallaspy @aronasorman @cpauya

rtibbles:
Fixes #4401 

Also removes KA Serve command to avoid confusion :)


benjaoming:
@rtibbles yes I know, I'm suggesting it as a means for development so you don't have to run `collectstatic` every time you change a file.......

benjaoming:
Looking good!! Found no references for kaserve in codebase, nor docs.

PR #4445, opened 2015-09-17
Disallow python 3

MCGallaspy:
See https://github.com/learningequality/installers/issues/152


benjaoming:
Looks good, test failures are irrelevant. Thanks for adding this extra check.

PR #4446, opened 2015-09-17
Persist annotations

### LINKED ISSUE #4335: Persist annotations when using SQLitedict
See https://github.com/learningequality/ka-lite/pull/4306#discussion_r38388364.

rtibbles:
Fixes #4335 

Persist annotations in the database for the database backed content cache.

Delete database for content cache when the cache is invalidated.


aronasorman:
Changes look good. Testing.

aronasorman:
How will I know that this is working? That 'Preloading content data' is now blazing fast?

aronasorman:
Because it is.

aronasorman:
After downloading a new video, the "Converting" message appeared and it's taking quite a while for the topic tree to appear (in Learn tab). Was this the behaviour before?

aronasorman:
Interesting, with running `create_content_db`, it seems that only the 'Preloading content data' step was fast. After downloading a video and waiting for the topic tree to load, now all steps (exercise, content and topic tree data) are now fast as well. Do you have an idea why we have this behaviouir @rtibbles?

aronasorman:
Otherwise, everything else seems to be working.

aronasorman:
All expected behaviour. This is good to go.

PR #4450, opened 2015-09-18
Properly triggers login on API 403.

### LINKED ISSUE #4449: window.userView is undefined - 403 returns do not properly prompt for user login
Branch: 0.15.x
Ubuntu, Chrome

To prompt a login when an API call returns a 403, we rely on a globally defined userView. This appears to be undefined.

rtibbles:
Fixes #4449 

Currently, if you are logged out and an API request returns a 403 forbidden, then our modal login does not get triggered (it will produce a Javascript error instead).

To test this, go to developer console, resources, and delete the sessionId cookie while on the video download page.

### DIFF
--- a/kalite/distributed/static/js/distributed/utils/api.js
+++ b/kalite/distributed/static/js/distributed/utils/api.js
@@ -65,9 +65,8 @@ function handleFailedAPI(resp, error_prefix) {
         case 403:
             messages = {error: gettext("You are not authorized to complete the request.  Please login as an authorized user, then retry the request.")};
             if (window.statusModel) {
-                window.statusModel.fetch().success(function() {
-                    window.userView.login_start_open = true;
-                });
+                window.toggleNavbarView.userView.login_start_open = true;
+                window.statusModel.fetch();
             }
             break;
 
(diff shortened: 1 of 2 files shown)


kalite/distributed/static/js/distributed/utils/api.js
```diff
@@ -65,9 +65,8 @@ function handleFailedAPI(resp, error_prefix) {
         case 403:
             messages = {error: gettext("You are not authorized to complete the request.  Please login as an authorized user, then retry the request.")};
             if (window.statusModel) {
-                window.statusModel.fetch().success(function() {
-                    window.userView.login_start_open = true;
-                });
+                window.toggleNavbarView.userView.login_start_open = true;
+                window.statusModel.fetch();
```
66eli77: after deleting the sessionid, when the login modal shows up, if you wait a bit, this handleFailedAPI will be called periodically, causing multiple login modals displayed on top of each other.

rtibbles:
@66eli77 This should fix the issue!

66eli77:
cool, verified, merging

PR #4453, opened 2015-09-18
Don't autofocus on first element to make access to search page easier for keyboard users.

### LINKED ISSUE #4011: First result from search box "drop-down" gets selected by default
**Branch**: 0.14

**Actual behavior**: When user starts typing a query in the search form, a list of possible results appear in the drop-down, and the **first result gets selected by default**. This way, as soon as the user finishes typing and presses Enter, instead of landing on a Search result page, they land on the page with the content that was first on the drop-down list. 

search-focus

**Expected behavior**: It sounds like the Google "I'm feeling lucky", which I never used/liked, but at this point I'm not sure if this behavior was intended. Personally, I always prefer to see all the possible results and choose what I need. If user is like me, once they realize why they are not getting the result page, they have to retype the query, actively deselect the first drop-down result with the up arrow key, and than press Enter.  

**Steps to reproduce**:

1. Starts typing a query in the search form, and press Enter.

rtibbles:
Fixes #4011

### DIFF
--- a/kalite/distributed/static/js/distributed/search/views.js
+++ b/kalite/distributed/static/js/distributed/search/views.js
@@ -90,7 +90,7 @@ var AutoCompleteView = BaseView.extend({
         this.$el.html(this.template({search_url: window.Urls.search()}));
 
         this.$("#search").autocomplete({
-            autoFocus: true,
+            autoFocus: false,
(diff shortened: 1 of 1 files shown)


jtamiace:
does exactly what it says

PR #4457, opened 2015-09-19
Fixes exercise wrapping to prevent overflow after watching a video.

### LINKED ISSUE #4418: Exercises pane no longer centered.
**Branch:**  0.15.x

**Expected Behaviour:**  The left pane must align horizontally with the right pane.

**Current Behaviour:** The left pane align odd to the right pane. 
screenshot 2015-09-11 17 47 07

**Steps to Replicate**
>1.  Navigate to the Learn Tab
>1. Select a topic on the Topic tree
>1. Notice that the exercise page view odd.

rtibbles:
Fixes #4418 as far as I think is feasible for 0.15.x

Masks the issue by wrapping exercises in the same content box as other content.


MCGallaspy:
Looks great, but the title is a bit lower contrast now:

ss2

I can take this off your hands, should be an easy fix.

MCGallaspy:
I'm just going to remove the opacity of the title:
ss2

PR #4465, opened 2015-09-22
Add get_cookie to base module & global scope

MCGallaspy:
Add it to the global scope, so that it can be accessed by inline JS
  in central server templates.

@rtibbles is this what you were thinking?

### DIFF
--- a/python-packages/fle_utils/django_utils/command.py
+++ b/python-packages/fle_utils/django_utils/command.py
@@ -153,7 +153,7 @@ def call_outside_command_with_output(command, *args, **kwargs):
 
     # build the command
     if kalite_dir:
-        kalite_bin = os.path.join(kalite_dir, "bin/kalite")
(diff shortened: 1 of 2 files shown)


python-packages/fle_utils/django_utils/command.py
```diff
@@ -153,7 +153,7 @@ def call_outside_command_with_output(command, *args, **kwargs):
 
     # build the command
     if kalite_dir:
-        kalite_bin = os.path.join(kalite_dir, "bin/kalite")
+        kalite_bin = os.path.join(kalite_dir, "bin", "kalite")
```
MCGallaspy: C'mon guys... why ya gotta be so unix-centric?

MCGallaspy: Causes test failure on the central-server, which imports this module

rtibbles:
Yep

Issue #4468, opened 2015-09-22
Can't switch to another language to download dubbed videos


aronasorman:
branch: 0.15.x

In the download videos page, we have this: 

![image](https://cloud.githubusercontent.com/assets/191955/10028052/2cc90cc8-611f-11e5-8e2f-32a57b0b8920.png)

Expected behaviour: Clicking the "English" link will change the link to a dropdown box, allowing me to download dubbed videos for that language.

Current behaviour: Clicking said link will just reload the same page.

66eli77:
fixed in PR #4469

rtibbles:
Should be fixed by #4469.

PR #4469, opened 2015-09-23
fix issue #4468

66eli77:


kalite/updates/templates/updates/update_videos.html
```diff
@@ -11,6 +11,7 @@
 
 {% block headjs %}{{ block.super }}
     <script type="text/javascript" src="{% static 'js/updates/bundles/bundle_update_videos.js' %}"></script>
+    <script type="text/javascript" src="{% static 'js/distributed/utils/get_params.js' %}"></script>
```
rtibbles: What is the purpose of this?

rtibbles:
Can you post a screenshot of how this looks now?

kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -299,7 +299,8 @@ $(function() {
 
 
     if ($("#download_language_selector option").length > 1) {
-        $("#language_choice_titlebar a").attr("onclick", "show_language_selector();");
+        // $('#language_choice_titlebar a').click(show_language_selector);
```
rtibbles: Definitely agree that it is better just to show the drop down rather than have the onclick behaviour. Maybe just delete this line instead of commenting it out?

66eli77:
@rtibbles do you know what's the design consideration for not display the language selector directly, instead asking user to click the "English" first then display the language selector?
For now, I made the drop down language selector directly accessible as this img shown 
screen shot 2015-09-22 at 11 39 23 pm

rtibbles:
I think it was a poor design consideration, @66eli77, so I think you are right to remove it. Much better with just the direct drop down selector.

rtibbles:
Fine to leave as is for now, but could you open an issue for cleaning up the template itself? We could just have the dropdown selector there alone, without the other stuff to impede it (and no need to run the 'show_language_selector' function).

66eli77:
sure. but it looks better and make more sense to not display a dropdown list when there are only one language. 
screen shot 2015-09-23 at 12 04 08 am

rtibbles:
Ah right, that's true. We could probably do that in the Django template though, and not fuss about with javascript for it!

kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -299,12 +300,12 @@ $(function() {
...
 
     if ($("#download_language_selector option").length > 1) {
-        $("#language_choice_titlebar a").attr("onclick", "show_language_selector();");
+        show_language_selector();
     }
 
     $("#download_language_selector").change(function() {
          var lang_code = $("#download_language_selector option:selected")[0].value;
-         window.location.href = setGetParam(window.location.href, "lang", lang_code);
+         window.location.href = get_params.setGetParam(window.location.href, "lang", lang_code);
```
rtibbles: That's more like it!

Issue #4429, opened 2015-09-15
No indicator in the topic tree for a specific video after a successful download.


djallado:
**Branch:** `Master`

**Current Behaviour:** After a successful download of a video, the indicator is not shown in the topic tree for that specific video download.
![screen shot 2015-09-10 at 5 49 02 pm](https://cloud.githubusercontent.com/assets/8664071/9812772/4d8cfc28-58b2-11e5-9df1-15e1e1646c12.png)

**Expected Behaviour:** The indicator must be shown after a successful video download.

**Step to Reproduce:**

1. Login as admin.
2. Click on Video tab, and select a video you want to download in a topic tree.
3. Click on Download new selected video button.

djallado:
The same issue I filed in https://github.com/learningequality/installers/issues/166.

66eli77:
I can confirm that the issue can be replicated.
when I reload the page, the indicator will show up correctly.
@rtibbles do you think this has anything to do with availability stamping?

rtibbles:
I don't think directly, no. Probably just the fancytree data not being updated on successful download.

66eli77:
this should be fixed in #4476
there is another sort of relating issue I going to deal with in #4477

rtibbles:
Fixed in #4476

PR #4476, opened 2015-09-23
fix for #4429

66eli77:
fix No indicator in the topic tree for a specific video after a successful download. #4429


66eli77:
@rtibbles feel free to pass the torch to @aronasorman

rtibbles:
Just tested this, @66eli77, and didn't see any change in behaviour.

66eli77:
@rtibbles you mean the indicator still not change accordingly after successful download or delete?

rtibbles:
Yeah,it stayed exactly the same.

rtibbles:
Oh,sorry, just tried again to double check. Worked this time. Must have had cached JS or something.

PR #4485, opened 2015-09-24
0.14.1

aronasorman:
Fixes:
- Fix assessment items not getting translated (#4482)
- Notify users that they need to restart their server after downloading a language pack (#4491)
- Generate a new secret key when the server starts (#4494)


benjaoming:
We are releasing this as 0.14.1

PR #4487, opened 2015-09-24
fix #4484

### LINKED ISSUE #4484: second attempt to download video always fail
the work around be refreshing the page or without refresh, click the cancel button

66eli77:


kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -178,7 +178,7 @@ $(function() {
                         // Best to assume offline, as online check returns much faster than offline check.
                         if(server_is_online){
                             $(".enable-when-server-online").removeAttr("disabled");
-                            base.updatesStart("videodownload", 5000, video_callbacks);
+                            // base.updatesStart("videodownload", 5000, video_callbacks); no idea why run this here?
```
rtibbles: @66eli77 we run this here in case there are any currently running download jobs going, so we start polling as soon as you load the page.

kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -396,7 +396,7 @@ function withNodes(nodeKey, callback, currentNode) {
 
 function setNodeClass(nodeKey, className) {
     withNodes(nodeKey, function(node) {
-        $(node.span).removeClass("unstarted partial complete").extraClasses(className);
+        // $(node.span).removeClass("unstarted partial complete").extraClasses(className);  no idea why run this here?
```
rtibbles: This is to reset any existing classes on the nodes, so that when we set the current progress class on the nodes, then it will change appropriately.

rtibbles:
I think the commented out code should be reinstated unless you have a very compelling reason why it needs to go.

kalite/updates/static/js/updates/bundle_modules/update_videos.js
66eli77: this line never gets run successfully because it should be "$(node.span).removeClass("unstarted partial complete").extraClasses = className;"
but the line below this line ("node.extraClasses = className;") should get the job done right?

kalite/updates/static/js/updates/bundle_modules/update_videos.js
66eli77: @rtibbles agree, this line is needed

PR #4488, opened 2015-09-24
Show warning when no videos available.

### LINKED ISSUE #4474: When no dubbed videos for a language, Javascript error occurs
Branch 0.15.x
Ubuntu/Chrome

Expected Behaviour: If no dubbed videos exist for a language (for example for the DEBUG language pack), then we should inform the user of this.
Current Behaviour: If no dubbed videos exist, the video update topic tree endpoint returns None, and causes fancytree to error out while trying to call properties of null.

image

rtibbles:
Fixes #4474.

When no videos are available for a language, show a warning to the user, rather than trigger a Javascript error.


66eli77:
looking good, merging

PR #4494, opened 2015-09-24
Generate a secret key file, and load SECRET_KEY from said file.

### LINKED ISSUE #4213: Generate a new SECRET_KEY during first time setup
Branch: all the branches

Expected behaviour: Arbitrary user doesn't know what their installation's secret key is.

Actual behaviour: Our secret key is not just shared across all installations, it's public as well! This means that any of the methods described [in this stack overflow answer](http://stackoverflow.com/a/15383766) is vulnerable to attack.

So we should generate a key that will always be read by KA Lite, even if a different settings object is specified.

aronasorman:
Fixes #4213. We already have said vulnerability wherein the SECRETT_KEY is both public and the same for each installation. With the introduction of sessions-in-cookies it's a double whammy.

So now the solution is to create a secret key file for each installation if it's not detected, and load that as the value of SECRET_KEY.

Assigning @jamalex, but @benjaoming please take a look too.

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -314,8 +314,20 @@ USE_I18N = getattr(local_settings, "USE_I18N", True)
 USE_L10N = getattr(local_settings, "USE_L10N", False)
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = getattr(local_settings, "SECRET_KEY",
-                     "8qq-!fa$92i=s1gjjitd&%s@4%ka9lj+=@n7a&fzjpwu%3kd#u")
+SECRET_KEY_FILE = getattr(local_settings,
+                          "SECRET_KEY_FILE",
+                          os.path.join(USER_DATA_ROOT, "secretkey.txt"))
+
+
+try:
+    with open(SECRET_KEY_FILE) as f:
+        SECRET_KEY = getattr(local_settings, "SECRET_KEY", f.read())
+except Exception as e:
+    sys.stderr.write("Error reading secret key file. Generating one now. Error was: %s\n" % e)
+
+    from ._utils import generate_secret_key, cache_secret_key
+    SECRET_KEY = generate_secret_key()
+    cache_secret_key(SECRET_KEY, SECRET_KEY_FILE)
 
 LANGUAGE_COOKIE_NAME = "django_language"
 
@@ -467,7 +479,7 @@ KEY_PREFIX = version.VERSION
 # Separate session caching from file caching.
 SESSION_ENGINE = getattr(
     local_settings, "SESSION_ENGINE", 'django.contrib.sessions.backends.signed_cookies' + (''))
-    
+
 # Expire session cookies whenever we close the browser.
 SESSION_EXPIRE_AT_BROWSER_CLOSE = True
 
--- a/kalite/settings/_utils.py
+++ b/kalite/settings/_utils.py
@@ -1,0 +1,17 @@
+import base64
+import sys
+import uuid
+
+
+def generate_secret_key():
+    key = base64.b64encode(str(uuid.uuid4()))
+    return key
+
+
+def cache_secret_key(key, key_file_path):
+    try:
+        with open(key_file_path, "w") as f:
+            f.write(key)
+            f.flush()
+    except Exception as e:
+        sys.stderr.write("Error writing secret key file. Error was: %s\n" % e)


kalite/settings/base.py
```diff
@@ -314,8 +314,22 @@
 USE_L10N = getattr(local_settings, "USE_L10N", False)
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = getattr(local_settings, "SECRET_KEY",
-                     "8qq-!fa$92i=s1gjjitd&%s@4%ka9lj+=@n7a&fzjpwu%3kd#u")
+SECRET_KEY_FILE = os.path.join(USER_DATA_ROOT, "secretkey.txt")
```
benjaoming: Yes, wondering what's going on here. Since `local_settings` is a dynamic object that may or may not be a module, the secret key is not all that reliable.

But is adding a file really that wise? What if it doesn't exist? I don't think that tests are gonna like this, nor do I think it's a viable pattern since settings should be load-able ALWAYS, especially since the `manage setup` command would depend on them.

Needs a fallback IMO.

kalite/settings/base.py
```diff
@@ -314,8 +314,22 @@
...
+
+
+def generate_secret_key():
```
benjaoming: A function does not belong in a settings module :)

benjaoming:
Why the backport for 0.14?

benjaoming:
Sorry, why is this not a backport but straight for 0.14?

aronasorman:
re: backport -- I just plan to cherry pick this single commit to 0.15. Unless you have other preferred method of propagating this change?

kalite/settings/base.py
aronasorman: re: file might not exist -- that's why we create the file if it doesn't exist. It's also done in `USER_DATA_ROOT` which we assume is always writeable by the server.

kalite/settings/base.py
aronasorman: OK, let me move this to another file and import it from settings.base.

aronasorman:
```
>>> settings.SECRET_KEY
'MjFiOTJmYzYtNGYyYS00MWViLTk5ZmMtZTUxZGU3NGU1NTZl'
```

kalite/settings/base.py
```diff
@@ -314,8 +314,22 @@
...
+    import uuid
+    import base64
+    key = base64.b64encode(str(uuid.uuid4()))
+    with open(SECRET_KEY_FILE, "w") as f:
+        f.write(key)
+        f.flush()
+
+if not os.path.exists(SECRET_KEY_FILE):
+    generate_secret_key()
```
jamalex: Rather than checking existence, I'd do a try-except around the full open-read-interpret block. That way, issues with format, etc will also trigger generating a new key.

jamalex:
Looks good on my side, but made one suggestion for robustness/pythonicness.

kalite/settings/base.py
```diff
@@ -314,8 +314,22 @@
...
+
+def generate_secret_key():
+    import uuid
+    import base64
+    key = base64.b64encode(str(uuid.uuid4()))
+    with open(SECRET_KEY_FILE, "w") as f:
+        f.write(key)
+        f.flush()
+
+if not os.path.exists(SECRET_KEY_FILE):
```
benjaoming: Ah sorry, didn't scroll this far to see that the file gets created here.

benjaoming: Should be `os.path.isfile` though :)

kalite/settings/base.py
aronasorman: Do you have any recommendation of what error to catch? Else I'm just catching `Exception`.

aronasorman:
@jamalex @benjaoming comments addressed. Main changes:

- moved secret key functions to its own file. it's under `kalite/project/settings/_utils.py`.
- changed the way we detect SECRET_KEY_FILE to catching exceptions and then generating secret keys from that.
- saving the secret key is optional. If we can't save it, then we just keep generating a new one every reboot. This will mean that users are logged out every restart, which @EdDixon wants. Might wanna add a flag for that moving forward.

aronasorman:
@benjaoming added `sys.stderr.write` calls.

jamalex:
Looks good to me -- might be good to test 1 or 2 weird edge cases before merging.

aronasorman:
Tried with `secretkey.txt`. It prints out this in stderr, then continues:

```
In the future, it is recommended not to keep your own settings module in the kalite code base but to put the file somewhere else in your python path, for instance in the current directory when running 'kalite --settings=my_module'.
  RemovedInKALite_v015_Warning
Error reading secret key file. Generating one now. Error was: [Errno 21] Is a directory: '/home/aron/src/github.com/aronasorman/ka-lite/secretkey.txt'
Error writing secret key file. Error was: [Errno 21] Is a directory: '/home/aron/src/github.com/aronasorman/ka-lite/secretkey.txt'
```

It's expected behaviour.

PR #4497, opened 2015-09-24
Don't accidentally translate "Invalidate caches" during language pack download.

### LINKED ISSUE #4473: Language switches during language pack download
Branch 0.15.x
Ubuntu/Chrome

Expected Behaviour: Language should always be in the currently selected language.
Current Behaviour: When downloading a language pack, the stage tracker switches to that language name when on the final step.

image

I think @jamalex had reported something similar too.

rtibbles:
Fixes #4473


aronasorman:
Good code. Testing.

aronasorman:
Fix fixed what it says it fixed.

PR #4502, opened 2015-09-25
Remove deprecated scripts.

### LINKED ISSUE #3627: Remove the deprecated scripts/ directory
Getting issues like #3624 should remind us that `scripts/` can be removed in 0.14

Any feelings about this?

I would suggest the following two replacements:

1. Functional installers + PyPi packages will take over the role of installing and removing KA Lite.
1. Documentation chapters handling how to add KA Lite to start up environments in case installers don't do that -- this is mainly relevant for Linux distributions, so it shouldn't be that long an example.

rtibbles:
Fixes #3627


MCGallaspy:
You know, I misspoke. [add_systemstart_task.bat](https://github.com/learningequality/installers/blob/1a6ed1e8eecbad3f68acabde2ec39197be98f042/windows/gui-packed/guitools.vbs#L60) and [remove_systemstart_task.bat](https://github.com/learningequality/installers/blob/1a6ed1e8eecbad3f68acabde2ec39197be98f042/windows/gui-packed/guitools.vbs#L72) are still used. Can we keep those in for now?

rtibbles:
Sure thing.

MCGallaspy:
Here's the screencast of the failing test: https://saucelabs.com/jobs/fe52c2ba43f648878d457a4b33a5ba63

Seems like a false negative to me, so I'm ready to merge if you agree.

rtibbles:
Yep, seems fine to merge, @MCGallaspy

PR #4504, opened 2015-09-25
Support for Coverage metrics and coveralls reports

benjaoming:
Taking over #4376

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,2 @@
+repo_token: YfchfjB9wbHxDS7MDadVe09f2StV0pds1
+service_name: circleci
--- a/circle.yml
+++ b/circle.yml
@@ -23,8 +23,17 @@ test:
     - cd sc-*-linux && ./bin/sc -u $SAUCE_USERNAME -k $SAUCE_ACCESS_KEY --tunnel-identifier $CIRCLE_BUILD_NUM-$CIRCLE_NODE_INDEX --readyfile ~/sauce_is_ready > sc_output.txt 2>&1:
         background: true
     - while [ ! -e ~/sauce_is_ready ]; do sleep 1; done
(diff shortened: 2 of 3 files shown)


rtibbles:
Doesn't seem to be running tests properly :/

benjaoming:
Re-running them... it worked in the former PR, no reason why it shouldn't work now.

rtibbles:
Must be that Friday afternoon feeling!

benjaoming:
Yes, Circle CI is tired after a long week.

rtibbles:
My assumption here is that it is interacting oddly with SauceConnect? That's the difference I see here between this test run and the previous one.

benjaoming:
Great, now what!? :(

rtibbles:
@benjaoming I will run your PR locally later in the week to debug.

benjaoming:
@rtibbles I've run the coverage commands locally, it's working fine on Coverage 3.7.1 -- I don't know if the problem is that output to stdout/stderr from the test commands is somehow not proxied by Coverage or disappears somewhere else.

benjaoming:
*) But locally, Coverage proxies test output fine, and it used to do so in the previous PR, too.

This sadly made the 21 day no-merge of #4376 even more costly. We should really merge things faster.

rtibbles:
Did you run it locally using SauceConnect? That seems to be the relevant difference, so it would be interesting to debug that.

benjaoming:
@rtibbles no but as you can see from the commits above, I tried removing sauceconnect from circle.yml without results :(

benjaoming:
Seeing that sauceconnect is also a part of `testing.base_environment`, there's still a bit of disabling work to do before verifying if SC is the reason.

benjaoming:
Should perhaps try this with Codecov.io

PR #4522, opened 2015-09-28
Generate a secret key file, and load SECRET_KEY from said file.

### LINKED ISSUE #4213: Generate a new SECRET_KEY during first time setup

aronasorman:
Fixes #4213.

Conflicts:
	kalite/settings/base.py

### DIFF
--- a/kalite/settings/_utils.py
+++ b/kalite/settings/_utils.py
@@ -1,0 +1,17 @@
+import base64
+import sys
+import uuid
+
+
+def generate_secret_key():
+    key = base64.b64encode(str(uuid.uuid4()))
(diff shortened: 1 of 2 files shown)


aronasorman:
Test passed. Merging.

PR #4523, opened 2015-09-28
Try to catch TransactionManagementErrors during database teardown.

rtibbles:
TransactionManagementErrors appear to be appearing again.

This is an attempt to catch them where they arise - during teardown of the database in the behave tests.

### DIFF
--- a/kalite/testing/base_environment.py
+++ b/kalite/testing/base_environment.py
@@ -16,6 +16,7 @@ from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from django.contrib.auth.models import User
 from django.core.management import call_command
 from django.db import connections
+from django.db.transaction import TransactionManagementError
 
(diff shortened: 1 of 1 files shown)


kalite/testing/base_environment.py
```diff
@@ -168,7 +169,10 @@ def database_teardown(context):
     setup/teardown done by TestCases in order to achieve consistent isolation.
     """
     for alias in connections:
-        call_command("flush", database=alias, interactive=False)
+        try:
+            call_command("flush", database=alias, interactive=False)
+        except TransactionManagementError:
```
MCGallaspy: Needs to be `except TransactionManagementError as e:`

rtibbles: Whoops.

rtibbles:
Ironically enough, no TransactionManagementError got thrown during this test, it seems.

Issue #4459, opened 2015-09-21
Makefile for common build tasks


benjaoming:
As per discussion on how we build and distribute installers, a lot of tasks can be outlined and unified in a simple Makefile.

One principle as per discussion: The Makefile has to be readable.

Common tasks outlined in learningequality/installers#158

@cpauya @MCGallaspy I'm hoping you can adopt the file for the installers in 0.15 -- is that reasonable? I guess it's a matter of having make available on the build system and then replacing all previous steps with `make sdist`, grabbing the resulting sdist (we'll need a way to automatically determine the name) and proceeding with the installer build as usual.

cpauya:
I'd like to use this and clean-up the installer scripts but I'm not sure we can squeeze this into 0.15.x @benjaoming :smile_cat:.   If we aim stability for that milestone, I think it's best to punt the use of this in the installers for 0.16.x.

MCGallaspy:
Agree with @cpauya, as per our discussion, I find it preferable to make no changes to the Windows installer build process in the limited time remaining -- let it be a project for the next cycle.

benjaoming:
Yeah, as @MCGallaspy mentioned we had a quick discussion about how to unify the installers more.

In 0.15, we introduce the Makefile and use it in the .deb process.

For 0.16, we use the Makefile for all installers -- and introduce the concept of an import/export for KA_LITE home directories such that languagepacks, database, and video downloads can be bundled for a preseeded installation. #4311

rtibbles:
Status, @benjaoming?

benjaoming:
The Makefile is a cool new and modern thing we dig now! :D

Issue #4375, opened 2015-09-04
Add documentation to sdist


benjaoming:
MANIFEST.in.dist should include the documentation.

The debian dist is built from an sdist, so that would immediately fix it. Windows already adds it in another way.

@cpauya does the OSX installer build from a PyPi sdist or do you create an sdist with `setup.py sdist`? Because in that case, fixing this issue will likely also add documentation to the OSX installer.

aronasorman:
Not right now, although eventually it will, so it would be good to have it working through PyPi.

benjaoming:
@aronasorman okay but does it use the output of `python setup.py sdist` because that was my impression...

aronasorman:
No, it does not use the `sdist` yet.

benjaoming:
Thanks, I got the link from Cyril to the setup.sh script :)

rtibbles:
Status?

benjaoming:
Well, the docs (html output like the stuff on RTD) are bundled in, and part of the sdist. What's remaining now is that they are served in urls.py?

It's a waste to ship the docs without making the accessible to users. And it's not that hard to add, let this be a release blocker, I shall attend to it.

MCGallaspy:
It's already the case. However, the link only appears in the user menu when you're logged in.

MCGallaspy:
The docs are served as static files: https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/settings/base.py#L406

MCGallaspy:
Link is created here: https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/distributed/static/js/distributed/user/hbtemplates/navigation.handlebars#L47

MCGallaspy:
Admittedly there's probably a much better way to present it.

benjaoming:
Maybe we can assign this to someone who will come up with a suggestion of how to refer the user to the bundled docs?

MCGallaspy:
I'll do it.

benjaoming:
Just a quick note: Currently, we show documentation link to learners.. that doesn't make much sense?

![screenshot from 2016-02-15 11 49 30](https://cloud.githubusercontent.com/assets/374612/13046636/330da44e-d3da-11e5-8e49-1eda66cbb463.png)

MCGallaspy:
> that doesn't make much sense?

Who should be able to see it then?

MCGallaspy:
[...]

Issue #4348, opened 2015-09-02
Tweaks to make to embedded documentation


jamalex:
- [ ] Provide an easy way to link back to the KA Lite interface itself -- currently, no escape. Perhaps (eventually) embedding the docs directly within the KA Lite interface (having a view under /docs/ that maps paths, loads the appropriate HTML docs file, and renders it into the main page context in the content area)
- [ ] Allow link to docs even when not logged in
- [ ] Don't include "install guide"? Maybe? Or at least link to /static/html/usermanual/userman_main.html as the default page. Most of the sections other than "usermanual" are actually largely not so useful on an installed offline server, so we may want to just ship the "usermanual" portion?
- [x] Don't include "View page source" or "View on Github" links, to avoid confusion.
- [ ] Don't include links in "Built with Sphinx using a theme provided by Read the Docs." text at bottom (or delete whole line)
- [x] Change "Copyright 2015, FLE" to "Copyright 2015, Learning Equality, licensed under a Creative Commons Attribution-ShareAlike 4.0 International License".
- [x] Remove online badges

Ad Removing online badges:

In 0.15, we started using an ReST README because it can be reused for the docs so we don't maintain the same stuff in multiple places. It would be nice if the offline build had the section with badges from online sources disabled, though

http://sphinx-doc.org/config.html#conf-tags
http://sphinx-doc.org/markup/misc.html#including-content-based-on-tags

MCGallaspy:
I'd like to first address these in 0.15, and then create a patch for 0.14.x time permitting.

benjaoming:
#4375 addresses that documentation is actually not part of what's currently distributed. That needs to be solved.

@MCGallaspy The release notes actually mention that we are going to have documentation for OSX/deb for 0.14 soon.

benjaoming:
> Don't include "View page source" or "View on Github" links, to avoid confusion.

The "Edit on Github" is IMO very important. Having a very very low bar for contributing to the documentation is important.

The documentation is intended for administrators / devs, so I think they can understand that it's pointing to an online location. Even if they don't have internet access, it'll just signal that this is an open source thing.

MCGallaspy:
> The release notes actually mention that we are going to have documentation for OSX/deb for 0.14 soon.

[...]

PR #4526, opened 2015-09-28
Add Makefile and distribute docs in sdist

benjaoming:
# Change affecting 0.15 release

Docs are included in the sdist and built automatically by running `make sdist`.

PR for #4348 (partly) and #4375

# Change affecting build process

The PyPi and .deb for 0.15 will be released by using a new Makefile. Yay!

#4459 

Makefile does this:

```
➜  ka-lite git:(dist-docs) make help
clean - remove all build, test, coverage and Python artifacts
clean-build - remove build artifacts
clean-pyc - remove Python file artifacts
clean-test - remove test and coverage artifacts
lint - check style with pep8
test - run tests quickly with the default Python
test-all - run tests on every Python version with tox DISABLED
assets - build all JS/CSS assets
coverage - check code coverage quickly with the default Python
docs - generate Sphinx HTML documentation, including API docs
release - package and upload a release
dist - package locally
install - install the package to the active Python's site-packages
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,82 @@
+.PHONY: clean-pyc clean-build docs clean
+
+help:
+	@echo "clean - remove all build, test, coverage and Python artifacts"
+	@echo "clean-build - remove build artifacts"
+	@echo "clean-pyc - remove Python file artifacts"
+	@echo "clean-test - remove test and coverage artifacts"
+	@echo "lint - check style with pep8"
+	@echo "test - run tests quickly with the default Python"
+	@echo "test-all - run tests on every Python version with tox DISABLED"
+	@echo "assets - build all JS/CSS assets"
+	@echo "coverage - check code coverage quickly with the default Python"
+	@echo "docs - generate Sphinx HTML documentation, including API docs"
+	@echo "release - package and upload a release"
+	@echo "dist - package locally"
+	@echo "install - install the package to the active Python's site-packages"
+
+clean: clean-build clean-pyc clean-test
+
+clean-build:
+	rm -fr build/
+	rm -fr dist/
+	rm -fr .eggs/
+	rm -fr dist-packages/
+	rm -fr dist-packages-temp/
+	find . -name '*.egg-info' -exec rm -fr {} +
+	find . -name '*.egg' -exec rm -f {} +
+
+clean-pyc:
+	find . -name '*.pyc' -exec rm -f {} +
+	find . -name '*.pyo' -exec rm -f {} +
+	find . -name '*~' -exec rm -f {} +
+	find . -name '__pycache__' -exec rm -fr {} +
+
+clean-test:
+	rm -fr .tox/
+	rm -f .coverage
+	rm -fr htmlcov/
+
+lint:
+	pep8 kalite
+
+test:
+	python setup.py test
+
+test-all:
+	@echo "Not supported yet"
+	# tox
+
+coverage:
+	coverage run --source kalite_gtk setup.py test
+	coverage report -m
+	coverage html
+	open htmlcov/index.html
+
+docs:
+	# rm -f docs/ka-lite.rst
+	# rm -f docs/modules.rst
+	# sphinx-apidoc -o docs/ ka-lite-gtk
+	$(MAKE) -C docs clean
+	$(MAKE) -C docs html
+	# open docs/_build/html/index.html
+
+assets: clean
+	# Necessary because NPM may have wrong versions in the cache
+	npm cache clean
+	npm install --production
+	node build.js
+	bin/kalite manage compileymltojson
+
+release: clean docs assets
+	python setup.py sdist --formats=gztar,zip upload --sign
+	python setup.py sdist --formats=gztar,zip upload --sign --static
+	ls -l dist
+
+dist: clean docs assets
+	python setup.py sdist
+	python setup.py sdist --static
+	ls -l dist
+
+install: clean
+	python setup.py install
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,9 @@ help:
 	@echo "clean-pyc - remove Python file artifacts"
 	@echo "clean-test - remove test and coverage artifacts"
 	@echo "lint - check style with pep8"
-	@echo "test - run tests quickly with the default Python"
-	@echo "test-all - run tests on every Python version with tox DISABLED"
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -162,6 +171,11 @@ def gen_data_files(*dirs):
     gen_data_files('static-libraries')
 )
 
+data_files += map(
+    lambda x: (os.path.join(kalite.ROOT_DATA_PATH, x[0]), x[1]),
+    gen_data_files('docs/_build/html', optional=True)
+)
+
```
benjaoming: Notice optional=True which is a new thing for `gen_data_files`

Makefile
```diff
@@ -0,0 +1,82 @@
...
+	@echo "coverage - check code coverage quickly with the default Python"
+	@echo "docs - generate Sphinx HTML documentation, including API docs"
+	@echo "release - package and upload a release"
+	@echo "dist - package locally"
+	@echo "install - install the package to the active Python's site-packages"
+
+clean: clean-build clean-pyc clean-test
+
+clean-build:
+	rm -fr build/
```
MCGallaspy: The equivalent command on windows is `rd /s /q build`

MCGallaspy: Though it seems you can use `rm` if you're using the [MSYS shell](http://www.mingw.org/wiki/msys) instead of cmd.exe.

Makefile
```diff
@@ -0,0 +1,82 @@
...
+	rm -fr dist/
+	rm -fr .eggs/
+	rm -fr dist-packages/
+	rm -fr dist-packages-temp/
+	find . -name '*.egg-info' -exec rm -fr {} +
```
MCGallaspy: Seems like MSYS also provides `find` with the same interface.

Makefile
```diff
@@ -0,0 +1,82 @@
...
+	find . -name '*~' -exec rm -f {} +
+	find . -name '__pycache__' -exec rm -fr {} +
+
+clean-test:
+	rm -fr .tox/
+	rm -f .coverage
+	rm -fr htmlcov/
+
+lint:
+	pep8 kalite
```
MCGallaspy: Windows user would need to install pep8 and ensure it's on the path (by default it's not) in order for this to work.

Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+test:
+	python setup.py test
+
+test-all:
+	@echo "Not supported yet"
+	# tox
+
+coverage:
+	coverage run --source kalite_gtk setup.py test
```
MCGallaspy: Is that a script from [the coverage package on PyPI](https://pypi.python.org/pypi/coverage/3.7.1)? Same caveat as `pep8`.

Makefile
```diff
@@ -0,0 +1,82 @@
...
+	coverage report -m
+	coverage html
+	open htmlcov/index.html
```
MCGallaspy: `open` not available on Windows as far as I can tell. Seems like the closest equivalent is just to execute the html file directory, then Windows will open it in the default browser.

MCGallaspy: Or more likely whatever program is associated to html files... may not necessarily be the browser.

Makefile
```diff
@@ -0,0 +1,82 @@
...
+	$(MAKE) -C docs clean
+	$(MAKE) -C docs html
+	# open docs/_build/html/index.html
+
+assets: clean
+	# Necessary because NPM may have wrong versions in the cache
+	npm cache clean
+	npm install --production
+	node build.js
+	bin/kalite manage compileymltojson
```
MCGallaspy: Won't work, since Windows can't associate `bin/kalite` with python and doesn't understand the `#!` notation.

MCGallaspy: Alternative is to use `kalite.bat`

Makefile
benjaoming: It's the same for all I guess: "pip install pep8"

Makefile
benjaoming: Oh yeah this part is still copy paste stuff, needs to be adapted when I port the contents of circle.yml

Makefile
benjaoming: so `python bin/kalite` would work?

MCGallaspy: Actually, it appears MSYS interprets the `#!` correctly as long as python.exe is on the path.

MCGallaspy:
I think the takeaway is that we can probably get this working with a little effort on Windows using MinGW.

MCGallaspy:
BTW, who is assigned to this? What part of #4348 does it address? Is this a release-blocker? I thought I was just supposed to give this a sanity check, but if you really wish to include it in 0.15.x, then I would also like to make sure the makefile is usable on Windows by testing it out.

aronasorman:
@rtibbles and I can sign off on this.

Makefile
aronasorman: Can you add `pep8` to `requirements_dev.txt` @benjaoming?

MCGallaspy:
Are one of you going to make sure it's usable on Windows? Or do you want me to do that?

aronasorman:
You'll be the best one for that @MCGallaspy. I can test this on the Mac as well. 

But getting parity across all platforms isn't important for me at the moment, since this will only be used by the deb installer for 0.15. So I wouldn't be too strict at this point.

benjaoming:
@aronasorman correct, the Makefile is not part of the release as pointed out in the description of the PR.

However, the setup.py changes are... but theyre not very risky.

benjaoming:
I'm having issues with a constant `pkg_resources.DistributionNotFound: ka-lite==0.15.0` running kalite commands on Circle.

Logging in with SSH, I can reissue `pip install -e .` and it works. I don't get why it doesn't initially work. I've tried rebuilding without a cache.

@aronasorman any clues here?

aronasorman:
@benjaoming I've seen that error during 0.14, what I did was to install it first in my virtualenv. Maybe try adding that line in circle.yml?

MCGallaspy:
Re: my earlier comment, I'll sign off on this.

benjaoming:
@MCGallaspy that's cool, once it's merged we can go into more details about Win compat

@aronasorman I'm not sure what line? The circle.yml is identical to the functional one in 0.15.x, and it has enabled the virtualenv at the point of issuing `pip install -e .`.

I'm gonna create a new PR in case Circle has a messed up cache. Then I can also clean up commit history.

PR #4533, opened 2015-09-29
Add FAQ entry on installing language packs offline.

### LINKED ISSUE #4049: Document the ability to install language packs offline
(From call with Team4Tech today)

Right now, [according to our documentation](http://field.learningequality.org/develop/docs/usermanual/userman_admin.html#language-packs), you need to have Internet access to download language packs. However, we actually have a management command (`kalite manage languagepackdownload --from-file=FILE`) that allows for this to be done offline, and I was asked today how to do that.

Two things are needed here:
- Additional documentation, possibly as a separate "doing installations offline" section also linked from both the Download Videos and the Download Language Packs sections. Basically, we need to tell people what files to copy, where to put the files after installing, and what commands to run.
- Links to the zip files for the language packs for each language. Might be easiest to just hardcode these into the docs for now, unless we want to finish fixing up https://hub.learningequality.org/languages/dashboard instead (@aronasorman?)

This isn't a release blocker, but will be important to do by late July for Team4Tech's deployments.

aronasorman:
Fixes #4049.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -35,6 +35,50 @@ I would like to download the videos for KA Lite via BitTorrent, is this possible
 
 Yes! Please see the instructions for ":ref:`bulk-video-downloads`".
 
+How can I install a language pack without a reliable internet connection?
+-------------------------------------------------------------------------
+
+To install, a language pack on a computer without internet access, look up the language code for the language you wish to download:
+
+    ================ ======
+     Language name    Code
+    ================ ======
+     Arabic           ar
+     Bulgarian        bg
+     Burmese          my
+     Danish           da
+     English          en
+     French           fr
+     German           de
+     Hindi            hi
+     Kannada          kn
+     Lao              lo
+     Polish           pl
+     Portuguese, BR   pt-BR
+     Spanish          es
+     Swahili          sw
+     Tamil            ta
+     Xhosa            xh
+     Zulu             zul
+    ================ ======
+
+Once you've determined the right language code, you can download the right language pack using this url::
+
+  http://keep.learningequality.org/media/language_packs/0.15/<code>.zip
+
+where <code> is replaced by the right language code. You can now carry around that zip file to computers you want to install the language pack to.
+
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -35,6 +35,50 @@ I would like to download the videos for KA Lite via BitTorrent, is this possible
...
+
+where <code> is replaced by the right language code. You can now carry around that zip file to computers you want to install the language pack to.
+
+You can install the language pack with the following command (for Debian and OS X)::
+
+  kalite manage languagepackdownload --from-file=<path to zip file> -l<language code> --commandline
+
+Or for Windows::
+
+  <path to KA Lite installation>\bin\windows\kalite.bat languagepackdownload --from-file=<path to zip file> -l<language code> --commandline
```
MCGallaspy: Is this interactive? `kalite.bat` currently has a known issue where it can't properly handle user input.

aronasorman: Not interactive, the `--commandline` flag is for just allowing us to run this thing in the terminal.

MCGallaspy:
Whenever you give an example like this: `<path to KA Lite installation>\bin\windows\kalite.bat languagepackdownload --from-file=<path to zip file> -l<language code> --commandline`

Where the parts enclosed in the angle brackets (`<` & `>`) are meant to be replaced, I would also provide an example after the replacement is made -- the notation be unfamiliar to some people. (I can dig up some old emails where this was the case, if you want some proof!)

aronasorman:
I concur. Let me add that.

docs/faq.rst
```diff
@@ -35,6 +35,54 @@ I would like to download the videos for KA Lite via BitTorrent, is this possible
...
+
+An example invocation for Windows would be::
+
+  C:\Program Files\KA Lite\ka-lite\bin\windows\kalite.bat languagepackdownload --from-file="C:\Downloads\es.zip" -les --commandline
```
aronasorman: @MCGallaspy example here.

Issue #4458, opened 2015-09-19
Catch sigterm and shutdown cleanly


benjaoming:
How to here:

https://nattster.wordpress.com/2013/06/05/catch-kill-signal-in-python/

Use case example:

> I have a desktop entry to “start KA-Lite server on localhost” but I’ve noticed that if KA-Lite server is running and then user logs out or powers down, I get a status message from KA-Lite next time saying  “unclean shutdown (7)”. Do you recommend I create another desktop entry to shut down the server and hope users will remember to click it before logging off?  Seems like the logoff KILL process isn’t catching KA-Lite.

aronasorman:
Perhaps you meant `SIGTERM`? We can pobably use the `atexit` module to handle shutdowns cleanly.

benjaoming:
Could be that atexit is better than said link -- any opinions?

rtibbles:
We are currently using the 'atexit' module for closing browserify while running the dev server - I imagine it is probably more cross platform also.

benjaoming:
Seems like this part of `cherrypy.process.wspbus` might be the issue:

    def block(self, interval=0.1):
        """Wait for the EXITING state, KeyboardInterrupt or SystemExit.
        
        This function is intended to be called only by the main thread.
        After waiting for the EXITING state, it also waits for all threads
        to terminate, and then calls os.execv if self.execv is True. This
        design allows another thread to call bus.restart, yet have the main
        thread perform the actual execv call (required on some platforms).
        """
        try:
            self.wait(states.EXITING, interval=interval, channel='main')
        except (KeyboardInterrupt, IOError):
            # The time.sleep call might raise
            # "IOError: [Errno 4] Interrupted function call" on KBInt.
            self.log('Keyboard Interrupt: shutting down bus')
            self.exit()
        except SystemExit:
            self.log('SystemExit raised: shutting down bus')
            self.exit()
            raise

Notice the last `raise`, it reraises SystemExit, we have to handle that in kalitectl...

benjaoming:
Sigterm is already caught, it's not the culprit of the original problem. Killing KA Lite does not create "Unclean shutdown."

rtibbles:
Fixed in #4536

PR #4536, opened 2015-09-29
Catch SystemExit and to shutdown cleanly #4458

benjaoming:
I don't believe the original report can be recreated by simply issuing a kill. Not even `kill -9`.

This was the only thing that came to mind.

#4458


rtibbles:
This looks safe. Merging.

PR #4537, opened 2015-09-29
initialize color palette implementation

66eli77:
It's a working prototype, just throw here for progress tracking and idea sharing.
@rtibbles I assign you since you are most familiar with the overall design, but feel free to assign others if you know someone's also interested.


66eli77:
this feature is basically done, we can now dynamically change the colors for kalite system-wide.

rtibbles:
Note: For modern browsers, there is a color picker included, so we could just use the:
```Modernizr.inputtypes.color``` to check if we need a color picker and load one as a polyfill if necessary.

**Comment by rtibbles** [comment]
I am also not convinced that we need LESS here to do what we want to do. It should be simple enough to create the CSS that we want to insert using a handlebars template, without having to load in LESS to do the processing.

**Comment by 66eli77** [comment]
@rtibbles the Modernizr included in our node_modules seems do not have support for inputtypes

**Comment by rtibbles** [comment]
2 possibilities here, @66eli77 - either we should upgrade our Modernizr, or more likely, you need to 'activate' the Modernizr test. It is a slightly persnickety process using it within browserify.

See here to see how to include specific Modernizr tests: https://github.com/learningequality/ka-lite/blob/develop/kalite/distributed/static/js/distributed/bundle_modules/base.js#L13

**Comment by 66eli77** [comment]
Cool, thanks @rtibbles , that worked out.
I'm facing this require("a_minified_js") problem again, watchify just keep complains about some "error" in the js library. Do you know any workaround?

File: kalite/distributed/static/js/distributed/color_palette/color_palette.js [hunk 1]
```diff
@@ -0,0 +1,127 @@
...
+            k_bg_clr: $("#my_background").val(),
+            k_acn_clr: $("#my_accent").val(),
+            k_act_clr: $("#my_action").val(),
+            k_hln_clr: $("#my_headline").val(),
+            k_byt_clr: $("#my_bodytext").val(),
+        }));
+    }
+});
+
+$(function(){
```
rtibbles: Could some of this logic be encapsulated in another Backbone View? We can then just delegate the initialization of that view to the ```$()``` wrapper.

Would also be good to put the ```$()``` in a separate file, so we can expose the colour picker code as a pluggable module, separate from its in page instantiation.

kalite/distributed/static/js/distributed/color_palette/color_palette.js
```diff
@@ -0,0 +1,127 @@
...
+    template: colorPaletteTemplate,
+    tagName: "style", 
+
+    initialize: function() {
+        this.render();
+    },
+
+    render: function(){
+        this.$el.html(this.template({
+            k_bg_clr: $("#my_background").val(),
```
rtibbles: Where are these colours getting picked from? Would be good to pull a lot of the code out from below and include it in this view, so that we can use Backbone events to handle the value changes etc.

benjaoming:
@66eli77 could you retarget this PR for the `develop` branch?

The test errors are the old timeout errors that we got rid of so hopefully you can make this work easily...

benjaoming:
Closing this as we've refocused to Kolibri and the feature is much more likely to be implemented there...

Issue #4545, opened 2015-09-30
Delete my content rating button visibility 


radinamatic:
Delete my content rating button visibility 

**Branch**: 0.15 (**ka-lite_0.15~b2-0ubuntu1_all.deb** installer  on Ubuntu 15.04, but it's not an installer issue)

**Current Behavior**: 

`Delete my review` button is visible even if the user did not leave any review (yet):

![responsive-cr](https://cloud.githubusercontent.com/assets/1457929/10201489/994437ae-67ac-11e5-945b-c15031a7ed05.png)

**Expected Behavior**:
I believe the user experience would be better if the button appeared only **after** they actually left the "review".

MCGallaspy:
A valid point.

rtibbles:
Fixed in #4548

Issue #4527, opened 2015-09-28
content_rating test not happy


benjaoming:
Any chance getting this fixed for 0.15? It's a test so not a dangerous fix, just that it'd be nice to get rid of these inconclusive / random test failures.

Circle CI output...

```
  Scenario: User alters text feedback       # kalite/distributed/features/content_rating.feature:18
    Given I have filled out a feedback form # kalite/distributed/features/steps/content_rating.py:113
    Given I have filled out a feedback form # kalite/distributed/features/steps/content_rating.py:113 8.351s
      Assertion Failed: FAILED SUB-STEP: When I fill out the form
      Substep info: Traceback (most recent call last):
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1173, in run
          match.run(runner.context)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1589, in run
          self.func(context, *args, **kwargs)
        File "/home/ubuntu/ka-lite/kalite/distributed/features/steps/content_rating.py", line 108, in impl
          enter_star_ratings(context)
        File "/home/ubuntu/ka-lite/kalite/distributed/features/steps/content_rating.py", line 159, in enter_star_ratings
          rate_id(context, id_)
        File "/home/ubuntu/ka-lite/kalite/distributed/features/steps/content_rating.py", line 179, in rate_id
          rating_el.click()
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/selenium/webdriver/remote/webelement.py", line 65, in click
          self._execute(Command.CLICK_ELEMENT)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/selenium/webdriver/remote/webelement.py", line 402, in _execute
          return self._parent.execute(command, params)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 175, in execute
          self.error_handler.check_response(response)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 166, in check_response
          raise exception_class(message, screen, stacktrace)
      StaleElementReferenceException: Message: Element is no longer attached to the DOM
[...]

PR #4548, opened 2015-10-01
Rating delete hide

MCGallaspy:
Quick fix to #4545. Hope it looks safe. Can re-target to develop if desired.

@benjaoming I also take a stab at #4527. See commit message.


benjaoming:
Trying a rebuild on the successful test :D Nobody is that weird :D

rtibbles:
Looks decent.

PR #4551, opened 2015-10-01
annotate_content_models has been optimized and memory overhead reduce…

gwong89:
…d by 40% from original

@rtibbles  Me and @aronasorman  were wondering if the update check is necessary since it looks like all contents are not empty? line 433 in content_models.py

here  is the gist with the new improvment

https://gist.github.com/gwong89/bc11bbd727eeb715e292


aronasorman:
Can you post the latest gist @gwong89?

aronasorman:
63 MB vs. 130 MB. Nice improvement :)

kalite/topic_tools/content_models.py
```diff
@@ -16,13 +16,14 @@
...
 import sqlite3
+import itertools
 
 from peewee import Model, SqliteDatabase, CharField, TextField, BooleanField, ForeignKeyField, PrimaryKeyField, Using, DoesNotExist, fn
 from playhouse.shortcuts import model_to_dict
 
 from .settings import CONTENT_DATABASE_PATH, CHANNEL
 from .annotate import update_content_availability
-
+from memory_profiler import profile
```
aronasorman: Please remove all the memory_profiler calls.

kalite/topic_tools/content_models.py
```diff
@@ -416,19 +429,19 @@ def recurse_availability_up_tree(node, available):
...
                     for item in items:
-                        if item.kind != "Topic":
-                            item_data = unparse_model_data(model_to_dict(item, recurse=False))
-                            item_data.update(update)
-                            item_data = parse_model_data(item_data)
-                            for attr, val in item_data.iteritems():
-                                setattr(item, attr, val)
-                            item.save()
-                            recurse_availability_up_tree(item, update.get("available", False))
+                            #item_data = unparse_model_data(model_to_dict(item, recurse=False))
```
aronasorman: @rtibbles we removed this section as we assumed that the only thing we need to update here is the availability, not the rest of the fields.

Come to think of it, we don't have the `update` call here @gwong89.

aronasorman:
We may need to clean this PR a bit.

kalite/topic_tools/content_models.py
```diff
@@ -126,11 +125,8 @@ def wrapper(*args, **kwargs):
 
         if dicts and output:
             if expanded:
-                output = map(unparse_model_data, output.dicts())
-            else:
-                output = [item for item in output.dicts()]
-
-        return output
+                for dic in output:
```
aronasorman: We don't need to change this.

kalite/topic_tools/content_models.py
```diff
@@ -395,11 +390,21 @@ def annotate_content_models(db=None, channel="khan", language="en", ids=None, **
     Our ids can be duplicated at the moment, so this may be several content items per id.
     When a content item has been updated, propagate availability up the topic tree.
     """
-    content_models = get_content_items(ids=ids, channel=channel, language=language)
-    updates_dict = update_content_availability(content_models)
+    def iterator_content_items(ids=None, **kwargs):
+            if ids:
```
aronasorman: 8 spaces. We use 4 spaces across the project.

kalite/topic_tools/content_models.py
```diff
@@ -126,9 +125,10 @@ def wrapper(*args, **kwargs):
 
         if dicts and output:
             if expanded:
-                output = map(unparse_model_data, output.dicts())
+                for dic in output:
+                    output = map(unparse_model_data, output.dicts())
             else:
-                output = [item for item in output.dicts()]
+                    output = [item for item in output.dicts()]
```
rtibbles: Looks like too much indent to me.

kalite/topic_tools/content_models.py
```diff
@@ -416,19 +425,14 @@ def recurse_availability_up_tree(node, available):
...
-            for path, update in updates_dict.iteritems():
+            for path, update in content_models:
                 if update:
                     # We have duplicates in the topic tree, make sure the stamping happens to all of them.
-                    items = Item.select().where(Item.path == path)
+                    items = Item.select().where((Item.path == path) & (Item.kind != "Topic"))
+
                     for item in items:
-                        if item.kind != "Topic":
-                            item_data = unparse_model_data(model_to_dict(item, recurse=False))
```
rtibbles: Deleting these lines makes this whole thing a noop. You aren't actually updating any of the content nodes.

kalite/topic_tools/content_models.py
```diff
@@ -395,11 +394,21 @@ def annotate_content_models(db=None, channel="khan", language="en", ids=None, **
...
+    def iterator_content_items(ids=None, **kwargs):
+        if ids:
+            items = Item.select().where(Item.id.in_(ids)).dicts().iterator()
+        else:
+            items = Item.select().dicts().iterator()
 
-    with Using(db, [Item]):
+        mapped_items = itertools.imap(unparse_model_data, items)
+
+        updated_mapped_items = update_content_availability(mapped_items)
```
rtibbles: As mentioned verbally, we should modify ```update_content_availability``` to be a generator too, otherwise this will just cause inflation again.

rtibbles:
I fear the main memory gains here so far are from the fact the database is not being modified at all.

kalite/topic_tools/annotate.py
```diff
@@ -30,14 +30,15 @@ def update_content_availability(content_list, language="en"):
...
+    # turn this whole function into a generator
     try:
         contents_folder = os.listdir(django_settings.CONTENT_ROOT)
     except OSError:
         contents_folder = []
 
     subtitle_langs = {}
-
-    updates = {}
+    updates = set()
```
rtibbles: If we're just yielding an ```update``` object on every iteration, we probably don't want this here, as it will be making a reference to each yielded object and prevent them from being garbage collected.

kalite/topic_tools/annotate.py
```diff
@@ -150,6 +148,6 @@ def update_content_availability(content_list, language="en"):
                 update["available"] = False
 
         # Path is the only unique key available.
-        updates[content.get("path")] = update
-
-    return updates
+        updates.update(content.get("id"))
```
rtibbles: See above.

kalite/topic_tools/content_models.py
```diff
@@ -171,9 +173,9 @@ def get_content_items(ids=None, db=None, **kwargs):
     """
     with Using(db, [Item]):
         if ids:
-            values = Item.select().where(Item.id.in_(ids))
+            values = Item.select().where(Item.id.in_(ids)).iterator()
```
rtibbles: These changes should be reverted.

kalite/topic_tools/content_models.py
```diff
@@ -416,19 +427,15 @@ def recurse_availability_up_tree(node, available):
                     recurse_availability_up_tree(parent, available)
 
 
-            for path, update in updates_dict.iteritems():
+            for path, update in content_models:
                 if update:
                     # We have duplicates in the topic tree, make sure the stamping happens to all of them.
-                    items = Item.select().where(Item.path == path)
+                    
+                    items = Item.select().where((Item.path == path) & (Item.kind != "Topic"))
```
rtibbles: Not sure why this is necessary, as path is a unique key.

kalite/topic_tools/content_models.py
```diff
@@ -416,19 +427,15 @@ def recurse_availability_up_tree(node, available):
...
+                    Item.update("available", True).where((Item.path == path) & (Item.kind != "Topic"))
```
rtibbles: You are updating things irregardless of the contents of ```update``` here. It is possible that something should be flagged as unavailable.

Also, there is other data stored in update, aside from availability.

rtibbles: See the code you have deleted for more information.

kalite/topic_tools/tests/annotate_tests.py
```diff
@@ -0,0 +1,82 @@
...
+ u'{"sha": "00808cbc066910be1220235faa7d93a9126f6a57", "live": true, "id": "x1855395b96b1e34f"}',
+ u'{"sha": "90db973ee0fd319c596410e215a76a651601901d", "live": true, "id": "x477b6212a24d08da"}',
+ u'{"sha": "f6cf4d0675de30afee8cab929b97000e2c41fbc8", "live": true, "id": "xfc8946e7c80f2800"}',
+ u'{"sha": "4f522fb965ab3805f56250db187fefadb281edd2", "live": true, "id": "x14784d8f428fa949"}',
+ u'{"sha": "849c88aa2122ea5a40bb39540acf551d3ac67728", "live": true, "id": "x3f2d3b6cb53f67a4"}',
+ u'{"sha": "a9197ffb73ad014d477c7eb1eed41f9ea025774b", "live": true, "id": "xb6e923d2f396f5ab"}',
+ u'{"sha": "446eef76c02fc29049012312c704c12b4b54ba8d", "live": true, "id": "x9ef4ca7e914c87ec"}']
+    import json
+    stuff = json.dumps(thedict)
+    import pdb; pdb.set_trace()
```
rtibbles: This is going to cause you some issues during testing...

kalite/topic_tools/tests/annotate_tests.py
```diff
@@ -0,0 +1,82 @@
...
+ u'{"sha": "bb340ca13b104a9b9d1065b74a2fdfa423f817d1", "live": true, "id": "xa756d02df7435e1a"}',
+ u'{"sha": "f65a9c4a6a8df260021fe424d562e8ff6d1741bd", "live": true, "id": "x6b9db70231ff254d"}',
+ u'{"sha": "00808cbc066910be1220235faa7d93a9126f6a57", "live": true, "id": "x1855395b96b1e34f"}',
+ u'{"sha": "90db973ee0fd319c596410e215a76a651601901d", "live": true, "id": "x477b6212a24d08da"}',
+ u'{"sha": "f6cf4d0675de30afee8cab929b97000e2c41fbc8", "live": true, "id": "xfc8946e7c80f2800"}',
+ u'{"sha": "4f522fb965ab3805f56250db187fefadb281edd2", "live": true, "id": "x14784d8f428fa949"}',
+ u'{"sha": "849c88aa2122ea5a40bb39540acf551d3ac67728", "live": true, "id": "x3f2d3b6cb53f67a4"}',
+ u'{"sha": "a9197ffb73ad014d477c7eb1eed41f9ea025774b", "live": true, "id": "xb6e923d2f396f5ab"}',
+ u'{"sha": "446eef76c02fc29049012312c704c12b4b54ba8d", "live": true, "id": "x9ef4ca7e914c87ec"}']
+    import json
```
rtibbles: This should be imported in the global scope.

kalite/topic_tools/content_models.py
```diff
@@ -416,19 +427,15 @@ def recurse_availability_up_tree(node, available):
...
                     for item in items:
```
aronasorman: It turns out we don't need this for loop anymore; the IDs passed into this function are already updated through the `update_content_availability` call. Instead we need to make sure that each unique path is passed into the `recurse_availability_up_tree` function.

kalite/topic_tools/tests/annotate_tests.py
```diff
@@ -0,0 +1,82 @@
...
+    expected = {
+      "thepath" :  {'all_assessment_items': [],
+              'description': 'test',
+              'title': 'testing '}      
+    }
+    with Using(self.db, [Item]):
+      actual = update_content_availability([unparse_model_data(model_to_dict(self.item))])
+      self.assertEqual(expected, actual)
+
+  '''def test_khan_exercises_availiability_true(self):
```
aronasorman: Can you remove these commented out tests?

kalite/topic_tools/tests/annotate_tests.py
```diff
@@ -0,0 +1,82 @@
...
+from playhouse.shortcuts import model_to_dict
+import json
+
+class AnnotateTestCase(TestCase):
+  TITLE = "testing "
+  AVAILABLE = False
+  KIND = "Exercise"
+  
+  @set_database
+  def setUp(self, db=None):
```
aronasorman: Your text editor keeps changing its indentation. Now you're using 2 spaces here. Before you were using 8 spaces. Can you sort this out and set this to 4 spaces?

kalite/topic_tools/tests/annotate_tests.py
```diff
@@ -0,0 +1,82 @@
...
+              'title': 'testing '}      
+    }
+    with Using(self.db, [Item]):
+      actual = update_content_availability([unparse_model_data(model_to_dict(self.item))])
+      self.assertEqual(expected, actual)
+
+
+  def test_update_content_availability_false(self):
+    expected = {
+      "thepath" :  {'all_assessment_items': [],
```
aronasorman: The way you indent things makes things harder to read in general.

kalite/topic_tools/content_models.py
```diff
@@ -416,19 +427,18 @@ def recurse_availability_up_tree(node, available):
...
-                        if item.kind != "Topic":
-                            item_data = unparse_model_data(model_to_dict(item, recurse=False))
-                            item_data.update(update)
-                            item_data = parse_model_data(item_data)
-                            for attr, val in item_data.iteritems():
-                                setattr(item, attr, val)
-                            item.save()
-                            recurse_availability_up_tree(item, update.get("available", False))
+                    item = Item.get().where(Item.path == path)
+                    if item.kind != "Topic":
```
aronasorman: @rtibbles we removed the `for` loop here thanks to your tip of `path` being unique. It also turns out that we only save about 3 MB by moving this to the DB, so you were right that we don't save much memory here.

kalite/topic_tools/annotate.py
```diff
@@ -64,10 +64,8 @@ def update_content_availability(content_list, language="en"):
         exercise_templates = []
 
     for content in content_list:
-
         # Some nodes are duplicated, but they require the same information
         # regardless of where they appear in the topic tree
-        if content.get("id") not in updates:
```
aronasorman: @rtibbles we removed the tracking of `ids`, as per what we talked about before.

kalite/topic_tools/annotate.py
```diff
@@ -148,8 +144,5 @@ def update_content_availability(content_list, language="en"):
             # Content is currently flagged as available, but is not. Flag as unavailable.
             if content.get("available") and "available" not in update:
                 update["available"] = False
-
-        # Path is the only unique key available.
-        updates[content.get("path")] = update
-
-    return updates
+    
+    yield content.get("path"),update
```
aronasorman: @gwong89 we normally put spaces between a `comma` and the variable.

kalite/topic_tools/content_models.py
```diff
@@ -126,9 +127,10 @@ def wrapper(*args, **kwargs):
 
         if dicts and output:
             if expanded:
-                output = map(unparse_model_data, output.dicts())
+                for dic in output:
+                    output = map(unparse_model_data, output.dicts())
```
aronasorman: Why does this need to be in a for loop? You're doing unnecessary work here.

aronasorman:
A lot of indentation fixes, some comments about logic.

rtibbles:
This file still needs to be removed:
```kalite/distributed/static/django_js_reverse/js/reverse.js```

File: kalite/distributed/views.py [hunk 23]
```diff
@@ -71,7 +71,7 @@ def check_setup_status_wrapper_fn(request, *args, **kwargs):
 
 
 @render_to("distributed/learn.html")
-def learn(request):
+def learn(request, db=None):
```
rtibbles: This change is unnecessary, please remove.

kalite/topic_tools/annotate.py
```diff
@@ -64,10 +65,8 @@ def update_content_availability(content_list, language="en"):
         exercise_templates = []
 
     for content in content_list:
-
         # Some nodes are duplicated, but they require the same information
         # regardless of where they appear in the topic tree
-        if content.get("id") not in updates:
             update = {}
```
rtibbles: Shouldn't this section get dedented then - if the 'if' statement is removed?

kalite/topic_tools/content_models.py
```diff
@@ -395,11 +396,21 @@ def annotate_content_models(db=None, channel="khan", language="en", ids=None, **
...
+
+        mapped_items = itertools.imap(unparse_model_data, items)
+        updated_mapped_items = update_content_availability(mapped_items)
 
+        for path, update in updated_mapped_items:
+            yield path, update
+
+    with Using(db, [Item]):
+        content_models = iterator_content_items(ids=ids, channel=channel, language=language)
         with db.atomic() as transaction:
```
rtibbles: With this whole block as an atomic transaction, what are the memory implications?

rtibbles:
These look like the right changes. One thing that needs to be removed, and I would like to see a rerun of the memory profile here too.

kalite/distributed/views.py
gwong89: noted and done

kalite/topic_tools/annotate.py
gwong89: noted and done

kalite/topic_tools/content_models.py
gwong89: heres a gist where it does a memory profiler on the change that me and aron worked compared to the original you had on updating content. The difference is very minor

https://gist.github.com/gwong89/69e196741cab984be86a

kalite/topic_tools/tests/annotate_tests.py
```diff
@@ -0,0 +1,81 @@
...
+            'description': 'test',
+            'title': 'testing '}      
+        }
+    
+        with Using(self.db, [Item]):
+            actual = dict(update_content_availability([unparse_model_data(model_to_dict(self.item))]))
+            self.assertEqual(expected, actual)
+
+
+    def test_update_content_availability_false(self):
```
MCGallaspy: @rtibbles this test and the one above are identical. Also, what are they testing in plain English? Is this a unit test for `update_content_availability`? Maybe we can spice it up a notch.
image

PR #4552, opened 2015-10-01
Causes syntax error when page is loaded

MCGallaspy:
Contains a minimal example of how to replicate an error.

So bizarre. @tk21 discovered this.

To replicate: start the server, middleware will intercept your request and render the homepage instead (this function is identical to the homepage view function), but you'll see the following error in the js console (which throws a wrench in everything): `Uncaught SyntaxError: Unexpected token <`

It's referring to the very first character in the document: `<!DOCTYPE HTML>`

@rtibbles @aronasorman @jamalex any idea what's up?

### DIFF
--- a/kalite/facility/middleware.py
+++ b/kalite/facility/middleware.py
@@ -16,8 +16,17 @@ def error_view(request):
     return {}
 
 class ErrorMiddleware:
-    def process_request(self, request):
-        return error_view(request)
+    def process_response(self, request, response):
+        # could also look at the request object, e.g.
+        # for the request to http://localhost/blah
+        # request['PATH_INFO'] is '/blah'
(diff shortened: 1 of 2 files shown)


rtibbles:
Is this while running with kalite.project.settings.dev? It is possible that this may be due to Django DEBUG toolbar middleware if so.

MCGallaspy:
Ah hah, obvious in hindsight, all requests are being caught by this, including some js assets.

MCGallaspy:
I'm looking at `dynamic.js` in particular, whatever that is.

MCGallaspy:
@tk21 got it figured out... see last commit message.

MCGallaspy:
Not really a bug.

PR #4554, opened 2015-10-02
Update the OSX documentation.

cpauya:
Hi @aron / @MCGallaspy - this updates the docs on "greyed-out assessment items on OSX during upgrade" and the uninstallation process on OSX.

This hopefully fixes https://github.com/learningequality/installers/issues/201

### DIFF
--- a/docs/installguide/install_all.rst
+++ b/docs/installguide/install_all.rst
@@ -27,6 +27,16 @@ Mac/OSX
 #. When prompted that KA-Lite has been started, click on the logo icon again and select "Open in Browser" menu option - this should launch KA-Lite on your preferred web browser.
 #. Login using the administrator account you have specified during setup.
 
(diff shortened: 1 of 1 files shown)


aronasorman:
Is there a way we can detect an upgrade @cpauya?

cpauya:
No, I haven't encountered a way to detect an upgrade yet @aronasorman.  However we are planning to use [Sparkle](http://sparkle-project.org/) in the app after we've moved to 0.16.x.  Haven't ready at [Packages](http://s.sudre.free.fr/Software/Packages/about.html) if it can detect an upgrade but we can create a script to check if there's already an existing `~/.kalite/` folder.

aronasorman:
Ok, let's keep this unfortunate bug for now and see that we get to the new installer for 0.16.

Issue #4535, opened 2015-09-29
Are migrations possibly blocked by stale content types?


benjaoming:
From `bin/kalite start`:

```
Running migrations for main:
 ! Migration main:0039_log_timestamps should not have been applied before main:0039_auto__del_assessmentitem__add_contentrating__add_unique_contentrating_ but was.
 - Migrating forwards to 0040_auto__del_unique_contentrating_content_kind_content_id_user__add_uniqu.
 > main:0039_auto__del_assessmentitem__add_contentrating__add_unique_contentrating_
 > main:0040_auto__del_unique_contentrating_content_kind_content_id_user__add_uniqu
The following content types are stale and need to be deleted:

    main | assessmentitem

Any objects related to these content types by a foreign key will also
be deleted. Are you sure you want to delete these content types?
If you're unsure, answer 'no'.

    Type 'yes' to continue, or 'no' to cancel: 
```

MCGallaspy:
Something's wrong with migrations. It kept trying to delete a table (main.assessment_items) that didn't exist, and now it thinks there's a stale content type? What about this is a blocker? The model does not even exist.

MCGallaspy:
I think we need to either 1) amend past migrations to remove this crud or 2) create a new migration that handles this.

benjaoming:
Did a migration for this model exist in our dev version at some point?

MCGallaspy:
That seems likely.

rtibbles:
I believe so, but I am damn sure we didn't release it.

rtibbles:
Was this during an upgrade, @benjaoming?

MCGallaspy:
Yeah, what branch is this? `0.15.x`? I'll try to resolve it.

benjaoming:
This was 0.15.x -- running on some kind of dev environment. Sorry about the lacking info from my side.

benjaoming:
Gonna test the PR, I'll merge when I verify it :)

benjaoming:
Had issues with ghost migrations from a previous develop branch. Those migrations were never part of 0.14 or 0.14.1. Mike cleaned up the migration path.

Everything looks good!

PR #4558, opened 2015-10-02
Remove & consolidate migrations

MCGallaspy:
Hopefully fixes https://github.com/learningequality/ka-lite/issues/4535, though I couldn't replicate that specific error. Definitely weirdness when upgrading from 0.14, though.


benjaoming:
Checked the master branch and master for these ghost migrations previously reported, they're definitely not there. Did a database setup of 0.14.1 -- and then migrated this one on top. No problems.

Then did a clean migration. Worked.

Then did from master (0.14.0). Worked.

PR #4560, opened 2015-10-02
15 into develop merge conflict resolution

MCGallaspy:
Let's run the tests!


MCGallaspy:
Re: the last two commits (https://github.com/MCGallaspy/ka-lite/commit/5ca6df059c6876986615c56791cfc49449af9b68 and https://github.com/MCGallaspy/ka-lite/commit/d7929318a496826c2b022612c012c651c546417c)

There's a quick fix: amend the test so that it has updated knowledge of languagepackdownload details. But what we're really trying to test is that stuff doesn't break when a non-English language is used. I have two ideas, but I want to check here before I get too far down that rabbit hole:

1. Create a fake language pack for testing purposes, and run the tests again using that language pack. Upshot is that we really do test that everything works well when using a different language, but we would be running the tests twice. Plus ideally program logic doesn't depend on the language setting, so it might be overkill.
2. In cases where we've identified a bug that *did* depend on the language and want to write a regression test, use a fake language pack for *those particular tests* and set it using e.g. a settings_override decorator.

Any preference here? Or am I missing the point, and I should just do what's most expedient?

rtibbles:
I think this looks good. And if the tests are passing, I think in this case that is a good sign. Let's merge this.

MCGallaspy:
Oh, I never did run the tests with a language pack activated.

rtibbles:
It's in develop, open an issue targeted for 0.16?

Issue #4471, opened 2015-09-23
Language selector not available for logged in users - seems problematic.


rtibbles:
Branch: 0.15.x
Ubuntu/Chrome

Expected Behaviour: Previously selected language should be retained on page navigation. Language selector should be available on every page.

Current Behaviour: Language defaults back to English, no language selector available.

![image](https://cloud.githubusercontent.com/assets/1680573/10053452/029178f6-61e1-11e5-8485-d9a0d2c3b004.png)

rtibbles:
Woops, sorry this is a login issue. Seems like we hide the language selector when you login, even as an admin.

rtibbles:
@jtamiace Can you comment on whether this is expected behaviour? Hiding language selector for logged in users?

jtamiace:
That doesn't seem right. I think it makes sense for the language selector to always be available, even for logged out users. You can only toggle between language packs that have been downloaded on a particular local server right?

rtibbles:
Yeah - I assume we must be doing this on purpose though, so was wondering when/how that happened.

rtibbles:
Ahh... it is flagged as 'not-admin-only'. So, it gets hidden for admins.

rtibbles:
https://github.com/learningequality/ka-lite/blob/0.15.x/kalite/distributed/templates/distributed/partials/_footer.html#L13

Should admins be able to switch language?

aronasorman:
@rtibbles the original reasoning for that was they already control the default language for everyone. But for consistency, it might be good to just have everyone be able to switch their own language. But then there might be confusion as to what language the admin is switching currently -- the default language or just their current language?

rtibbles:
Yeah - although if someone has more than one language installed, they have already managed to make it to the language pack download page, and hence have seen the 'default language' setting there, so they have some reason not to think that, at least.

radinamatic:
One argument in favor of admins having the ability to switch languages for themselves, is being able to ascertain if the content (videos & subtitles) is properly displayed in non-default language, **without** having to change the default language for all the users, or logging in as non-admins. 

Granted there might not be a wide user base for that case, but it would also make our testing and debugging a couple steps shorter... :smirk:

benjaoming:
Template code in `distributed/templates/distributed/partials/_footer.html` is to blame:

                
                {% if request.session.facility_user.is_teacher %}
                  <div class="teacher-only language-select">
                {% else %}
                  <div class="not-admin-only language-select">
[...]

PR #4577, opened 2015-10-05
Let everyone see the language selection #4471

benjaoming:
#4471


benjaoming:
The .not-admin-only CSS class doesn't even exist :) And I wouldn't know what it means :)

PR #4582, opened 2015-10-06
Put cache database in user-writable location

MCGallaspy:
@aronasorman @benjaoming let me know if this seems tenable. Will test it out on the deb by pulling source and setting `IS_SOURCE=False`, assuming this essentially replicates installing using deb.

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -146,7 +146,8 @@ else:
 # ...OR it's been installed in:
 # sys.prefix/share/kalite/content
 # sys.prefix/share/kalite/locales
-
+#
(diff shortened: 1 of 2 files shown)


MCGallaspy:
For posterity, the problem was that after installing using a .deb on raspberry pi, the running user didn't have sufficient permissions to write to the affected database.

benjaoming:
Looks good! Testing it is a matter of installing it to a real virtualenv rather than using pip editable mode (`-e`)

    workon YOUR_VIRTENV
    python setup.py sdist
    pip install dist/ka-lite-0.15.0.tar.gz
    cd somewhere/else
    kalite start --foreground

MCGallaspy:
This seems to work as expected. Recommend merge.

PR #4585, opened 2015-10-07
Avoid unnecessary dependencies for old dbbackup version

### LINKED ISSUE #4584: We are not pure python anymore
Branch: 0.15.x & develop

While trying to get KA Lite running on the Sandisk, we noticed that we tried compiling PyCrypto. Unlike other libraries PyCrypto does not have a pure python fallback, and we fail the installation.

I believe the trojan horse that introduced said library was `django-dbbackup`. Can you confirm @benjaoming?

Temp solution: fork `django-dbbackup`, find a way to remove the pycrypto dependency. Failing that, make the backup feature opt-out, and add another config for sandisk which disables this.

benjaoming:
Fixes #4584

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,5 @@ slimit==0.8.1
 six==1.9.0
 django-appconf==1.0.1
 python-dateutil==2.4.2
-django-dbbackup==2.0.4
+django-dbbackup==2.3.1
 sqlitedict==1.3.0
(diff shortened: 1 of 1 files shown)


aronasorman:
@benjaoming `django-dbbackup` is failing to install. @ralphiee22 encountered the same issue.

benjaoming:
Yeah seems the new maintainer team member has uploaded a broken release, I've added a PR for it, trying an earlier version.

aronasorman:
@ralphiee22 reported that this fixed the issue for the SanDisk.

PR #4586, opened 2015-10-07
Log whenever we fail to load the SQLite content cache.

aronasorman:


python-packages/fle_utils/general.py
```diff
@@ -270,10 +270,13 @@ def softload_json(json_filepath, default={}, raises=False, logger=None, errmsg="
 
 
 def softload_sqlite_cache(cache_filepath, raises=False):
+    from django.conf import settings
```
aronasorman: I put the django import here as this file is supposed to be unaware of django.

aronasorman:
Is this OK to merge @benjaoming?

PR #4587, opened 2015-10-07
Fix language pack availability issue in 0.14.1

MCGallaspy:
Language packs not shown as available, due to version comparison on distributed server. See my commit message/code.

See @aronasorman's note for context: https://github.com/fle-internal/ka-lite-central/commit/7a48ceca9eff#diff-b4c6913ca485fa33b3dfb6040efc3dddR680

We should cherry-pick this to `0.15.x` and `develop` as well.

Commit message:
```text
Change version comparison logic

The central server sends a short version (e.g. 0.14), which the distributed server
then compares to its long version (e.g. 0.14.1). The problem is that 0.14 compares
as less than 0.14.1, so language packs are shown as unavailable. Instead,
when comparing two mismatched-length versions, only consider the shortest version
available -- thus 0.14 and 0.14.1 would compare equal. Then the languagepack versions
will be compared, which is the actual intent.
```

@jamalex or @rtibbles any caveats here I should be aware of?

Due immediately.

### DIFF
--- a/kalite/updates/static/js/updates/update_languages.js
+++ b/kalite/updates/static/js/updates/update_languages.js
@@ -3,12 +3,18 @@ var installed_languages = [];
 var downloading = false;
 
 function version_comparison(v1, v2) {
-    // compare two version strings and return 1 if the first is higher than the second,
-    // -1 if the first is lower than the second, and 0 if they are equal
+    /*
+    compare two version strings and return 1 if the first is higher than the second,
+    -1 if the first is lower than the second, and 0 if they are equal.
+
+    :params v1, v2: Version strings expected format is either "N.N.N" or "N.N", where N is a positive integer.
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Um, I just had a total brain-fart with that commit message. Incoming amendment.

aronasorman:
Testing.

aronasorman:
Works.

benjaoming:
Just wanna clue in @EdDixon because this is the issue that caused the blue button to be gone. Now it's back in 0.14.2 which the above PR tracks.

screenshot from 2015-10-08 15 14 12

MCGallaspy:
Need to cherry-pick to 0.15 and develop as well.

rtibbles:
Is in 0.15. Should get merged to develop from there.

PR #4593, opened 2015-10-08
0.14.x into 0.15.x

rtibbles:


MCGallaspy:
The # of commits is small enough, I think it would be less hassle to list those we wish to cherry-pick to develop (incoming list):

rtibbles:
Sounds a better approach @MCGallaspy. Feel free to close this PR when you open yours with the cherry picks.

PR #4595, opened 2015-10-08
0.15.x cherry-picks from 0.14.x

MCGallaspy:
Supersedes https://github.com/learningequality/ka-lite/pull/4593. Most of the commits were already applied to 0.15.x.


rtibbles:
Looks fine, will merge when tests pass.

PR #4565, opened 2015-10-02
Include docs in sdist and add Makefile for common build, dev, test

benjaoming:
Taking over from #4526

# Change affecting 0.15 release

Docs are included in the sdist and built automatically by running `make sdist`.

PR for #4348 (partly) and #4375

# Change affecting build process

The PyPi and .deb for 0.15 will be released by using a new Makefile. Yay!

#4459 

Makefile does this:

```
➜  ka-lite git:(makefile-dist-docs) make help
clean - remove all build, test, coverage and Python artifacts
clean-build - remove build artifacts
clean-pyc - remove Python file artifacts
clean-test - remove test and coverage artifacts
lint - check style with pep8
test - run tests the default Python
test-bdd - run BDD tests only
test-nobdd - run non-BDD tests only
assets - build all JS/CSS assets
coverage - check code coverage quickly with the default Python
docs - generate Sphinx HTML documentation, including API docs
release - package and upload a release
dist - package locally
install - install the package to the active Python's site-packages

```

PR #4609, opened 2015-10-12
Change wording of warning

MCGallaspy:
Since it's been punted.


benjaoming:
Oh that was already part of #4565

PR #4610, opened 2015-10-12
0.15.x into develop

benjaoming:


circle.yml
```diff
@@ -10,23 +10,26 @@ dependencies:
...
   post:
     - if [[ ! -e sc-latest-linux/bin/sc ]]; then wget https://saucelabs.com/downloads/sc-latest-linux.tar.gz && tar -xzf sc-latest-linux.tar.gz &&  mv sc-*-linux sc-latest-linux; fi
 
 test:
   override:
-    - node build.js
-    - kalite manage init_content_items
+    - make assets
     - kalite start --traceback -v2
+    - kalite start --settings=kalite.project.settings.disk_based_cache --traceback -v2
```
MCGallaspy: I think the error is that there should only be one `kalite start` command here.

rtibbles: Indeed - the line above is now preferred, as the disk_based_cache as was is removed in develop.

benjaoming: It wasn't, but you mean to say it should?

PR #4618, opened 2015-10-13
automatically generate a manpage and put it in docs/

benjaoming:
I would like this included in 0.15, because then the sdist will come with a man page that the debian dist can include.

It's automatically generated from the docutils output using cli2man.

This does not affect the 0.15 code base at all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ static-updates
 # Documentation
 docs/_build
 docs/images/*
+docs/kalite.1.gz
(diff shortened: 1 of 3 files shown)


benjaoming:
Some sauce timeout in the test...

MCGallaspy:
This is it, don't try to slip any more PRs in. 0.15.x is now closed! Please respect that so that we can wrap up the build process.

PR #4621, opened 2015-10-13
Master into 0.15.x

benjaoming:
My suggestions for resolving merge conflicts.

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -466,9 +466,6 @@ SESSION_ENGINE = getattr(
 SESSION_COOKIE_AGE = 60 * 30     # 30 minutes
 SESSION_SAVE_EVERY_REQUEST = True
 
-# Expire session cookies whenever we close the browser.
-SESSION_EXPIRE_AT_BROWSER_CLOSE = True
-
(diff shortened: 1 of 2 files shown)


kalite/settings/base.py
```diff
@@ -466,6 +466,9 @@
 SESSION_COOKIE_AGE = 60 * 30     # 30 minutes
 SESSION_SAVE_EVERY_REQUEST = True
 
+# Expire session cookies whenever we close the browser.
```
rtibbles: I believe this is intentionally removed, as it does not work as a setting (seems to fail on most browsers).

kalite/updates/management/commands/languagepackdownload.py
```diff
@@ -101,7 +101,8 @@ def handle(self, *args, **options):
             self.next_stage(_("Invalidate caches"))
             caching.invalidate_all_caches()
 
-            self.complete(_("Finished processing language pack %(lang_name)s.") % {"lang_name": get_language_name(lang_code)})
+            self.complete((_("Finished processing language pack %(lang_name)s.") % {"lang_name": get_language_name(lang_code)}) +
```
rtibbles: This should be excluded also.

PR #4623, opened 2015-10-13
add missing entry for generated manpages

benjaoming:


MCGallaspy:
Context? What if the man pages don't exist because e.g. I ran the makefile on windows?

benjaoming:
We *could* also host the manpage in the repo so it doesn't need to be generated during `make dist` -- that seems more reasonable actually. Especially because the auto-generated manpage could do with a bit of manual adjustments anyways.

MCGallaspy:
Sounds fine to me.

benjaoming:
Cool, I'll do this for 0.16. For now, it means adding the .gz file and then we should also have a Makefile entry for building the manpages, perhaps together with `make docs` target, perhaps not.. will figure that out.

benjaoming:
Retarget for 0.16

PR #4630, opened 2015-10-16
Database backed content cleanup

### LINKED ISSUE #4507: Integrate video download status tracking with database backed content caches
Now that we have a database for this kind of thing - let's use it!

Deprecate old content download tracking, and use new database instead.

@jamalex Might need some input on design here - it might be helpful to annotate every node with 'download size', 'on disk size' and 'status' (including Topic nodes) in order to be able to do efficient loading of the data (i.e. we can just query one level of the topic tree at a time, rather than the whole thing).

### LINKED ISSUE #4508: Migrate assessment item database to new system
This is desirable because we will want to distribute a new assessment item database for each language, rather than having the translation happen on the fly using the Django mo files (which when loaded into memory increase memory usage).

rtibbles:
@aronasorman Opening this PR now so that you and @gwong89 can review the changes I am making and we can avoid collisions.

Fixes #4508, fixes #4507.

Also starts dismembering pieces on the distributed server that will eventually end up in content pack building.


kalite/topic_tools/annotate.py
```diff
@@ -50,97 +50,69 @@ def update_content_availability(content_list, language="en"):
...
-                } for lc in subtitle_lang_codes]
-
-            # Sort all subtitle URLs by language code
-            update["subtitle_urls"] = sorted(subtitle_urls, key=lambda x: x.get("code", ""))
-
-        with i18n.translate_block(language):
-            update["title"] = _(content.get("title"))
-            update["description"] = _(content.get("description")) if content.get("description") else ""
-
+            if (filename) in contents_folder or language in subtitle_lang_codes:
```
MCGallaspy: Why the redundant parentheses?

MCGallaspy:
I have no beef with this PR, but I do have a few questions/nice-to-haves:

* I've been a little out of the loop, so I wonder if it's true that `topic_tools.content_models` module now plays the role of being the interface to content for the rest of the app? If so, I would like to request docstrings that explain the parameters of functions a bit, like this:

```python
def update_parents(db=None, parent_mapping=None, channel="khan", language="en", **kwargs):
    """
    What this function does. Does it have any side effects, or is it pure (in the functional programming sense)?
    
    :param db: What is the significance/expected value for this argument? How is it used?
    :param parent_mapping: Same.
    :param channel: Same.
    :param language: Same.
    :param kwargs: Same.
    :return: What if anything does it return?
    """
```

* On that same point, using `@set_database` decorator for functions that take `db` kwarg seems a little misleading -- those functions don't really take a `db` kwarg at all. Consider using a context manager?

```python
import contextlib

@contextlib.contextmanager
def pick_db():
  # Do yo thang
  yield db
  # Finish it

def get_topic_node(content_id=None, **kwargs):
    with pick_db() as db: # "as" grabs whatever was yielded above
        # blah blah blah
```

* My main questions here: how does this affect setup? You indicated that `init_content_items` was maybe a provisional command -- so when we create distributions, how will we supply a content db that is aware of the assessment items we'll be bundling in? Will it be created on the fly when the server starts? If so it should be fast, so can we benchmark that process on our low-powered devices, like the sandisk device and the RPi?

If it's too slow, can we create a file and distribute it (kinda like what I'm suggesting in #4656)? Note that it may not be appropriate to distribute content-related items in the *source* distribution -- ideally we'd have some fast-ish process for unpacking all the relevant files that can be done at install time.

I like creating parity between the content types by handling videos & assessment items together, but it raises another issue at setup time. If a user installs "fresh" then all videos can be marked as unavailable initially. But if a user upgrades, then how will their existing videos be marked as available? Can we force availability stamping once at upgrade time (and if so how will we do it)?

MCGallaspy:
Is this D-E-D dead?

rtibbles:
It's just resting...

rtibbles:
Under active development, should have it in good form by EOD.

rtibbles:
Yes, the content_models.py now acts as the (implementation agnostic) interface to content metadata.

The point of using the decorator, rather than a context manager is so that the functions sometimes need direct access to the db object.

Also, the way you suggest doing the context manager here would not actually work for this use case, as the point is to be able to dynamically specify the particular database file on each function call, not to set it globally at module execution.

The other way to do this would be to use a context_manager and not explicitly pass in 'db', however, there are several functions that require explicit reference to the db kwarg, and hence it is passed into the functions.

```init_content_items``` will be removed when the central server content packs are ready. Content dbs will be bundled into content packs.

All awareness of which assessment items we are bundling will also be done within the content packs.

In the current implementation, annotation happens during the setup process, so immediately after upgrade everything is updated.

**Comment by MCGallaspy** [comment]
> Also, the way you suggest doing the context manager here would not actually work for this use case, as the point is to be able to dynamically specify the particular database file on each function call, not to set it globally at module execution.

Near as I can tell, that's not the case with `@contextmanager` -- the the code is executed every time a new context is defined, not when the module is loaded, and yields different mutable objects:

```python
from contextlib import contextmanager

@contextmanager
def context():
    is_this_global = ["a", "b", "c"]
    yield is_this_global

with context() as obj1:
    with context() as obj2:
        print "obj1 is obj2? %s " % (obj1 is obj2)
```
Output:
`obj1 is obj2? False`

That said, do what's expedient for you, my main purpose was to discover how this will affect the build process. If as you say annotation occurs during setup (and has acceptable performance on slow hardware) and missing content dbs are handled gracefully (so that there's no obligation in the build process to create content dbs) then all is well.

gwong89:
its not dead I'm going through the commits and marking some potential places that could use improvement

rtibbles:
This is now mergeable and ready to review. Doing self review now.

kalite/contentload/management/commands/init_content_items.py
```diff
@@ -56,14 +56,11 @@ def recurse_nodes(node, parent=""):
 
         node.pop("child_data", None)
 
-        child_availability = []
-
-        child_ids = [child.get("id") for child in node.get("children", [])]
+        total_files = 0
```
rtibbles: Adds properties up the topic tree to show how many files are available for download.

Note: This is going to be removed from distributed, and in the content packs, initial file sizes will also be added.

kalite/contentload/management/commands/init_content_items.py
```diff
@@ -133,9 +135,9 @@ def handle(self, *args, **kwargs):
         channel = kwargs["channel"]
         # temporarily swap out the database path for the desired target
         database_path = kwargs["database_path"] or CONTENT_DATABASE_PATH.format(channel=channel, language=language)
-        bulk_create = not kwargs["no_bulk_create"]
+        bulk_create = kwargs["bulk_create"]
 
-        if bulk_create and os.path.isfile(database_path):
```
rtibbles: Due to issues with the bulk creation, make this the default.

kalite/contentload/management/commands/unpack_assessment_zip.py
```diff
@@ -68,8 +68,8 @@ def handle(self, *args, **kwargs):
 def should_upgrade_assessment_items():
     # if assessmentitems.version doesn't exist, then we assume
     # that they haven't got assessment items EVER
-    if not os.path.exists(settings.KHAN_ASSESSMENT_ITEM_DATABASE_PATH) or not os.path.exists(settings.KHAN_ASSESSMENT_ITEM_VERSION_PATH):
```
rtibbles: No longer have a separate assessment items DB.

kalite/contentload/settings.py
```diff
@@ -21,7 +21,7 @@
 
 
 # Are assessment items distributed in the system-wide data directory?
-if settings.ASSESSMENT_ITEMS_SYSTEM_WIDE:
```
rtibbles: This setting relies on the assessment items DB path.

kalite/distributed/management/commands/initialize_kalite.py
```diff
@@ -35,18 +35,7 @@ def setup_server_if_needed(self):
             # Double check the setup process worked ok.
             assert Settings.get("database_version") == VERSION, "There was an error configuring the server. Please report the output of this command to Learning Equality."
 
-    def reinitialize_server(self):
-        """Reset the server state."""
-
-        # Next, call videoscan.
-        logging.info("Running videoscan.")
-        call_command("videoscan")
```
rtibbles: RIP videoscan.

kalite/distributed/management/commands/setup.py
```diff
@@ -393,8 +393,6 @@ def handle(self, *args, **options):
 
         else:
 
-            call_command("syncdb", interactive=False, verbosity=options.get("verbosity"), database="assessment_items")
```
rtibbles: As before, assessment item db is gooooone.

kalite/distributed/management/commands/setup.py
```diff
@@ -485,7 +480,7 @@ def handle(self, *args, **options):
             else:
                 start_script_path = kalite_executable
 
-            # Run videoscan, on the distributed server.
+            # Run annotate_content_items, on the distributed server.
```
rtibbles: @MCGallaspy see here for where annotation happens during setup.

kalite/i18n/base.py
```diff
@@ -0,0 +1,361 @@
+import errno
```
rtibbles: This refactor was required due to circular import issues.

kalite/project/settings/raspberry_pi.py
```diff
@@ -17,6 +17,4 @@
 
 ENABLE_CLOCK_SET = True
 
-DO_NOT_RELOAD_CONTENT_CACHE_AT_STARTUP = True
```
rtibbles: No longer required.

kalite/testing/testrunner.py
```diff
@@ -21,7 +21,7 @@
 from optparse import make_option
 
 from kalite.testing.base import DjangoBehaveTestCase
-from kalite.topic_tools.content_models import database_exists
+from kalite.topic_tools.base import database_exists
```
rtibbles: Required due to circular import issues.

kalite/topic_tools/annotate.py
```diff
@@ -50,97 +50,71 @@ def update_content_availability(content_list, language="en"):
...
-                    if exercise_template in exercise_templates:
-                        update["available"] = True
-                        update["template"] = exercise_template
-                else:
-                    exercise_template = os.path.join(language, exercise_file)
-                    if exercise_template in exercise_templates:
-                        update["available"] = True
-                        update["template"] = exercise_template
+            # Databases have been pre-filtered to only contain existing exercises
+            if database_exists(language=language, channel=channel):
```
rtibbles: We are going to assume that if the database is present then an exercise is available.

Come to think of it - if we are doing this annotation, then the database is present... so we can make this just be True regardless.

rtibbles: Template will be prestamped.

kalite/topic_tools/annotate.py
```diff
@@ -50,97 +50,71 @@ def update_content_availability(content_list, language="en"):
...
-                } for lc in subtitle_lang_codes]
-
-            # Sort all subtitle URLs by language code
-            update["subtitle_urls"] = sorted(subtitle_urls, key=lambda x: x.get("code", ""))
-
-        with i18n.translate_block(language):
-            update["title"] = _(content.get("title"))
-            update["description"] = _(content.get("description")) if content.get("description") else ""
-
+            if filename in contents_folder or language in subtitle_lang_codes:
```
rtibbles: Fallback to using a subtitled video with a different language video if available.

kalite/updates/__init__.py
```diff
@@ -1,46 +0,0 @@
-"""
```
rtibbles: This refactor was required due to circular import issues.

kalite/updates/api_views.py
```diff
@@ -17,31 +17,20 @@
...
 from fle_utils.internet.classes import JsonResponse, JsonResponseMessageError, JsonResponseMessageSuccess
 from fle_utils.orderedset import OrderedSet
-from kalite.i18n import get_youtube_id, get_video_language, lcode_to_ietf, delete_language, get_language_name
+from kalite.i18n.base import lcode_to_ietf, delete_language, get_language_name
 from kalite.shared.decorators.auth import require_admin
-from kalite.topic_tools.settings import TOPICS_FILEPATHS, CHANNEL
-
-
-
-def divide_videos_by_language(youtube_ids):
```
rtibbles: No longer required - we now only allow adding to download queue per language (this was actually de facto the only way to add stuff due to the front end interface, but our API supported a more liberal way).

kalite/updates/api_views.py
```diff
@@ -135,33 +124,14 @@ def start_video_download(request):
...
-    # One query per chunk
-    for chunk in break_into_chunks(youtube_ids):
-        video_files_needing_model_update = VideoFile.objects.filter(download_in_progress=False, youtube_id__in=chunk).exclude(percent_complete=100)
-        video_files_needing_model_update.update(percent_complete=0, cancel_download=False, flagged_for_download=True)
-
-    force_job("videodownload", _("Download Videos"), locale=request.language)
+    youtube_ids = get_download_youtube_ids(paths)
 
-    return JsonResponseMessageSuccess(_("Launched video download process successfully."))
+    queue = VideoQueue()
```
rtibbles: The VideoQueue is the interface to the video download queueing.

kalite/updates/api_views.py
```diff
@@ -235,99 +203,14 @@ def delete_language_pack(request):
     return JsonResponse({"success": _("Successfully deleted language pack for %(lang_name)s.") % {"lang_name": get_language_name(lang_code)}})
 
 
-def annotate_topic_tree(node, level=0, statusdict=None, remote_sizes=None, lang_code=None):
```
rtibbles: We don't need this any more (this was happening on the fly on every API request) - the data is stamped on content in the database.

kalite/updates/api_views.py
```diff
@@ -235,99 +203,14 @@ def delete_language_pack(request):
...
-def get_annotated_topic_tree(request, lang_code=None):
-    call_command("videoscan")  # Could potentially be very slow, blocking request... but at least it's via an API request!
+def get_update_topic_tree(request, lang_code=None):
 
+    parent = request.GET.get("parent")
     lang_code = lang_code or request.language      # Get annotations for the current language.
-    statusdict = dict(VideoFile.objects.values_list("youtube_id", "percent_complete"))
 
-    return JsonResponse(annotate_topic_tree(softload_json(TOPICS_FILEPATHS.get(CHANNEL), logger=logging.debug, raises=False), statusdict=statusdict, lang_code=lang_code))
+    return JsonResponse(get_topic_update_nodes(parent=parent, language=lang_code))
```
rtibbles: This returns layer by layer, so we can lazily load the topic tree into the front end.

kalite/updates/download_track.py
```diff
@@ -0,0 +1,63 @@
+import json
+
+from django.conf import settings
+logging = settings.LOG
+
+from .settings import VIDEO_DOWNLOAD_QUEUE_FILE
+
+class VideoQueue:
```
rtibbles: Class to manage video downloads. It works, but implementation comments/feedback appreciated here!

kalite/updates/management/commands/compute_video_sizes.py
```diff
@@ -1,46 +0,0 @@
-"""
```
rtibbles: We don't need this any more.

kalite/updates/management/commands/videodownload.py
```diff
@@ -11,12 +11,11 @@
 from django.utils.translation import ugettext as _
```
rtibbles: Edits in here are mostly to replace the use of the VideoFile model to track download progress.

kalite/updates/management/commands/videoscan.py
```diff
@@ -1,108 +0,0 @@
-"""
```
rtibbles: RIP. All this functionality is done by annotation now.

rtibbles: We might want to have a 'import content' button on the download page? Or we can just require people to run setup.

kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -103,101 +102,107 @@ var tree;
...
+        aria: true, // Enable WAI-ARIA support.
+        checkbox: true, // Show checkboxes.
+        debugLevel: 0, // 0:quiet, 1:normal, 2:debug
+        selectMode: 3,
+        source: {
+            url: window.Urls.get_update_topic_tree(),
+            data: { parent: "root"},
+            cache: false
+        },
+        postProcess: function(event, data) {
```
rtibbles: Moving some of the rendering coding to front end here.

kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -103,101 +102,107 @@ var tree;
...
+                    node["extraClasses"] = "complete";
+                }
+                if (node.kind == "Topic") {
+                    node["lazy"] = true;
+                    node["folder"] = true;
+                }
+                node["key"] = node["id"];
+            });
+        },
+        lazyLoad: function(event, data){
```
rtibbles: Use lazy loading for children.

manage.py
```diff
@@ -17,24 +17,10 @@
 
 if __name__ == "__main__":
     import warnings
-    
+
     # We are overriding a few packages (like Django) from the system path.
     #   Suppress those warnings
     warnings.filterwarnings('ignore', message=r'Module .*? is being added to sys\.path', append=True)
-
```
rtibbles: kaserve is gone...

MCGallaspy:
Now looking.

kalite/i18n/base.py
MCGallaspy: Was anything changed, or was this just a move?

kalite/topic_tools/annotate.py
```diff
@@ -50,97 +50,71 @@ def update_content_availability(content_list, language="en"):
                     else:
                         subtitle_langs[filename] = [lc]
 
-    # English-language exercises live in application space, translations in user space
```
MCGallaspy: Everything lives in user space now, right? If the content databases don't exist, they'll be created in USER_DATA_ROOT somewhere?

kalite/topic_tools/annotate.py
```diff
@@ -50,97 +50,71 @@ def update_content_availability(content_list, language="en"):
...
-                        }
-                    elif django_settings.BACKUP_VIDEO_SOURCE:
-                        update["available"] = True
-                        update["content_urls"] = {
-                            "stream": django_settings.BACKUP_VIDEO_SOURCE.format(youtube_id=dubbed_id, video_format=format),
-                            "stream_type": "{kind}/{format}".format(kind=content.get("kind").lower(), format=format),
-                            "thumbnail": django_settings.BACKUP_VIDEO_SOURCE.format(youtube_id=dubbed_id, video_format="png"),
-                        }
+            format = content.get("format", "")
+            filename = file_id + "." + format
```
MCGallaspy: So if there's no `format` the file name will just end in with `.`? Intentional?

kalite/topic_tools/base.py
```diff
@@ -39,27 +38,20 @@
...
+def database_exists(channel="khan", language="en", database_path=None):
+    path = database_path or settings.CONTENT_DATABASE_PATH.format(channel=channel, language=language)
+
+    return os.path.exists(path)
+
+def available_content_databases():
+    """
+    Generator to return the channel and language for every content database that exists in the system.
+    """
+    pattern = re.compile("content_([^_]+)_([^_]).sqlite")
```
MCGallaspy: Should it be `content_([^_]+)_([^_]+).sqlite`? [What's there doesn't match `content_en_khan.sqlite`](http://pythex.org/?regex=content_(%5B%5E_%5D%2B)_(%5B%5E_%5D).sqlite&test_string=content_en_khan.sqlite&ignorecase=0&multiline=0&dotall=0&verbose=0)

It's not entirely obvious, what order is the channel & language supposed to be returned in? `("khan", "en")` or `("en", "khan")`? Add to docstring or use named groups, or better yet both.

MCGallaspy:
Was hoping this would gracefully handle a missing content db, but I get the following error:
Cmd: `bin/kalite start --foreground`

```python
INFO:kalite:Annotating content models for language: en, channel: khan in databas
e located at D:\ka-lite\content_databases\content_khan_en.sqlite
  File "D:\ka-lite\python-packages\django\core\management\base.py", line 224, in
 run_from_argv
  File "D:\ka-lite\python-packages\django\core\management\base.py", line 263, in
 execute
  File "D:\ka-lite\kalite\distributed\management\commands\initialize_kalite.py",
 line 41, in handle
    self.setup_server_if_needed()
  File "D:\ka-lite\kalite\distributed\management\commands\initialize_kalite.py",
 line 34, in setup_server_if_needed
  File "D:\ka-lite\python-packages\django\core\management\__init__.py", line 161
, in call_command
    return klass.execute(*args, **defaults)
  File "D:\ka-lite\python-packages\django\core\management\base.py", line 263, in
 execute
  File "D:\ka-lite\kalite\distributed\management\commands\setup.py", line 486, i
n handle
    call_command("annotate_content_items")
  File "D:\ka-lite\python-packages\django\core\management\__init__.py", line 161
, in call_command
    return klass.execute(*args, **defaults)
  File "D:\ka-lite\python-packages\django\core\management\base.py", line 263, in
 execute
  File "D:\ka-lite\kalite\contentload\management\commands\annotate_content_items
.py", line 50, in handle
    annotate_content_models(database_path=database_path, channel=channel, langua
ge=language)
  File "D:\ka-lite\kalite\topic_tools\content_models.py", line 116, in wrapper
    output = function(*args, **kwargs)
  File "D:\ka-lite\kalite\topic_tools\content_models.py", line 542, in annotate_
content_models
    for path, update in content_models:
  File "D:\ka-lite\kalite\topic_tools\content_models.py", line 466, in iterator_
content_items
    items = Item.select().dicts().iterator()
  File "C:\Users\Mike\Envs\develop-0-16\lib\site-packages\peewee.py", line 2797,
 in iterator
    return iter(self.execute().iterator())
  File "C:\Users\Mike\Envs\develop-0-16\lib\site-packages\peewee.py", line 2787,
 in execute
    self._qr = ResultWrapper(model_class, self._execute(), query_meta)
  File "C:\Users\Mike\Envs\develop-0-16\lib\site-packages\peewee.py", line 2470,
 in _execute
    return self.database.execute_sql(sql, params, self.require_commit)
  File "C:\Users\Mike\Envs\develop-0-16\lib\site-packages\peewee.py", line 3199,
 in execute_sql
    self.commit()
  File "C:\Users\Mike\Envs\develop-0-16\lib\site-packages\peewee.py", line 3048,
 in __exit__
    reraise(new_type, new_type(*exc_args), traceback)
  File "C:\Users\Mike\Envs\develop-0-16\lib\site-packages\peewee.py", line 3191,
 in execute_sql
    cursor.execute(sql, params or ())
OperationalError: no such table: item
```

MCGallaspy:
I see that in addition to refactoring the content stuff, you've also made some considerable changes to the updates app -- can you summarize those changes? Or replay those changes onto a separate branch and open a new PR?

MCGallaspy:
In reference to https://github.com/learningequality/ka-lite/pull/4630#issuecomment-155562193, running the cmd twice successfully starts the server.

kalite/updates/download_track.py
MCGallaspy: inherit from object to get a new-style class: `class VideoQueue(object):`

MCGallaspy: See (docs on new-style classes)[https://docs.python.org/2/reference/datamodel.html#newstyle]

MCGallaspy: I had a similar idea, BTW, that @tk21 is now working on -- [see doc on assessment item download during setup](https://docs.google.com/document/d/1ALHs6Ry3vxbd5TS--HkL9fqEWcR7XNWzZACs7n_ALE0/edit#heading=h.low45iy0i5at). But now I'm also wondering how the plans for that are affected by using content DBs? Do we have a concept of downloading assessment items at runtime, e.g. during upgrade, or is that replaced by downloading a content db?

kalite/updates/download_track.py
```diff
@@ -0,0 +1,63 @@
...
+        and optionally, the language of the video.
+        """
+        files = [{"youtube_id": filename, "language": language} for filename in reversed(files)]
+        files.extend(self.queue)
+        self.queue = files
+        self.save()
+
+    def save(self):
+        """Save the current queue to disk"""
+        with open(VIDEO_DOWNLOAD_QUEUE_FILE, "w") as f:
```
MCGallaspy: What happens if several processes or threads try to access this file simultaneously?

MCGallaspy:
Given the size, might be good to get another reviewer here, too.

MCGallaspy:
In general, I think it would be really good to document what the expected behavior is, too. In this case it would be retroactive, but for future efforts it could be a big aid to review!

MCGallaspy:
Note on the updates app: in video download page, I can seemingly only expand or contract menu items with a double click, instead of single click. This makes a weird experience, as it also selects text.

MCGallaspy:
Another note: I used `init_content_items` to take the updates app for a test drive -- the estimated video size is always "0.0 MB", for any selection.

MCGallaspy:
Download progress bar is much improved, but displays video id instead of video title, which is a little funky:

ss2

MCGallaspy:
I'll finish reviewing the updates app later, though I'd like to note again that it might be helpful to open a separate PR for those changes, as well as write some documentation describing the app at a high level. The content stuff looks fine to me.

kalite/i18n/base.py
rtibbles: The only thing that was changed was unused functions were deleted, and things that relied on old content backend stuff was modified - don't think any new code is in this file though.

kalite/updates/download_track.py
rtibbles: @MCGallaspy the only change would be that the database would no longer be in the zip file. All the image resources would still be downloaded in the same way.

kalite/updates/management/commands/videoscan.py
rtibbles: @jamalex @jtamiace input appreciated!

kalite/topic_tools/annotate.py
rtibbles: Yes - inside 'content_database' folder.

kalite/topic_tools/annotate.py
rtibbles: Not hugely intentional - if something doesn't have a format, we're probably not going to find the file.

kalite/topic_tools/base.py
rtibbles: Woops, missed the second +.

Will do both.

kalite/updates/download_track.py
rtibbles: Bad stuff(TM). Should probably add a try except for IOError

kalite/updates/management/commands/videoscan.py
jamalex: > "import content" button
Is this a "rescan for content already in the content folder" or "load content from elsewhere into the content folder"? The latter is slated for Kolibri, the former would probably be good to have in KA Lite, if we're taking it out of startup.

rtibbles: Former. Will add.

kalite/topic_tools/annotate.py
rtibbles: On further thought, make this contingent on the existence of assessment_items.version - we are going to assume that if you have downloaded assessment items zip, then all exercises that can be available are available (ones not available in that language will not be in the content db at all).

rtibbles:
Assuming test pass (odd that the https://github.com/rtibbles/ka-lite/commit/0c7b1bb804c8aaf79409a1a9e5eef6a7e6649eb8 failed) this should address most of the main issues raised above.

MCGallaspy:
On the updates app:
* Deleting files has unexpected behavior:
  * Deleted file re-appears as downloaded after page refresh. Initiating a scan doesn't help.
  * Deleting a video and then selecting it again gives me the option to delete it -- not download it.
* Scan can only be initiated once -- page needs to be refreshed to scan again.
* Starting a download, cancelling it, and choosing a different set of videos to download seems to launch a download process for the original selection.
* Please write user documentation

rtibbles:
Addressed above issues.

PR #4637, opened 2015-10-19
fixes #4627 and also solves assessment unzipping issue with Windows

### LINKED ISSUE #4627: Menu UI drop down should go over existing content
While testing the 0.15 final release version I noticed that the menu partially covered by existing content as in the attached screen shot. I think the intention is that these would go over any existing content.

jtamiace:
navbar dropdown overlap fix


rtibbles:
We should merge this, but just a note that a lot of this code will be getting blown away by the refactor that @aronasorman and I are undertaking.

PR #4649, opened 2015-10-29
Update Makefile

### LINKED ISSUE #4648: cli2man fails on Windows
In particular running `make dist` (using msys) fails when trying to build the docs, because the `cli2man` script fails. Unclear which label to put on this. @benjaoming

MCGallaspy:
Remove cli2man for now, doesn't work on Windows. Fixes #4648.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,6 @@ docs:
 	# sphinx-apidoc -o docs/ ka-lite-gtk
 	$(MAKE) -C docs clean
 	$(MAKE) -C docs html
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Confirmed that it works great without this one weird line.

benjaoming:
Since cli2man is still a dev dependency, it'd be a waste to completely remove it :)

Sorry that it didn't work on Win, but fantastic if everything else is working!!

MCGallaspy:
Yeah, works great since all the various executables are available with e.g. MinGW. :+1:

Issue #4653, opened 2015-10-29
sdist needs to be updated to include content databases


MCGallaspy:
While tackling an issue with the Windows installer build process, I encountered the following issue. Currently on `develop` when I run `make dist` and install the output with `pip install <outfile>`, when I try to start the server I get the following error:

```sh
INFO:kalite:Annotating content models for language: en, channel: khan in databas
e located at C:\Users\Mike\.kalite\content\content_khan_en.sqlite
Traceback (most recent call last):
  File "C:\Python27\share\kalite\python-packages\django\core\management\base.py"
, line 224, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "C:\Python27\share\kalite\python-packages\django\core\management\base.py"
, line 263, in execute
    output = self.handle(*args, **options)
  File "C:\Python27\lib\site-packages\kalite\distributed\management\commands\ini
tialize_kalite.py", line 48, in handle
    self.setup_server_if_needed()
  File "C:\Python27\lib\site-packages\kalite\distributed\management\commands\ini
tialize_kalite.py", line 34, in setup_server_if_needed
    call_command("setup", interactive=False)
  File "C:\Python27\share\kalite\python-packages\django\core\management\__init__
.py", line 161, in call_command
    return klass.execute(*args, **defaults)
  File "C:\Python27\share\kalite\python-packages\django\core\management\base.py"
, line 263, in execute
    output = self.handle(*args, **options)
  File "C:\Python27\lib\site-packages\kalite\distributed\management\commands\set
up.py", line 491, in handle
    call_command("annotate_content_items")
  File "C:\Python27\share\kalite\python-packages\django\core\management\__init__
.py", line 161, in call_command
    return klass.execute(*args, **defaults)
  File "C:\Python27\share\kalite\python-packages\django\core\management\base.py"
, line 263, in execute
    output = self.handle(*args, **options)
  File "C:\Python27\lib\site-packages\kalite\contentload\management\commands\ann
otate_content_items.py", line 51, in handle
    annotate_content_models(database_path=database_path, channel=channel, langua
ge=language)
  File "C:\Python27\lib\site-packages\kalite\topic_tools\content_models.py", lin
e 103, in wrapper
    output = function(*args, **kwargs)
  File "C:\Python27\lib\site-packages\kalite\topic_tools\content_models.py", lin
e 398, in annotate_content_models
    content_models = get_content_items(ids=ids, channel=channel, language=langua
ge)
[...]

PR #4656, opened 2015-10-30
Add empty db to sdist

MCGallaspy:
Not touching content/assessment items right now. Fixes part of #4653. New behavior:

* Create empty db in `assets` make target, and include in sdist.
* If the user has no database at `settings.DEFAULT_DATABASE_PATH` and `settings.DB_TEMPLATE_FILE` is truthy, then the latter file will be copied to the former by the setup mgmt command.

The reason this works -- when the sdist is built `IS_SOURCE` is `True` (or it should be), so we create a db at `kalite/database/data.sqlite` (which is the `DEFAULT_DATABASE_PATH`). We simply preserve this file in the sdist, then when IS_SOURCE is `False` (and `DEFAULT_DATABASE_PATH` has a different value) we set `DB_TEMPLATE_FILE` to point to that file in the source tree.

### DIFF
--- a/kalite/__init__.py
+++ b/kalite/__init__.py
@@ -6,13 +6,8 @@ from version import *
 
 __version__ = VERSION
 
-
-# ROOT_DATA_PATH should point to the directory where the source files live, including the "docs" dir
-# TODO-BLOCKER(MCGallaspy): Use setuptools in the windows installer to avoid this nonsense.
-ROOT_DATA_PATH = os.environ.get(
-    "KALITE_ROOT_DATA_PATH",
-    os.path.join(sys.prefix, 'share', 'kalite')
-)
+# ROOT_DATA_PATH is *not* where the source files live. It's a place where non-user-writable data files may be written.
+ROOT_DATA_PATH = os.path.join(sys.prefix, 'share', 'kalite')
 
 
 # TODO: Burn down this function, the name is weird, it just checks if a
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -380,9 +380,13 @@ class Command(BaseCommand):
         # call_command("clean_pyc", interactive=False, verbosity=options.get("verbosity"), path=os.path.join(settings.PROJECT_PATH, ".."))
 
         # If a db template exists, copy it instead of creating a new one, since migrations take a long time.
-        if settings.DB_TEMPLATE_FILE and not os.path.isfile(settings.DEFAULT_DATABASE_PATH):
+        database_exists = os.path.isfile(settings.DEFAULT_DATABASE_PATH)
+        if settings.DB_TEMPLATE_FILE and not database_exists:
             print("Copying database file from {0} to {1}".format(settings.DB_TEMPLATE_FILE, settings.DEFAULT_DATABASE_PATH))
             shutil.copy(settings.DB_TEMPLATE_FILE, settings.DEFAULT_DATABASE_PATH)
+        elif database_exists:
+            print("Migrating existing database.")
+            call_command("migrate", merge=True, verbosity=options.get("verbosity"))
         else:
             print("Baking a fresh database from scratch.")
             call_command("syncdb", interactive=False, verbosity=options.get("verbosity"))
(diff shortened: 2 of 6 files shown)


MCGallaspy:
@rtibbles

kalite/__init__.py
```diff
@@ -6,13 +6,8 @@
...
 __version__ = VERSION
 
-
-# ROOT_DATA_PATH should point to the directory where the source files live, including the "docs" dir
-# TODO-BLOCKER(MCGallaspy): Use setuptools in the windows installer to avoid this nonsense.
-ROOT_DATA_PATH = os.environ.get(
-    "KALITE_ROOT_DATA_PATH",
-    os.path.join(sys.prefix, 'share', 'kalite')
-)
+# ROOT_DATA_PATH is *not* where the source files live. It's a place where non-user-writable data files may be written.
```
benjaoming: Cool, so by this, we get rid of the environ KALITE_ROOT_DATA_PATH and can stop worry about yet another configurable gotcha :) :+1:

kalite/distributed/management/commands/setup.py
```diff
@@ -379,9 +379,14 @@ def handle(self, *args, **options):
...
-        call_command("syncdb", interactive=False, verbosity=options.get("verbosity"))
-        call_command("migrate", merge=True, verbosity=options.get("verbosity"))
+        # If a db template exists, copy it instead of creating a new one, since migrations take a long time.
+        if settings.DB_TEMPLATE_FILE and not os.path.isfile(settings.DEFAULT_DATABASE_PATH):
+            print("Copying database file from {0} to {1}".format(settings.DB_TEMPLATE_FILE, settings.DEFAULT_DATABASE_PATH))
+            shutil.copy(settings.DB_TEMPLATE_FILE, settings.DEFAULT_DATABASE_PATH)
+        else:
+            print("Baking a fresh database from scratch.")
+            call_command("syncdb", interactive=False, verbosity=options.get("verbosity"))
+            call_command("migrate", merge=True, verbosity=options.get("verbosity"))
```
benjaoming: I think existing databases should always have `migrate` run?

```
    elif os.path.isfile(settings.DEFAULT_DATABASE_PATH):
        try:
        except Exception as e:
            logger.error("Encountered exception while migrating {}, exception was: {}".format(settings.DEFAULT_DATABASE_PATH, e))
            logger.error("Consider deleting your database {} and running kalite manage setup again.".format(settings.DEFAULT_DATABASE_PATH))
```

Not sure if it's robust enough... but I added the exception handling in case someone kills their database creation mid-time or in some other way end up with a corrupted DB.

benjaoming: Btw above is just draft code.

setup.py
```diff
@@ -358,6 +358,7 @@ def install_distributions(distributions):
         gen_data_files('dist-packages')
     )
 
+data_files += [(os.path.join(kalite.ROOT_DATA_PATH, "database"), [os.path.join("kalite", "database", "data.sqlite")])]
```
benjaoming: This shouldn't be needed because of `include_package_data=True`.

kalite/distributed/management/commands/setup.py
MCGallaspy: Yeah, you are correct about running migrate. Never had an exception raised during migration before, feel like trying to catch one might be premature.

setup.py
MCGallaspy: Noted.

kalite/distributed/management/commands/setup.py
benjaoming: Maybe it's because the whole migration is in one big transaction so it might not be possible to leave a corrupt DB file behind... not sure. I just added it to provide a resolution to the user in case something goes wrong.

setup.py
benjaoming: I think it's kind of this pattern... this from a commented out version though, because it describes a possible pitfall.. it's from kalite/legacy/blah_settings.py

```
# This is the standard method... but doesn't work because we cannot load
# kalite.i18n while loading settings because of its __init__.py
# from pkgutil import get_data
# I18N_DATA_PATH = get_data("kalite.i18n", "data")
```

MCGallaspy:
Please notice me!

benjaoming:
Ah, you fixed the migrate thing, awesome, that was my only comment here, not seeing any other reasons to not merge this... so merging!

PR #4669, opened 2015-11-16
Updated OSX version support note

metasean:


aronasorman:
Hi @metasean -- thanks for the contribution! Unfortunately we don't make pull requests to master directly, but I'll add your change to our `0.15.x` branch.

metasean:
@aronasorman - Thank you! (I had initially tried to make the pull request to the [develop](https://github.com/learningequality/ka-lite/pull/4667) branch, but when I got an unrelated failing test, it seemed best to try pushing to the master branch.)

aronasorman:
Yeah it happened here as well unfortunately -- looks like we're having a spate of nondeterministic test failures again. Either way your docs change shouldn't make tests fail!

Here's the updated PR for reference: https://github.com/learningequality/ka-lite/pull/4671/files

PR #4671, opened 2015-11-16
add os x supported versions

aronasorman:


benjaoming:
This looks harmless, @aronasorman , it should also be merged onto 0.16.x and develop right?

PR #4689, opened 2015-11-27
unbundle tastypie

benjaoming:
#3825

PR #4702, opened 2015-11-27
remove unused package mimeparse

benjaoming:
It's used in tastypie so tastypie will manage this dependency as of #4689


benjaoming:
This is allowed to fail!

PR #4705, opened 2015-11-27
unbundle dev requirement polib

benjaoming:


aronasorman:
I recommend not removing this -- I've made some hacks in this library for the central server.

benjaoming:
> *for the central server*

I've removed polib.py, so the central server could put this file somewhere in its path?

benjaoming:
@aronasorman just a head's up in case this breaks the central server once 0.17 comes out.

PR #4714, opened 2015-11-27
unbundle django-snippets

benjaoming:


benjaoming:
this doesnt work because django-snippets on pypi is a different project

benjaoming:
Done in #4822

PR #4718, opened 2015-12-01
Create zipfile with `make dist` command

MCGallaspy:
Building the Windows installer requires a zipfile here. @benjaoming

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,8 @@ release: clean docs assets
 	ls -l dist
 
 dist: clean docs assets
-	python setup.py sdist
-	python setup.py sdist --static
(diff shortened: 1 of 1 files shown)


aronasorman:
Looks good. But the format sdist was creating before wasn't a zip file?

MCGallaspy:
It defaults to a different format depending on your OS. So on Windows it defaults to a zipfile, but on OSX it defaults to a gzipped tarball. This option ensures it creates both.

aronasorman:
I'm OK with this then. Merge once the tests pass.

benjaoming:
:+1: 

I was saving it until someone found it necessary, as you can see, it was already part of the pypi release.

Issue #4716, opened 2015-11-30
Random BDD test failure


benjaoming:
Having opened a lot of PRs (25 in a day), I found that I had to run the same tests several times because of this BDD test failure:

```
  Scenario: Content is available        # kalite/distributed/features/learn_content.feature:9
    Given I open some available content # kalite/distributed/features/steps/learn_content.py:17
    Given I open some available content # kalite/distributed/features/steps/learn_content.py:17 4.449s
    Then I should see no alert          # kalite/distributed/features/steps/learn_content.py:33
    Then I should see no alert          # kalite/distributed/features/steps/learn_content.py:33 1.732s
      Assertion Failed: NoSuchElementException not raised
```

It occurs alone. Often after dark. It's very hard to spot because it's very camouflaged. Only a vague red stripe distinguishes it.

IMO we should get it fixed before releasing because it's going to be annoying for both this release and any future fixes.

@aronasorman please re-assign at will :)

MCGallaspy:
I noticed this too.

aronasorman:
I got a similar behaviour in the 0.15.1 PR.

MCGallaspy:
Here's the relevant steps as they appear on `develop` now:

```python
@given("I open some available content")
def step_impl(context):
    update_item(path="khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/", update={"available": True})
    context.browser.get(build_url(context, reverse("learn") + "khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/"))

@then(u'I should see no alert')
def impl(context):
    assert_no_element_by_css_selector(context, ".alert")
```

My only thought is that there is a race condition with `update_item` -- that the requested page does not reflect the updated db row. Not sure how to test this hypothesis. Any insight here @rtibbles?

MCGallaspy:
Strange, here's the screencast when it passes: https://saucelabs.com/jobs/b8d864eed3af4bea9daff749f63d81ae

Note the alert indicating the content was not found.

benjaoming:
```


  @as_learner
  Scenario: User alters a star rating       # kalite/distributed/features/content_rating.feature:12
    Given I have filled out a feedback form # kalite/distributed/features/steps/content_rating.py:115
    Given I have filled out a feedback form # kalite/distributed/features/steps/content_rating.py:115 7.714s
    When I alter a star rating              # kalite/distributed/features/steps/content_rating.py:43
[...]

PR #4721, opened 2015-12-02
Refactor behave_helper functions; add & amend test cases; fix update_item function

MCGallaspy:
Starts to work on #4716. See commit message for rationale. But still, the troubled test passes even though there's clearly an alert bubble there. So something else is up.

### DIFF
--- a/kalite/distributed/features/steps/learn_content.py
+++ b/kalite/distributed/features/steps/learn_content.py
@@ -9,22 +9,26 @@ from kalite.topic_tools.content_models import update_item
 
 TIMEOUT = 3
 
-@given("I open some unavailable content")
+
+@given(u"I open some unavailable content")
 def step_impl(context):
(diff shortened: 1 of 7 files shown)


MCGallaspy:
Hm, I think `update_item` is having concurrency issues.

rtibbles:
Trying to write to the database while it is being read?

MCGallaspy:
Turns out it wasn't about concurrency at all.

MCGallaspy:
Regarding the latest test failure: seems like for some amount of time topic_tools tests were not being run at all. From a previous test run:
```
WARNING:kalite:App with label topic_tools is missing a models.py module.
```

MCGallaspy:
See https://github.com/learningequality/ka-lite/pull/4551/files#r46506825

rtibbles:
Yeah these tests need to be fixed up.

I will edit them and make a PR to your branch here.

rtibbles:
Tada: https://github.com/MCGallaspy/ka-lite/pull/25

PR #4725, opened 2015-12-03
raspberry pi install instruction

j-schwartz:
downloading file is not named ka-lite-raspberry-pi

### DIFF
--- a/docs/installguide/install_all.rst
+++ b/docs/installguide/install_all.rst
@@ -108,7 +108,7 @@ It can be installed by downloading the latest .deb on the Pi and installing it::
     # Fetch the latest .deb
     sudo wget https://learningequality.org/r/deb-pi-installer-0-15 --no-check-certificate
     # Install the .deb
-    sudo dpkg -i ka-lite-raspberry-pi*.deb
+    sudo dpkg -i deb-pi-installer*
 
 You can also add the PPA, see :ref:`ppa-installation`, and then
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Truly? I followed this link (from the docs): https://learningequality.org/r/deb-pi-installer-0-15

And the filename is indeed `ka-lite-raspberry-pi-0.15.0`. Can you provide a link that downloads a different filename?

j-schwartz:
So live on http://ka-lite.readthedocs.org/en/latest/installguide/install_all.html#linux site right now you have a command line prompt that download 0-14:

image

if I try to change it, by editing the Github, I see the 0-15 prompt:

image

Running these later commands (with 0-15 on the Github) on a Raspian Pi, I get the following:

image

image

file gets saved as:
deb-pi-installer-0-15

image


running with the downloaded file:
image

MCGallaspy:
> right now you have a command line prompt that downloads 0-14:

Fixed. We'll formulate more accurate instructions, since the behavior of wget is seemingly to name the output file based on the url -- instead we'd like it to have the filename on the server (the one we indicated in the docs).

MCGallaspy:
Fixed in #4729.

benjaoming:
@j-schwartz thanks for the thorough report! @MCGallaspy's fix is now on http://ka-lite.readthedocs.org/en/0.15.x/installguide/install_all.html#raspberry-pi

PR #4741, opened 2015-12-08
Develop into central-develop

MCGallaspy:
We need to test out the central server and see what's going on. We slacked on continually merging into central-develop.


MCGallaspy:
This is a place-holder, but we can re-open as 0.16.x into central-develop when we fork.

PR #4743, opened 2015-12-08
Prepopulated db support for all dists

benjaoming:
The previous pattern wasn't firing even though a database was in kalite/database/data.sqlite

I made the necessary changes such that it should work now. Amongst the problems was that an empty file was created as the project was loaded. This might be something that Django does.

It also sets install_clean=True in cases where there's no database already in place or it's a 0-length file. I think we might be skipping questions about passwords in these cases, even though I would prefer that setup.py didn't do stuff like that :)

### DIFF
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -249,7 +249,6 @@ class Command(BaseCommand):
                     RemovedInKALite_v016_Warning
                 )
 
-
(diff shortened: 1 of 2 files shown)


MCGallaspy:
Can you re-target 0.16.x now? Sorry for the inconvenience.

benjaoming:
Ah yes :)

PR #4754, opened 2015-12-10
Add explicit upgrade instructions

### LINKED ISSUE #4748: Add instructions for upgrading KA Lite
The documentation has instructions for installation, but not upgrading specifically. It's actually quite simple to upgrade: just run the latest installer. Let's note that in the documentation.

MCGallaspy:
Fixes #4748. Think I should also add an FAQ question?

### DIFF
--- a/docs/installguide/install_all.rst
+++ b/docs/installguide/install_all.rst
@@ -4,8 +4,13 @@ Windows
 #. Download the KA Lite `Windows <https://learningequality.org/r/windows-installer-0-15>`_ installer.
 #. Double-click the downloaded .exe file, and the wizard window will appear to guide you through the process of installing KA Lite on your server.
 
-Upgrading KA Lite over an existing installation is easy -- just run the installer and follow the prompts!
+Upgrade
+_______
+
+Upgrading KA Lite in Windows over an existing installation is easy -- just run the installer and follow the prompts!
 You don't need to uninstall your old KA Lite installation first.
+You can follow the prompts to either keep your existing data or delete old data and start over.
+See the :doc:`release notes <release_notes>` for critical upgrade information for specific versions.
 
 Mac OS X
 ========
@@ -50,6 +55,8 @@ To upgrade an existing KA Lite installation.
     #. Restart the server, login to the web app, then check the sidebar contents.
 
     This will enable the greyed-out assessment items on the sidebar.
+
+See the :doc:`release notes <release_notes>` for critical upgrade information for specific versions.
 
 
 Linux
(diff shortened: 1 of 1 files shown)


docs/installguide/install_all.rst
```diff
@@ -7,6 +7,13 @@ Windows
 Upgrading KA Lite over an existing installation is easy -- just run the installer and follow the prompts!
 You don't need to uninstall your old KA Lite installation first.
 
+Upgrade
+_______
+
+To upgrade KA Lite on Windows, simply download the latest installer and run it.
+You can follow the prompts to either keep your existing data or delete old data and start over.
```
benjaoming: I don't think this is true in all cases. It's true if you run `kalite manage setup` because it's interactive. But I think that the case of the deb is that it'll restart the system service, and the process will be non-interactive. In case the user chose to download assessment items during the initial installation, those assessment items will be re-downloaded and the original ones will be deleted and replaced.

MCGallaspy: It's true for the Windows installer -- though now that I'm looking at it closer I see we already had a note about upgrading. Maybe verify what I wrote under the linux section?

MCGallaspy:
Modifying this commit to keep it simple.

benjaoming:
Now it looks much better. I don't know if the re-downloading of assessment items should be mentioned. To be frank, I don't have time to study the process and document it, maybe @EdDixon could try installing the 0.15.0 and then upgrading to 0.15.1 and say if the docs are okay from his experience?

EdDixon:
@benjaoming just a heads up @radinamatic is taking care of testing this cycle while I work on correctional potentials and implementations of Kolibri.

radinamatic:
Ok, @benjaoming @EdDixon, I'll try that testing scenario (0.15 to 0.15.1) and see what happens with assessment items, and if it needs further or more detailed documentation.

benjaoming:
@radinamatic @EdDixon did you try the 0.15 -> 0.15.1 upgrade with a .deb install?

benjaoming:
Mainly @radinamatic :)

radinamatic:
Shoot, my appologies @benjaoming, I've been tota11y sucked up in the Windows vortex so far, and forgot about this one. :stuck_out_tongue_closed_eyes: :crying_cat_face: 

I'll try to get it done on Monday!

radinamatic:
@benjaoming 

I tested the upgrade from `ka-lite-bundle-0.15.0.deb` to `ka-lite-bundle-0.15.1.deb` from pantry, but since both are already bundled with exercises, I haven't seen anything that warrants red flag, docs wise... Maybe I missed something else regarding the assessment items for this release, and I since don't see anything specified in the [release notes](http://ka-lite.readthedocs.org/en/0.15.x/installguide/release_notes.html), I'm not sure at which point the "re-downloading of assessment items" should occur...? Let me know if I should use other .deb installers!

benjaoming:
@radinamatic re-downloading of assessment items should be a question during the upgrade process IMO... but not for the "bundle" .deb because it already has the assessment items (hence "bundled"). It's the one that does not bundle them that should prompt for a download...

radinamatic:
This one: 
https://launchpad.net/~learningequality/+archive/ubuntu/ka-lite/+files/ka-lite_0.15.0-0ubuntu4_all.deb ?

benjaoming:
@radinamatic that one exactly!

Issue #4668, opened 2015-11-14
cannot export "All" data when device not registered


66eli77:
Branch: develop
OS: Mac
reproduce: do not register the device, then login as admin, go to "MANAGE" page and click "export data", then keep the "facility" dropdown list to select "All", click "Export". You will get an empty output file.

the problem seems happen in this line:
https://github.com/learningequality/ka-lite/blob/develop/kalite/control_panel/api_resources.py#L102
if device is not registered, you will get a None zone_id

66eli77:
fixed by #4756

PR #4756, opened 2015-12-12
Export defaults to all users

MCGallaspy:
"Fixes" #4668 -- however I'd like to take a closer look and see if this whole page can be polished up a bit. Consider this PR a statement of intent.

### DIFF
--- a/kalite/control_panel/api_resources.py
+++ b/kalite/control_panel/api_resources.py
@@ -105,7 +105,6 @@ class ParentFacilityUserResource(ModelResource):
 
         # They must have a zone_id, and may have a facility_id and group_id.
         # Try to filter from most specific, to least
-        facility_user_objects = []
         if group_id:
             facility_user_objects = FacilityUser.objects.filter(group__id=group_id)
         elif facility_id:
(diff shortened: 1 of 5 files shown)


MCGallaspy:
See commit message, perhaps more could be done:

```text
Refactor export view, fix bug where Facility selector is disabled …
I dug through the commit messages to determine that the intent was
to disable the Facility selector when the requesting user is a
FacilityUser (not an admin). Instead of a conditional in the
View's render function, I tried to make the intent more explicit
by exposing a new context variable and putting the disable logic
in the View's is_disabled function.

Also fixes a bug where the Facility selector is erroneously
disabled when some resources are selected.
```

benjaoming:
@MCGallaspy You stated an intention to polish up something in the description? Is that still relevant or should this be merged in time for 0.16.4?

MCGallaspy:
I don't intend to push any more changes for now, but I have identified this page as an area for potential improvement.

benjaoming:
Would you say the current changes are merge-able?

MCGallaspy:
Yes, but as always a second pair of eyes helps.

benjaoming:
Agreed, as I'm not the expert in this department of the codebase, perhaps @66eli77 can have a quick look?

benjaoming:
Perhaps @66eli77 can have a quick look?

benjaoming:
Perhaps @66eli77 can have a VERY QUICK quick look? :)

66eli77:
Yes, looking

66eli77:
seems fixed the issue described in #4668 , merged

benjaoming:
Coolios, at least then there's a new feature in 0.16.6 :)

PR #4757, opened 2015-12-15
First pass on the content pack retriever command.

aronasorman:
First pass, although there's one interesting bug on the content db side:

screen shot 2015-12-14 at 18 25 29

### DIFF
--- a/kalite/distributed/management/commands/retrievecontentpack.py
+++ b/kalite/distributed/management/commands/retrievecontentpack.py
@@ -8,6 +8,8 @@ Usage:
   contentpackretrieve -h | --help
 
 """
+import json
+import os
 import urllib
 import tempfile
 import shutil
@@ -15,6 +17,9 @@ import zipfile
 
 from django.core.management.base import BaseCommand
 
+from fle_utils.general import ensure_dir
+
+from kalite.i18n.base import lcode_to_django_lang, get_po_filepath, get_locale_path
 from kalite.topic_tools import settings
 
 from kalite.version import SHORTVERSION
@@ -36,8 +41,26 @@ class Command(BaseCommand):
         with tempfile.NamedTemporaryFile() as f:
             zf = download_content_pack(f, lang)
 
-            extract_catalog_files(zf)
+            extract_catalog_files(zf, lang)
             extract_content_db(zf, lang)
+            extract_content_pack_metadata(zf, lang)
+
+
+def extract_content_pack_metadata(zf, lang):
+    # stub for now, until we implement metadata creation on the maker side.
+    metadata_path = os.path.join(get_locale_path(lang), "{lang}_metadata.json".format(lang=lang))
+    barebones_metadata = {
+        "code": lang,
+        'software_version': SHORTVERSION,
+        'language_pack_version': 1,
+        'percent_translated': 100,
+        'subtitle_count': 0,
+        "name": "DEBUG",
+        'native_name': 'DEBUG',
+    }
+
+    with open(metadata_path, "wb") as f:
+        json.dump(barebones_metadata, f)
 
 
 def download_content_pack(fobj, lang):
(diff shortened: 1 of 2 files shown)


aronasorman:
The bug in here is that the topic tree is somehow rendering the very last level first.

aronasorman:
Turns out there's the parent foreign key isn't getting saved on the content db. Will investigate why.

aronasorman:
Fixed now: 
screen shot 2015-12-15 at 09 09 54

@rtibbles do we want to run the full `annotate_content_items` after fetching the content pack?

rtibbles:
Yes.

aronasorman:
`annotate_content_items` never finishes. (busybeaver)

Will investigate why.

aronasorman:
`annotate_content_items` finishes, it just takes a very long time on my system (at least 15 minutes).

rtibbles:
Interesting, that is significantly longer than previously.

Is something up with the parent mappings?

aronasorman:
Doesn't seem like something is off, as in the end most of the videos are marked as unavailable (as they should be). Maybe because I mark everything as available during content pack creation? Could we have a slower code path when that is the default?

rtibbles:
Ah, right. All non-exericise content should be defaulted to False during content pack creation. Should be quicker.

aronasorman:
Makes sense. Defaulting to False now, then testing.

aronasorman:
4 seconds to annotate everything!

rtibbles:
That is more like it.

MCGallaspy:
If this is meant to replace the existing assessment item stuff, then I have some points to make (from the standup):

1. Clean up assessment.zip code & settings like this: https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/contentload/settings.py#L24
2. Handle content database locations in the same way as we did with assessment items, in terms of where KA Lite looks for them (first in user's data directory, then is system data directory). @benjaoming is that correct? I was looking for the settings that handle this, but I'm not sure if KA Lite still does this.

Let's please try to keep the build process simple by not changing it. Here's how khan_assessment.zip is currently handled in the Windows installer build process, ideally content db workflow is exactly the same:
1. Create sdist -- assessment items not included.
2. Create windows installer -- khan_assessment.zip bundled in but not unpacked.
3. During installation, after source is unpacked, call `kalite manage unpack_assessment_zip` with bundled zip as the target.
4. Voila.

benjaoming:
For Debian/Ubuntu: The assessment zip is unpacked during the build process of ka-lite-bundle. This means that each individual file is unpacked into the source tree that goes into the binary .deb and installed and tracked by dpkg during installations and upgrades.

I would happily get rid of ka-lite-bundle alltogether and move to a scenario where the assessment items are installed through a separate data package, ideally bundled with content and language packs. Those packages could be created just as with the assessment zips. They should be installed in user space and be cross-platform.

Having separate commands for downloading and installing assessment items and content packs into user space is a possible first step towards a unified `import` command. I welcome this! :+1:

aronasorman:
@MCGallaspy Yeah, we'll take that into account. Right now the process should be like this during installation:

1. `sdist` is installed.
2. Run `kalite manage retrievecontentpack file <filepath>` to extract that specific content pack. That will place content DBs in the value of `CONTENT_DATABASES_PATH`, and the assessment resources with whatever variable `unpack_assessment_zip` was using.

benjaoming:
> Create windows installer -- khan_assessment.zip bundled in but not unpacked.

This isn't possible for the .debs as you can only include files that are also installed on the target system. So I would have to install the .zip somewhere, unpack it and then remove a file that dpkg installed, which is probably going to confuse dpkg and make it complain about "this file was removed blah blah" during upgrades.

Suggesting not to make .zip data part of the .deb but instead a separate thing to download. Or create .zips with a .zip and a .deb inside to unpack into the same destination, from where I can make the installer *look* for a .zip in such predefined location (i.e. `cwd`), then giving the user an option to choose a local file or a remote download location through the debconf interface.

kalite/distributed/management/commands/retrievecontentpack.py
```diff
@@ -0,0 +1,102 @@
...
+        json.dump(barebones_metadata, f)
+
+
+def download_content_pack(fobj, lang):
+    url = CONTENT_PACK_URL_TEMPLATE.format(
+        version=SHORTVERSION,
+        code=lang,
+    )
+
+    urllib.urlretrieve(url, filename=fobj.name)
```
aronasorman: Why this rather than `requests.get(stream=True)`? Because the bundled version of `requests` doesn't support `stream`.

rtibbles: Also, requests streaming is one of the few janky things about the whole library, and you have to iterate over the chunks to write it to a file.

aronasorman: Yeah, would've been nice if they exposed a file-like interface instead if `stream` was True.

aronasorman:
Note: the english content pack looks OK. I'm still ironing out the kinks for other languages, but this should be ready to go in this state.

aronasorman:
Aaaand git accidentally pushed to 0.16.x.

Issue #4752, opened 2015-12-10
No matching distribution found for django-tastypie==0.12.2dev0


amodia:
**Branch:** `develop`
**Current behavior:** When running `make dist` command it will spawn this error:
```
python setup.py sdist --formats=gztar,zip --static
This is a static build... invoking pip to put static dependencies in dist-packages/
No handlers could be found for logger "pip.index"
Traceback (most recent call last):
  File "<pyrun>/pyrun2.7.py", line 823, in <module>
  File "<pyrun>/pyrun2.7.py", line 699, in pyrun_execute_script
  File "setup.py", line 307, in <module>
    install_distributions(STATIC_REQUIREMENTS)
  File "setup.py", line 301, in install_distributions
    command.run(opts, distributions)
  File "/Users/richard/installers/osx/temp/pyrun-2.7/lib/python2.7/site-packages/pip/commands/install.py", line 294, in run
    requirement_set.prepare_files(finder)
  File "/Users/richard/installers/osx/temp/pyrun-2.7/lib/python2.7/site-packages/pip/req/req_set.py", line 334, in prepare_files
    functools.partial(self._prepare_file, finder))
  File "/Users/richard/installers/osx/temp/pyrun-2.7/lib/python2.7/site-packages/pip/req/req_set.py", line 321, in _walk_req_to_install
    more_reqs = handler(req_to_install)
  File "/Users/richard/installers/osx/temp/pyrun-2.7/lib/python2.7/site-packages/pip/req/req_set.py", line 461, in _prepare_file
    req_to_install.populate_link(finder, self.upgrade)
  File "/Users/richard/installers/osx/temp/pyrun-2.7/lib/python2.7/site-packages/pip/req/req_install.py", line 250, in populate_link
    self.link = finder.find_requirement(self, upgrade)
  File "/Users/richard/installers/osx/temp/pyrun-2.7/lib/python2.7/site-packages/pip/index.py", line 571, in find_requirement
    'No matching distribution found for %s' % req
pip.exceptions.DistributionNotFound: No matching distribution found for django-tastypie==0.12.2dev0
make: *** [dist] Error 1
.. Abort!  Error/s encountered running 'make dist'.
```

The build server on OS X installer will fail because of this error.

/cc: @benjaoming

arceduardvincent:
Hi @benjaoming and @cpauya I found out some hints in this issue. 
We are trying to invoke pip to put static dependencies in ` dist-packages/` however it is failing to get output from pip.

[...]

PR #4761, opened 2015-12-16
Fix for #4752 until current tastypie master is released

benjaoming:
#4752


benjaoming:
If push comes to shove and there isn't a new tastypie released for 0.17, we have to upload our own hotfixed version on PyPi. But I won't do that quite yet :) We could also dump tastypie back into our codebase, but honestly I'd rather chew off my right leg :D

cpauya:
Thanks for this @benjaoming! :+1:

PR #4763, opened 2015-12-16
Scramble up docopt more to get better parsing

### LINKED ISSUE #4039: Get docopt to understand "kalite manage django_command --option value" with spaces instead of '='

MCGallaspy:
Half-fixes #4039.

commit msg:
```text
I messed around with our docopt function some more to
try to get it to parse long --options with whitespace,
like `kalite start --port 8008`. Seems to work!

Works ok. Maybe doesn't work for short options like `-x 123`
with whitespace, but kalitectl.py itself will complain.

Also included is the script is used to test docopt.
```

This needs to be tested "in the wild" a bit, to check that it indeed parses commands correctly and doesn't just pass on my limited test set.

### DIFF
--- a/docopt_test.py
+++ b/docopt_test.py
@@ -1,0 +1,270 @@
+"""
+We messed around with the internals of docopt in order to get parsing *just right*, but we paid the price because
+it only parses *almost right* -- it's got issues with "--options that have whitepsace".
+
+This script is meant to put our messed-around-with docopt function through it's paces.
+
+Usage: run from the command line. Outputs a bunch of debug info, and raises an assertion error if docopt fails
+  to parse something correctly. Additionally, raises an error if docopt quietly parses something we *don't* want
+  it to parse.
+"""
+import kalitectl
+
+from kalitectl import docopt, DocoptExit
+
+valid_commands = [
+    # Sanity check -- these commands worked before and should continue to work
+    (
+        "kalite start",
+        {
+            '--debug': False,
+            '--foreground': False,
+            '--help': False,
+            '--port': None,
+            '--settings': None,
+            '--skip-job-scheduler': False,
+            '--version': False,
+            '--watch': False,
+            'COMMAND': None,
+            'DJANGO_OPTIONS': [],
+            'diagnose': False,
+            'manage': False,
+            'restart': False,
+            'shell': False,
+            'start': True,
+            'status': False,
+            'stop': False,
+            'test': False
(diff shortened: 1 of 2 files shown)


aronasorman:
The current docopt implementation does not like `es-ES` as a positional argument.

MCGallaspy:
What is the full command that fails?

MCGallaspy:
Also, by *current* you mean on `develop` or `0.16.x`, or from this branch?

aronasorman:
0.16.x. Command:

```
kalite manage languagepackdownload es-ES something
```
Arguments don't really matter. The critical thing here is the dash in `es-ES`, followed by another positional argument.

The error I get:
```
Please only use --option=value or -x123 patterns. No spaces allowed between option and value. The option parser gets confused if you do otherwise.

Will be fixed in a future release.[70307 refs]
[52419 refs]
```

MCGallaspy:
Tried it on this branch? I'll give it a shot.

benjaoming:
I've been thinking if we need to approach this way of recognizing stuff with docopt in a radically different way in which we can retain docopt, yet get around its problems of understanding our `--any option -for --django management "commands" --is --okay` by simply filtering out that part before we switch to docopt parsing?

`[kalite manage blah]` -< that part is for doctopt
`kalite manage blah [--but --this --part --is --not]` <- so we just filter out this part manually before even querying docopt.

`kalite start --settings blah` <- this will work because we explicitly define everything that `start` is allowed to be passed.

Does that make sense? I should probably write code instead of suggesting stuff here :)

MCGallaspy:
@benjaoming sounds okay

benjaoming:
Cool, I'll make amendments to the issue regarding this...

benjaoming:
No, nevermind, just please merge this if it makes life easier, close existing issue and I'll work on this as a separate thing.

MCGallaspy:
I think you have the right idea @benjaoming, I'd rather get it right the first time than merge something half-baked. Will work on it.

MCGallaspy:
Closing, let's take a different approach suggested in #4766.

PR #4770, opened 2015-12-19
0.16.x into develop

MCGallaspy:


MCGallaspy:
Let's see what our robot test overlords say.

PR #4776, opened 2015-12-28
Parallelize bdd tests to use all 4 circle nodes

MCGallaspy:
Opening this after parallelization changes made on circle site -- previously it wasn't using all four VMs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,10 +47,10 @@ test:
 	bin/kalite manage test $(KALITE_TEST_LABELS) --bdd-only
 
 test-bdd:
-	bin/kalite manage test --bdd-only
+	bin/kalite manage test $(KALITE_TEST_LABELS) --bdd-only
 
 test-nobdd:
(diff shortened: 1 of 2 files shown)


MCGallaspy:
Doesn't seem to honor the env variables...

MCGallaspy:
Bingo.

MCGallaspy:
Looks like sauce is struggling with 4 concurrent connections?

MCGallaspy:
Sadly this isn't the no-brainer improvement I thought it would be. But I'll keep it open for comments for a little while.

aronasorman:
One con for using all four is that only one pull request's tests can run. If the benefits aren't huge, then I prefer if we continue just using two containers for each test run.

aronasorman:
I like Circle's UI change: https://circleci.com/gh/learningequality/ka-lite/1734

They now show which container took the longest time. Based on this we're still hamstrung by container 1 (BDD tests).

PR #4779, opened 2016-01-07
Extract metadata from content pack.

aronasorman:
screen shot 2016-01-07 at 14 36 47
Previously we generated our own dummy metadata. Now, extract the one
that comes with the content pack. 

Also, there's a fix in the content pack generation side. We now properly include interface translations. See https://github.com/fle-internal/content-pack-maker/commit/f1041d22ab2525e84f6630c2ddeda3f9aefe7970

### DIFF
--- a/kalite/distributed/management/commands/retrievecontentpack.py
+++ b/kalite/distributed/management/commands/retrievecontentpack.py
@@ -1,4 +1,3 @@
-import json
 import os
 import urllib
 import tempfile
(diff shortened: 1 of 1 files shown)


aronasorman:
Only adds translation metadata though. Subtitles and dubbed video counts will come soon once they're implemented on the content pack repo.

rtibbles:
Should be testable soon. Merging for further testing on develop.

PR #4786, opened 2016-01-20
changed css to less files from the distributed folder

rayykim:
no more css but less files


.gitignore
```diff
@@ -127,3 +128,16 @@ ghostdriver.log
 *.swp
 /data/khan/assessmentitems.json.version
 file_map.json
+kalite/bin/*
```
benjaoming: It seems you're ignoring an awful lot of files that are misplaced on your local system :) Perhaps figure out what causes that?

benjaoming:
We should be moving to LESS or SASS. I think it's okay to move KA Lite to LESS, but we need to ensure the dev/dist process is in place:

1) Guide for devs: How to change LESS source
2) Update our [Makefile](https://github.com/learningequality/ka-lite/blob/master/Makefile#L79) to add LESS compiling to `make assets`.

We should use a PyPi LESS compiler. I only know of the NPM-based one, and it sucks to force devs to use this.

Any experiences with [lesscpy](https://github.com/lesscpy/lesscpy)? I would suggest adding it to test requirements and put the command in `make assets`.

benjaoming:
Closing this now due to inactivity.

Community: Feel free to pick it up for future versions in case you can finalize it.

PR #4611, opened 2015-10-12
Cleanup port stuff

benjaoming:
Fix for:

#4608
#4599

Rids this warning during kalite start:

    WARNING:kalite:App with label django_cherrypy_wsgiserver is missing a models.py module.

Removes lots of stale code.

PR #4795, opened 2016-01-29
Cleanup port stuff 0.16.x

benjaoming:
#4611 was targeted for develop but actually intended for 0.16 -- I didn't see that. I rebased my local branch, so now I've cherry-pick'ed everything from the original PR in here for 0.16.x

@MCGallaspy since you reviewed the original one, I'm assigning you :)


MCGallaspy:
I'm assuming that when I signed off on it the first time it was okay -- doesn't look like anything substantial has changed and tests pass!

Issue #4802, opened 2016-02-02
No content database on first run of dist


benjaoming:
Creating a dist and installing it, renders a KA Lite without a content db:

```
make dist
mkvirtualenv test
workon test
pip install dist/ka-lite-0-16a1.tar.gz
```

@rtibbles I can try fixing this if you give me a hint about what's wrong below:

```
➜  ~  kalite start
Running 'kalite start' as daemon (system service)

Stand by while the server loads its data...

Error reading secret key file. Generating one now. Error was: [Errno 2] No such file or directory: '.kalite/secretkey.txt'
2016-02-02 02:39:19,419 - INFO:kalite: Setting up KA Lite; this may take a few minutes; please wait!

                                     
   _   __  ___    _     _ _          
  | | / / / _ \  | |   (_) |         
  | |/ / / /_\ \ | |    _| |_ ___    
  |    \ |  _  | | |   | | __/ _ \   
  | |\  \| | | | | |___| | ||  __/   
  \_| \_/\_| |_/ \_____/_|\__\___|   
                                     
https://learningequality.org/ka-lite/
                                     
         version 0.16.a1
                                     
WARNING:root:It's recommended that you install Python version 2.7.9. Your version is: 2.7.6

Cleaning *pyc files (if writable) from: .virtualenvs/kalite_pip/share/kalite/dist-packages
Cleaning *pyc files (if writable) from: .virtualenvs/kalite_pip/share/kalite/python-packages
(Re)moving database file to temp location, starting clean install. Recovery location: /tmp/tmpuoTjGT
Copying database file from .virtualenvs/kalite_pip/lib/python2.7/site-packages/kalite/database/data.sqlite to .kalite/database/data.sqlite
Found bundled assessment items
Generating 2048-bit RSA encryption key (may take a few minutes; please wait)...
Done!
Device 'test' has been successfully initialized.
Annotating availability of all content, checking for content in this directory: (.kalite/content)
2016-02-02 02:39:23,478 - INFO:kalite: Annotating content for language: en, channel: khan
INFO:kalite:Annotating content for language: en, channel: khan
2016-02-02 02:39:23,479 - ERROR:kalite: No content database file found
ERROR:kalite:No content database file found
```

benjaoming:
I see the problem.

The content database is not part of the dist, but it's user data. It's being generated as user data with the Makefile, but we need to copy it into a distributed location.

[...]

PR #4804, opened 2016-02-02
Preseed dist with content item db

benjaoming:
Fix for #4802

### DIFF
--- a/MANIFEST.in.dist
+++ b/MANIFEST.in.dist
@@ -10,7 +10,7 @@ include requirements.txt
 recursive-include kalite *.html *.txt *.png *.js *.css *.gif *.less *.otf *.svg *.woff *.eot *.ttf *.zip *.json *.handlebars
 
 # Get the empty DBs -- make sure they are properly generated!
-include kalite/database/data.sqlite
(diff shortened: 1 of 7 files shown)


benjaoming:
@rtibbles would you be able to comment on the most recent test failure?

rtibbles:
Seems to be flakiness in the Item Rating Behave tests - @MCGallaspy any ideas?

rtibbles:
Otherwise, nothing hugely problematic here, except that content data should be coming from an external source, not our old json files.

If we want to just have a content database to seed the dist with, we should probably just have a dummy fixture - as none of the content will be available, except the Khan Exercises (not Perseus).

MCGallaspy:
> Seems to be flakiness in the Item Rating Behave tests - @MCGallaspy any ideas?

I guess it times out waiting for some condition to be fulfilled. Same step passes in another test. I say merge anyway, and we'll address the test separately.

benjaoming:
There were issues about this approach, so we need to implement content DBs from the content packs before this can be used. Manually merging stuff from this PR into #4865.

PR #4806, opened 2016-02-02
[WIP] remove unused commands

benjaoming:
#2908


benjaoming:
I think this should be okay, is i18nize_templates used!?

aronasorman:
Internally, yes. It's for checking if we've marked all templates for i18n. It should be moved to a standalone script, but let's keep it for now to make my life easier :)

benjaoming:
@aronasorman Hmmm... okay, you want to beg for the script's life :D

I get the usecase, but the script is treating symptoms, not causes.. adding localization tags to strings should be an instinct of a programmer. We don't have a script that converts tabs into spaces, neither :)

I could also copy in stuff that I think makes my life easier, but if we want codebase to be easier to manage, stuff like that script should go. The script is nice but should sit in own my_utils django app. Not something for the whole team to manage, and not something to distribute in our software onto user platforms.

aronasorman:
I agree, that's the same rationale for moving our content pack building scripts into another repo.

> I get the usecase, but the script is treating symptoms, not causes.. adding localization tags to strings should be an instinct of a programmer.

Yes, but at the end of the day someone has to check and make sure user-facing strings get marked, and the burden of checking them fell on my shoulders these past few releases. I will do the same for this release, and having `i18nize_templates` makes the process much faster.

> but if we want codebase to be easier to manage, stuff like that script should go.

I agree with this sentiment, but if you really want to push for this change to happen in 0.16, then let's wait until the QA process starts and I can use said script for fixing i18n issues.

Issue #3974, opened 2015-07-02
Remove settings.PROJECT_PATH


benjaoming:
This setting causes user data to be accessed or generated in application directories and application data to be mis-managed.

Ideally, this is replaced by individual settings.

For instance, this:

```
all_playlists = json.load(open(os.path.join(settings.PROJECT_PATH, 'playlist/playlists.json')))
```

Should be:

```
all_playlists = json.load(open(settings.PLAYLIST_PATH))
```

_Or_ by accessing files as package data:

```
from pkg_resources import resource_filename
all_playlists = resource_filename("playlist", "data/playlists.json")
```

Overview of problem:

```
✗ ack-grep --python PROJECT_PATH kalite

kalite/playlist/management/commands/checkplaylistentries.py
22:        all_playlists = json.load(open(os.path.join(settings.PROJECT_PATH, 'playlist/playlists.json')))

kalite/basetests/tests.py
6:PROJECT_PATH = os.path.realpath(os.path.join(THIS_PATH, "..", ".."))
11:        os.path.join(PROJECT_PATH),
12:        os.path.join(PROJECT_PATH, "kalite"),
13:        os.path.join(PROJECT_PATH, "python-packages"),
315:        content_path = os.path.realpath(os.path.join(PROJECT_PATH, "content"))
320:        khan_path = os.path.realpath(os.path.join(PROJECT_PATH, "data", "khan"))
325:        khan_path = os.path.realpath(os.path.join(PROJECT_PATH, "data", "khan"))
351:        scripts_path = os.path.realpath(os.path.join(PROJECT_PATH, "scripts"))

kalite/distributed/management/commands/apacheconfig.py
34:            "project_path": settings.PROJECT_PATH,

kalite/distributed/management/commands/nginxconfig.py
76:        self.stdout.write(config_template % {"root_path": os.path.realpath(settings.PROJECT_PATH + "/../"),

kalite/distributed/management/commands/setup.py
357:        # call_command("clean_pyc", interactive=False, verbosity=options.get("verbosity"), path=os.path.join(settings.PROJECT_PATH, ".."))
462:                        os.path.join(settings.PROJECT_PATH, "..", "bin", kalite_executable))
465:                        os.path.join(settings.PROJECT_PATH, "..", "bin", "windows", "kalite.bat"))

kalite/distributed/management/commands/screenshots.py
184:            self.output_path = os.path.join( os.path.realpath(os.path.join(settings.PROJECT_PATH, '..')),

[...]

PR #4807, opened 2016-02-02
Remove unused management commands and content data

rtibbles:
This PR cleans up now defunct management commands, and removes all content data from the repo.

This content data is now generated on the content pack maker repo.

Putting up for review, to show what would be removed. But, let's not merge it until the dust has settled on #4802 .


benjaoming:
Great stuff, but `init_content_items` is supposed to be staying, no?

rtibbles:
This is why I put it up for review - this is what I want to remove. If init_content_items stays, then all the json files have to stay too, and some other stuff.

benjaoming:
@rtibbles could you open up this PR but for develop as @iCHAIT is intending to factor out some of the stuff in  #3974 that may already be factored out by this PR?

rtibbles:
Wilco.

PR #4814, opened 2016-02-04
Channel context data refactored

### LINKED ISSUE #4812: Move data/khan/images to static library
As we are facing out this "data files" location from the dist, the 0.16a1 misses 3 image files.

screenshot from 2016-02-04 22 24 48

benjaoming:
Fixes #4812 and #4815


rtibbles:
Does what it says on the tin.

PR #4818, opened 2016-02-06
Adds node sort order

### LINKED ISSUE #4811: No sort order on content items
Currently there is no way for determining which order items at the same level and under the same parent in the topic tree should be sorted.

This was not apparent previously, due to the particular idiosyncratic way in which init_content_items sorted through the nodes. Due to this, the order of the ids (which I believe defines the default sort order) happened to coincide with the original sort order of the nodes.

This will no longer be the case when the content packs are imported.

rtibbles:
Fixes #4811

### DIFF
--- a/kalite/topic_tools/content_models.py
+++ b/kalite/topic_tools/content_models.py
@@ -18,7 +18,8 @@ import json
 import sqlite3
 import itertools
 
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Do you have to make a migration for this?

rtibbles:
No, it's not in the django db.

aronasorman:
@rtibbles you forgot to add the import for `FloatField`: https://circleci.com/gh/learningequality/ka-lite/1783

rtibbles:
This is what happens when I code during all hands.

PR #4819, opened 2016-02-06
Content optimization

gwong89:
finished benchmark process ICP/RCP implementation via sockets

on command line, calling --benchmark will run the both the client and server, either wise it won't


benchmark.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+
+
+def benchmark_client(f):
+  """
+  decorator function for benchmarkclient.py
+  The wrapper starts the benchmark server
+  the arguement f is the function that is passed into the decorator
+  (which is the function that will be benchmarked)
```
aronasorman: Follow PEP 257: https://www.python.org/dev/peps/pep-0257/

kalite/topic_tools/content_models.py
```diff
@@ -50,6 +50,8 @@ class Item(Model):
     youtube_id = CharField(null=True)
     size_on_disk = IntegerField(default=0)
     remote_size = IntegerField(default=0)
+    __slots__ = ["title","description","available","files_complete","total_files","kind",
```
aronasorman: Can you remove this?

timer.py
```diff
@@ -0,0 +1,16 @@
+import time
+
```
aronasorman: What's this file? Please remove.

benchmark.py
```diff
@@ -0,0 +1,27 @@
+import benchmarkclient, benchmarkserver
```
aronasorman: separate into different import lines

aronasorman: delete

benchmark.py
```diff
@@ -0,0 +1,27 @@
...
+  """decorator function for benchmarkclient.py
+  
+
+  f --  the function that decorator wraps
+  The wrapper starts the benchmark server
+  and benchmarks the function that decorator is wrapped in
+  """
+  
+  def wrapper(*args, **kwargs):
+    if not kwargs['benchmark']:
```
aronasorman: 4 spaces

aronasorman:
You committed your `benchmarkresults.csv`.

benjaoming:
For now, this does not really seem mature to merge:

1. Merge conflicts
1. socket.error exceptions are thrown away
1. The CLI interferes with the `start` command, I think benchmarking is something we should do outside of the main CLI as it doesn't apply to normal users.
1. I lack some kind of guidance to using the command and understanding the results.

Please feel free to re-open an improved PR! However, more resources should be dedicated to Kolibri, so I would recommend benchmarking that one instead.

As for KA Lite, we probably won't even react if there are concerns about performance. It "seems fine" judging from its adoption.. but of course anyone spotting an optimization and wanting to submit a PR should feel absolutely welcomed :)

benjaoming:
Also, thanks for the effort, @gwong89 !

Issue #4728, opened 2015-12-03
No apparent order to coach report headings


EdDixon:
While testing 15.1 on Ubuntu 14 LTS, I noticed that the coach report for the student seemed very out of sequence with how the events actually happened which were in the sequence of the sidebar design for this section. The session they were done in was uninterrupted so I suspect this could be a bug or feature request not sure which.  I expected them to be listed in the chronological order they were done in. 
![2015-12-03_14-47-32](https://cloud.githubusercontent.com/assets/4349806/11575506/a7c7e204-99ce-11e5-9f04-0d0371d6f687.png)

AOS002:
I Would like to work on this bug. It would be great if someone could guide me through this.

MCGallaspy:
Just noticed this one -- perhaps a good beginner exercise to get familiar with the codebase?

shubhsingh594:
hi i am new to this community , and this would be a good opportunity for me to get familiar with the codebase. i would like to work on this ticket.

MCGallaspy:
Hi @shubhsingh594, sure thing! Take a look at our [getting started documentation](https://github.com/learningequality/ka-lite/wiki/Getting-started) to set up a development environment. Then I would suggest pulling the latest stable branch (`0.15.x` at this time) and try to replicate the issue. The code for coachreports lives in the [`kalite/coachreports` directory](https://github.com/learningequality/ka-lite/tree/0.15.x/kalite/coachreports). Then perhaps @rtibbles can offer guidance on which api endpoint generates the headings, and whether it would be more appropriate to sort the api response before its sent to the client, or to sort on the client.

MCGallaspy:
@shubhsingh594 feel free to pop into our hipchat room or email the dev mailing list for more guidance as well. Email is probably better if you're outside the US, since our hipchat room is only active during daytime hours in California.

rtibbles:
The get_learner_logs function inside api_views.py is the first place to look for this. The content headers are defined by the dicts in the "output_objects" list.

shubhsingh594:
hi all, 
i have submitted a pull request - https://github.com/learningequality/ka-lite/pull/4821 .
i request you to review the code and suggest changes. Thank you !
@MCGallaspy  @rtibbles

shubhsingh594:
[...]

PR #4821, opened 2016-02-08
ticket no:#4728

shubhsingh594:
set the ordering of the Model, to sort them by 'completion_timestamp' and also changed 'dict()' to 'OrderedDict()' to preserve the ordering.

### DIFF
--- a/kalite/coachreports/api_views.py
+++ b/kalite/coachreports/api_views.py
@@ -1,5 +1,6 @@
 from math import ceil
 import datetime
+from collections import OrderedDict
 
 from django.utils.translation import ugettext as _
 from django.db.models import Q, Sum, Avg
@@ -101,7 +102,7 @@ def learner_logs(request):
             topic_objects = log_objects.filter(latest_activity_timestamp__gte=start_date, latest_activity_timestamp__lte=end_date)
             if topic_objects.count() == 0:
                 topic_objects = log_objects
-            objects = dict([(obj[id_field], get_content_cache(language=lang).get(obj[id_field], get_exercise_cache(language=lang).get(obj[id_field]))) for obj in topic_objects]).values()
+            objects = OrderedDict([(obj[id_field], get_content_cache(language=lang).get(obj[id_field], get_exercise_cache(language=lang).get(obj[id_field]))) for obj in topic_objects.order_by('completion_timestamp')]).values()
         output_objects.extend(objects)
         output_logs.extend(log_objects)
 


iCHAIT:
@shubhsingh594   Doesn't the [doc](https://github.com/learningequality/ka-lite/wiki/Coding-guidelines-and-conventions#python-26-restrictions) mention that we can't use OrderedDict from the collections module since it's not present in Python 2.6.

I think you should have a look at [these methods](https://stackoverflow.com/questions/1617078/ordereddict-for-older-versions-of-python).

@MCGallaspy I think we should have tests to cover up for Python 2.6 as well (if we are not removing support for 2.6), for instance this particular PR passes all the tests even though it is not 2.6 compatible.
Another question why do we have `.travis.yml` in the codebase if we are not Travis CI?

rtibbles:
@iCHAIT We will be removing support for Python 2.6 in 0.16, so we are not motivated to reenable testing at this time.

Previously, due to the slowness of Python 2.6, we had got a lot of false positives from our integration tests that made running the test suite not feasible.

iCHAIT:
Fair enough :+1:

MCGallaspy:
I've added updating the wiki to my TODO list.

MCGallaspy:
@shubhsingh594, can you rebase onto `0.16.x` and open a new PR targeting that branch? I feel remiss, I had suggested replicating on `0.15.x` branch because it's stable, but our upcoming release is being developed on `0.16.x` (which is quite unstable at the moment). I should have qualified my earlier comment!

MCGallaspy:
Thanks again @shubhsingh594, just want to make some similar comments here, in addition to my earlier comment about re-targeting the PR to 0.16.x. These are essentially the same comments I have in your other PR, so when you re-target please address them.

1. Please write a regression test for this issue, meaning write a test which fails if your fix is not in place and succeeds after you've made the fix. I'll post more guidance on this below.
2. Please use a more descriptive name than "ticket:#nnnn" -- the PR name should indicate explicitly what it's fixing, like "Order coachreport table headings". It makes searching tickets and PRs much easier!
3. Please reference the issue using the phrase "Fixes #4728" in the body of the PR. It causes the ticket to be closed automatically when the PR is merged.

Later today when I get a chance I'll write a few notes about writing tests.

MCGallaspy:
The `learner_logs` view can be tested using the Django test client, by accessing the associated url and inspecting the result. For an example of how this is already done look at `kalite.coachreports.tests.coachreport_tests.InternalAPITests` (on the `0.16.x` branch). For example, one test method in that TestCase looks like this:

```python
    def test_learner_log_endpoint(self):
        response_keys = ["logs","contents","learners","page","pages","limit"]
        self.client.login(username='admin', password='admin')
        api_resp = json.loads(self.client.get("%s?facility_id=%s" % (self.reverse("learner_logs"), self.facility.id)).content)
        for key in response_keys:
            assert key in api_resp, "{key} not found in learner log API response".format(key)
        self.client.logout()
```

First the `client.login` method is called to be able to access the endpoint (it's restricted), then the `client.get` method is called to get a response, the response is decoded from JSON to a python dict, and then the response is inspected.

Write a regression test in that same module (`coachreport_tests`) which *fails* when the returned `"contents"` value is unordered, but succeeds when it's ordered. This may require you to create some test data in the `setUp` method of your test case. Please let me know if you require more guidance!

benjaoming:
@shubhsingh594 's comment in the original issue:

> okay, while tackling this ticket. i noticed that if there is just one learner. sorting his activities in chronological order makes sense, but if there are multiple learners then all of them may have unique order of activities. At such point it would make more sense to just order the activities in lexicographical order rather than in chronological order. 

Will this be addressed or should we close this PR and try a different approach? One learner does also sound odd......

rtibbles:
Let's close and sort by sort_order instead.

PR #4824, opened 2016-02-09
Do not show unavailable content in sidebar

### LINKED ISSUE #3997: minor UI suggestion
As someone opening this program for the first time, I was confused by why some of the modules at the end were grayed out. I assumed that it was uninstalled programs or something of the nature, but then I was unsure about where to get these files and whether I needed them, and if something was wrong. 

screen shot 2015-07-02 at 4 07 11 pm

MCGallaspy:
Fixes #3997 -- is this the right fix @rtibbles? Any suggestions for writing a test for this? I'm thinking to write a browser test. Perhaps we can take this as an opportunity to experiment with a "stateful" sidebar for writing robust browser tests in KA Lite, kind of like what I suggested for Kolibri?

Screenshots incoming.

### DIFF
--- a/kalite/testing/base_environment.py
+++ b/kalite/testing/base_environment.py
@@ -21,7 +21,7 @@ from peewee import Using
 
 from kalite.testing.base import KALiteTestCase
 from kalite.testing.behave_helpers import login_as_admin, login_as_coach, logout, login_as_learner
-from kalite.topic_tools.content_models import Item, set_database
+from kalite.topic_tools.content_models import Item, set_database, update_item
 
 from securesync.models import Zone, Device, DeviceZone
 
@@ -56,19 +56,14 @@ def setup_content_paths(context, db):
     # These paths are "magic" -- the success or failure of actually visiting the content items in the browser
     # depends on these specific values.
     context.unavailable_content_path, context.available_content_path = (
-        "khan/math/arithmetic/addition-subtraction/two_dig_add_sub/adding-whole-numbers-and-applications-1/",
-        "khan/math/early-math/cc-early-math-counting-topic/cc-early-math-counting/counting-with-small-numbers/",
+        "khan/foo/bar/unavail",
+        "khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/",
     )
+
+    update_item(path="khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/",
+                update={"available": True})
+
     with Using(db, [Item], with_transaction=False):
-        context._available_item = Item.create(
-            title="available_item",
-            description="Bingo",
-            available=True,
-            kind="Video",
-            id="2",
-            slug="avail",
-            path=context.available_content_path
-        )
         context._unavailabe_item = Item.create(
             title="Unavailable item",
             description="baz",
@@ -87,7 +82,6 @@ def teardown_content_paths(context):
     :param context: A behave context, which keeps a reference to the Items so we can clean them up.
     :return: None.
     """
-    context._available_item.delete_instance()
     context._unavailabe_item.delete_instance()
 
 
--- a/kalite/testing/base_environment.py
+++ b/kalite/testing/base_environment.py
@@ -9,7 +9,6 @@ import shutil
 import sauceclient as sc
 import socket
 
-from behave import *
 from httplib import CannotSendRequest
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException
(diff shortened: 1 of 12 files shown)


rtibbles:
The other alternative is to just filter the topic tree API endpoint to filter on available - that might be slightly preferable, only because it stops us from delivering data that won't be used to the client.

So https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/topic_tools/content_models.py#L239 would become:
`
        ).join(Parent, on=(Item.parent == Parent.pk)).where((selector) & (Item.available == True))
`

MCGallaspy:
Used to look like this:
ss2


Now looks like this: 
ss2

MCGallaspy:
> The other alternative is to just filter the topic tree API endpoint to filter on available - that might be slightly preferable, only because it stops us from delivering data that won't be used to the client.

It's also much easier to test an endpoint -- will do!

MCGallaspy:
Ah, somehow I've managed to rebase a random commit into here. Will rebase again to remove.

MCGallaspy:
@rtibbles ready for review

MCGallaspy:
Screenshot above is still accurate.

MCGallaspy:
Based on the screencast, this appears to be a true error *unrelated* to the content rating stuff, but somehow related to content availability. I noticed something similar when I double-checked my branch -- I had to run a "scan" from the management console for anything to appear in the sidebar. Advice on how to approach this @rtibbles?

rtibbles:
Hrm - we should probably do something a little better here in the 'Given I open some available content' step. Perhaps explicitly set the content as available?

'Scan' simply runs the management command 'annotate_content_items' which used to be run during tests - this seems to have vanished somewhere.

I would recommend adding a run of annotate_content_items into the Makefile under the assets directive.

MCGallaspy:
Hmm, `annotate_content_items` does not seem to have done the trick on the build server, despite working well enough locally -- indeed, *more* tests are now failing.

rtibbles:
Rebuilding without cache to see if that is the issue here.

MCGallaspy:
You may be right that availability has affected these tests as well, let me take another look at that.

MCGallaspy:
@rtibbles for the latest 2 commits, I'm not sure this is the right fix, but I was hoping you could give me some guidance if this approach is worth pursuing -- my goal is to display *something* even when invalid content is requested, either because the url corresponds to a nonexistent path, or due to the earlier changes of this PR because the content is unavailable. This might also make testing a little nicer, if we can create new content items on the fly and be assured that they'll get "rendered".

MCGallaspy:
BTW, be wary if you pull this, I'll probably change the commit history later (to at least use better commit messages).

MCGallaspy:
Unavailable content now looks like this:
ss4

@rtibbles once tests pass this is ready for review.

MCGallaspy:
Okay @rtibbles, now the tests are failing because the "available" content is in reality *not* available -- in other words, in the `content_item` api view:

```python
@api_handle_error_with_json
def content_item(request, channel, content_id):
    language = request.language

    # Hardcode the Brazilian Portuguese mapping that only the central server knows about
    if language == "pt-BR":
        language = "pt"

    content = get_content_item(channel=channel, content_id=content_id, language=language)
```

`content` is `None`. I would expect this to exhibit the same behavior as before this PR, and I also see no connection between the return value of `get_content_item` (to which I made no changes) and the original change of this PR in `get_topic_nodes`. It also passes on my machine, so I think the underlying issue is that I'm not reasoning correctly about content on the build server vs my installation. Any advice?

kalite/testing/base_environment.py
```diff
@@ -34,11 +35,54 @@ def after_all(context):
...
+    context.unavailable_content_path, context.available_content_path = (
+        "khan/foo/bar/unavail",
+        "khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/",
+    )
+
+    update_item(path="khan/math/arithmetic/addition-subtraction/basic_addition/addition_1/",
+                update={"available": True})
+
+    with Using(db, [Item], with_transaction=False):
+        context._unavailabe_item = Item.create(
```
rtibbles: I assume this typo might be causing some issues somewhere?

kalite/testing/base_environment.py
```diff
@@ -34,11 +35,54 @@ def after_all(context):
...
+
+
+def teardown_content_paths(context):
+    """
+    The opposite of ``setup_content_urls``. Removes content items created there.
+
+    :param context: A behave context, which keeps a reference to the Items so we can clean them up.
+    :return: None.
+    """
+    context._unavailabe_item.delete_instance()
```
rtibbles: Hrm, perhaps not.

rtibbles: @MCGallaspy one thing, this deletion is not wrapped in the Using context manager, which is causing an error during teardown.

Not the cause of the test errors, but would be good to cleanup.

MCGallaspy: Noted.

benjaoming:
All these BDD test failures look exactly like the ones that I've been getting rid of simply by rebuilding so I'm trying another rebuild for this.

Great work btw!

rtibbles:
@benjaoming No, these test errors were different.

I opened a PR to @MCGallaspy's branch fixing these specific issues, which he then merged, but it seems to have disappeared...

rtibbles:
:boom:

Issue #4662, opened 2015-11-10
My Progress show error message when learner has 0 points


66eli77:
branch: develop
OS: Mac
reproduce: create a new student account and go to "My Progress" with 0 point, you will see this scary error message
![screen shot 2015-11-10 at 3 31 46 pm](https://cloud.githubusercontent.com/assets/5577205/11079336/635f34c0-87c0-11e5-9c7a-5cef17957681.png)

expected behavior: should not see anything

shubhsingh594:
I would like to work on this issue.
I also noticed that this issue has been resolved in the branch - 0.15.x

MCGallaspy:
If it's resolved, then there's nothing to work on! Can you identify the PR that should close this?

MCGallaspy:
I did a cursory search for the closing PR, couldn't find it -- we'll re-open this issue if it comes up again.

shubhsingh594:
although the issue seems to be resolved in the branch - 0.15.x ( i wasn't able to reproduce the error and was getting the required blank progress report). The problem still persists in the branch - 0.16.x .

shubhsingh594:
i have submitted a pull request to the develop branch https://github.com/learningequality/ka-lite/pull/4825 
@MCGallaspy  @66eli77 , please review the code and suggest changes. thankyou :)

radinamatic:
Original issue fixed in #4832, tested on both Alpha4 deb on Ubuntu & build 150 on W7. 

However, while testing I found a curious thing: message for the user appears on both FF & Chromium on Ubuntu, and Chrome on Windows, but is missing on FF on Windows:

![workspace 1_002](https://cloud.githubusercontent.com/assets/1457929/13339545/dec13434-dc29-11e5-9737-b7a7928cf3f0.png)

(Win screenshot coming)

Heading "Progress report" seems to load at first, but then disappears as the message for the user is displayed. This might be an a11y issue, so even if there's nothing to report, I believe page H1 should appear above the user message.

Filing all this as a separate issue.

PR #4825, opened 2016-02-09
ticket:#4662

shubhsingh594:
The get_content_parents() was returning a None object which when passed to the enumerate() was throwing an exception. 
checking whether user_playlists is None and when it is replacing it with an empty list. resolves the error

### DIFF
--- a/kalite/coachreports/models.py
+++ b/kalite/coachreports/models.py
@@ -65,6 +65,9 @@ class PlaylistProgress(PlaylistProgressParent):
 
         # Build a list of playlists for which the user has at least one data point
         user_playlists = get_content_parents(ids=exercise_ids+video_ids)
+	
+	if user_playlists is None:
+		user_playlists = list()
 
         # Store stats for each playlist
         user_progress = list()


MCGallaspy:
Hi @shubhsingh594, thanks for the contribution! You've done a great job, but I have several points to address before I can merge:

1. I know this process is not well documented, but please target changes with the milestone `0.16.0` to the `0.16.x` branch. We'll then pull these changes into develop later. (I'll add a note to the getting started documentation about this shortly.)
2. Please write a regression test for this issue, meaning write a test which *fails* if your fix is not in place and succeeds *after* you've made the fix. I'll post more guidance on this below.
3. Please use a more descriptive name than "ticket:#4662" -- the PR name should indicate explicitly what it's fixing, like "Fix error in PlaylistProgress.user_progress method". It makes searching tickets and PRs much easier!
4. Please reference the issue using the phrase "Fixes #4662" in the body of the PR. It causes the ticket to be closed automatically when the PR is merged.

MCGallaspy:
Testing Model methods is generally much simpler than testing views. You'll need to create a new submodule in the `kalite.coachreports.tests` module, which I recommend calling `model_tests`. Then create a new class which inherits from `kalite.testing.base.KALiteTestCase`. Import the `PlaylistProgress` model, and create a test case which *fails* when your fix is not applied, but succeeds afterwards. The blank `model_tests` module should look something like this:

```python
from kalite.coachreports.models import PlaylistProgress
from kalite.testing.base import KALiteTestCase

class PlaylistProgressTestCase(KALiteTestCase):
    def test_user_progress(self):
        """ Write a docstring which describes the issue you're testing """
        pass  # Write your test method here
```

Again, let me know if you'd like more guidance on this.

PR #4829, opened 2016-02-10
Replace deprecation warning by raising exception

### LINKED ISSUE #4496: Remove the Removed_in_KALite_016 Warning
All warnings should be cleared before release.

MCGallaspy:
Fixes #4496. Decided it would be safest to raise an exception, rather than both trying to remove the warnings *and* refactor old code to remove cruft -- will open separate issue for that.


benjaoming:
Wuhuu!!

PR #4831, opened 2016-02-10
remove github link on offline version of docs

benjaoming:
#4348

### DIFF
--- a/docs/kalite-theme/breadcrumbs.html
+++ b/docs/kalite-theme/breadcrumbs.html
@@ -17,7 +17,7 @@
             <a href="{{ pathto('_sources/' + sourcename, true)|e }}" rel="nofollow"> View page source</a>
           {% endif %}
         {% endif %}
-        {% if not display_github %}
+        {% if "" and not display_github %}
           <a href="http://github.com/learningequality/ka-lite/blob/{{ theme_git_branch }}/sphinx-docs/{{ pagename }}.rst" class="fa fa-github">
               View on Github
           </a>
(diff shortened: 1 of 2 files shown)


docs/kalite-theme/breadcrumbs.html
```diff
@@ -13,11 +13,11 @@
             <a href="https://bitbucket.org/{{ bitbucket_user }}/{{ bitbucket_repo }}/src/{{ bitbucket_version}}{{ conf_py_path }}{{ pagename }}{{ source_suffix }}" class="fa fa-bitbucket"> Edit on Bitbucket</a>
           {% elif show_source and source_url_prefix %}
             <a href="{{ source_url_prefix }}{{ pagename }}{{ source_suffix }}">View page source</a>
-          {% elif show_source and has_source and sourcename %}
+          {% elif "" and show_source and has_source and sourcename %}
```
MCGallaspy: I'd say either remove these branches entirely or update this LOC: https://github.com/learningequality/ka-lite/issues/4348#issuecomment-182539950

benjaoming:
@MCGallaspy please merge me

kitten

MCGallaspy:
I had a comment -- no response?

benjaoming:
Yes, the following commit?

MCGallaspy:
I was referring to this: https://github.com/learningequality/ka-lite/pull/4831/files#r52510801

I meant to either remove the conditional block (which always evaluates to false, as far as I can tell) or update the branch to 0.16.x instead of removing it entirely. No reason to leave dead code lying around.

benjaoming:
> I meant to either remove the conditional block (which always evaluates to false, as far as I can tell)

I left it on purpose and made it always evaluate to false on purpose, because viewing source and linking to Github are still interesting features and I've previously discussed the Github link as an interesting feature even for offline users, so they can be made aware of the possibility to contribute online. Hence, I left it disabled. It's a small issue, it's been fixed, it doesn't make the codebase any more or less complicated as it's just a bit of third-party template code that we've disabled.

Good point about the branch, I PR is wrongly targetted for develop and not 0.16.x

MCGallaspy:
Okay, I didn't understand your reasoning but now I do.

PR #4832, opened 2016-02-10
Fixerror in topic_tools/content_models.py get_content_parents() method

shubhsingh594:
get_content_parents() should return an empty list when passed and empty list of ids.

Fixes https://github.com/learningequality/ka-lite/issues/4662

### DIFF
--- a/kalite/topic_tools/content_models.py
+++ b/kalite/topic_tools/content_models.py
@@ -344,7 +344,11 @@ def get_content_parents(ids=None, **kwargs):
         parent_values = Item.select(
             Parent
         ).join(Parent, on=(Item.parent == Parent.pk)).where(Item.id.in_(ids)).distinct()
+	if parent_values is None:
+  	    parent_values = list()
         return parent_values
+    else:
+	return list()
 
 
 
--- a/kalite/topic_tools/content_models.py
+++ b/kalite/topic_tools/content_models.py
@@ -344,11 +344,11 @@ def get_content_parents(ids=None, **kwargs):
         parent_values = Item.select(
             Parent
         ).join(Parent, on=(Item.parent == Parent.pk)).where(Item.id.in_(ids)).distinct()
-	if parent_values is None:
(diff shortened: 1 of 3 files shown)


MCGallaspy:
Is this meant to replace #4825? If so, please close that PR. It also appears that you've created a *different* fix, changing `get_content_parents` instead of `user_progress`. Why?

Otherwise it seems that you've addressed my comments from #4825, so I'll review this as soon as I have the opportunity!

shubhsingh594:
the user_progress() is fine , inside the method it calls a get_content_parents(), which returns a None object when an empty list of ids is passed to it instead of returning an empty list(), so i thought i would be better to just make changes to get_content_parents(), where the problem actually lies instead of making changes to user_progress().

kalite/topic_tools/content_models.py
```diff
@@ -344,7 +344,11 @@ def get_content_parents(ids=None, **kwargs):
         parent_values = Item.select(
             Parent
         ).join(Parent, on=(Item.parent == Parent.pk)).where(Item.id.in_(ids)).distinct()
+	if parent_values is None:
```
MCGallaspy: Oh, it looks like you used tabs -- please use 4 spaces instead.

MCGallaspy:
Other than the fact that you've used tabs instead of 4 spaces, this looks fine! Change that and I'll merge. Thanks for the contribution. I'm looking into adding some basic linting to catch those kinds of errors in the test suite right now.

I also noticed a new javascript error on that page:

```text
Uncaught ReferenceError: show_message is not defined
  collection.fetch.success @ views.js:96
  options.success @ backbone.js:1027
  fire @ jquery.js:3099
  self.fireWith @ jquery.js:3211
  done @ jquery.js:8264
  (anonymous function) @ jquery.js:8605
```

However, I think that should be addressed as a separate issue.

rtibbles:
This looks like the right fix to me also - good work!

PR #4836, opened 2016-02-12
Add streak_progress to values, since we access it

### LINKED ISSUE #4787: Upgrading from 0.13.x to dev version causes error on facility management page
On `develop`

Steps to reproduce (I did this on Windows, but I expect similar results on other OSes):
1. Install KA Lite 0.13.x
2. Create a student account and log some activity -- presumably you'd need to get a streak on an exercise at least.
3. Upgrade to 0.16.0.
4. Log in as an admin and go to the facility management page.

I see the following error, in place of the normal management console:
```txt
Error

Sorry, but there was an error while trying to load the page.
(KeyError: 'streak_progress') 
```

Screenshot:
virtualbox_ie11 - win7_25_01_2016_10_01_02

MCGallaspy:
Fixes #4787. A mercifully simple fix! However, it suggests we may wish to visit pages after having generated some appropriate exercise logs and assert that no errors are found -- thoughts @rtibbles?

### DIFF
--- a/kalite/control_panel/views.py
+++ b/kalite/control_panel/views.py
@@ -472,7 +472,7 @@ def _get_user_usage_data(users, groups=None, period_start=None, period_end=None,
             Q(end_datetime__gte=period_start) & Q(end_datetime__lte=period_end)
         login_logs = login_logs.filter(login_q1)
     # Force results in a single query
-    exercise_logs = list(exercise_logs.values("exercise_id", "user__pk"))
+    exercise_logs = list(exercise_logs.values("exercise_id", "user__pk", "streak_progress"))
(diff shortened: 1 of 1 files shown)


rtibbles:
Yes, would be good to have some Log file fixtures before initiating behave testing? Can you open an issue for that?

MCGallaspy:
I'm not sure what you're suggesting exactly -- just to try to load fixtures, as a sort of assertion about the Model schema? IMO it might be more helpful (and less implementation dependent) to test the view. Do you think there are any useful tests we can write using the Django test client -- that is, without depending on javascript behavior?

rtibbles:
I meant just testing the view with some actual log data loaded into the test database.

PR #4838, opened 2016-02-12
Lint codebase, add linting step to tests

MCGallaspy:
Don't be afraid of the big diff! This should be really safe. I added a linting step to the tests -- it seems to have caught some legitimate errors. I also cleaned up the codebase both by using PyCharm's automatic features and a little ol' flake8 + elbow grease. See also:

`Ignore some linting errors, clean up others`
```text
I made a value judgement for which errors to ignore based on how quickly
I thought I could clean them up. We should more methodically clean up the
codebase later, by removing the ignored errors one by one.
```

Will maybe need some advice on the errors. @aronasorman @rtibbles


rtibbles:
My only real fear of the big diff is that I can't review it due to the size.

Seems to have a legitimately blocking issue in the test failure though?

MCGallaspy:
> Seems to have a legitimately blocking issue in the test failure though?

Yes, that would a bunch of `undefined name` errors I'm now looking at.

MCGallaspy:
> My only real fear of the big diff is that I can't review it due to the size.

I propose some combination of pulling to compare locally with your favorite diff tool and blind faith.

rtibbles:
If Circle passes it, I'm happy to take the rest on blind faith.

MCGallaspy:
I'm noticing a pattern of stuff breaking because my clean up removed unqualified imports.

benjaoming:
@MCGallaspy I don't think this is realistic. Normally you'd do a cleanup of this proportion right after a release when the desk is clean (i.e. no open PRs).

I hope you know my stance is otherwise very much pro-linting and cleaning up the codebase :)

How about if we manage to land a functional prerelease of 0.16 with all PRs and release blockers closed, then we can do the cleanup and another prerelease?

MCGallaspy:
You raise a good point. I decided to open this now for two reasons, despite the scariness of a big diff prior to release:

1. I wanted to communicate our coding style with new contributors in a more transparent way.
2. Since we've talked about this before but never implemented it, I'm afraid that if I don't do it now it will get delayed indefinitely.

This PR has also identified a few legitimate errors merely by linting (several undefined names). If we close this, then we should incorporate those fixes separately. So do my stated reasons stand up, or should I close this in your opinion?

benjaoming:
@MCGallaspy Have you had a look at how well it merges into develop at this point?

MCGallaspy:
No, but I'm sure there will be numerous merge conflicts. That's a direct result of parallel development on 0.16.x and develop for so long.

rtibbles:
That said, it might be easier to merge an unlinted 0.16.x into develop, and then lint develop instead?

benjaoming:
I agree with @rtibbles and if we're all in the same boat, dancing the same dance to the same rhythm, we can make this happen very soon.

Another option would be:

1. Merge all PRs for 0.16.x
1. Merge all PRs for develop
1. Lint both

MCGallaspy:
Not sure if it's realistic to merge everything and *then* lint, given our history of always having about a half dozen PRs open at any given time. I think we'll just have to choose the least painful time to lint and do it, there won't be a perfect opportunity.

benjaoming:
> given our history of always having about a half dozen PRs open at any given time

Let "merge all PRs for develop" be the goal but not a criterion :) If we can assess that merge conflicts caused are easier to clean up in individual PRs than finalizing all of them, we should do that!

benjaoming:
@MCGallaspy closing this, let's talk about the timing as we assess how tight the deadline for releasing 0.16 is, but I don't feel confident about the amount of issues to add in the larger set of lint changes for 0.16... would rather do it on `develop` right after releasing 0.16 final.

MCGallaspy:
Sure, wanted to keep this open as a reference for what *could* be done, especially regarding which linting errors to ignore.

Issue #4841, opened 2016-02-13
Progress report Error


shubhsingh594:
branch - 0.16.x

When i click on the blue button, to view the details of the progress. An error is thrown (screen shot attached) , the PlaylistDetailReport API also doesnt return a JSON object to the request.

solving this issue is a precursor to solving - https://github.com/learningequality/ka-lite/issues/4785 as in this issue the branch 0.15.x is sending a JSON response to the request but it is not being rendered.

Steps to reproduce:

Logged in as a learner
    Click on user name tab at the top-right corner.
    Go to My Progress.
    Click on the blue button below a progress bar.

![error_blue_button](https://cloud.githubusercontent.com/assets/15982349/13026486/fd40e16e-d24e-11e5-9b2a-7b8ed68d4c05.png)
![error_from_console](https://cloud.githubusercontent.com/assets/15982349/13026488/00052dba-d24f-11e5-9888-eb4d2f3de06c.png)

this is the trace from the console : -

SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
Stack trace:
require<[428]</</jQuery.parseJSON@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:58320:9
handleFailedAPI@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:30071:28
require<[42]</Backbone.ajax/<@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:26912:13
require<[428]</</jQuery.Callbacks/fire@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:53897:10
require<[428]</</jQuery.Callbacks/self.fireWith@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:54009:7
done@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:59064:1
require<[428]</</</<.send/callback/<@http://127.0.0.1:8000/static/js/distributed/bundles/bundle_common.js:59403:1
 bundle_common.js:30078:17

shubhsingh594:
i noticed that this issue has been solved. closing this.

Issue #4785, opened 2016-01-19
Individual progress report not showing completed exercises/videos


jinchuika:
- **Branch**: master (0.15.1)
- **Expected behavior**: when clicking on the blue button below the progress bar for exercises or videos, it should show what exercises are started or completed for that specific topic (see screenshot below).
- **Current behavior**: it shows a loading image and remains the same. The exercises and videos are not shown (see screenshot below).

### Steps to reproduce:
*Logged in as a learner*

1. Click on user name tab at the top-right corner
2. Go to My Progress
3. Click on the blue button below a progress bar

*Logged in as a teacher*

1. Go to Teach tab
2. Open the Tabular Report
3. Click on learner's name
4. Click on the blue button below a progress bar

**Expected behavior**  
![selection_002](https://cloud.githubusercontent.com/assets/5230176/12424329/42d81710-be95-11e5-96a0-06d423b96f30.png)

**Current behavior**
![bug](https://cloud.githubusercontent.com/assets/5230176/12424338/4f96069c-be95-11e5-8153-26d9f856b888.png)

## HOW IT CAN BE SOLVED (or avoided, at least)
This error occurred when I was installing the 0.15.1 version from both PPA and .deb file. But it didn't happened when I installed the 0.15.0 version and THEN updated to the 0.15.1 version.

MCGallaspy:
@jinchuika thanks for the detailed report!

MCGallaspy:
@rtibbles any thoughts on this?

rtibbles:
Looks like it may be an error on the PlaylistDetailReport API endpoint, as the data is not fetching.

@jinchuika if you are able to get any information from the Javascript Console, that would also help!

jinchuika:
Sorry @rtibbles, but I'm not able to get the info at the moment.  

However, I realized that running ```kalite manage setup``` to rebuild the database from the 0.15.1 version will also break the progress report.

shubhsingh594:
hi all, 
Can i take this ?

rtibbles:
@shubhsingh594 Yes - bear in mind, it is possible that this may work fine in English, but be problematic in another language, so test carefully!

jinchuika:
Some other weird stuff I found:

1. If you create the database using `kalite manage setup` in version 0.15.0 and then upgrade to 0.15.1, it will be displayed, but only in English (like it was before #4650 was solved).
[...]

PR #4842, opened 2016-02-13
Fix error in PlaylistDetail API, it is now returning a JSON object.

shubhsingh594:
Fixes https://github.com/learningequality/ka-lite/issues/4841

From here on we can continue solving https://github.com/learningequality/ka-lite/issues/4785 , for branch 0.16.x.

### DIFF
--- a/kalite/topic_tools/content_models.py
+++ b/kalite/topic_tools/content_models.py
@@ -61,7 +61,10 @@ class Item(Model):
         super(Item, self).__init__(*args, **kwargs)
 
 class AssessmentItem(Model):
-    id = CharField(max_length=50, primary_key=True)
+    id = CharField(max_length=50)
+    # looks like peewee doesn't like a primary key field that's not an integer.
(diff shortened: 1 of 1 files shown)


shubhsingh594:
changes made -

1) user_progress_detail() has a language attribute being passed to it, which was not being received.
2) playlist doesn't have a key "children", this attribute is probably getting lost when converting from model to dictionary inside the function get_topic_node().
3) entry was just declared but not defined.

MCGallaspy:
Why did you close this? I simply hadn't had the opportunity to review it yet.

shubhsingh594:
i was a few commits behind the upstream repository. i recently pulled the latest copy, and noticed that the user_progress_detail() has undergone some changes.
hence this fix would not work, and infact may cause conflicts. i would make a new PR soon. :)

issues : -
still, a language parameter is being passed to user_progress_detail() which is not being received, and checked. That problem should be resolved.

PR #4851, opened 2016-02-16
Fix errors discovered during linting

MCGallaspy:
If we forgo #4838, this at least fixes the errors I discovered while linting. Shall we merge this, or should I finish #4838? @rtibbles tagging you since you weighed in on the original as well.

### DIFF
--- a/kalite/shared/decorators/auth.py
+++ b/kalite/shared/decorators/auth.py
@@ -196,7 +196,7 @@ def require_superuser(handler):
 
     """
     def require_superuser_wrapper_fn(request, *args, **kwargs):
-        if getattr(request.user, is_superuser, False):
+        if getattr(request.user, "is_superuser", False):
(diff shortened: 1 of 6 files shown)


rtibbles:
I vote to merge this, so we have these at least, then we can follow @benjaoming's proposed strategy in #4838

kalite/shared/decorators/auth.py
```diff
@@ -196,7 +196,7 @@ def require_superuser(handler):
 
     """
     def require_superuser_wrapper_fn(request, *args, **kwargs):
-        if getattr(request.user, is_superuser, False):
+        if getattr(request.user, "is_superuser", False):
```
benjaoming: Judging from the graveness of the error, this decorator can be much in use :)

benjaoming:
Yes, this is great!

PR #4853, opened 2016-02-16
Create top-level nav item for docs

### LINKED ISSUE #4375: Add documentation to sdist

MCGallaspy:
Fixes #4375 by adding a top-level item to the menu for docs if they are detected. Appears regardless of whether the user is logged in or not. Now also links directly to the user manual section, and I changed the display name accordingly. @radinamatic your opinion on this?

Screenshot:
screenshot

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,7 +8,7 @@ dependencies:
     - "data"
     - "sc-latest-linux"
   override:
-    - pip install -r requirements_test.txt
+    - pip install -r requirements_sphinx.txt
     - pip install -e .
     # This cannot be done because pip on Circle doesn't understand our sdist
     # - make sdist
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ lint:
 test:
 	bin/kalite manage test --bdd-only
 
-test-bdd:
+test-bdd: docs
 	bin/kalite manage test --bdd-only
 
 test-nobdd:
--- a/kalite/distributed/features/steps/ui_regression.py
+++ b/kalite/distributed/features/steps/ui_regression.py
@@ -1,11 +1,13 @@
-from behave import *
+from behave import given, then, when
 from django.core.urlresolvers import reverse
 
-from kalite.testing.behave_helpers import *
+from kalite.testing.behave_helpers import find_id_with_wait, assert_no_element_by_id, \
+    elem_is_visible_with_wait, build_url, login_as_coach
 
 DROPDOWN_MENU_ID = "username"
 NAVBAR_TOGGLE_ID = "navbar_toggle"
 LOGOUT_LINK_ID = "nav_logout"
+DOCS_LINK_ID = "nav_documentation"
 
 
 @given("I'm logged in as a coach")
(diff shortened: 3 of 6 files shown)


radinamatic:
More immediate access to the content relevant for the user - I'd say it's a winner, @MCGallaspy!

MCGallaspy:
This failure must be unrelated to the changes:

```python

  Scenario: Content is available        # kalite/distributed/features/learn_content.feature:10
    Given I open some available content # kalite/distributed/features/steps/learn_content.py:14
    Given I open some available content # kalite/distributed/features/steps/learn_content.py:14 2.374s
    Then I should see no alert          # kalite/distributed/features/steps/learn_content.py:31
    Then I should see no alert          # kalite/distributed/features/steps/learn_content.py:31 6.401s
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1173, in run
          match.run(runner.context)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1589, in run
          self.func(context, *args, **kwargs)
        File "/home/ubuntu/ka-lite/kalite/distributed/features/steps/learn_content.py", line 35, in impl
          assert_no_element_by_css_selector(context, ".alert")
        File "/home/ubuntu/ka-lite/kalite/testing/behave_helpers.py", line 95, in assert_no_element_by_css_selector
          _assert_no_element_by(context, By.CSS_SELECTOR, css_value, wait_time)
        File "/home/ubuntu/ka-lite/kalite/testing/behave_helpers.py", line 85, in _assert_no_element_by
          EC.presence_of_element_located((by, value))
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/selenium/webdriver/support/wait.py", line 91, in until_not
          raise TimeoutException(message)
      TimeoutException: Message:

Link to your job: https://saucelabs.com/jobs/f9090b82c09f489cb5e91413998dd57e
```

MCGallaspy:
From another thread (@benjaoming):

>If you can add a Help tab and only show it to logged-in users who aren't learners, and as @radinamatic mentions deeplink to the User Manual, that'd be sufficient IMO?


Good point, will address.

MCGallaspy:
However, I'm wondering if we should also show the link to non-logged-in users?

MCGallaspy:
I went ahead and made the suggested changes, @benjaoming. The link is *not* shown to non-logged-in users.

MCGallaspy:
Ah, test likely fails because documentation does not exist.

circle.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
 test:
   override:
     - make assets
+    - make docs
```
benjaoming: Since the make target `test-bdd` now depends on docs, you can actually just add it as a dependency of the said target in the Makefile:

```
test-bdd: docs # <- that one
        bin/kalite manage test --bdd-only
```

benjaoming:
Yeah, this is super nice! Makes the docs extremely visible to the users that the docs work really well for in the current circumstances.

As for non-logged in users... hm. Good question, but I would skip it because most users who aren't logged in a Learners in the bigger picture. But there could be a link in the first-run "create admin account" modal.

benjaoming:
Looks great! Sorry if I was slow at merging it -- Github doesn't create direct notifications of subscribers when new commits arrive in a PR, so just post a whatever comment about the updates to makes sure subscribers are pinged.

PR #4859, opened 2016-02-17
Remove docs GitHub link for offline users

### LINKED ISSUE #4348: Tweaks to make to embedded documentation

benjaoming:
Targeting #4831 for 0.16.x instead, fixes #4348

### DIFF
--- a/docs/kalite-theme/breadcrumbs.html
+++ b/docs/kalite-theme/breadcrumbs.html
@@ -17,7 +17,7 @@
             <a href="{{ pathto('_sources/' + sourcename, true)|e }}" rel="nofollow"> View page source</a>
           {% endif %}
         {% endif %}
-        {% if not display_github %}
+        {% if "" and not display_github %}
           <a href="http://github.com/learningequality/ka-lite/blob/{{ theme_git_branch }}/sphinx-docs/{{ pagename }}.rst" class="fa fa-github">
               View on Github
           </a>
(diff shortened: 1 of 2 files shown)


MCGallaspy:
See https://github.com/learningequality/ka-lite/pull/4831#issuecomment-185152489 for explanation.

PR #4861, opened 2016-02-17
fixes error javascript error- SyntaxError: [sprintf] unexpected placeholder.

shubhsingh594:
Fixes https://github.com/learningequality/ka-lite/issues/4785

 '%' needs to be escaped in the playlist-progress-details.handlebars

@MCGallaspy @rtibbles


shubhsingh594:
Fixes https://github.com/learningequality/ka-lite/issues/4841
 
changes made 
1) user_progress_detail() has a language attribute being passed to it, which was not being received.
2) playlist doesn't have a key "children", this attribute is probably getting lost when converting from model to dictionary inside the function get_topic_node(). there fore using get_topic_nodes(), which returns a list of containing the children of the parent id passed.

MCGallaspy:
Looks great! Thanks for the contribution! Only comment I have is to write a test -- looks like there aren't any tests for PlaylistProgressDetailResource api endpoint, but you can very simply assert that it doesn't raise an error like so:

```python
class PlaylistProgressResourceTestCase(FacilityMixins, KALiteClientTestCase):
    def setUp(self):
        # Create data used in the test, like a user
        # FacilityMixins provides a create_student method.

    def test(self):
        base_url = reverse('api_dispatch_list', kwargs={'resource_name': 'playlist_progress_detail'})
        # create a setUp method that creates the user object and just pick a playlist_id value that already exists
        url = base_url + '?' + urllib.urlencode({'user_id': 'foo', 'playlist_id': 'bar'})
        resp = self.client.get(url)
        # Make some assertion about the resp.content to ensure it's error free
```

Write your tests in this file: https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/coachreports/tests/coachreport_tests.py

Let me know if you need any more guidance!

shubhsingh594:
my commit is timing out on circle ci test.
this command seems to be a problem.

while [ ! -e ~/sauce_is_ready ]; do sleep 1; done

@MCGallaspy @rtibbles

benjaoming:
@shubhsingh594 I think this problem might be related to the Sauce setup not receiving jobs from your repo (https://circleci.com/gh/shubhsingh594/ka-lite/8)

This is a flaw from using Sauce but if you deactivate your own repo's registration with Circle, I think it will work by having the tests run through the PR on learningequality's repo. Try removing your Circle repo and adding another commit

shubhsingh594:
@benjaoming  thanks i will try this approach.

PR #4872, opened 2016-02-19
Miscellaneous fixes

MCGallaspy:
## Summary

A few random issues that I noticed while addressing other issues:

* A "style" issue, with a function that name shadowed itself with three different objects sharing the same name!
* A javascript error

I thought about the best way to create a regression test for the js error -- we need a js testing framework. It's sort of unrelated, but I decided to add some tests for the otherwise untested api endpoint as a "consolation prize".

## TODO

- [x] ~~Has documentation been written/updated?~~
- [x] ~~New dependencies (if any) added to requirements file~~
- [x] ~~Add an entry to CHANGELOG.rst~~

### DIFF
--- a/kalite/main/tests/api_tests.py
+++ b/kalite/main/tests/api_tests.py
@@ -1,9 +1,49 @@
 """
 """
+import json
+from django.core.urlresolvers import reverse
+
 from ..models import VideoLog, ExerciseLog
 from kalite.facility.models import Facility, FacilityUser
-from kalite.testing.base import KALiteTestCase
+from kalite.testing.base import KALiteTestCase, KALiteClientTestCase
 from kalite.testing.client import KALiteClient
+
+
+class ContentItemApiViewTestCase(KALiteClientTestCase):
+    def test_sanity(self):
+        resp = self.client.get(reverse('content_item', kwargs={
+            'channel': 'khan',
+            'content_id': 'subtraction_1',
+        }))
+        expected = json.loads('{"available": true, "kind": "Exercise", "description": "Subtract small numbers. All answers are four or less.", "parent": {}, "title": "Subtraction within five", "extra_fields": "{\\\"curated_related_videos\\\":[\\\"x2fefc435\\\"],\\\"display_name\\\":\\\"Subtraction within five\\\",\\\"prerequisites\\\":[\\\"addition_1\\\"],\\\"uses_assessment_items\\\":false,\\\"file_name\\\":\\\"subtraction_1.html\\\",\\\"all_assessment_items\\\":[],\\\"name\\\":\\\"subtraction_1\\\"}", "youtube_id": null, "messages": [], "slug": "subtraction_1", "files_complete": 0, "total_files": 0, "pk": 31, "path": "khan/math/arithmetic/addition-subtraction/basic_addition/subtraction_1/", "size_on_disk": 0, "sort_order": 0.0, "id": "subtraction_1", "remote_size": 0}')  # noqa
+        self.assertDictEqual(json.loads(resp.content), expected)
+
+    def test_template_key_not_in_extra_fields(self):
+        """
+        A regression test -- hard to test the JS, but at least we can assert which keys we expect to find and which
+        we do not.
+        """
+        resp = self.client.get(reverse('content_item', kwargs={
+            'channel': 'khan',
+            'content_id': 'subtraction_1',
+        }))
+        resp_dict = json.loads(resp.content)
+        extra_fields = json.loads(resp_dict['extra_fields'])
+        self.assertNotIn('template', extra_fields)
+
+    def test_file_name_key_in_extra_fields(self):
+        """
+        A regression test -- hard to test the JS, but at least we can assert which keys we expect to find and which
+        we do not.
+        """
+        resp = self.client.get(reverse('content_item', kwargs={
+            'channel': 'khan',
+            'content_id': 'subtraction_1',
+        }))
+        resp_dict = json.loads(resp.content)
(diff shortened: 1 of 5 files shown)


MCGallaspy:
Afraid I'm not too sure why these errors occur: 

```
FAIL: test_file_name_key_in_extra_fields (kalite.main.tests.api_tests.ContentItemApiViewTestCase)
  File "/home/ubuntu/ka-lite/kalite/main/tests/api_tests.py", line 45, in test_file_name_key_in_extra_fields
    self.assertIn('file_name', extra_fields)
AssertionError: 'file_name' not found in {u'display_name': u'Subtraction within five', u'name': u'subtraction_1', u'prerequisites': [u'addition_1'], u'uses_assessment_items': False, u'seconds_per_fast_problem': 4.0, u'basepoints': 10.0, u'exercise_id': u'subtraction_1'}

FAIL: test_sanity (kalite.main.tests.api_tests.ContentItemApiViewTestCase)
  File "/home/ubuntu/ka-lite/kalite/main/tests/api_tests.py", line 19, in test_sanity
    self.assertDictEqual(json.loads(resp.content), expected)
AssertionError: {u'available': False, u'kind': u'Exercise', u'descriptio[1608 chars]': 0} != {u'available': True, u'kind': u'Exercise', u'description[640 chars]': 0}
Diff is 2373 characters long. Set self.maxDiff to None to see it.

Ran 205 tests in 382.071s

FAILED (failures=2, skipped=20)
Destroying test database for alias 'default'...
```

@rtibbles perhaps these are the wrong assertions to make? Would you expect the `content_item` endpoint to return different values depending on the runtime environment?

benjaoming:
It's a persistent test failure though, tried a rebuild. Latest 0.16.x works. You could try a rebase, otherwise there must be some kind of causality.

benjaoming:
Crossing out TODO items renders this:

screenshot from 2016-02-22 03 04 04

kalite/main/tests/api_tests.py
```diff
@@ -1,10 +1,50 @@
...
+from django.core.urlresolvers import reverse
+
 from ..models import VideoLog, ExerciseLog
 from kalite.facility.models import Facility, FacilityUser
-from kalite.testing.base import KALiteTestCase
+from kalite.testing.base import KALiteTestCase, KALiteClientTestCase
 from kalite.testing.client import KALiteClient
 
+
+class ContentItemApiViewTestCase(KALiteClientTestCase):
```
rtibbles: I think for theses tests, you would be better off writing a topic node to the database first, and then retrieving it, then you can be sure about what you find.

MCGallaspy: Okay, I'll do that. I don't understand in general how to know *what* content is available on the test server vs available to me locally. In this specific instance, I also have no idea what's supposed to be in the "extra fields", which was the cause of the original error.

kalite/main/tests/api_tests.py
```diff
@@ -1,9 +1,43 @@
...
+            "id": "3",
+            "slug": "my_content",
+            "path": "khan/my_content",
+            "extra_fields": json.dumps({}),
+            "size_on_disk": 0,
+            "sort_order": 0,
+            "remote_size": 0
+        }
+        with Using(db, [Item], with_transaction=False):
+            self.item = Item.create(**item_attributes)
```
MCGallaspy: @rtibbles I still expect this test to fail, but I'm not really sure how create a "real" content item. The one I create here is not returned by the content_item endpoint. Is this documented anywhere?

benjaoming:
Since this is the only PR merged into 0.16a4, I would suggest rebasing. Test failures are different from the latest build on this branch.

rtibbles:
@MCGallaspy Looks like you are expecting an unavailable response, but getting a 'does not exist' response, am I right?

```
AssertionError: {u'available': False, u'kind': u'Video', u[347 chars]ent'} != {'available': True, 'description': 'A desc[183 chars]ent'}
- {u'available': False,
?  -             ^^^^

+ {'available': True,
?               ^^^

-  u'description': u'This content is unavailable. Either it must be downloaded, or the url is incorrect.',
-  u'id': u'unavailable_content',
+  'description': 'A description!',
+  'extra_fields': '{}',
+  'id': '3',
-  u'kind': u'Video',
?  -        -

+  'kind': 'Video',
-  u'messages': [{u'warning': u'This content was not found! You must login as an admin/coach to download the content.'}],
-  u'path': u'unavailable_content',
-  u'slug': u'unavailable_content',
-  u'title': u'Unavailable Content'}
+  'path': 'khan/my_content',
+  'remote_size': 0,
+  'size_on_disk': 0,
+  'slug': 'my_content',
+  'sort_order': 0,
+  'title': 'My Cool Item'}
```

rtibbles:
Added a fix here:

https://github.com/MCGallaspy/ka-lite/pull/28

Main issue is that 'id' is a text field, not a numeric pk.

MCGallaspy:
> Looks like you are expecting an unavailable response, but getting a 'does not exist' response, am I right?

No, I was expecting an **available** response. In reality I'd like to make some reasonable assertion about `extra_fields`, since that was the cause of the original issue. But I've fallen short in this test because I'm not sure what assertions *can* be made about it. Instead I hope that having a test for this endpoint will at least raise the alarm for unintended consequences of future changes.

rtibbles:
Yeah, I think I correctly interpreted it when I tried fixing your test locally.

MCGallaspy:
Rebased onto better, more error-free `0.16.x` branch!

rtibbles:
Much success!

Issue #4840, opened 2016-02-12
Javascript error on "My Progress" page


MCGallaspy:
Currently on `0.16.x` (see [this comment](https://github.com/learningequality/ka-lite/pull/4832#issuecomment-183450219)):

To reproduce, log in as a student (possibly with 0 points) and navigate to the "My Progress" page from the user's dropdown menu.
Observe the following error in the console:

```
Uncaught ReferenceError: show_message is not defined
  collection.fetch.success @ views.js:96
  options.success @ backbone.js:1027
  fire @ jquery.js:3099
  self.fireWith @ jquery.js:3211
  done @ jquery.js:8264
  (anonymous function) @ jquery.js:8605
```

shubhsingh594:
can i take this ?

rtibbles:
Yes, @shubhsingh594 - first check that the JS module that defines 'show_message' is being imported, and if it is, make sure that show_message is being referenced as a method of it. My guess is that it probably just needs to be properly namespaced.

PR #4877, opened 2016-02-19
fixes error show_message is not defined

shubhsingh594:
Fixes https://github.com/learningequality/ka-lite/issues/4840


MCGallaspy:
Bingo, looks good! Generally I would require a test, but since this is a trivial error I don't find it necessary. Thanks for the contribution!

PR #4881, opened 2016-02-21
Manpages in sdist

benjaoming:
## Summary

Taking over from #4623 so it targets 0.16.x

The manpage is created when running `make release` and NOT `make dist` so the latter retains its Windows compatibility.

screenshot from 2016-02-22 00 31 34


benjaoming:
Fails for unrelated reasons (selenium timeout) -- already reviewed by @MCGallaspy -- doing a self-merge.

PR #4884, opened 2016-02-22
Do not create a .zip archive for sdist, tar.gz is enough

benjaoming:
## Summary

Save time and space. The .zip files are not necessary for pip to work.

## Reviewer guidance

Please comment if you **do** find them necessary on PyPi.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,9 +12,12 @@ help:
 	@echo "assets - build all JS/CSS assets"
 	@echo "coverage - check code coverage quickly with the default Python"
 	@echo "docs - generate Sphinx HTML documentation, including API docs"
-	@echo "release - package and upload a release"
-	@echo "dist - package locally"
+	@echo "release - package and upload a release, options: format=[gztar,zip]"
+	@echo "dist - package locally, options: format=[gztar,zip]"
 	@echo "install - install the package to the active Python's site-packages"
+
+# used for release and dist targets
+format?=gztar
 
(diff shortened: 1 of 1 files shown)


MCGallaspy:
I do find it necessary on Windows.

MCGallaspy:
I can always build it myself, but I just now switched to downloading the sdist zipfile from PyPI -- it would be nice to keep it that way.

benjaoming:
Aha! I thought you were building it with `make dist`, I'll make sure it builds zips for PyPi only instead...

aronasorman:
Merge conflicts.

benjaoming:
There we go, resolved

MCGallaspy:
Actually, I couldn't find a good way to download the zipfile (as opposed the tarball) from pypi using `pip download` -- is there an option for specifying the format? If not, it may yet be useful to keep this in the Makefile, so that the buildserver can continue to generate the sdist itself.

benjaoming:
Okay, I was assuming you were doing the other thing.

Would it be okay to add an argument to the dist target, something like `make dist format=zip` ?

In that way, when we make the dist on other systems, we won't needlessly generate zip files, and when we create them on Windows, we won't be needlessly generating tarballs. And you can even do `make dist format=zip,tarball`

benjaoming:
http://stackoverflow.com/a/2826068/405682

MCGallaspy:
Sure, or perhaps we can control it with an environment variable? That would be fine.

PR #4865, opened 2016-02-18
Retrieving en-minimal content pack + remove stale code

benjaoming:
## Summary

Make the sdist backwards-compatible by shipping an English content db. After backwards-compat is asserted, remove all the stale code and JSON data.

This PR led me to fix unrelated issues:

 - Add better logging because I was finding it difficult to hunt down problems in JSON views
 - Update tests to match new content db
 - Fix an error in video download (because ultimately I needed to see that the full scope was working)
 - Hotfix (removing check `total_files!=0`) check because it was breaking the video update topic tree view

## TODO

- [x] Update `make assets`
- [x] Merge in #4804 to use the English content db as a database template
- [x] Added proper logging so we can read everything in server.log in the same format (settings.LOGGING)
- [x] Remove stale code and data files (as @rtibbles had previously suggested but retargetted for 0.17)

## Reviewer guidance

Check that there are no regressions in removed functionality.

## Issues addressed

#4864 
#3327 (logging)
#4719 (prettify unpacking assessment items)

## Documentation

No

PR #4894, opened 2016-02-22
Remove unnecessary steps from Makefile

MCGallaspy:
From the commit message:
```
Remove unnecessary steps from Makefile

From the assets directive, we don't use the `init_content_items` or `annotate_content_items` commands while making
the sdist anymore -- this should be replaced with a post-install call to `retrievecontentpack`.
```

@benjaoming is this true?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,8 +87,6 @@ assets:
 	npm install --production
 	node build.js
 	bin/kalite manage compileymltojson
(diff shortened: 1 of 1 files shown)


benjaoming:
@MCGallaspy please see #4865 : https://github.com/learningequality/ka-lite/pull/4865/files

PR #4904, opened 2016-02-23
Rm unused commands (+ kalite.testing and kalite.basetests)

benjaoming:
## Summary

Re-target #4806 for develop

Updates are in release notes, here's a dump of the currently removed commands:

 * `kalite manage gitmigrate`
 * `kalite manage katest`
 * `kalite manage initdconfig`
 * `kalite manage nginxconfig`
 * `kalite manage apacheconfig`
 * `kalite manage todolist`
 * `kalite manage i18nize_templates`
 * `kalite manage benchmark`
 * `kalite manage createmodel`
 * `kalite manage modifymodel`
 * `kalite manage readmodel`
 * `kalite manage runcode`
 * `kalite manage unpack_asessment_zip`
 * `kalite manage create_dummy_language_pack`
 * `kalite manage generate_blacklist`


benjaoming:
Self-merging this because we should have enough time before 0.17 releases to address if I was too eager removing any of these commands.

The most notable removal is probably `runcode` -- but it's not mentioned in the docs, and not used in the code base. I therefore find it safe to remove and hope our efforts in supporting users will not have to rely on distributing this rather odd command as opposed to user scripts and instructions for running code in `kalite shell`.

PR #4909, opened 2016-02-24
Let commands actually set their progress log.

### LINKED ISSUE #4817: Not able to Download videos.
I have set up ka-lite on my local machine, when I try to download a video from the Videos Tab it shows size of selected videos as 0.0 MB - 

screen shot 2016-02-05 at 10 41 50 pm

Moreover, when I hit the download button, it shows a banner at the top and nothing happens, it does not show any progress bar that I can see in the ka-lite DEMO.

screen shot 2016-02-05 at 10 42 38 pm

Checked out from develop branch, the 0.16.x branch also gives the same problem.

rtibbles:
## Summary

Removes the seemingly superfluous foreground flag on UpdatesCommands - as it is set during init, can't be modified post hoc after handle (as suggested in the comment that implemented it).

## Reviewer guidance

This is a reversion of a previous change that caused this bug.

## Issues addressed

Fixes #4817.

@benjaoming please comment as to why the foreground flag is necessary if I am removing it in this way (and just letting the process_log be None), and I will try to update this to account for that use case, but for now, it just seemed to block the intended use of these UpdatesCommands.

### DIFF
--- a/kalite/distributed/management/commands/retrievecontentpack.py
+++ b/kalite/distributed/management/commands/retrievecontentpack.py
@@ -73,6 +73,8 @@ class Command(UpdatesStaticCommand):
 
     def handle(self, *args, **options):
 
+        self.setup(options)
+
         operation = args[0]
         self.minimal = options.get('minimal', False)
         self.foreground = options.get('foreground', False)
--- a/kalite/updates/management/commands/classes.py
+++ b/kalite/updates/management/commands/classes.py
@@ -40,22 +40,24 @@ class UpdatesCommand(LocaleAwareCommand):
 
     def __init__(self, process_name=None, *args, **kwargs):
 
+        self.process_name = process_name or self.__class__.__module__.split(".")[-1]
+
+        super(UpdatesCommand, self).__init__(*args, **kwargs)
+
+    def setup(self, options):
+
         # Should be set by command's handle() method since there's no clear
         # call structure defined. Respected by the methods that update stuff
         # in the database.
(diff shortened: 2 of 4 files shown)


benjaoming:
@rtibbles Creating the content db template happens from command line, so it's supposed to run in the foreground and not put data in data.sqlite (which isn't even created at that point).

I introduced the --foreground flag especially for `make assets`, which invokes:

    bin/kalite manage retrievecontentpack download en --minimal --foreground --template

I'm not sure why it has affect other things, apologies for that. But please don't remove it, consider this commit:

https://github.com/rtibbles/ka-lite/commit/1b53758f126560d921a8544cef1ce115c88837c8#diff-9b6a9dbe0240241084068eb7080204e8

rtibbles:
OK, will work to make it function as intended - however, from what I can tell, the foreground flag is actually doing nothing, because the options never get parsed anywhere?

So, in the retrievecontentpacks instance, you actually want it *not* to be foreground, because you want it to run without creating process logs, right? In which case, we should maybe have a 'background' flag that gets invoked instead, and prevents the process logs from being accessed?

rtibbles:
How about this, @benjaoming? Setting the foreground flag should now prevent it from doing any process_log related malarkey, and each Class now has to call the setup method in handle to make sure the process logs are there if needed.

benjaoming:
Apparently a blunder in that I set `self.foreground` to False disregarding my own setup argument, good intention half delivery :) Thanks for fixing it!

PR #4912, opened 2016-02-24
Don't load facilities into session object.

### LINKED ISSUE #4720: Don't load all facilities into session (ever? at least for central server...)
- The following code, in kalite/facility/middleware.py, loads data for all facilities into the session object:

```Python
# To enable simplified login, also list the id and names of all facilities in the session object.
# This keeps it cached so that the status api call can return this to the client side
# without significantly increasing DB load on every status call.
request.session["facilities"] = [{"id": id, "name": name} for id, name in Facility.objects.values_list("id", "name")]
```

- The session object is persisted, either in the DB or in temp files.

- On the central server, we have 6K+ facilities, and many sessions

- The central server DB is now 7GB+, and we are no longer doing backups because the 500GB backup drive overflowed.

@rtibbles Is that code still required? I can't see any references to `request.session["facilities"]` elsewhere. If it *is* required, we should find a way to disable it on the central server, where it definitely isn't required.

rtibbles:
## Summary

Don't load facilities into session object.
Don't even bother caching a list of facilities on the central server.

No regression test, as the main offense of this issue was on the central server, not distributed.

## Issues addressed

Fixes #4720

### DIFF
--- a/kalite/facility/api_resources.py
+++ b/kalite/facility/api_resources.py
@@ -1,7 +1,6 @@
 import datetime
-import os
 from dateutil.tz import tzlocal
-
+from django.db.models.signals import post_save
 from tastypie import fields
 from tastypie.http import HttpUnauthorized
 from tastypie.resources import ModelResource, ALL_WITH_RELATIONS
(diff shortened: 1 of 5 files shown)


benjaoming:
This works, I checked the codebase both Python and HTML that the session isn't used anywhere else and it isn't.

But there are the two others, `facility_count` and `facility_exists` - would it be possible to get rid of those, too?

In .py files:

```
➜  ka-lite git:(no-zip-dist) ack-grep --python "session.*facility_count" kalite      
kalite/facility/decorators.py
53:        elif request.session["facility_count"] == 1:
57:        elif request.session["facility_count"] > 0:

kalite/facility/views.py
187:        "singlefacility": request.session["facility_count"] == 1,

kalite/facility/middleware.py
12:def refresh_session_facility_info(request, facility_count):
15:    request.session["facility_count"] = facility_count
16:    request.session["facility_exists"] = request.session["facility_count"] > 0
64:            refresh_session_facility_info(request, facility_count=Facility.objects.count())
➜  ka-lite git:(no-zip-dist) ack-grep --python "session.*facility_exists" kalite
kalite/distributed/views.py
51:            elif not request.session["facility_exists"]:
59:            elif not request.session["facility_exists"]:

kalite/facility/decorators.py
94:        if not request.session["facility_exists"]:

kalite/facility/middleware.py
16:    request.session["facility_exists"] = request.session["facility_count"] > 0
```

In HTML templates: No usage of these sessions found.

rtibbles:
Possibly, I have no recollection what they are used for, so didn't touch.
Will investigate.

benjaoming:
They are harmless, but just really lame :)

rtibbles:
Managed to remove `"singlefacility": request.session["facility_count"] == 1,` - but, I don't want to change the others because it will add db hits on every request due to the middleware usages, and in some decorators, so will leave for now.

rtibbles:
Think this is ready to merge, @benjaoming.

PR #4919, opened 2016-02-26
Propagates topic selection to tabular coach report.

### LINKED ISSUE #4869: Coach Reports - tabular view pulls up full range of data regardless of  topic selection
Branch/Build: 0.16.x, OSX 10.10

Steps to Reproduce:
1. Sign in as admin/coach
2. Go to Teach tab
3. Do what is needed to fetch some real data (different facility, group, dates, etc)
4. Check the tabular view
5. Now do some topic selection
6. Check tabular view again

Expected Results:
Tabular view should have different data depending on usage of topic selection or "default overview"

Actual Results:
Tabular view shows the same full range of data both times

rtibbles:
## Summary

Ensures that when the topic selector is used in the coach reports, this also restricts the data used in the tabular report to just those topics.

I have not written any tests, because we need Javascript unit tests to properly check this. Sorry, the integration tests are just too flakey and difficult to implement at this juncture.

## Reviewer guidance

To test this, generate some data using `generaterealdata`, go to the coach reports, look at the tabular report.

Select a topic from the topic drop down. The report view will refresh. Show the tabular report again. Now only content items in the topic drop down will display.

## Issues addressed

Fixes #4869.
Also puts in the KA Lite side fix for #4728 - sort_order is being set incorrectly by content packs though, so this change is not yet visible.

### DIFF
--- a/kalite/coachreports/api_views.py
+++ b/kalite/coachreports/api_views.py
@@ -13,6 +13,22 @@ from kalite.facility.models import FacilityUser
 from kalite.shared.decorators.auth import require_admin
 from kalite.topic_tools.content_models import get_topic_contents, get_topic_nodes, get_leafed_topics, get_content_parents
 
+
+def unique_by_id_and_kind_sort(seq):
+
+    seq.sort(key=lambda x: x.get("sort_order", 0))
+    seen = {}
+    result = []
+    for item in seq:
+        marker = item.get("id") + item.get("kind")
+
+        if marker in seen:
+            continue
+        seen[marker] = 1
+        result.append(item)
+    return result
+
+
 def get_learners_from_GET(request):
     learner_ids = request.GET.getlist("user_id")
 
@@ -106,6 +122,8 @@ def learner_logs(request):
             objects = dict([(item.get("id"), item) for item in get_topic_nodes(ids=[obj[id_field] for obj in topic_objects]) or []]).values()
         output_objects.extend(objects)
         output_logs.extend(log_objects)
+
+    output_objects = unique_by_id_and_kind_sort(output_objects)
 
     return JsonResponse({
         "logs": output_logs,
--- a/kalite/coachreports/api_views.py
+++ b/kalite/coachreports/api_views.py
@@ -15,6 +15,12 @@ from kalite.topic_tools.content_models import get_topic_contents, get_topic_node
 
 
 def unique_by_id_and_kind_sort(seq):
+    """
+    Due to the fact that we have duplicate content items for the same content id in our topic tree, as the way that
+    we have implemented duplication of content across the topic tree.
+    :param seq: an iterator of content items.
+    :return: A unique, sorted list of content items.
+    """
 
     seq.sort(key=lambda x: x.get("sort_order", 0))
     seen = {}
@@ -126,7 +132,9 @@ def learner_logs(request):
     output_objects = unique_by_id_and_kind_sort(output_objects)
 
     return JsonResponse({
+        # All learner log objects for each content item.
         "logs": output_logs,
+        # All content items for which logs are being returned.
         "contents": output_objects,
         # Sometimes 'learners' gets collapsed to a list from the Queryset. This insures against that eventuality.
         "learners": [{
--- a/kalite/coachreports/api_views.py
+++ b/kalite/coachreports/api_views.py
@@ -74,7 +74,7 @@ def learner_logs(request):
 
     end_date = request.GET.get("end_date")
 
-    topic_ids = request.GET.getlist("topic_id", [])
+    topic_ids = json.loads(request.GET.get("topic_ids", "[]"))
 
     learners = get_learners_from_GET(request)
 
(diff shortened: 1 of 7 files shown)


MCGallaspy:
> Sorry, the integration tests are just too flakey and difficult to implement at this juncture.

What test *would* you write in this case, ideally?

rtibbles:
That when you click the drop down on the topic selector, it then reduces the displayed topics in the tabular report.

One thing I can do is test the API endpoint instead, so I will do that.

Meanwhile - any insight into why this behave test for content rating keeps failing on me, @MCGallaspy?

MCGallaspy:
Yeah, looking at [the screencast](https://saucelabs.com/jobs/ad2ec1ca4b1f4246b47412011c812d85) you can see that the sidebar never closes. Can't interact with widget, hence the test fails. Can we figure out why the sidebar never closes occasionally?

rtibbles:
Sounds like a recurrence of the bug that was reported before but we couldn't replicate.

MCGallaspy:
Testing the endpoint would be great. I think we can iteratively improve the testability of our app, so if you think it's useful I'll write an integration test for coachreports in this case.

kalite/coachreports/api_views.py
```diff
@@ -13,6 +13,22 @@
...
 from kalite.topic_tools.content_models import get_topic_contents, get_topic_nodes, get_leafed_topics, get_content_parents
 
+
+def unique_by_id_and_kind_sort(seq):
+
+    seq.sort(key=lambda x: x.get("sort_order", 0))
+    seen = {}
+    result = []
+    for item in seq:
+        marker = item.get("id") + item.get("kind")
```
MCGallaspy: Small thing, but is it possible for id + kind to be the same for two different items, even when id and kind are separately different? You could instead use a tuple.

rtibbles: I can't imagine a circumstance where they could be equal, unless an item had a blank "kind" field.

kalite/coachreports/api_views.py
```diff
@@ -107,6 +123,8 @@ def learner_logs(request):
         output_objects.extend(objects)
         output_logs.extend(log_objects)
 
+    output_objects = unique_by_id_and_kind_sort(output_objects)
```
MCGallaspy: I'm a little confused, but I think this kind of question could be cleared up by a few simple examples/client tests of this endpoint. Why does it *matter* that output_objects are sorted and unique?

Also, it seems that `output_objects` is a dictionary mapping topic ids to some value, but both `output_objects` and the key in the JsonResponse below (`"contents"`) are pretty vague.

MCGallaspy:
This seems to work as intended. If you intend to address my comments about the test, then I'll wait to merge. Otherwise we're ready. Let me know and I'll click the button. I *do* think it would be a good idea to write integration tests for coach reports, but I'm willing to tackle it separately.

rtibbles:
Yeah, I'll get the API Tests working a bit more in depth.

MCGallaspy:
Build was cancelled? I'd say try rebasing anyway, since there's a critical fix on 0.16.x now.

rtibbles:
Yeah, this was where I was trying to log errors for the sidebar stuff. Will rebase and rerun.

kalite/coachreports/tests/coachreport_tests.py
```diff
@@ -139,3 +141,114 @@ def test_playlist_progress(self):
...
+        self.exercise2.delete()
+        self.topic1.delete()
+        self.topic2.delete()
+
+    def test_learner_log_topic_filters(self):
+
+        self.client.login(username='admin', password='admin')
+        api_resp_1 = json.loads(self.client.get("%s?facility_id=%s" % (self.reverse("learner_logs"), self.facility.id)).content)
+        api_resp_2 = json.loads(self.client.get("%s?facility_id=%s&topic_ids=%s" % (self.reverse("learner_logs"), self.facility.id, json.dumps([self.topic1.id]))).content)
+        assert len(api_resp_2["contents"]) < len(api_resp_1["contents"])
```
MCGallaspy: Looking good! Still 2 things. Again, happy to merge as-is but I think some small easy improvements could be made.

1. Use the TestCase [assertion methods](https://docs.djangoproject.com/en/1.9/topics/testing/tools/#assertions) for pretty output.
2. Instead of asserting about the *length*, can you make assertions about the *content*? E.g.
```python
self.assertEqual(api_resp_2["contents"], ["foo", "bar"])
```
Or do you think that is likely to change frequently from version to version, and hence maybe *too* specific?

rtibbles: Or make an assertion about the actual contents in the test below - see `test_learner_log_topic_filters_contents_id`, they are separated into multiple tests, as you asked I limit myself to a single assertion per test.

MCGallaspy: > I make an assertion about the actual contents in the test below

Ah I see. My mistake!

rtibbles:
@MCGallaspy I think this is ready to merge.

PR #4925, opened 2016-02-26
Add more ignore patterns when making messages

aronasorman:
## Summary

Add more ignore patterns for our `makemessages`, so it doesn't try to process random stuff like `node_modules` and `dist-packages-temp`, which will make it error out due to some files' non-ASCII nature.

## Reviewer guidance

Dead simple PR.

### DIFF
--- a/kalite/i18n/management/commands/makemessages.py
+++ b/kalite/i18n/management/commands/makemessages.py
@@ -11,16 +11,27 @@ from django.conf import settings
 from django.core.management.commands import makemessages
 
 
+IGNORE_PATTERNS = [
+    "*/node_modules*",
+    "*dist-packages*",
+    "*python-packages*",
+    "*/kalite/static*",
+    "*bundle*.js",
+    "*/ka-lite/build/*",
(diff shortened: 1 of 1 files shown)


kalite/i18n/management/commands/makemessages.py
```diff
@@ -11,16 +11,27 @@
...
 
+IGNORE_PATTERNS = [
+    "*/node_modules*",
+    "*dist-packages*",
+    "*python-packages*",
+    "*/kalite/static*",
+    "*bundle*.js",
+    "*/ka-lite/build/*",
+    "*/js/i18n/*.js",
+    "*/ka-lite/docs/*"
```
aronasorman: I'm withholding the internationalization of docs for now. Takes a bit of thinking through before we should go ahead with it.

aronasorman: This ignore pattern is mainly for `gettext` not to see the text in RST's generated buttons and links. We need more integration for the docs itself.

Issue #4927, opened 2016-02-27
'Search completed message' when Search is complete


manipalsingh013:
## Summary
Search result should display some message when nothing was found for that search so as to make user clear that search is complete. Some message like "Search completed, no content was found for your search. Try something else."

Branch: **0.15.x**
**Ubuntu 14.04**

## How to reproduce
1. Login as some user
2. Search for something, maybe some random string for which nothing exists in content.

rtibbles:
Fixed in #4928

PR #4928, opened 2016-02-27
shows a message when the search returns no content

shubhsingh594:
Fixes https://github.com/learningequality/ka-lite/issues/4927


kalite/main/api_views.py
```diff
@@ -34,6 +34,9 @@ def search_api(request, channel):
 
     matches, exact, pages = search_topic_nodes(query=query, channel=channel, language=request.language, page=1, items_per_page=15, exact=False)
 
+    if matches==list():
```
rtibbles: A more Pythonic way to do this would be to test matches for emptiness, rather than seeing if it equals an empty list.

In Python empty objects are False-y, so we could just do:
```
if not matches:
```

rtibbles:
:+1:

PR #4936, opened 2016-02-29
Change message for the user while they wait for server response

radinamatic:
## Summary

Maybe there is a "real" `case 0` that warrants the previous message, but in 3 months of testing I haven't seen it. Every time this message appears, it's about waiting for the server response at logging in and out, and not really needing to "try again later". 

I can't even grab a screenshot, it resolves too fast on my dev setup, which kind of proves a point... :)
Let me know what you think @rtibbles @aronasorman

### DIFF
--- a/kalite/distributed/static/js/distributed/utils/api.js
+++ b/kalite/distributed/static/js/distributed/utils/api.js
@@ -57,7 +57,7 @@ function handleFailedAPI(resp, error_prefix) {
     var messages = {};
     switch (resp.status) {
         case 0:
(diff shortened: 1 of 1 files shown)


rtibbles:
I can't think of a real case 0, as that's not a valid http response code anyway, so I think this change should be very low impact.

radinamatic:
CircleCI seems to disagree... :unamused:

rtibbles:
I fail to see how your PR is related to that test failure.

PR #4939, opened 2016-02-29
Allow positional arguments with dashes inside

### LINKED ISSUE #4935: Arg parsing bug affecting language pack downloading
## Summary

I can't download language packs with a dash in their name, like `es-ES`.

Output:

```
$ kalite manage retrievecontentpack local 'es-ES' '~/src/github.com/fle-internal/content-pack-maker/out/es-ES.zip'
Please only use --option=value or -x123 patterns. No spaces allowed between option and value. The option parser gets confused if you do otherwise.

Will be fixed in a future release.[70307 refs]
[52419 refs]
```

## Branch or installer method

0.16.x

```
INSERT TRACEBACK, LOG MESSAGES ETC. HERE
```

## How to reproduce

See log pasted above.

## Screenshots

*If applicable*

For admins who want to install a sublanguage, they'll have to do it through a roundabout way, running `kalite shell` and then `call_command(`retrievecontentpack`...)

MCGallaspy:
## Summary

Fixes #4935:
```text
Allow positional arguments with dashes inside

It choked on valid commands like:
`kalite manage retrievecontentpack local es-ES foo.zip`
Because of the dash in the middle of a positional argument.
So only match if there is preceding whitespace.

Also remove redundant non-capturing group -- negative lookaheads
don't capture anyway.
```

Remove reference to nonexistent CHANGELOG.rst in our PR template, too.

## TODO

- [x] Has documentation been written/updated? [[See the warning message, comments]]

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,6 @@ If not all TODOs are marked, this PR is considered WIP (work in progress)
 - [ ] Have **tests** been written for the new code? If you're fixing a bug, write a regression test (or have a really good reason for not writing one... and I mean **really** good!)
 - [ ] Has documentation been written/updated?
 - [ ] New dependencies (if any) added to requirements file
-- [ ] Add an entry to CHANGELOG.rst
 
 ## Reviewer guidance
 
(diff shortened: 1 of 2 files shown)


aronasorman:
Can confirm that this command now works as expected:

```
kalite manage retrievecontentpack local pt-BR ~/src/github.com/fle-internal/content-pack-maker/out/pt-BR.zip
```

aronasorman:
Will wait 'til the test is fixed, or otherwise merged by someone else.

PR #4940, opened 2016-03-01
Remove that custom pt-BR -> pt mapping.

aronasorman:
## Summary

For 0.15 we made a hack that manually maps `pt-BR` to `pt`. This hack is not needed anymore, in fact it actively hinders `pt-BR` from working properly by causing annotation to happen on a nonexistent DB.

### DIFF
--- a/kalite/topic_tools/content_models.py
+++ b/kalite/topic_tools/content_models.py
@@ -104,12 +104,7 @@ def set_database(function):
     """
 
     def wrapper(*args, **kwargs):
-        # Hardcode the Brazilian Portuguese mapping that only the central server knows about
-        # TODO(jamalex): BURN IT ALL DOWN!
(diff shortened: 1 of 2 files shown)


aronasorman:
tada:

screen shot 2016-02-29 at 17 09 53

kalite/topic_tools/content_models.py
```diff
@@ -108,9 +108,6 @@ def wrapper(*args, **kwargs):
         # TODO(jamalex): BURN IT ALL DOWN!
```
rtibbles: You missed a spot, @aronasorman.

aronasorman:
Spot begone!

rtibbles:
:boom:

aronasorman:
Merging this now, so we can build an sdist for the Google Brazil folks!

Issue #4937, opened 2016-02-29
Videos downloaded in a non-english language pack aren't showing up in the topic tree.


aronasorman:
## Summary

When I download a video in a non-english lang pack, it's not showing up in the topic tree.

## Branch or installer method

*Please specify if you are using the current `develop` branch or which installation source you were using when the issue occurred.*

0.16.x

## How to reproduce

1. Download a language pack, activate said language
2. Download a video through the video download interface
3. Scroll through topic tree, trying to find the said video in vain

radinamatic:
And the video topic tree is still not showing localized (this is after installing the latest `es-ES-zip`:


![actualizar_videos](https://cloud.githubusercontent.com/assets/1457929/13405401/76fe968c-df1e-11e5-934a-32837e21d759.png)

rtibbles:
Yeah, I think I see what is wrong here, trying a fix now!

rtibbles:
This is fixed by #4938.

PR #4942, opened 2016-03-02
Prevent TypeErrors getting thrown on partial downloads.

rtibbles:
## Summary

Prevents Javascript TypeErrors from being thrown.

## Reviewer guidance

UI is unaffected, this just ensures against unnecessary and misleading JS errors in the console.

@MCGallaspy this was reported by @radinamatic as part of her follow up on #4937 - that issue is solved, but these residual errors could be misleading in debug reports.

### DIFF
--- a/kalite/updates/static/js/updates/bundle_modules/update_videos.js
+++ b/kalite/updates/static/js/updates/bundle_modules/update_videos.js
@@ -415,10 +415,12 @@ function getSelectedStartedMetadata(data_type) {
 }
 
 function setNodeClass(node, className) {
-    if (node.extraClasses != className) {
-        node.extraClasses = className;
-        if (node.parent !== null) {
-            updateNodeClass(node.parent);
+    if (node !== null) {
+        if (node.extraClasses != className) {
+            node.extraClasses = className;
+            if (node.parent !== null) {
+                updateNodeClass(node.parent);
+            }
         }
     }
 }


kalite/updates/static/js/updates/bundle_modules/update_videos.js
```diff
@@ -415,10 +415,12 @@ function getSelectedStartedMetadata(data_type) {
 }
 
 function setNodeClass(node, className) {
-    if (node.extraClasses != className) {
-        node.extraClasses = className;
-        if (node.parent !== null) {
-            updateNodeClass(node.parent);
+    if (node !== null) {
```
MCGallaspy: Under what conditions is `node === null` true?

rtibbles: Two reasons:

1) the youtube_id for the video being downloaded will not represent a node in the tree, as the tree is lazy loaded, and so not all videos are present.
2) the youtube_id being passed is undefined, and so again the node is null.

This check could happen before we get to this function call, but this seemed the neatest place to check.

MCGallaspy: > This check could happen before we get to this function call, but this seemed the neatest place to check.
 
That's not consistent with `updateNodeClass`, where a similar check is made before calling the function: https://github.com/rtibbles/ka-lite/blob/prevent_type_error/kalite/updates/static/js/updates/bundle_modules/update_videos.js#L421

Same with `recurNodeParent`: https://github.com/rtibbles/ka-lite/blob/prevent_type_error/kalite/updates/static/js/updates/bundle_modules/update_videos.js#L486

Do you feel that affects the future readability of this module? If not, I'll merge. Otherwise, I'd say let's establish the convention that these functions *must* take a non-null node (or conversely add similar null-guards to the other functions which have node parameters).

rtibbles: I don't think it does affect future readability of this code, as the code is badly constructed, poorly written, and non-idiomatic JS.

The best fix here is to :fire: down the code and start from scratch. Maybe we can with @benjaoming's 0.17 reworking.

benjaoming:
I think @rtibbles 's comment seems legit.. and I think we've established that this also fixes the missing "Cancel download" button. #4949

PR #4943, opened 2016-03-02
Migrate to VTT subtitle format

### LINKED ISSUE #4166: Migrate subtitles to VTT format
Migrate subtitles to VTT format

*milestone to be decided*
See #4072 and #4151 for context.

VideoJS version 4.11 that we currently have running on develop is the [last one that supports the](https://github.com/videojs/video.js/issues/1956#issuecomment-85617607) **SRT** format for subtitles. Given that the upgrade to `videojs` version 5 (once it is released) is recommended because of the a11y improvements, we'll have to implement the migration to **VTT** format prior to that.

**Resources**:
http://www.webvtt.org/
https://github.com/captioning/captioning/wiki#convert-a-subrip-file-srt-into-a-webvtt-file-vtt

MCGallaspy:
## Summary

Migrate to VTT subtitle format. They're included in the content packs, so we just have to serve them (and tweak a few things). Upgrade to the latest video.js as well.

## TODO

## Reviewer guidance

Native text track support seems to be sparse or broken. Look at the commit messages for a description of the issue.

I would like guidance on writing a test for this. One thing that might not be too brittle, assuming we stick to video.js, would be to load a video page and assert that the "CC" button appears. Alternatively, we could use the video.js api to assert that the expected [textTracks](http://docs.videojs.com/docs/api/player.html#MethodstextTracks) are present.

Both would involve installing a content pack on the test server and downloading at least one video. Thoughts on this approach?

## Issues addressed

Fixes #4166.

## Screenshots

Check out the snazzy new interface.

image

### DIFF
--- a/kalite/testing/base_environment.py
+++ b/kalite/testing/base_environment.py
@@ -289,7 +289,7 @@ def _make_video(context):
 
 
 def _teardown_video(context):
-    context.video.delete_instance()
+    delete_instances([context.video.id])
     try:
         os.remove(context._subtitle_file_path)
     except WindowsError as e:
--- a/kalite/testing/base_environment.py
+++ b/kalite/testing/base_environment.py
@@ -3,6 +3,7 @@ environment.py defines setup and teardown behaviors for behave tests.
 The behavior in this file is appropriate for integration tests, and
 could be used to bootstrap other integration tests in our project.
 """
+import json
 import os
 import tempfile
 import shutil
@@ -19,9 +20,11 @@ from django.db import connections
 from django.db.transaction import TransactionManagementError
 from peewee import Using
 
+from kalite.i18n.base import get_srt_path, get_srt_url
 from kalite.testing.base import KALiteTestCase
 from kalite.testing.behave_helpers import login_as_admin, login_as_coach, logout, login_as_learner
-from kalite.topic_tools.content_models import Item, set_database, annotate_content_models
+from kalite.topic_tools.content_models import Item, set_database, annotate_content_models, create, get, \
+    delete_instances
 
 from securesync.models import Zone, Device, DeviceZone
 
@@ -77,6 +80,7 @@ def setup_content_paths(context, db):
             slug="unavail",
             path=context.unavailable_content_path
         )
+
 
 @set_database
 def teardown_content_paths(context, db):
@@ -154,6 +158,9 @@ def setup_local_browser(context):
 def before_scenario(context, scenario):
     database_setup(context)
 
+    if "uses_video_with_subtitles" in context.tags:
+        _make_video(context)
+
     if "registered_device" in context.tags:
         do_fake_registration()
 
@@ -183,6 +190,9 @@ def before_scenario(context, scenario):
 
 
 def after_scenario(context, scenario):
+    if "uses_video_with_subtitles" in context.tags:
+        _teardown_video(context)
+
     if context.logged_in:
         logout(context)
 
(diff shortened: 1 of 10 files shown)


MCGallaspy:
@rtibbles willing to review this one?

rtibbles:
Yes.

rtibbles:
@MCGallaspy one quick way to do a test would be to assert that the subtitle track itself was requested from the server e.g. `http://127.0.0.1:8008/static/srt/en/subtitles/y2-uaPiyoxc.vtt` - I don't know how easy that is to check, but it might be easier than inspecting video-js.

Fine with whichever approach is easier for you - and I also agree it's a good thing to have a test of.

rtibbles:
Also, aside from the test, this is ready to merge.

MCGallaspy:
What makes testing this challenging is that both a video must be available, and a content pack should be installed (for subtitles). Both of these things are potentially very time consuming. Is there a workaround? Even the en-minimal content pack takes a nontrivial amount of time to download & install, and I'm not sure that it includes subtitles.

radinamatic:
> Both of these things are potentially very time consuming

Welcome to my world... :grinning: 
I keep a copy of contenpacks in a shared network folder from where it's accessed by different VMs. It usually works OK, a part from this week when they were re-compiled 3 times.

Time for installing them over and over is lost forever, I'm afraid... :disappointed:

rtibbles:
My suggestion would be to fake the subtitle file, force annotation for that
single file and then test that the file request is made. If in doubt, mock,
mock, mock.

MCGallaspy:
> Welcome to my world... :grinning: 

Yeah, I've accumulated a lot of downtime for the same reason... However my sense is that a lot of prior design choices have locked us in to this model. ¯\\\_(ツ)\_/¯

MCGallaspy:
@richard take a look at the last commit. The problem with the test is that instead of getting the content item I just created (in base_environment), I'm getting the unavailable content warning:

ss3

Any idea again how to *actually* get the content item I want? Do I need to make it a child of the "root" Item (the one with path "khan")? Should I create such an Item, or will it always exist?

kalite/testing/base_environment.py
```diff
@@ -245,3 +254,42 @@ def do_fake_registration():
...
+    lang_code = "en"
+    youtube_id = "my_cool_id"
+    item_dict = {
+        "title": "Subtitled Video",
+        "description": "A video with subtitles",
+        "available": True,
+        "kind": "Video",
+        "id": "video_with_subtitles",
+        "slug": "video_with_subtitles",
+        "path": "khan/video_with_subtitles",
```
rtibbles: Yes, you would need to add a 'parent' field here that referenced the root node. It is the only node with the parent key as null.

There is currently no 'get' method in content_models.py but you could make one following the patterns here: https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/topic_tools/content_models.py#L508 - obviously, no need for parse_item_data as we want the `Item` object itself.

MCGallaspy: So follow up -- is there any expectation that such an Item will exist on the test server? Or would it only exist if a content pack is installed?

rtibbles: Yep, `make assets` gets run before tests (https://github.com/learningequality/ka-lite/blob/0.16.x/circle.yml#L21), meaning that this line (https://github.com/learningequality/ka-lite/blob/0.16.x/Makefile#L100) is executed, and a minimal language pack is available.

MCGallaspy:
See the last commit:

`Add subtitle availability test`
```
Making this test *possible* was actually quite challenging.
Changes:
1. Added the object `window._kalite_debug` to our base js bundle,
   so that I could track the state of the video player in tests
   in order to avoid race conditions.
2. In video/views.js, publish the state using `_kalite.debug` and
   avoid a divide-by-zero error.
3. In base_environment, create and teardown a video for testing
   subtitle availability. It was a bit tricky to figure out
   exactly *how* to do this -- I had to inspect a real video to
   figure out the expected format of extra_fields, and it turns
   out that a video must also have a parent object in order
   to be able to navigate to it/load it properly.
4. Add a new function to `behave_helpers` specific to the learn
   page. This sort of gets at the idea of page objects (see
   the Selenium docs), wherein one defines an API for each
   page in the app being tested.
5. Created a couple functions in content_models in order to
   create the video mentioned above.
```

rtibbles:
@MCGallaspy can you rebase against latest 0.16.x to include the test fix I just merged?

kalite/topic_tools/annotate.py
```diff
@@ -95,13 +91,6 @@ def update_content_availability(content_list, language="en", channel="khan"):
                         "stream_type": "{kind}/{format}".format(kind=content.get("kind").lower(), format=format),
                         "thumbnail": thumbnail,
                     }
-            elif django_settings.BACKUP_VIDEO_SOURCE:
```
rtibbles: Why is this deleted? This is needed for the demo server.

rtibbles:
@MCGallaspy this is looking good - great work on getting a test to run on this. If you can just reinstate the codeblock that I flagged above, this will be ready to merge.

kalite/topic_tools/annotate.py
MCGallaspy: Didn't appear to be used anywhere, but I'll put it back.

MCGallaspy: Actually there's an error in this block of code. See my commit message:

```
That setting is referenced nowhere else, and furthermore the block
of code uses an undefined name "dubbed_id". So if executed, it would
result in an error.

It's not obvious to me what value *should* go there, hence removing it.
```

rtibbles: Ah, right. I'll merge this and fix it up in another PR.

PR #4970, opened 2016-03-08
Update points if there is server-client discrepancy

### LINKED ISSUE #4643: Points seen on dropdown is added twice as we earn them
This is seen on develop branch.

Steps to reproduce.
1) Login as learner. 
2) See start any excercise and before answering note the point you have.
3) Solve a question. Note your points again. It will be equal to previous points plus twice the points earned in dropdown.

MCGallaspy:
## Summary

Fixes error in points display. For exercise log sync events, only update the points display if there is a discrepancy between the points reported by the server and the points known to the client.

## TODO

Here's my really good reason. Testing this using selenium would take the following steps:

1. Create new student account and set the points to some appropriate initial value.
2. Navigate to an exercise.
3. Interact with the exercise to gain points.
4. Make assertion about observed points in user menu.

This is challenging because we don't provide a testing api for interacting with content nor a reliable way to determine the "state" of the page in order to avoid race conditions (e.g. polling the `#points` element too soon or not waiting long enough after a page is loaded). I also wrote a [similar test that was removed](https://github.com/learningequality/ka-lite/commit/bc13993fa319eec9084773f71cc81413e8c2a3d7#diff-5460c9ea5f5d69ca7164a26b56b092a9L407). I think I can write a better, less error-prone test now in about a day. I'm on the fence as to whether that's a good use of time right now. Dear reviewer, what do you think?

## Issues addressed

Fixes #4643

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -677,10 +677,22 @@ var ExerciseWrapperBaseView = BaseView.extend({
     },
 
     update_total_points: function(data) {
-        // update the top-right point display, now that we've saved the points successfully
-        if (this.log_model.has("points")) {
-            window.statusModel.update_total_points(this.log_model.get("points") - this.status_points);
-            this.status_points = this.log_model.get("points");
+        /*
+            Update the top-right point display, now that we've saved the points successfully.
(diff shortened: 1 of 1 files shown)


rtibbles:
I think it is probably not a great use of our time. Given that we have already accepted that we need manual testing, I would add this to the list of things that an exercise tester should check for.

PR #4976, opened 2016-03-09
Ensure kalite is initialized in base mgmt command

MCGallaspy:
## Summary

```text
The class `UpdatesCommand` is the base class for
the `retrievecontentpack` command. Since this could
potentially be run *before* setup, it produces an error
when the database is accessed by the `UpdateProgressLog`
Model methods -- the database wouldn't exist!

`initialize_kalite` ensures that no such error is raised
by creating a database.
```
## Issues addressed

Addresses an issue with the windows installer -- it will fail to unpack the content pack at install time because of this.

### DIFF
--- a/kalite/updates/management/commands/classes.py
+++ b/kalite/updates/management/commands/classes.py
@@ -3,12 +3,16 @@
 from datetime import datetime
 from optparse import make_option
 
-from django.conf import settings; logging = settings.LOG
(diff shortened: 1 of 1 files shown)


MCGallaspy:
I'm told the correct solution is to run `retrievecontentpack --foreground`.

PR #4978, opened 2016-03-09
Do one-pass finds in Makefile

MCGallaspy:
## Summary

Simple: Do one-pass finds in Makefile. Otherwise it takes multiple passes, which can be much slower.


MCGallaspy:
No rush on this -- more of a suggestion.

PR #4992, opened 2016-03-12
Update manpage for 0.16

benjaoming:
Wouldn't expect any further changes to the `kalite` command, so let's get the updated manpage in.


MCGallaspy:
What's changed about this binary file?

MCGallaspy:
No diff at all? I'm not able to easily view this on Windows. Can you explain how the man page is installed?

benjaoming:
The manpage binary is built using a command that doesn't work on Windows. That was the reason for including it directly. It reflects any changes made in the usage of `kalite`.

PR #4996, opened 2016-03-15
Makes video.js reload with flash only when the player errors.

rtibbles:
## Summary

Makes video.js reload with flash only when the player errors.
Also independently imports vtt.js and uses the videojs no-vtt to prevent unnecessary cdn loads.

## TODO

## Reviewer guidance

## Issues addressed

Independently imports vtt.js, uses videojs-novtt.


kalite/distributed/static/js/distributed/video/views.js
```diff
@@ -68,12 +77,28 @@ var VideoPlayerView = ContentBaseView.extend({
...
             this.player = window.player = _V_(player_id, video_player_options);
+            this.player.ready(function() {
+                videoPlayer = this;
+
+                videoPlayer.on('error', function() { // error event listener
+                    // dispose the old player and its HTML
+                    videoPlayer.dispose();
+
+                    self.flash_only = true;
+                    self.render();
```
MCGallaspy: Wouldn't you have to set `this._loaded = false;` as well?

rtibbles: Yes, just fixed..

kalite/distributed/static/js/distributed/video/views.js
```diff
@@ -68,12 +77,29 @@ var VideoPlayerView = ContentBaseView.extend({
                 "playbackRates": [0.5, 1, 1.25, 1.5, 2],
                 "html5": {
                     nativeTextTracks: false
+                },
+                "techOrder": this.flash_only ? ["flash"] : ["html5", "flash"],
+                flash: {
+                    swf: window.sessionModel.get("STATIC_URL") + "distributed/js/distributed/video/video-js.swf"
```
MCGallaspy: The url is also wrong -- remove the leading `distributed`

MCGallaspy:
1. `video-js.swf` is fetched locally (and not from a CDN) if you fix the url. :+1: 
2. VTTs don't work if the `vtt.js` file isn't fetched from a CDN, so somehow it's not the right fix.
3. On Windows XP+FF, the `loadedmetadata` event is never triggered at all -- I started trying to figure out how we could initialize the player without listening for that event, as I suspect we could get the correct flash fallback behavior if we do that.

Do you want to continue with this PR and take over the issue? Or should I? If the latter, input on disentangling player initialization from `loadedmetadata` event would be appreciated.

MCGallaspy:
> loadedmetadata event is never triggered at all

The point being that `initialize_player` is never executed.

MCGallaspy:
BTW, if we can't resolve the issues on XP then we need to document a workaround (like using chrome browser).

rtibbles:
I won't be able to fix this today. I would recommend closing this pr and
opening a new one with the right changes.

PR #4997, opened 2016-03-15
Self-host all video.js dependencies

### LINKED ISSUE #4993: Videos/subtitles not displayed in some browsers (including Firefox)
## Summary

Confirmed on both Beta2 in Ubuntu & build 176 on Windows XP. Windows 8.1 displays video but do not load subtitles. IE11 also has problems with subs, but does display video. Chrome does everything :+1: 

@MCGallaspy 
## Branch or installer method

Beta2 in Ubuntu & build 176 on Windows 

Win8 console error.

```
TypeError: w.default.WebVTT.StringDecoder is not a function
require<[581]</</</<[127]</x()  bundle_video.js:16
require<[581]</</</<[127]</O/</<()  bundle_video.js:16

```
## Screenshots

Win XP
ie8 - winxp ready for ka lite running - oracle vm virtualbox_258

Win8.1
ie11 - win8 1 ready for ka lite running - oracle vm virtualbox_290

Ubuntu 14.04
ubuntu 14 04 ready for ka lite running - oracle vm virtualbox_282

MCGallaspy:
## Summary

...instead of getting them from a CDN. Replaces  #4996 and addresses the first two points in https://github.com/learningequality/ka-lite/pull/4996#issuecomment-197000885

However, regarding point 3
>On Windows XP+FF, the loadedmetadata event is never triggered at all -- I started trying to figure out how we could initialize the player without listening for that event, as I suspect we could get the correct flash fallback behavior if we do that.

Something is still not quite right.

## TODO

Setting aside for now, but will revisit point 3 shortly.

## Issues addressed

Will close #4993 when point 3 is addressed, either by finding the right fix or documenting a workaround.


MCGallaspy:
Problem seems to be that on windows `node_modules` is now flat (so the `require` stmt resolves correctly) whereas on the build server it's nested. This must be a recent-ish change on Windows, as it wasn't always the case.

kalite/i18n/base.py
```diff
@@ -89,7 +89,7 @@ def get_srt_path(lang_code=None, youtube_id=None):
     if lang_code:
         srt_path = os.path.join(srt_path, lcode_to_django_dir(lang_code), "subtitles")
     if youtube_id:
-        srt_path = os.path.join(srt_path, youtube_id + ".srt")
+        srt_path = os.path.join(srt_path, youtube_id + ".vtt")
```
MCGallaspy: Not so certain about these changes. Somehow this slipped by me in previous PRs, but it's used for instance to build correct paths/urls in tests. Should I create new functions, instead of modifying these @aronasorman?

kalite/distributed/static/js/distributed/video/views.js
```diff
@@ -68,12 +78,38 @@ var VideoPlayerView = ContentBaseView.extend({
...
+                "techOrder": this.flash_only ? ["flash"] : ["html5", "flash"],
+                flash: {
+                    swf: window.sessionModel.get("STATIC_URL") + "js/distributed/video/video-js.swf"
                 }
             };
             if( this.data_model.get("content_urls").thumbnail ) {
                 video_player_options['poster'] = this.data_model.get("content_urls").thumbnail;
             }
             this.player = window.player = _V_(player_id, video_player_options);
+            this.player.ready(function() {
```
rtibbles: May as well delete this codeblock if unused. Without the callback inside, it is pointless.

kalite/i18n/base.py
rtibbles: Might be worth changing the names to 'get_vtt_path' to be more semantic?

MCGallaspy: This snippet from the docstring gives me pause:

```
Both central and distributed servers must make these available
    at a web-accessible location.

    Now, they share that location, which was published in 0.10.2, and so cannot be changed
    (at least, not from the central-server side)
```

kalite/distributed/static/js/distributed/video/views.js
MCGallaspy: Yeah, still haven't found a good solution. So removed.

kalite/i18n/base.py
MCGallaspy: Changed, but let's get @aronasorman or @jamalex to weigh in.

MCGallaspy:
Rebased, good to go.

rtibbles:
Merge conflict.

MCGallaspy:
Rebased again.

kalite/i18n/base.py
aronasorman: @MCGallaspy we purged the i18n functions from the central server. You can change these at will now.

rtibbles: Sounds like a green to go.

PR #4998, opened 2016-03-16
Languagepack progress bar fix

### LINKED ISSUE #4955: Language pack donwload progress bar...
## Summary

...is still missing. I can't find any issue regarding this, and it's a definite release blocker!

Delete this if I'm blind... :stuck_out_tongue_closed_eyes:

MCGallaspy:
## Summary

Fixes #4955. But while looking in to this I went down a dead end with the updates management command base classes: https://github.com/learningequality/ka-lite/blob/0.16.x/kalite/updates/management/commands/classes.py#L41

As far as I can tell, it's not possible to pass kwargs to a command's constructor using the `call_command` function (or possibly at all). For instance, it doesn't seem possible to set the `process_name`, which was the first thing I attempted. Instead changes to the back-end (like in this case changing the management command that's being called) require changes to the front-end. Is this worth a refactor?

## TODO

Test incoming.

## Issues addressed

Fixes #4955.

### DIFF
--- a/kalite/updates/static/css/updates/update_languages.css
+++ b/kalite/updates/static/css/updates/update_languages.css
@@ -47,7 +47,7 @@
     font-weight: bold;
 }
 
-#languagepackdownload-progressbar {
+#retrievecontentpack-progressbar {
     margin-top: 25px;
     margin-bottom: 25px;
 }
--- a/kalite/updates/static/js/updates/bundle_modules/update_languages.js
+++ b/kalite/updates/static/js/updates/bundle_modules/update_languages.js
@@ -229,13 +229,13 @@ function start_languagepack_download(lang_code) {
     messages.clear_messages();  // get rid of any lingering messages before starting download
     $("#get-language-button").prop("disabled", true);
     downloading = true;
-    // tell server to start languagepackdownload job
+    // tell server to start retrievecontentpack job
     api.doRequest(
         Urls.start_languagepack_download(),
         { lang: lang_code }
     ).success(function(progress, status, req) {
         base.updatesStart(
-            "languagepackdownload",
+            "retrievecontentpack",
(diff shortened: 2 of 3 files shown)


aronasorman:
Tested. Working. Merging.

aronasorman:
Actually, won't merge yet. Can you also remove this notification, since a restart is no longer needed:

screen shot 2016-03-17 at 11 18 18

aronasorman:
Also, are you planning to write tests for this?

MCGallaspy:
>Can you also remove this notification, since a restart is no longer needed:
Yes

For tests, I'm hesitant -- the language pack list depends AFAICT on external variables, namely the central server. So it's not super isolated. The rest of the functionality depends on the list. I would like to assert that the download bar appears or something, but if it breaks when the central server is offline or when factors outside of this app change, it's brittle. Your thoughts?

aronasorman:
> For tests, I'm hesitant -- the language pack list depends AFAICT on external variables, namely the central server.

I concur. If you can think of a way to isolate this functionality with a unit test, then try mocking or using vcr.py. Otherwise, merge at will.

66eli77:
I just pulled and make sure the changes are present. But I didn't see the progress bar when downloading language pack (Spain). Did I miss anything? @MCGallaspy @aronasorman

MCGallaspy:
@66eli77 maybe so... worked for me, seems to have worked for @aronasorman. Perhaps you didn't rebuild the javascript bundles?

66eli77:
could it be that I started a language pack download process before the code change, and I stop the server and git pull the changes, and do the download again?

radinamatic:
I just pulled latest 0.16 and can confirm that the progress bar is present. @66eli77 try rebuilding as @MCGallaspy mentioned.

PR #5009, opened 2016-03-17
Add extra info about windows task-tray program -- 0.15.x

MCGallaspy:
## Summary

Companion to #5007. Rebasing seems to have picked up some extra commits as well.


aronasorman:
I'm assuming that we won't be needing a new release after merging this.

PR #5010, opened 2016-03-17
Open docs link in a separate window (accessible)

### LINKED ISSUE #4977: Improving UX of KA Lite --> User Manual 0.16
## Summary

There is no button to direct back to the ka lite application from the user manual. A user could just use the browser's back button, but the change in look and feel is abrupt.

user manual

Possible solution is to add an html tag in this link that allows the user manual to open in a new tab 
(```<a target="_blank"></a>```)
Further improvement to the UX of this solution is to add an alert modal to let the user know they will be directed away from the ka lite app, so that they can confirm or cancel that request.

## Branch or installer method

 - Installer: Windows
 - Version: 0.16

## How to reproduce

1. Be logged in as admin
2. Click User Manual tab

radinamatic:
## Summary

Fixes #4977. 

Tab link now has both a title & target attribute with the appropriate icon and accessible info for the screenreader users. I renamed it to `DOCS` since `USER MANUAL` seemed too big for me visually, although I have considered `HELP`too...

@MCGallaspy 

## Screenshots

docs-new-window

### DIFF
--- a/kalite/distributed/static/js/distributed/user/hbtemplates/navigation.handlebars
+++ b/kalite/distributed/static/js/distributed/user/hbtemplates/navigation.handlebars
@@ -13,9 +13,11 @@
 {{#if docs_exist}}
   {{#if is_admin}}
     <li role="menuitem" id="nav_documentation">
-      <a role="menuitem" tabindex="-1" id="documentation_a" href="/static/html/usermanual/userman_main.html">
-        {{_ "User Manual" }}
-       </a>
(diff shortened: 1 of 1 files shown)


MCGallaspy:
The test is flaky. The fix is good, so merging.

radinamatic:
No idea where the test came from, wasn't me - scout's honor!
(I figured now was not the right time for me to try learning to write proper tests... :wink:)

MCGallaspy:
Yeah, by flaky test I mean something which fails randomly. Has nothing to do with your changes in this case.

PR #5013, opened 2016-03-17
Remove unnecessary assertion in test

MCGallaspy:
## Summary

Seems to have been added by a revert-revert, however
it seems to be the wrong assertion to make in this case.
Causes flakiness if the modal is dismissed *first*.

### DIFF
--- a/kalite/distributed/features/steps/superuser_create.py
+++ b/kalite/distributed/features/steps/superuser_create.py
@@ -85,12 +85,6 @@ def step_impl(context):
 
 @then("the modal will dismiss")
 def impl(context):
-    for id_ in (PASSWORD_CONFIRM_ID, PASSWORD_ID, USERNAME_ID):
-        try:
-            is_border_red(context, id_)
-            raise RedBorderException("The border should not be red for the element with id #{0}".format(id_))
-        except AssertionError:
-            pass  # The border should _not_ be red, so the above function _should_ raise an exception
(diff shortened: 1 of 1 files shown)


66eli77:
it should not cause any problem if modal is dismissed first because https://github.com/learningequality/ka-lite/blob/master/kalite/distributed/static/js/distributed/user/views.js#L101
just hide the model, means the test can still find the elements in the DOM and evaluate the color. and if the color is red, we should raise an exception.
I just quickly ran the test and didn't encounter any issue.
Did you actually encounter the flakiness issue? @MCGallaspy

MCGallaspy:
>Did you actually encounter the flakiness issue?

Yes, I did.

> just hide the model, means the test can still find the elements in the DOM and evaluate the color. and if the color is red, we should raise an exception.

Just because you can do it, should we? Is it a useful assertion to make, if the modal will *also* just be hidden anyway?

66eli77:
In theory, there should not be issue caused by this assertion.
the assertion just make sure when the modal isn't hidden, it's not because of wrong password or username.

66eli77:
ok, I find the reason for the flaky issue
https://github.com/learningequality/ka-lite/blob/master/kalite/distributed/static/js/distributed/bundle_modules/homepage.js#L24
the modal does get removed from the DOM once superuser is successfully created

PR #5016, opened 2016-03-17
Only download videos not already downloaded

### LINKED ISSUE #4990: Mismatch between number of videos said will be downloaded and number then queued
## Summary

I was upgrading an installation with all the 0.15 videos to 0.16, and wanting to download just the new videos. At first, the download button claimed only 411 new videos to download, but once the download began, it showed 7000+ in the queue.

It seems to be queuing all selected videos, rather than just the ones that are new?

## Branch or installer method

 - Branch: 0.16.x

## How to reproduce

1. Run 0.16.x branch from source, with content directory already containing many videos
1. Go to video download page, and select all topics, and start download.

## Screenshots

image

image

MCGallaspy:
## Summary

Filters the requested ids to download. I think this could potentially increase the runtime of this endpoint significantly, since each youtube_id hits the database -- but maybe it could be improved by a big constant factor if all the content items are fetched at once? Alternatively, if the requests are batched on the front-end. Thoughts?

## TODO

- [ ] ~~Have **tests** been written for the new code? If you're fixing a bug, write a regression test (or have a really good reason for not writing one... and I mean **really** good!)~~

## Issues addressed

Fixes #4990.

### DIFF
--- a/kalite/updates/api_views.py
+++ b/kalite/updates/api_views.py
@@ -128,7 +128,7 @@ def start_video_download(request):
 
     lang = json.loads(request.body or "{}").get("lang", "en")
 
-    youtube_ids = get_download_youtube_ids(paths, language=lang)
+    youtube_ids = get_download_youtube_ids(paths, language=lang, downloaded=False)
 
     queue = VideoQueue()
 
@@ -150,7 +150,7 @@ def delete_videos(request):
 
     lang = json.loads(request.body or "{}").get("lang", "en")
 
-    youtube_ids = get_download_youtube_ids(paths, language=lang)
+    youtube_ids = get_download_youtube_ids(paths, language=lang, downloaded=True)
 
     num_deleted = 0
 
--- a/kalite/topic_tools/content_models.py
+++ b/kalite/topic_tools/content_models.py
@@ -12,9 +12,9 @@ the dictionary, or with many of the fields collapsed into an 'extra_fields' key.
 All functions return the model data as a dictionary, in order to prevent external functions from having to know
 implementation details about the model class used in this module.
 """
-import json
-
 import itertools
+import json
+import os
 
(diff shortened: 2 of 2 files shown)


kalite/updates/api_views.py
```diff
@@ -129,10 +130,12 @@ def start_video_download(request):
     lang = json.loads(request.body or "{}").get("lang", "en")
 
     youtube_ids = get_download_youtube_ids(paths, language=lang)
-
+    ids_to_queue = {id_: title for (id_, title) in youtube_ids.iteritems()
```
rtibbles: You are right, this is not the right fix. See the function above that retrieves the youtube_ids in the first place. Let's modify that instead. It is only used to retrieve ids for downloads, so you can add the filter there.

I would also suggest filtering on 'files_complete === 1', so that we fetch additional videos.

As noted by @radinamatic in a recent issue, we annotate videos as available if we have the English video + relevant language subtitle, so being 'available' is not a guarantee that the video has been downloaded.

MCGallaspy: How can I ensure that a file isn't downloaded *again* if it's already on disk? Is this information contained somehow in `extra_fields`? Or can I infer it from `files_complete`? What is the significance of that field?

rtibbles: files_complete shows how many files have been downloaded. Each video
currently only has one associated file.

For a water tight check, ensuring that files_complete is less than
total_files would only give you undownloaded content.

But, do this in the method above to reduce database queries rather than
filtering after the fact.

MCGallaspy:
@rtibbles in the interest of time, I'm compromising my test policy.

rtibbles:
I still don't think checking the file system is the right fix. We should, perhaps, check that at download, but the point of flagging these things in the database is so that we don't have to constantly check the file system.

I have implemented a fix for this in my fix for the download availability issue for different languages, so will close this when that is opened.

rtibbles:
Closing in preference to #5045.

Issue #4980, opened 2016-03-09
Content not translated on student Progress Report


rtibbles:
## Summary

Content titles and descriptions are not translated when viewing them on the progress report.

## Branch or installer method

 - Installer: Debian
 - Version: 0.16.x beta1

## How to reproduce

- Login as learner with some learning history.
- Go to progress report.
- Switch to e.g. Spanish.

## Screenshots

![image](https://cloud.githubusercontent.com/assets/1680573/13654158/470728c4-e60b-11e5-82e6-01293d374252.png)

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

Very confusing for non-English speaking learners.

radinamatic:
Thanks, noted that on Windows too in #4951, never got around to file the issue.

66eli77:
confirm the issue with French language pack, I'm working on this issue

rtibbles:
Fixed in #5018

PR #5018, opened 2016-03-18
Translation on Progress Report

66eli77:
## Summary

pass in the language keyward argument when retrieve content nodes on progress report so that language pack setting will take effect

## TODO

## Issues addressed

#4980

### DIFF
--- a/kalite/coachreports/models.py
+++ b/kalite/coachreports/models.py
@@ -64,7 +64,7 @@ class PlaylistProgress(PlaylistProgressParent):
         video_ids = list(set([vid_log["video_id"] for vid_log in user_vid_logs]))
 
         # Build a list of playlists for which the user has at least one data point
-        user_playlists = get_content_parents(ids=exercise_ids+video_ids)
+        user_playlists = get_content_parents(ids=exercise_ids+video_ids, language=language)
 
         # Store stats for each playlist
         user_progress = list()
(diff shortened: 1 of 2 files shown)


rtibbles:
This is the right fix. Test?

I would suggest doing an API only test, with the request language set - and then use mock.patch to check that language argument gets passed into the `get_content_parents` and `get_topic_nodes` functions.

66eli77:
do you aware that mock.patch not working on two or more imports deep calls @rtibbles

rtibbles:
Really? Seems to work fine for me - e.g. https://github.com/rtibbles/kolibri/blob/frontend_asset_loading/kolibri/utils/test/test_webpack_utils.py#L100

66eli77:
for some reason, this return false
```
@patch('kalite.topic_tools.content_models.get_content_parents')
    def test_language_setting(self, get_content_parents):
        base_url = self.reverse('api_dispatch_list', kwargs={'resource_name': 'playlist_progress'})
        url = base_url + '?' + urllib.urlencode({'user_id': self.student.id, '_': '1458605610477'})
        resp = self.client.get(url)
       return get_content_parents.called
```
but this return True
```
@patch('kalite.coachreports.models.PlaylistProgress.user_progress')
def test_language_setting(self, user_progress):
        base_url = self.reverse('api_dispatch_list', kwargs={'resource_name': 'playlist_progress'})
        url = base_url + '?' + urllib.urlencode({'user_id': self.student.id, '_': '1458605610477'})
        resp = self.client.get(url)
       return user_progress.called
```

do you have any insights about this?

66eli77:
nvm, I figured it out, it should be @patch('kalite.coachreports.models.get_content_parents')

rtibbles:
Merging as is in the interests of time. Pity we couldn't properly mock the language in the request though!

PR #5033, opened 2016-03-23
Pinned node module dependencies in package.json to specific versions.

jamalex:
Our node module dependencies were specified with "floating" version numbers (starting with `~` or `^`), meaning they would be automatically upgraded to newer version even without our codebase changing. In theory, this could cause problems, and in practice, we saw the first of these problems arise with `bootstrap-datepicker` not behaving properly on the coach report pages due to a background version change.

This PR pins all modules to their latest versions as of today. These are the versions we've been building and testing with, and hence are the safest versions to pin.


rtibbles:
Outrageous! Changing so many dependencies this close to release!

rtibbles:
Merging.

Issue #4989, opened 2016-03-11
Syncing error - {'exercise_id': [u'This field cannot be blank.']}


radinamatic:
## Summary

Trying to sync with central server gives an error

## Branch or installer method

0.16

## Traceback or relevant snippet from server.log
Nothing in Console or the server.log, errors on sync page:
```
Checking purgatory for unsaved models...
Purgatory is model-free! Congrats!
Initiating SyncSession...
Syncing models...
Exception uploading models (in api_client): ValidationError, , ()
Uploaded : 0 (0 failed, 1 error(s))
Downloaded : 0 (0 failed, 0 error(s))
Closing session... (Total uploaded: 0, Total downloaded: 0, Total errors: 1)
Checking purgatory once more, to try saving any unsaved models...
Purgatory is model-free! Congrats!
Upload error: {'exercise_id': [u'This field cannot be blank.']}
Completed with 1 errors.
```

## How to reproduce

1. Do some exercises/videos as Learner
1. Login as Admin and try to sync with central server

## Screenshots

![local_device_syncing](https://cloud.githubusercontent.com/assets/1457929/13706487/92351cac-e7a5-11e5-8f12-c1c50cfbebb4.png)

jamalex:
Can you check whether an ExerciseLog or AttemptLog with empty exercise_id
exists locally. That shouldn't be the case, but would explain this.

Also, stuff involving the central server isn't quite ready to test; we're
still working on integrating the latest to there.

rtibbles:
Fixed in #5039

PR #5039, opened 2016-03-24
Ensure attempt logs are given exercise_id, and inserted in right order.

### LINKED ISSUE #5017: Red crosses for incorrect exercises will vanish if the user refresh the page.
## Summary

This is in a `Functional Testing Checklist` of every installers.
`Do some exercise items incorrectly, refresh the page and make sure both incorrect and correct points (green check marks and red crosses are still there).`
When testing this the red crosses for incorrect exercises, will vanish if the learner refresh the page.

I tested this in OS X Yosemite and using Google Chrome browser.

## Branch or installer method

 - Version: 0.16.x

or:

 - Branch: 0.16.x

## How to reproduce

1. Login as learner
1. Do some exercise items incorrectly.
1. Refresh the page.  

## Screenshots
screen shot 2016-03-18 at 2 51 24 pm
Taking exercises with 3 incorrect answers.

screen shot 2016-03-18 at 2 56 36 pm
Red crosses will vanish after a refresh.

jamalex:
## Summary

Due to some refactoring, we were no longer setting `exercise_id` when saving an `AttemptLog`. This means there was no way of know what exercise it was for, which was messing up coach reports, syncing (it was a required field, and hence caused failed validation), and exercises (refreshing the page would make attempts disappear).

Minimal changes here, but they fix all of these issues.

## Issues addressed

Fixes #5017 and #4989.

### DIFF
--- a/kalite/distributed/static/js/distributed/exercises/views.js
+++ b/kalite/distributed/static/js/distributed/exercises/views.js
@@ -550,10 +550,10 @@ var ExerciseWrapperBaseView = BaseView.extend({
         }
 
         var defaults = {
-            exercise_id: this.options.exercise_id,
+            exercise_id: this.log_model.get("exercise_id"),
             user: window.statusModel.get("user_uri"),
-            context_type: this.options.context_type || "",
-            context_id: this.options.context_id || "",
+            context_type: "exercise",
+            context_id: "",
             language: "", // TODO(jamalex): get the current exercise language
             version: window.statusModel.get("version"),
             seed: seed,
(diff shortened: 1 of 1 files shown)


kalite/distributed/static/js/distributed/exercises/views.js
```diff
@@ -564,7 +564,7 @@ var ExerciseWrapperBaseView = BaseView.extend({
 
         this.current_attempt_log = new Models.AttemptLogModel(data);
 
-        this.attempt_collection.add(this.current_attempt_log);
+        this.attempt_collection.unshift(this.current_attempt_log);
```
jamalex: Attempt marks (X and checks) were being added to the lefthand side, but then showing back up on the right when refreshed. This fixes that.

rtibbles:
Looks like the right fix to me.

PR #5045, opened 2016-03-25
Availability

### LINKED ISSUE #4990: Mismatch between number of videos said will be downloaded and number then queued

### LINKED ISSUE #4994: English videos annotated as dubbed in other languages
## Summary

Downloaded a few English videos, but after installing a Spanish language pack, I saw that the same videos appear as already downloaded when I switched the default language to Spanish. Once a deleted 2 of them and re-downloaded, they were correct, dubbed version. The same thing repeated when I re-tested with subsequent French language pack - videos appear as downloaded, when played it is clear that it's the English version, so one needs to go back, delete and re-download to get the dubbed version.     

## Branch or installer method

Both Beta2 in Ubuntu & build 176 on Windows. 

```
No visible error in the console and nothing of consequence in the server.log.
```

## How to reproduce

1. Download a couple of English videos.
1. Install language pack & switch the default language.
1. Check the Manange > Videos page and you'll see the same videos (test the play to confirm they are in English).
1. Delete and re-download, and you'll get the dubbed version.

rtibbles:
## Summary

Properly stamps availability flags, files_complete etc. on different languages, and ensures that the subtitle fallback annotation works properly.

Also filters retrieval of youtube_ids for downloading and deletion by whether there are any files to download/delete.

Finally, prevents overwriting/download of videos that already exist in the content dir.

## TODO

## Issues addressed

Fixes #4994, fixes #4990, and supersedes #5016.

### DIFF
--- a/kalite/main/tests/base.py
+++ b/kalite/main/tests/base.py
@@ -11,7 +11,7 @@ from django.core.cache.backends.filebased import FileBasedCache
 from django.core.cache.backends.locmem import LocMemCache
 
 from kalite.testing.base import KALiteTestCase
-from kalite.topic_tools.content_models import get_random_content
+from kalite.topic_tools.content_models import get_random_content, update_item
 from securesync.models import Device
 
 
@@ -56,5 +56,6 @@ class MainTestCase(KALiteTestCase):
         fake_content_file = os.path.join(settings.CONTENT_ROOT, "%s.mp4" % youtube_id)
         with open(fake_content_file, "w") as fh:
             fh.write("")
+        update_item(update={"files_complete": 1, "available": True, "size_on_disk": 12}, path=content["path"])
         self.assertTrue(os.path.exists(fake_content_file), "Make sure the content file was created, youtube_id='%s'." % youtube_id)
         return (fake_content_file, content["id"], youtube_id, path)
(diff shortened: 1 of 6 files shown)


rtibbles:
Genuine test failure. Fixing.

rtibbles:
Issue here was that we were updating every Video node on annotation, not just those that needed. As @jamalex has previously noted, this can take 45 minutes for every video node.

MCGallaspy:
`killall --wait sc` this command timed out... odd, but otherwise the build passes.

rtibbles:
Also this:
```
FAIL: test_delete_existing_video_file (kalite.updates.tests.api_tests.TestAdminApiCalls)
  File "/home/ubuntu/ka-lite/kalite/updates/tests/api_tests.py", line 77, in test_delete_existing_video_file
    self.assertFalse(os.path.exists(self.fake_video_file), "Video file should not exist on disk.")
AssertionError: True is not false : Video file should not exist on disk.

```
Can't replicate locally.

rtibbles:
@MCGallaspy Tests pass!

PR #5058, opened 2016-03-28
Revert "Return no data if no data."

aronasorman:
Fixes the tests.


aronasorman:
@rtibbles please see if this change doesn't adversely affect anything else.

rtibbles:
This looks right to me. Waiting on tests.

PR #5082, opened 2016-04-12
Fix critical bug in filtering attempt log collection request properly

jamalex:
Critical bug fix for 0.16.2, as discussed with @aronasorman and @rtibbles.

Problem: `data_model.get("exercise_id")` was `undefined`, and hence when we made a request for attempt logs, it didn't filter by exercise, and returned the last 10 of *any* attempt logs from the user.

@radinamatic, I will send you an sdist momentarily, for testing.


rtibbles:
This looks like the right fix to me.

rtibbles:
Happy to merge when @radinamatic has confirmed fix.

PR #5086, opened 2016-04-13
Default kalite user should now have a shell

benjaoming:
## Summary

Since https://github.com/learningequality/installers/issues/368 is getting fixed, we can make this slight simplification.

### DIFF
--- a/bin/kalite
+++ b/bin/kalite
@@ -102,7 +102,7 @@ if os.name == "posix" and os.path.isfile("/etc/ka-lite/username"):
             ).format(username))
             sys.stderr.write((
                 "To run the command as the default user, run this instead:\n\n"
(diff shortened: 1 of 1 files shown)


benjaoming:
Sorry about the new silly branch.. the change is pretty straight-forward so possibly just merge and delete the branch after.

MCGallaspy:
You tested this out?

benjaoming:
I've tried the command manually, yes:

screenshot from 2016-04-18 19 25 52

If you upgrade from a previous debian installation that already created a shell-less 'kalite' user, then the message will be mis-leading... but I would rule that people who have installed previous versions should be aware of the issue.

benjaoming:
Small change, self-merging. Is that even a word?

PR #5088, opened 2016-04-15
Fixes #5087: Properly parse kernel's release number string in ifcfg m…

### LINKED ISSUE #5087: Error parsing kernel release number in ifcfg module
## Summary

When I install ka-lite in my ARM device, I can't get it to start without crashing, due to a parsin problem in the ifcfg module, which is printing this error instead:
```
Running 'kalite start' as daemon (system service)

Stand by while the server loads its data...

[WARNING] [2016-04-15 15:45:35,777] kalite: Channel khan does not exist.
  File "/usr/share/kalite/kalitectl.py", line 925, in <module>
    port=arguments["--port"]
  File "/usr/share/kalite/kalitectl.py", line 527, in start
    addresses = get_ip_addresses(include_loopback=False)
  File "/usr/share/kalite/python-packages/fle_utils/internet/functions.py", line 94, in get_ip_addresses
    ips = [iface.get("inet") for iface in ifcfg.interfaces().values()]
  File "/usr/share/kalite/python-packages/ifcfg/__init__.py", line 62, in interfaces
    parser = get_parser()
  File "/usr/share/kalite/python-packages/ifcfg/__init__.py", line 42, in get_parser
    if float(kernel) < 3.3:
ValueError: invalid literal for float(): 3.10-trunk-meson8
```

Looking at [the relevant code](https://github.com/learningequality/ka-lite/blob/master/python-packages/ifcfg/__init__.py#L31) it seems to me that there are two issues:
  * The code assumes that the kernel's release number contains the minor revision, which is true in all the other environments I tried (e.g. it's `'4.3.3-301.fc23.x86_64'` on my Intel laptop), but not necessarily true everywhere (it's `'3.10-trunk-meson8'` in my ARM device).
  * The [comparison to determine whether the kernel is old](https://github.com/learningequality/ka-lite/blob/master/python-packages/ifcfg/__init__.py#L37) is broken, since for instance `float('3.10')` will evaluate to the `3.1` number, resulting in the wrong conclusion.

## Branch or installer method

The crash is happening both in the stable version installable with `pip install ka-lite-static` and the one from the **master** branch on Linux machines, as long as the output of `uname -r` returns a release number string where the major revision number part is not a valid number (e.g. containing non-digits).

The wrong comparison against the 3.3 version seems to be wrong regardless of the underlying system.

## How to reproduce

1. Run `kalite start` on  Linux where `uname -r` reports a non-integer major revision number string 
2. See how it crashes with the backtrace mentioned above

Alternatively replace [this line](https://github.com/learningequality/ka-lite/blob/master/python-packages/ifcfg/__init__.py#L33) with something `full_kernel = "3.10-trunk-meson8"` to replicate my environment and you should get the crash too.

mariospr:
## Summary

Fix this parsing problem, so that kalite will start without crashing on platforms where the kernel's release number does not contain a minor revision number and the major revision number does not parse to an integer. Also, fix the wrong comparison to determine if the kernel is < 3.3.

*Short description*

Do not assume that kernel release number strings come always in a 'x.y.z' format and only use the first two digits (kernel's version and major revision), sanitizing the second one to make sure that we parse its numeric value properly.

Also, compare the kernel's version and major revision separately, to avoid jumping into the wrong conclusion if float('ver.major_rev') does not result in the expected value (e.g. float('3.10') -> 3.1).

## Issues addressed

  * [Error parsing kernel release number in ifcfg module #5087](https://github.com/learningequality/ka-lite/issues/5087)

### DIFF
--- a/python-packages/ifcfg/__init__.py
+++ b/python-packages/ifcfg/__init__.py
@@ -1,5 +1,6 @@
 
 import platform
+import re
 from . import meta
 from . import parser
 from . import tools
@@ -31,10 +32,12 @@ def get_parser(**kw):
     if not parser:
         distro = kw.get('distro', platform.system())
         full_kernel = kw.get('kernel', platform.uname()[2])
-        kernel = '.'.join(full_kernel.split('.')[0:2]) 
-        
+        split_kernel = full_kernel.split('.')[0:2]
+        kernel_version = int(split_kernel[0])
+        kernel_major_rev = int(re.match('\d+', split_kernel[1]).group())
+
         if distro == 'Linux':
-            if float(kernel) < 3.3:
+            if kernel_version < 3 and kernel_major_rev < 3:
                 from .parser import Linux2Parser as LinuxParser
             else:
                 from .parser import LinuxParser


MCGallaspy:
Hi @mariospr, really great job! Thanks a lot for bringing this error to our attention.

To give you some context, KA Lite has previously included 3rd party packages for distribution in the python-packages directory. We've started moving away from this practice because it has some problems, however we haven't finished the process of unbundling 3rd party code yet. If we modify the source of the `ifcfg` module here, it means that we'll have trouble upgrading if/when a new official release of this module is made.

I'd like to suggest the following course of action:
1. Open a PR fixing this exact same bug in the python-ifcfg project: https://github.com/learningequality/python-ifcfg/blob/master/ifcfg/__init__.py#L37
2. After that's merged and a release is made, submit a PR to KA Lite unbundling this code and adding a new dependency to our `requirements.txt` file for the updated ifcfg package. (I can offer guidance with unbundling the code.)

I realize this is a slower process than accepting this PR, but following these practices will help make KA Lite much more maintainable in the future. In the meantime, are you able to work around this bug? How critical is a fix for your particular use case?

mariospr:
> Hi @mariospr, really great job! Thanks a lot for bringing this error to our attention.

No problem, I just stumbled upon it while trying to run KA Lite in our ARM platform and thought it would be nice to report it back upstream.

> To give you some context, KA Lite has previously included 3rd party packages for distribution in the python-packages directory. We've started moving away from this practice because it has some problems, however we haven't finished the process of unbundling 3rd party code yet. If we modify the source of the ifcfg module here, it means that we'll have trouble upgrading if/when a new official release of this module is made.

That makes sense to me, thanks for the detailed explanation. To be honest I'm totally new to KA Lite (and kind of new to python too), so I'm not surprised I missed the fact that stuff under `python-packages/` are in fact, well... "python packages" :-). So, this basically means that, apparently, I mistakently confused KA Lite as being the upstream project for this issue, when it's the ifcfg package.

All this just to say that I perfectly understand the situation and I agree with you it does not make any sense to fix this here, and I'll follow your advice to fix it in the real upstream.

> I'd like to suggest the following course of action:
> 1. Open a PR fixing this exact same bug in the python-ifcfg project: https://github.com/learningequality/python-ifcfg/blob/master/ifcfg/__init__.py#L37

Done: https://github.com/learningequality/python-ifcfg/pull/1

> 2. After that's merged and a release is made, submit a PR to KA Lite unbundling this code and adding a new dependency to our requirements.txt file for the updated ifcfg package. (I can offer guidance with unbundling the code.)

I'll do that once the other PR is merged, and eventually ask for advice if needed (thanks!)

> I realize this is a slower process than accepting this PR, but following these practices will help make KA Lite much more maintainable in the future. 

No worries. I perfectly understand that's the right thing to do, thanks for pointing it out.

> In the meantime, are you able to work around this bug? How critical is a fix for your particular use case?

Not really a problem yet to be honest, since we are still exploring KA Lite and haven't done anything about it yet :-). But if I need the fix I believe it should be as easy as applying my current patch downstream while until we eventually catch up with the upstream version.

Now closing this pull request, but keeping issue #5807 open while the whole thing is not fixed

PR #5095, opened 2016-04-20
Add options object to reference org_id from to ZoneSelectView in data…

rtibbles:
## Summary

Properly reference object to get org_id


jamalex:
Looks good!

PR #5101, opened 2016-04-25
0.16.3

benjaoming:
## Summary

Release of 0.16.3


MCGallaspy:
Do we need to update the home page to reflect this release? (By uploading new installers and updating download redirect links?)

benjaoming:
I would attend to this, however there was a severe bug in the Raspberry Pi and so I'm going to hold back until done releasing a 0.16.3-ubuntu2 :)

https://github.com/learningequality/installers/issues/394

MCGallaspy:
Okay, but please make sure you loudly announce when you're ready for release so that we can make the necessary preparations. I don't want the releases to get out of sync with the homepage, or for the various platform releases to get out of sync with one another (e.g. 0.16.5 available on Ubuntu but 0.16.1 is the latest for Windows). It's not going to do much good to create a release but not make the installers widely available.

benjaoming:
@MCGallaspy the intention of this release have been stated in chat, email, standup, various issues.... but with all the priorities at present, I've played low-key, trying to assume full responsibility for the release in the various channels that the release affects. Hence I wrote:

> @here 0.16.3 is on pypi and will be in the PPAs shortly.. this is only a release that's relevant to very few people, such as the Endless team.. so no need to make a big fuzz about it such that people start to needlessly upgrade :simple_smile:

I partly agree that it's nice to have installers follow suit in terms of version numbers, but I also don't wanna send a signal that people should drop everything and build installers that don't serve much more purpose than incrementing a version number :)

In other words, don't spend time building a Windows installer for 0.16.3, the source code change does not benefit the Windows platform.. but 0.16.4 will arrive very soon and that one *will* contain important fixes.

Moving forwards, I've pinged @aronasorman regarding a mailing list for installer maintainers, so we can have ONE single channel to signal when to build installers and not have it left open for interpretation through various issues, chats, pings and emails.

MCGallaspy:
> I partly agree that it's nice to have installers follow suit in terms of version numbers, but I also don't wanna send a signal that people should drop everything and build installers that don't serve much more purpose than incrementing a version number :)

At least for Windows, *building* the installer is automatic and releasing is not now a large effort. What concerns me is that there is not a release procedure in place, and as you suggested I think a proliferation of communication channels is a weakness here. Perhaps this is a personal failing, but sometimes I can't keep track of which patch release is important/unimportant for the various platforms, or what is expected of me during a release (like updating links) and not having one definitive source to check makes it harder for me.

Issue #5099, opened 2016-04-21
Perseus tool in geometry section not working


rifatbd:
## Summary
In the geometry section (more specifically in Transformations-Congruence-Similarity section) when doing exercise no operation can be done on shapes. Like rotating the shape, or transform or operations like those.

## Branch or installer method

 - Installer: Windows installer from pantry (win7 64 bit)
 - Version: 0.16.2 

## Traceback or relevant snippet from server.log

server.log is empty

## How to reproduce

1. Start Ka lite & naviagte to /learn/khan/math/basic-geo/transformations-congruence-similarity-geo/transformations-basics/performing-reflections-on-the-coordinate-plane/
2. Try to do the "Reflect" operation by clicking the button underneath the image
3. Nothing happens

## Screenshots

![capture1](https://cloud.githubusercontent.com/assets/18570759/14703090/cc4fe4b8-07cf-11e6-99d8-0279a2e03d65.PNG)
![capture2](https://cloud.githubusercontent.com/assets/18570759/14703089/cc4f17f4-07cf-11e6-80ad-0070dba9ef7a.PNG)
![capture3](https://cloud.githubusercontent.com/assets/18570759/14703093/cc50ab96-07cf-11e6-8486-e5b4cc12ad4c.PNG)
![capture4](https://cloud.githubusercontent.com/assets/18570759/14703091/cc500f9c-07cf-11e6-81f9-1528a86d7e01.PNG)
![capture5](https://cloud.githubusercontent.com/assets/18570759/14703092/cc504fac-07cf-11e6-8eb9-422544bba589.PNG)

benjaoming:
@rtibbles what's your take on this? You mentioned as a possible dirty resolution to #5036 that exercises could be excluded.

benjaoming:
I find that lots of geometry exercises are broken, almost anyone that I click. I guess these are using the "more advanced" features of Perseus.

Using: Latest English content pack + 0.16.3 release.

Here's another example of an error:

![screenshot from 2016-04-26 16 07 27](https://cloud.githubusercontent.com/assets/374612/14821064/99a7daea-0bc9-11e6-9230-7db02447a7f0.png)

rifatbd:
@benjaoming how can i install the 0.16.3 release? I cant see any installer for windows. Or there is some other method.

benjaoming:
[...]

PR #5110, opened 2016-04-30
Add woff2 files to MANIFEST and sdist

benjaoming:
## Summary

As mentioned by @rifatbd in #5099, we are missing woff2 files. They need to be in our MANIFEST in order to end up in the sdist.


benjaoming:
@rifatbd could you clarify what the missing fonts result in? Like e.g. components that aren't rendered because of this?

rifatbd:
@benjaoming i will check and let you know..

Issue #4875, opened 2016-02-19
Questions in learn exercise never end


manipalsingh013:
When using exercise from learn, questions never end as they start repeating from the beginning and keep repeating.

Branch : **0.16.x**
OS: **Ubuntu 14.04**

Steps to reproduce:
1. Login as a Learner
2. Choose any exercise from learn
3. Keep answering question

Expected Behavior: Once all the question in exercise are answered by learner, he/she should get a notification regarding exercise completed.
Current Behavior: Question never end, they start to repeat from beginning and keep repeating.
![exercise](https://cloud.githubusercontent.com/assets/13970876/13178558/7ce9a770-d745-11e5-92e3-18d44f109665.png)

manipalsingh013:
I noticed that Progress bar is full i.e Quiz is complete but all questions are not covered till then, repetition starts few questions later(in some exercises)

manipalsingh013:
@rtibbles I would like to work on this, but what should be the fix?
Is it fine giving learner some alert?or it should be something else?

j-schwartz:
Hi Guys - I'm getting the same behavior on 0.16.5 for all exercises... This would seem like a bigger deal, no?

rtibbles:
I guess I am missing the point here - when the student has been judged as 'mastered' the exercise, they should get a notification.

If a student choose to continue practicing, then they will be able to continue to take the exercise for extra practice if they so wish.

I guess what I am saying is: this is intended behaviour, but perhaps the user experience is not as good as it could be if people don't realize they have achieved mastery?

j-schwartz:
Ah, what would the notification look like?  Just the green bar filling up?  I guess in the past I felt like when a subject was mastered they were moved on (or at least presented an option to move on) to the next video.

It seems like a bigger issue because when a user finishes watching a video, there is an error that doesn't automatically move them on to the exercises, so this seemed to me at least, to be a similar error.

https://github.com/rachelproject/rachelplus/issues/56

rtibbles:
There should be, and perhaps something is awry here, a notification bar (blue in colour) that appears above the exercise frame, saying "You have finished this exercise".

Note: none of these messages will appear if no user is logged in, or the user is either a coach or an admin - these messages will only appear for student accounts.

The error in the link above looks like the error that might occur if watching a video as an admin (when it tries to save the VideoLog as an admin, it has no student user field, so cannot save the file).

[...]

PR #5124, opened 2016-05-25
Do not save unconditionally when video ends; only if learner logged in.

jamalex:
Addresses the error message issue in #4875.

Tested with admin and no user: now, no error message.
With logged-in learner, still saves the log at the end of the video.


rtibbles:
Ah, that's why it doesn't happen for exercises, but does for videos!

Issue #5126, opened 2016-05-25
Unable to download videos when m2crypto is available


mariospr:
## Summary

When trying to run KA Lite 0.16.5 while having version 0.21.1 of the python-m2crypto package installed in my Ubuntu/Debian-like systems, I could not download any video from the administration interface, apparently becaus of "peer did not return certificate" errors happening all the time:
![dev _25_05_2016_14_37_30](https://cloud.githubusercontent.com/assets/2197259/15555102/1cd6cc44-22bf-11e6-8c7b-1bd60fa3da26.png)

A quick test using newer versions of python-m2crypto (tried 0.22 and 0.24 for now) made the error go away, but still there are **no videos available after the process finishes**, even if the page says that all videos have been "downloaded successfully".

Finally, monitoring the contents of `~/.kalite/content` with `watch` while downloading the videos I've observed something weird going on, regardless of the version of python-m2crypto being available: KA Lite definitely starts downloading the video files (I see them growing on disk) but then deletes each of them systematically after finishing the download, leaving the computer with no videos in the end. Notice that I can see the "peer did not return certificate error" with m2crypto == 0.21, while no error message is shown with m2crypto >= 0.22, but perhaps the same error is still there and is behind this mysterious deletions?

Last, just for the sake of testing a bit more, I've tried the same procedure on my Fedora 23 machine after installing the m2crypto package (version 0.22) and I observed the same phenomena: it looks like it's downloading (I can see files growing in `~/.kalite/content`, but there are no videos in the end).

## Branch or installer method

I could reproduce this with version 0.16.5 of KA Lite installed via 2 different ways:
  * Using the debian package from [this PPA](https://launchpad.net/~learningequality/+archive/ubuntu/ka-lite)
  * Installing with `sudo pip install ka-lite-static`

Also, I could reproduce this in three different OS's:
  * Endless OS (Debian like system) (32bit and 64bit) 
  * Debian Testing (64 bit)
  * Fedora 23 (64 bit)

## Traceback or relevant snippet from server.log

No relevant logs seen (happy to collect more info if somebody guides me, though)

## How to reproduce

1. Install latest stable version KA Lite (0.16.5 atm) from PIP or from the debian package [from here](https://launchpad.net/~learningequality/+archive/ubuntu/ka-lite) 
2. Make sure you have the python-m2crypto (Debian/Ubuntu) or m2crypto (fedora) package installed
3. Launch KA Lite with `kalite start` and open a browser at http://localhost:8008
[...]

PR #5138, opened 2016-06-06
unbundle requests and use requests.get instead of urllib.urlretrieve

benjaoming:
## Summary

Fixes issues with m2crypto not able to handle https connections.

## TODO

## Reviewer guidance

If there's any concern about including this PR in 0.16.6, please state it in the discussion, but I would caution that the old download method was probably less reliable than using the latest version of  `requests`  @jamalex 

## Issues addressed

#5126


kalite/distributed/models.py
```diff
@@ -1,8 +0,0 @@
-import requests
-from kalite.version import *
-
-# Modify Request's default user agent to include the KA Lite version number
-# (this is done in this file because putting it in __init__.py causes circular imports,
-# and putting it in urls.py doesn't get loaded by management commands)
-base_headers = requests.defaults.defaults["base_headers"]
-base_headers["User-Agent"] = ("ka-lite/%s " % VERSION) + base_headers["User-Agent"]
```
benjaoming: What components of KA Lite should report the version in the user agent? Monkey-patching requests is impossible in recent versions as the default headers aren't accessible but hard-coded @jamalex

benjaoming:
This is where the requests library is used currently, I don't see any reason of concern, but should we change the User-Agent on any of these requests @jamalex ?

```
kalite/distributed/static/js/distributed/perseus/ke/build/update_title_tags.py
9:import requests
13:    r = requests.get("http://www.khanacademy.org/api/internal/exercises")

kalite/distributed/static/js/distributed/perseus/ke/build/close-old-issues.py
10:import requests
39:        r = requests.get(url, auth=github_auth)
52:    r = requests.get("https://api.github.com/repos/Khan/khan-exercises/issues/"
65:    r = requests.post("https://api.github.com/repos/Khan/khan-exercises/"
73:    except requests.HTTPError:
83:    r = requests.post("https://api.github.com/repos/Khan/khan-exercises/"
91:    except requests.HTTPError:

kalite/i18n/management/commands/create_dummy_language_pack.py
13:import requests
57:        resp = requests.get(url)
59:    except requests.ConnectionError as e:

kalite/i18n/management/commands/setup_in_context_translations.py
5:import requests
40:        resp = requests.get(url)

kalite/i18n/tests/create_dummy_language_pack_tests.py
1:import requests
17:    @patch.object(requests, "get", autospec=True)

kalite/static/js/distributed/perseus/ke/build/update_title_tags.py
9:import requests
13:    r = requests.get("http://www.khanacademy.org/api/internal/exercises")

kalite/static/js/distributed/perseus/ke/build/close-old-issues.py
10:import requests
39:        r = requests.get(url, auth=github_auth)
52:    r = requests.get("https://api.github.com/repos/Khan/khan-exercises/issues/"
65:    r = requests.post("https://api.github.com/repos/Khan/khan-exercises/"
73:    except requests.HTTPError:
83:    r = requests.post("https://api.github.com/repos/Khan/khan-exercises/"
91:    except requests.HTTPError:

kalite/contentload/management/commands/unpack_assessment_zip.py
1:import requests
47:            r = requests.get(ziplocation, prefetch=False)

kalite/contentload/tests/unpack_assessment_zip.py
4:import requests
52:    @patch.object(requests, "get")
60:        self.assertEqual(get_method.call_count, 0, "requests.get was called even if we should've skipped!")
66:    @patch.object(requests, "get", autospec=True)

```

kalite/distributed/models.py
jamalex: This is to help with diagnosing errors (and calculating version stats) based on our log files on the central servers. Please see if there's some alternative way to set the default, even if it involves some monkey-patching.

benjaoming:
Confirmed working on Raspberry Pi with m2crypto 0.21.1 installed as a dependency of ka-lite-raspberry-pi.

python-packages/fle_utils/internet/download.py
```diff
@@ -1,97 +1,81 @@
...
-    """Return response from url; retry up to 5 times for server errors.
-    When returning an error, return human-readable status code.
-
-    codes: server-error, client-error, unexpected-error
-    """
-    assert max_retries >= 0, "max_retries must be non-negative."  # guarantees response will never be None
-
-    response = None
-    last_error = None
-    for retries in range(1, 1 + max_retries):
```
jamalex: You seem to have deleted all the error handling code here? That stuff was pretty critical, added to handle very common cases of 1) connections dropping out, and 2) video URLs actually being 404's, despite being in the KA API. We need to keep these checks, as otherwise it will become very brittle.

jamalex:
> This is where the requests library is used currently, I don't see any reason of concern, but should we change the User-Agent on any of these requests @jamalex ?

Most of the ones you list there don't seem important (as they're not hitting our central server, so we won't be getting logs with the user agent in it for those), but that doesn't seem to be an exhaustive list. Here are some that are more important:
https://github.com/learningequality/ka-lite/blob/develop/python-packages/fle_utils/internet/functions.py#L22
https://github.com/learningequality/ka-lite/blob/develop/python-packages/securesync/api_client.py#L14
https://github.com/learningequality/ka-lite/blob/develop/python-packages/securesync/devices/views.py#L100
(I know these are under `python-packages`, but these aren't actually external python packages; that code is only used in the context of KA Lite)

Might be good to pull out this header calculation logic into a reusable function:
https://github.com/learningequality/ka-lite/pull/5138/files#diff-5e841025f58a948893e991da20eeebaaR64
(though not sure the best way to avoid having securesync and fle_utils depend on kalite...)

kalite/distributed/models.py
benjaoming: I can't set the default, but I can add the headers explicitly where we use requests. I'm guessing this is only relevant on our own servers?

python-packages/fle_utils/internet/download.py
benjaoming: `make_request` isn't called anywhere!

python-packages/fle_utils/videos.py
```diff
@@ -36,18 +38,18 @@ def download_video(youtube_id, download_path="../content/", download_url=OUTSIDE
...
             raise URLNotFound("Video was not found, tried: {}".format(url))
 
-        path, response = download_file(thumb_url, thumb_filepath, callback_percent_proxy(callback, start_percent=95, end_percent=100))
+        response = download_file(thumb_url, thumb_filepath, callback_percent_proxy(callback, start_percent=95, end_percent=100))
         if (
                 not os.path.isfile(thumb_filepath) or
-                "Content-Length" not in response or
-                not str(len(open(thumb_filepath, "rb").read())) == response["Content-Length"]):
+                "content-length" not in response.headers or
+                not len(open(thumb_filepath, "rb").read()) == int(response.headers['content-length'])):
```
benjaoming: Expand and look at line 55-68, that's where all the error handling essentially takes place, even a catch-all-exceptions block :/

python-packages/fle_utils/internet/download.py
jamalex: Ah, cool, looks like it was made redundant at some point but never cleaned up... until now.

benjaoming:
>  (though not sure the best way to avoid having securesync and fle_utils depend on kalite...)

They already did when kalite was monkey-patching requests :)

benjaoming:
Thanks for searching through fle_utils btw, adding that change as well.

python-packages/fle_utils/videos.py
jamalex: Cool, looks like it should cover the bases, though it seems it doesn't do "retries" at all, but that wasn't due to your changes. We maintain a list of `failed_youtube_ids`, but don't use it for anything other than getting a count of failures. Not ideal, but not a big problem.

kalite/distributed/models.py
jamalex: Yeah, only relevant for stuff going to our servers.

jamalex:
> They already did when kalite was monkey-patching requests :)

Kind of, though implicitly, and optionally. They would still function fine without kalite. More like a "hook". :D

jamalex:
> Thanks for searching through fle_utils btw, adding that change as well.

Cool -- there were a bunch of other uses of `requests` in there, but they didn't seem so critical. The ones I listed are the places that hit the central server in ways we may want to examine the logs for.

benjaoming:
> They would still function fine without kalite.

If the intention was to have correct user agent stats, you would be failing but much more severely than a circular import, you'd have wrong stats and think they were right! So all this functionality really depends on the monkey patch firing.

benjaoming:
@jamalex done in 5884508

jamalex:
Ok, looks good to me! Merging now, but please make sure data syncing also still works (with a basic check -- registering and creating a local record and syncing it up, with no errors), in addition to the video downloading, as both depend on requests.

benjaoming:
@jamalex 

Tested:

 * One-click Device registration OK
 * Sync'ing OK
 * Video download OK

screenshot from 2016-06-08 13 43 24

jamalex:
:boom: :+1: :zap:

Issue #5143, opened 2016-06-08
Make ROOT_DATA_PATH configurable


mariospr:
## Summary

While trying to bundle KA Lite as a [flatpak bundle](http://flatpak.org) I realized that setting the `KALITE_DIR` variable to point to the alternative root directory (`/app/share/kalite`, for the case of flatpak) was not enough, since `kalite start` would fail with an error on startup, since it would still look into `/usr/share/kalite` when copying the static media files.

By looking into [kalite/__init__.py](https://github.com/learningequality/ka-lite/blob/master/kalite/__init__.py), it seems the problem is that the `ROOT_DATA_PATH` is always using python's `sys.prefix` to construct the data path, which will point to the non-existent `/usr/share/kalite` in this particular case.

So, even though I agree one option to prevent this could be to avoid doing that copy of static media, I think it would be nice if we could construct the `ROOT_DATA_PATH` variable considering the value of the `KALITE_DIR` variable, if it exists.

## Branch or installer method

I'm using a development environment setup on top of my Fedora 23 machine using virtualenv and  [this instructions](https://github.com/learningequality/ka-lite/wiki/Getting-started), with the **`0.16.x` branch**.

## Traceback or relevant snippet from server.log

On first run of `kalite start` with KA lite installed in a non-standard location (`/app/share/kalite` for flatpak), it will fail with this error:
```
[...]
[INFO] [2016-06-08 15:23:40,480] root: Copying static media...
Traceback (most recent call last):
  File "/app/share/kalite/python-packages/django/core/management/base.py", line 224, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/app/share/kalite/python-packages/django/core/management/base.py", line 263, in execute
    output = self.handle(*args, **options)
  File "/app/lib/python2.7/dist-packages/kalite/distributed/management/commands/initialize_kalite.py", line 42, in handle
    self.setup_server_if_needed()
  File "/app/lib/python2.7/dist-packages/kalite/distributed/management/commands/initialize_kalite.py", line 36, in setup_server_if_needed
    call_command("setup", interactive=False)
  File "/app/share/kalite/python-packages/django/core/management/__init__.py", line 161, in call_command
    return klass.execute(*args, **defaults)
  File "/app/share/kalite/python-packages/django/core/management/base.py", line 263, in execute
    output = self.handle(*args, **options)
  File "/app/lib/python2.7/dist-packages/kalite/distributed/management/commands/setup.py", line 494, in handle
[...]

PR #5144, opened 2016-06-08
Make ROOT_DATA_PATH consider the KALITE_DIR environment variable

mariospr:
## Summary

Make ROOT_DATA_PATH consider the KALITE_DIR environment variable

Otherwise, kalite start will fail to find the static media files that it will try to copy on the first run, causing assets to be unresolvable when loading the web UI.

## TODO

  * I'm unsure how to write a test for this, but I checked it manually. Any advice?
- [X] Has documentation been written/updated?
  * Not needed: it's all about using KALITE_DIR in a missing place, which is already documented
- [X] New dependencies (if any) added to requirements file
  * Not needed

## Issues addressed

https://github.com/learningequality/ka-lite/issues/5143

### DIFF
--- a/kalite/__init__.py
+++ b/kalite/__init__.py
@@ -7,4 +7,7 @@ from version import *
 __version__ = VERSION
 
 # ROOT_DATA_PATH is *not* where the source files live. It's a place where non-user-writable data files may be written.
-ROOT_DATA_PATH = os.path.join(sys.prefix, 'share', 'kalite')
+if 'KALITE_DIR' in os.environ:
+    ROOT_DATA_PATH = os.environ['KALITE_DIR']
+else:
+    ROOT_DATA_PATH = os.path.join(sys.prefix, 'share', 'kalite')


benjaoming:
@cpauya @aronasorman 

My thoughts are that the idea of `KALITE_DIR` was to indicate if a flat structure like the Github repo exists. The only usage is to add python-packages, dist-packages and kalite/ to sys.path. In any other case these directories are assumed to be in the same folder as kalitectl.py, which is always in the same folder (typically `/usr/share/kalite`)

In that case, the contents of ROOT_DATA_PATH is -- as @mariospr argues - always the same directory.

So I'd agree to this change :)

cpauya:
We're good with this too. 👍

kalite/__init__.py
```diff
@@ -7,4 +7,11 @@
 __version__ = VERSION
 
 # ROOT_DATA_PATH is *not* where the source files live. It's a place where non-user-writable data files may be written.
-ROOT_DATA_PATH = os.path.join(sys.prefix, 'share', 'kalite')
+ROOT_DATA_PATH = None
```
aronasorman: Can you remove this initialization of `ROOT_DATA_PATH`, and move the comment closer to the `if` statement? That makes it easier to find where `ROOT_DATA_PATH` is meaningfully set.

aronasorman:
Agree, if they are the same directory, we should make it explicit!

Just one change needed, then ready to merge.

kalite/__init__.py
mariospr: Sure. I guess I was wearing my C hat when I did this, since it felt weird to me not to have a common place to initialize this variable outside of the if branches, but it's probably fine in python. I'll change that.

mariospr:
Thanks for the feedback. PR updated, let me know if there's anything else I need to do, otherwise please feel free to merge.

aronasorman:
PR merged, thanks for your contribution!

Issue #4982, opened 2016-03-09
Obsolete error message on video download page


rtibbles:
## Summary

We should no longer be warning people to limit downloads on Raspberry Pi.

## Branch or installer method

 - Installer: Debian
 - Version: 0.16.x beta1

## How to reproduce

- Login as admin.
- Go to video download page
- See message.

## Screenshots

![image](https://cloud.githubusercontent.com/assets/1680573/13654516/b31c29b8-e60d-11e5-9cd9-bface89e5424.png)

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

People might believe our warning.

radinamatic:
How many videos can one actually download concurrently on RPi?

radinamatic:
Not seeing this warning on Ubuntu with Beta2...

benjaoming:
Done and released!

PR #5149, opened 2016-06-11
Remove useless warning message that no longer serves a purpose #4982

benjaoming:
## Summary

I think it's about time we let go of this odd warning message as we've intended to. Only took 5 seconds to find and remove :)

## Issues addressed

#4982


benjaoming:
@rtibbles Any elaboration needed here?

PR #5150, opened 2016-06-11
Allow logins during LOCKDOWN #5117

### LINKED ISSUE #5117: Activating lockdown causes page redirect error
## Summary

Activating the setting LOCKDOWN = True in settings.py to prevent users from accessing exercises and videos without logging in causes a page redirect error ERR_TOO_MANY_REDIRECTS.

## Branch or installer method
 - Installer: Debian
 - Version: 0.16.5

## How to reproduce

1. Add the line LOCKDOWN = True to settings.py
2. Restart ka-lite

## Screenshots

screenshot from 2016-05-11 17 03 28

Capturing student activity data is really important to us. When students are not supervised they sometimes watch videos or do exercises without logging in, and their progress is not recorded. This would be a good way to 'force' students to log in.

benjaoming:
## Summary

We hardcode-specify which pages are allowed during lockdown :/

Tested that this enables both first-run admin creation and subsequent logins, after which the user can view content etc.

This is critical to fix because reporter has said so, and I believe @Aypak :)

## Fixes

#5117

### DIFF
--- a/kalite/distributed/middleware.py
+++ b/kalite/distributed/middleware.py
@@ -6,16 +6,21 @@ from django.core.urlresolvers import reverse
 
 
 class LockdownCheck:
+
     def process_request(self, request):
         """
         Whitelist only a few URLs, otherwise fail.
         """
-        if (settings.LOCKDOWN
-            and not request.is_logged_in
-            and request.path not in [reverse("facility_user_signup"), reverse("dynamic_js"), reverse("dynamic_css")]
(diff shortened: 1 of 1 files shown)


benjaoming:
Assigning loads of devs to show that lockdown was broken and fixed, I'm not sure anyone was aware of that.

benjaoming:
Not sure how this went under everyone's radar, but it's a simple fix to a critical issue, so self-merging before releasing...

PR #5153, opened 2016-06-12
Add release notes for 0.16.5 and 0.16.6

benjaoming:
## Summary

Forgot this in 0.16.5, let's not do that for 0.16.6

### DIFF
--- a/data/version.yml
+++ b/data/version.yml
@@ -1,3 +1,36 @@
+0.16.6:
+  release_date: "2016/06/14"
+  git_commit: ""
+  new_features: []
+  bugs_fixed:
+    all:
+      - Allow logins during LOCKDOWN #5117
+      - Remove useless warning message that no longer serves a purpose #4982
+      - Make ROOT_DATA_PATH consider the KALITE_DIR environment variable #5143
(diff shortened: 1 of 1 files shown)


benjaoming:
Stuff has been failing consistently for some hours now, but this PR was okay before I did a force-push with a non-essential change.

data/version.yml
```diff
@@ -1,3 +1,36 @@
+0.16.6:
+  release_date: "2016/06/14"
+  git_commit: ""
```
aronasorman: Don't forget to update this once we're sure of a release!

aronasorman:
Just one comment, merge when ready.

data/version.yml
```diff
@@ -1,3 +1,36 @@
...
+  new_features: []
```
aronasorman: Do you want to mention an updated content pack?

data/version.yml
benjaoming: I actually had a commit hash in the previous commit, but it was not the final one so removed it and force-pushed :)

I'll make amends during the release process.

benjaoming:
Adding all this over in #5158

Issue #4934, opened 2016-02-29
Search paginator not working properly


manipalsingh013:
## Summary
Although the next arrow key of paginator seems disabled after clicking for a few times but user can still click on it and go to empty page. 

**Branch: 0.15.x
Ubuntu 14.04
Firefox**

## How to reproduce
1. Login as some user
2. Search for something for which some results appear(eg 'count')
3. Click on next button of paginator(arrow button)
4. Keep clicking next until you find that mouse cursor changed to unclickable when hovering on next button.
5. Click again on next button.
6. It takes you to an empty page.

## Screenshots
![paginator](https://cloud.githubusercontent.com/assets/13970876/13385792/0c5b9ad0-decb-11e5-9678-7b4c2c1f5e0d.png)

MCGallaspy:
I'm not sure I consider this a critical bug for 0.16.0, but we can take a look at fixing it on `develop` for future releases.

benjaoming:
Fixed in #5155

PR #5155, opened 2016-06-13
Remove links to search pages that don't exist

benjaoming:
Search paginator not working properly #4934


benjaoming:
Tested this manually and it works, so self-merging :)

Issue #2003, opened 2014-04-28
Wrong IP address displayed for syncing


bcipolli:
Branch 0.11.1

On a distributed server, IP address of syncing shows `kalite.learningequality.org`.  

When the device being showed is the local device, we should show the local device IP.  When it's the remote device, we should show the remote IP.

![image](https://cloud.githubusercontent.com/assets/4072455/2818240/969d5922-cee2-11e3-87f2-1e95c276b822.png)

PR #5156, opened 2016-06-13
Since we're clearly not displaying an IP address for the local device..

benjaoming:
## Summary

Quick fix for #2003 - since we're clearly not displaying an IP address but a host name, let's write that so it isn't confused with the local device IP

We apparently have issues that can be fixed in 2 seconds :)

## Issues addressed

#2003

### DIFF
--- a/kalite/control_panel/templates/control_panel/device_management.html
+++ b/kalite/control_panel/templates/control_panel/device_management.html
@@ -134,7 +134,6 @@
                     <thead>
                         <tr class="header-footer-bg">
                             <th width="250px">{% trans "Sync Date" %}</th>
-                            <th width="250px">{% trans "Device IP Address" %}</th>
                             <th width="200px">{% trans "# Models Uploaded" %}</th>
                             <th width="200px">{% trans "# Models Downloaded" %}</th>
                             <th width="200px">{% trans "# Errors" %}</th>
@@ -144,7 +143,6 @@
                     {% for sync_session in session_pages %}
                         <tr>
                             <td>{{ sync_session.timestamp }}</td>
-                            <td>{{ sync_session.ip }}</td>
                             <td>{{ sync_session.models_uploaded }}</td>
                             <td>{{ sync_session.models_downloaded }}</td>
                             <td>{{ sync_session.errors }}</td>


jamalex:
Looks like a good improvement. One thing: we try not to change translatable strings in patch releases, because our language packs (containing .MO files) are meant to work across all versions of a minor release (e.g. 0.16.x). I'll let @aronasorman chime in on whether he thinks that's a blocker here, and merge as appropriate.

aronasorman:
These things do get translated: https://crowdin.com/translate/ka-lite/28886/en-es#q=IP

And since this is in our frontend, I prfer that we don't merge this now. Maybe for 0.17?

benjaoming:
Erh, I don't think it's appropriate to keep a 2 year old issue open in order to retain a misleading column heading for the sake of not translating "upstream hostname" which is a technical term on a page intended for techies anyways :)

jamalex:
Actually, note that this same template is used in the ka-lite-central control panel, where "Device IP Address" is a more accurate term than "Upstream hostname". I would say we either change to something more general, applicable to both contexts, such as "Syncing IP/hostname", or just delete the column from the table.

image

benjaoming:
@jamalex fixed + rebased

jamalex:
Great, thanks! I still defer to @aronasorman regarding targeting for this; it looks like he would prefer we target this to develop, for 0.17, to avoid breaking the 0.16.x string freeze.

benjaoming:
@jamalex I'm aware of the central server but hardly found it worthwhile to bother with... I'm sitting in a steam roller driving 80 mph cruising towards <200 open issues :)

benjaoming:
I prefer just delete the column so it doesn't have to be translated then.

benjaoming:
Gosh this isn't supposed to be easy :)

jamalex:
> I prefer just deleting the column so it doesn't have to be translated then.

I would be fine with that -- I doubt that column provides much value to end users, and I mostly interact with data from that column via the shell/management commands.

jamalex:
> @jamalex I'm aware of the central server but hardly found it worthwhile to bother with... I'm sitting in a steam roller driving 80 mph cruising towards <200 open issues :)

.jpg?itok=5Bd4Zngf)

benjaoming:
Goodbye misleading column, goodbye 2 year old issue :)

benjaoming:
It's another PR where I just remove code :)

jamalex:
> It's another PR where I just remove code :)

Issue #5118, opened 2016-05-13
Downloading Ubuntu KA Lite Installer


waltonrp:
## Summary

The link on the Install Guide appears to be bad, this page:
http://ka-lite.readthedocs.io/en/0.16.x/installguide/install_all.html#main-method-ubuntu-debian-deb

This link:
Download the latest .deb installer, and run this command:

Takes me to a page not found error.

I'm assuming that you are wanting to point to in installer on this page:
http://pantry.learningequality.org/downloads/ka-lite/0.16/installers/debian/

Can you confirm that the Ubuntu installer I should download is this:
ka-lite-bundle_0.16.5.deb 
or another file?

## Branch or installer method

*Please specify if you are using the current `develop` branch or which installation source you were using when the issue occurred.*

 - Installer: Windows/OSX/Debian (?)
 - Version: 0.15.1 (?)

or:

 - Branch: develop (?)

## Traceback or relevant snippet from server.log

```
INSERT TRACEBACK, LOG MESSAGES ETC. HERE
```

## How to reproduce

1. Steps
1. To
1. Reproduce

## Screenshots

*If applicable*

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

*If applicable*

radinamatic:
@waltonrp Thank you for reporting this! :)

Yes, the latest Debian/Ubuntu installer is `ka-lite-bundle_0.16.5.deb` in the:
http://pantry.learningequality.org/downloads/ka-lite/0.16/installers/debian/

I believe @aronasorman @benjaoming or @jamalex will update those redirects asap.

benjaoming:
What a silly issue, @radinamatic I have no idea how it went unnoticed. Please feel welcome to make more noise, but I have to admit my overview of issues might have been clouded lately. Will get up to speed!

Link is fixed, thanks @waltonrp for pointing that out!!!

radinamatic:
Message received @benjaoming, I'll try to be more "noisy"! 😄

PR #5158, opened 2016-06-14
Docs version string substitution and release notes

benjaoming:
## Summary

Adds proper version string substitution in our docs so we can hopefully avoid dead links when doing future releases.

Adds a new syntax for link substitution, for instance you can link to a github issue like this:

    :url-issue:`5127`

## Issues addressed

#5118 #3573 #4723 

## Screenshots

screenshot from 2016-06-14 15 14 49


benjaoming:
Merging this so it will build on RTD.

It's failing tests even though it introduces no changes to `kalite` itself, it's the same timeout messages that started popping up yesterday.

Issue #4622, opened 2015-10-13
Forward admin user to Manage tab after device registration


radinamatic:
Forward admin user to Manage tab after device registration - small improvement.

**Branch**: develop  

**Current Behavior**: 
After the successful device registration, admin user is left sitting duck on the confirmation page.

![forward-to-manage](https://cloud.githubusercontent.com/assets/1457929/10470531/24d02d26-720e-11e5-9570-6368f19ae18a.png)

I suggest that implementing the automatic forward to the Manage tab after, say 10 seconds, would be a better UX.

**Steps to reproduce**:

1. Log in as Admin on a new device.
2. Register device.
3. Enjoy watching the `Congratulation` message.

radinamatic:
Very minor issue, punt or close @benjaoming!

benjaoming:
I noticed this, too. I agree very much that it should be fixed.

benjaoming:
Alright, done :)

PR #5160, opened 2016-06-14
Device registration completed

benjaoming:
## Summary

~~I was unable to install requirements, not sure what the difference is, maybe Travis has been running another version of Pip and this.~~

~~Furthermore, fixed an issue since South fails to find migrations and thus if we try to migrate a fresh database, stuff would break.~~

Finally, that button I was trying to add :)

## Issues addressed

#4622 

screenshot from 2016-06-14 20 24 56

### DIFF
--- a/docs/installguide/release_notes.rst
+++ b/docs/installguide/release_notes.rst
@@ -1,6 +1,15 @@
 Release Notes
 =============
 
+0.17 (unreleased)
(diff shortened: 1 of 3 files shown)


benjaoming:
> Coverage decreased (-2.7%) to 43.711% 

No, it doesn't :)

PR #5168, opened 2016-06-17
Mention mailing list in docs

benjaoming:
Lots of subscribers, let's get more! :)


benjaoming:
Hurrying this so we can get it in the embedded docs, too

jamalex:
Looks good. This can be.... oh.

PR #5176, opened 2016-06-18
0.16.x into develop

benjaoming:
Merge conflict is simply version.py version bump, no real issue.


benjaoming:
Interesting to see if the persistent test failure in 0.16.x persists.

benjaoming:
Test failures don't persist, that is indeed intewesting

Issue #5157, opened 2016-06-13
Learner stats: Wrong login count and session tracking


benjaoming:
## Summary

Received the following report:

> After running the same tests as I had in IE, Chrome appears to be having the same issues, including no Student/Coach/Group Login and Login Times being recorded.

> Below is a screenshot showing that students who have watched videos and completed exercises still do not have a Login of Login Time.

![image001](https://cloud.githubusercontent.com/assets/374612/16011936/9fe2c490-3187-11e6-92bd-268de36ceaf6.png)

I can confirm this, using our latest 0.16.x branch and a clean installation.

## Branch or installer method

 - Branch: 0.16.x

## How to reproduce

1. Create a learner account
1. Login
1. Solve exercise
1. Login as manager
1. Go to facility
1. See that there is no login count for the learner.

jamalex:
:+1: on this, from a report from Liberia today; this probably shouldn't hold back the 0.16.6 release (since that's coming tomorrow), but is an important bug to address soon.

j-schwartz:
Hey @benjaoming - I know we lose you soon, if we haven't already.  Just curious if you think you'll be able to attack this prior to taking off.  Any idea?

radinamatic:
@j-schwartz, we lost @benjaoming already, at least until the end of the month... :stuck_out_tongue: 

#5180 has already been merged, so you can check out the repo directly, try the 0.16.7 debian pre-release on PPA, or the latest Win build in Dungeon!

benjaoming:
@j-schwartz -- this will be included in 0.16.7 as the PR is already merged, shall close this issue now :)

j-schwartz:
Fantastic, many thanks Ben

Issue #4288, opened 2015-08-25
generaterealdata produces extremely large numbers for time spent logged in


sethachoi:
When using the generaterealdata command, the logged in hours are ridiculously large (like, in the millions total), although all the other data looks somewhat reasonable. 

Steps to reproduce:
1. type bin/kalite manage generaterealdata
2. launch kalite
3. observe giant numbers

seen on current develop as of aug 25, 2015, and as early as aug 18, 2015 for certain.

benjaoming:
Seems like it's not just the generation of data that's broken for login stats but possibly both coach report and logging as well #5157

radinamatic:
After #4999 & #5015, I eventually stopped using the command.

PR #5180, opened 2016-06-24
Bug fixes in learner stats

benjaoming:
## Summary

This PR addresses several issues in our learner stats:

 - Not using real count, but assuming each login session is 1 login (will produce very bad results if `USER_LOG_MAX_RECORDS_PER_USER` is more than 1 day)
 - Fixes `generaterealdata` so it doesn't generate a 6-month login for every exercise solved and triggers the `pre_save` signal.
 - In case of overlapping summaries, we retain previous summaries when deleting stuff
 - `UserLogSummary.get_period_start_datetime` was called with `end_datetime`, producing wrong start times of sessions.
 - Counting a simple login correctly: Doing this on 0.16.6 produces no results: Login on a learner account, then log out. The login count is not incremented. After this PR, the login count is incremented exactly by 1.

Note: I will *not* write tests for this because I don't have the resources time-wise, and if KA Lite is indeed in "legacy mode", we should leverage from that decision and assume that once all this is fixed, it won't break again. Otherwise, I would want to rewrite the whole thing rather than writing tests for something that's as complicated as Brexit :)

## TODO

- [x] Have **tests** been written for the new code? *No, and the won't.*
- [x] Has documentation been written/updated? *Release notes should be added*
- [x] New dependencies (if any) added to requirements file *None*
- [x] A simple login/logout produces a new UserLog and increments current UserLogSummary by 1.
- [x] Figure out why logging in and logging out produces a `UserLogSummary` with `total_seconds=0`

## Reviewer guidance

If anyone has can recall all this UserLog stuff, please feel welcome. Otherwise, it's a matter of live-testing.

## Issues addressed

#5157
#4288 

## Screenshots

This shows the effect of ignoring total_seconds when displaying login counts... yes, there is an error in the data, but we don't care because the database still says the user logged in.

screenshot from 2016-06-24 21 16 37


benjaoming:
Update on the nature of the issue:

If I create an open `UserLog` entry in the past, e.g. an entry that isn't yet summarized, and then login again, that entry will be completely overwritten, yet not summarized.

benjaoming:
Hmm... pressing Logout calls the same view twice always...

```
[WARNING] [2016-06-24 21:42:57,714] kalite: test: Had to BEGIN a user log entry, but ENDING(1)! @ 2016-06-24 21:42:57.712011
[INFO] [2016-06-24 21:42:58,224] django.request: HTTP Request /securesync/api/user/logout/ - Response: 200
[WARNING] [2016-06-24 21:42:58,629] kalite: test: Had to BEGIN a user log entry, but ENDING(1)! @ 2016-06-24 21:42:58.625851
[INFO] [2016-06-24 21:42:59,118] django.request: HTTP Request /securesync/api/user/logout/ - Response: 200
```

benjaoming:
I'm trying to add a call to `UserLog.end_user_activity` in our logout view and finding the nature of 2 things broken.

 1. It fails to find an open user log even though there is one when inspecting the database before calling it and
 2. The login view is called twice:

```
[WARNING] [2016-06-24 21:42:57,714] kalite: test: Had to BEGIN a user log entry, but ENDING(1)! @ 2016-06-24 21:42:57.712011
[INFO] [2016-06-24 21:42:58,224] django.request: HTTP Request /securesync/api/user/logout/ - Response: 200
[WARNING] [2016-06-24 21:42:58,629] kalite: test: Had to BEGIN a user log entry, but ENDING(1)! @ 2016-06-24 21:42:58.625851
[INFO] [2016-06-24 21:42:59,118] django.request: HTTP Request /securesync/api/user/logout/ - Response: 200
```

benjaoming:
Next bug:

1. Log in
2. Start a new browser session and login as same user.
3. Log correctly displays: `[WARNING] [2016-06-24 22:32:27,737] kalite: test: had to END activity on a begin(1) @ 2016-06-24 22:32:27.735524`
4. New `UserLog` is created, but previous log is NOT summarized in `UserLogSummary`

benjaoming:
Another bug in `UserLog`

1. Log in, a fresh UserLog is created with `start_datetime == last_active_datetime`
1. Do stuff, like watch a video or solve an exercise
1. Observe that `last_active_datetime` never changes.

This behavior results in 0 login time because a new login session will use `last_active_datetime` to calculate `total_seconds` in `UserLog`.

benjaoming:
**Sub-bug 1** can be solved by assuming some maximum duration, like an abandoned sessions duration if 0 is always set to 5 minutes. Because we don't wanna just assume the student sat at the computer all the time until logging in again... seems obviously flawed.

**Sub-bug 2** is solved by calling `update_user_activity` where it's required.

benjaoming:
**Impact assessment**: Needless to say, since we **haven't** collected data in previous versions (?), then there is no way of fixing this. But I would leave hope that the actual login count is present, and that this was an error in the coach report and not the data.

benjaoming:
The only place where I can find that `update_user_activity` is called, is in `update_video_log`, trying to confirm that this would work...

benjaoming:
Okay, the behavior has never been implemented, I found the remnants in an unused class method `update_video_log` and an intention that all calls should be made through this. They weren't, so I've re-established user logging by *always* updating the current log whenever an exercise or video is logged.

benjaoming:
@rtibbles 

Yup, it's a file we distribute to have `man kalite` on Linux. I've possibly updated `cli2man` on my local system because I was helping the maintainer test a new release. It's harmless, it works fine, try doing `main docs/kalite.1.gz` and you'll see it :)

benjaoming:
Are you good about merging this now then?

We'll only have #5181 left as a true release blocker.

rtibbles:
Let's just merge for now, even though it breaks the build. Can you manually confirm that the content rating functionality is intact?

benjaoming:
@rtibbles already confirmed manually in another PR.

radinamatic:
Looking good in deb `0.16.7b1` release on Ubuntu 16.04!

ubuntu16 ready running - oracle vm virtualbox_512

benjaoming:
Thanks for testing @radinamatic !

PR #5184, opened 2016-06-25
0.16.x into develop

benjaoming:


benjaoming:
Test failures do not make any sense compared to the fact that this is a docs change.

Also test failures are new, random failures that I haven't seen before. Could be related to content packs. @aronasorman @rtibbles

Issue #5169, opened 2016-06-17
Broken tests in 0.16.x branch


benjaoming:
I've manually checked that the feedback form is displaying, however this test is still failing.

```
[INFO] [2016-06-17 18:26:48,722] django.request: HTTP Request /api/content/khan/undefined_entry_id/ - Response: 200
    Given I open some available content # kalite/distributed/features/steps/learn_content.py:14 1.368s
    Then I see a feedback form          # kalite/distributed/features/steps/content_rating.py:40
[INFO] [2016-06-17 18:26:48,876] django.request: HTTP Request /api/videolog/ - Response: 200
    Then I see a feedback form          # kalite/distributed/features/steps/content_rating.py:40 60.567s
      Traceback (most recent call last):
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1173, in run
          match.run(runner.context)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1589, in run
          self.func(context, *args, **kwargs)
        File "/home/ubuntu/ka-lite/kalite/distributed/features/steps/content_rating.py", line 42, in impl
          feedback_form_container = find_id_with_wait(context, RATING_CONTAINER_ID, wait_time=60)
        File "/home/ubuntu/ka-lite/kalite/testing/behave_helpers.py", line 222, in find_id_with_wait
          return _find_elem_with_wait(context, (By.ID, id_str), **kwargs)
        File "/home/ubuntu/ka-lite/kalite/testing/behave_helpers.py", line 267, in _find_elem_with_wait
          EC.presence_of_element_located(by)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/selenium/webdriver/support/wait.py", line 75, in until
          raise TimeoutException(message, screen, stacktrace)
      TimeoutException: Message: 
      Stacktrace:
          at <anonymous class>.FirefoxDriver.prototype.findElementInternal_ (file:///tmp/anonymous7166039749452953088webdriver-profile/extensions/fxdriver@googlecode.com/components/driver-component.js:10770)
          at <anonymous class>.FirefoxDriver.prototype.findElement (file:///tmp/anonymous7166039749452953088webdriver-profile/extensions/fxdriver@googlecode.com/components/driver-component.js:10779)
[...]

PR #5185, opened 2016-06-25
Remove static file that should not be tracked. Add to git ignore.

benjaoming:
## Summary

Experimenting with fix of broken tests #5169


benjaoming:
Trying to rebuild without caching...

rtibbles:
This test failure makes no sense, because it is meant to time out in order to assert that there is no such element....

benjaoming:
@rtibbles the messed up thing is that it doesn't happen at all in `develop`, I was looking for an explanation and tried this out... obviously doesn't work.. so might as well close it just in case it causes a new bug in 0.16.x?

rtibbles:
Closing then.

PR #5189, opened 2016-06-28
allow for CONTENT_ROOT to be set with an environment variable,

xaiki:
## Summary

allow for CONTENT_ROOT to be set with an environment variable

## TODO

## Reviewer guidance

## Issues addressed

## Documentation

*If the PR has documentation, link the file here (either .rst in your repo or if built on Read The Docs)*

*They're nice. :)*

We use this at endlessm.com to allow sharing content whilst keeping the databases separate (we merge them on launch).

Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -281,7 +281,7 @@ else:
 # database with assessment items.
 
 # Content path-related settings
-CONTENT_ROOT = os.path.realpath(getattr(local_settings, "CONTENT_ROOT", os.path.join(USER_DATA_ROOT, 'content')))
+CONTENT_ROOT = os.getenv('KALITE_CONTENT_ROOT', os.path.realpath(getattr(local_settings, "CONTENT_ROOT", os.path.join(USER_DATA_ROOT, 'content'))))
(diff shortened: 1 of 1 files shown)


benjaoming:
Related to #5144 I guess -- and thanks for another contribution.

It seems reasonable, but if you want more, I would suggest a custom settings module, that scales much better than having to do a PR everytime ;)

Since KA Lite's settings are generally unmanagable, hardly documented and scattered, it seems like adding to worse by wrapping oddly chosen stuff in `os.getenv` :)

benjaoming:
Btw. because of the way settings are derived from CONTENT_ROOT, then there is some idea to setting it in an env var.

The method using a settings module is somewhat described in the docs, namely by explicitly setting all the derived settings as well! :)

benjaoming:
Oh sorry, didn't see this was going into `master`

benjaoming:
@xaiki - you have to make the PR for the branch named `0.16.x`, then we can have it released for 0.16.7.

Issue #5181, opened 2016-06-24
Coach reports: Mastery percent shown as 10,000%


benjaoming:
## Summary

1. New database
1. `generaterealdata` (using #5157)
1. Observe 10,000% as below

![image001](https://cloud.githubusercontent.com/assets/374612/16011936/9fe2c490-3187-11e6-92bd-268de36ceaf6.png)

## Branch or installer method

 - Branch: 0.16.x

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

Coach reports will seem rather obscure!?

benjaoming:
Bugs are fixed, however 0.16.7 is not yet out but will be soon... closing this.

PR #5192, opened 2016-06-28
Summarize all exercises' individual streak progress

benjaoming:
## Summary

 - Fixes display error in which we multiplied 100 on a summary that was already stored as a percentage in the database.
 - Furthermore, it only counted exercises solved, which seems wrong, because in that case it would only count "100%" for each exercise that was solved, and thus an average "100%" mastery would always be displayed even though other exercises were incomplete.

## TODO

- [x] N/A: Have **tests** been written for the new code? If you're fixing a bug, write a regression test (or have a really good reason for not writing one... and I mean **really** good!)
- [x] N/A: Has documentation been written/updated?
- [x] N/A: New dependencies (if any) added to requirements file

## Reviewer guidance

This is a matter of something that was super-broken before and hardly explained. I won't be able to address any comments due to absence, so someone else would have to take over the PR.

## Issues addressed

#5181

screenshot from 2016-06-28 22 55 29

### DIFF
--- a/kalite/control_panel/templates/control_panel/partials/_groups_table.html
+++ b/kalite/control_panel/templates/control_panel/partials/_groups_table.html
@@ -87,8 +87,8 @@
                                     <td>{{ group.total_logins }}</td>
                                     <td>{{ group.total_hours|floatformat }} {% trans "hour(s)" %}</td>
                                     <td>{{ group.total_videos }}</td>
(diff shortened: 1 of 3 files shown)


rtibbles:
Looks good - I had originally tried to fix these before we had added the 'latest_activity_timestamp' fields, so I think that's why this mismatch may have occurred.

Issue #5200, opened 2016-07-10
Indicate duration of video scan


benjaoming:
Since it takes 10+ hours to scan a full download of all videos, we should indicate this in the docs.

Even better, optimize the scanning function so it doesn't take 10+ hours.

kenny55d:
The 10 hour figure doesn't apply to all hardware; I've found that it only takes about that long on the raspberry Pi2 and Pi3, which are the least powered devices I've ever run ka-lite 0.16.x on. 

I guess it's down to processing, ram and perhaps storage I/O speeds, so that figure drops significantly, the more powerful the device.

jamalex:
@rtibbles and I got as far as narrowing this down to some very weird issues in peewee, when writing to the database (and doing some recursive tree updates). We must be able to knock it down further -- I may take a stab at it later today.

spatiald:
FYI On our [RACHEL-Plus](https://racheloffline.org/), it takes about 5-6 hours.

Proc: Intel Atom Processor E3815, 1.46GHz
Mem: 2GB DDR3L-1067
Drive: 8GB eMMC

jamalex:
I've got some changes in the works that (on my computer) speed things up around 9x. Still doing some testing, and will open a PR soon. Hopefully can go into a 0.16.8 patch release sometime this month.

jamalex:
On my laptop, with the full set of videos:

- Old code: 1 hour 8 min
- New code: 6 mins 24 seconds

So, the speedup looks to be around 10x. I'll open a PR.

spatiald:
That's amazing...look forward to testing. Thanks!

On Jul 11, 2016, at 01:21, Jamie Alexandre <notifications@github.com> wrote:

On my laptop, with the full set of videos:

   - Old code: 1 hour 8 min
   - New code: 6 mins 24 seconds

So, the speedup looks to be around 10x. I'll open a PR.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/learningequality/ka-lite/issues/5200#issuecomment-231644452>,
or mute the thread
<https://github.com/notifications/unsubscribe/AFfEa4YJS1bVYHZDoicnOYZMZPV8t020ks5qUdLpgaJpZM4JI3pO>
.

jamalex:
The changes are now in place in #5201.

If anyone wants to test this out prior to the prerelease, you can try temporarily replacing `kalite/topic_tools/content_models.py` with the contents of:
[...]

PR #5201, opened 2016-07-11
Optimizations to annotate_content_models for a 10x speedup

jamalex:
This is to ameliorate the slowness identified in https://github.com/learningequality/ka-lite/issues/5200. 

On my laptop, with the full set of videos:

- Old code: 1 hour 8 min
- New code: 6 mins 24 seconds

So, the speedup looks to be around 10x.

The main sources of the speedup:

- `.annotate` was super slow in peewee; now we use list comprehensions
- we recalculated shared ancestors numerous times (non-deduped recursion); now, we queue up and then recurse level-by-level, only ever updating each node at most once


aronasorman:
No concerns. Merging.

PR #5202, opened 2016-07-13
Allow logins during lockdown

Aypak:
## Summary

The Lockdown feature was still not working as expected. Students were still not able to login when lockdown activated. @benjaoming made a fix that got rid of page redirect error that was previously present when lockdown was activated

## Issues addressed
#5150 

Students were not able to login when lockdown was activated because /securesync/api/user/login was not whitelisted. Tested and working after this change.


benjaoming:
Agreeing, that seems like the likely cause, thanks for fixing and confirming functionality restored!

Issue #5140, opened 2016-06-07
picture not displaying


harishgangwar:
Pictures are not displaying while doing the math geometry exercise. 
Version : Kalite 0.16
OS : Window XP
Browser : Chrome , Firefox

![ka1](https://cloud.githubusercontent.com/assets/19798337/15856915/2711a244-2cd6-11e6-8787-39af01d0f40d.JPG)
![ka2](https://cloud.githubusercontent.com/assets/19798337/15856926/2e8cd3c2-2cd6-11e6-8a65-73b28fa4543e.JPG)

radinamatic:
On Windows 7 & 0.16.5, image displays correctly, will confirm asap if I can replicate on XP.

harishgangwar:
I am facing same issues with other exercises too. I inspected one more exercise in chrome and noted that svg file has been called and opening properly in new tab. It  seems there is some issue in fetching the svg files in kalite. 
 
![ka3](https://cloud.githubusercontent.com/assets/19798337/15930778/5663e8de-2e72-11e6-906e-714d42d64a42.JPG)

benjaoming:
@harishgangwar thanks for another great detailed issue report! Are you using the new content pack that @aronasorman uploaded?

benjaoming:
@harishgangwar 

I'm not able to reproduce this with the latest 0.16.6 alpha, but it looks from your screen shots like this should work. When you say:

> It seems there is some issue in fetching the svg files in kalite. 

...what do you mean exactly? Because it does a "200 OK" so your browser receives the SVG file. However, I see there is "text/html" in the content-type header, and when I make this request to my local Ubuntu-based ka-lite server, I receive "image/svg+xml". I'm wondering if cherrypy (the HTTP server bundled into KA Lite) uses the host system to resolve the content type and create the header.

harishgangwar:
@benjaoming  I used chrome , Safari and mozilla browser for same exercise but picture could not be displayed in any of browser. I am having the assessment zip file bundled with content pack available on http://ka-lite.readthedocs.io/en/latest/installguide/install_all.html.  
The file is there in 'inspect window' but somehow not displaying.

benjaoming:
@harishgangwar is it possible for you to install KA Lite on a non-xp device and then connect to it from the affected Windows XP browser?

That way, we can see if it's something related to HTTP content type headers and if running the KA Lite server on a different device fixes it...

harishgangwar:
[...]

PR #5218, opened 2016-07-27
Try to add well-known content types because of Windows XP

benjaoming:
## Summary

According to cherrypy docs, we can add a dictionary of content types:

https://cherrypy.readthedocs.io/en/3.2.6/refman/lib/static.html#cherrypy.lib.static.staticdir

I hope this *supplements* the already well-known mimetypes and doesn't mean that we won't get content type headers for all the other files.

@harishgangwar

## TODO

## Reviewer guidance

Let's ensure that already known MIME types are still supported :)

## Issues addressed

#5140

### DIFF
--- a/kalite/distributed/cherrypyserver.py
+++ b/kalite/distributed/cherrypyserver.py
@@ -12,6 +12,9 @@ from django.core.handlers.wsgi import WSGIHandler
 
 __all__ = ['DjangoAppPlugin']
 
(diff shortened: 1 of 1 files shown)


benjaoming:
Confirming now that content-type header is not affected for PNG image is not affected by the change:

screenshot from 2016-07-28 01 26 55

PR #5219, opened 2016-07-27
Disable unpack assessment zip command tests

benjaoming:
## Summary

It seems that annotating the content db while running BDD tests concurrently will fail. This is okay, tests run locally.

## TODO

## Reviewer guidance

None

## Issues addressed

#5216


benjaoming:
Now tested `unpack_assessment_zip` locally and found a bug which is also fixed in this PR.

benjaoming:
#5138 introduced a bug in `unpack_assessment_zip` because `prefetch=False` is now `stream=True`

benjaoming:
Tests now work, since only tests are affected by this PR, I'm assuming low-risk (no-risk) and merging.

PR #5221, opened 2016-07-27
0.16.x into develop (merge conflicts)

benjaoming:
@aronasorman please avoid bumping versions in 0.16.x for prereleases as it makes merging to develop a nuisance :)


benjaoming:
New BDD errors now occur, they seem likely, will solve in a different PR.

Issue #5032, opened 2016-03-23
`kalite manage setup` erroneously warns about installing as root user


indirectlylit:
## Summary

Following the [getting started instructions](https://github.com/learningequality/ka-lite/wiki/Getting-started), running

    kalite manage setup --username=admin --password=pass --noinput

prints:

```
-------------------------------------------------------------------
WARNING: You are installing KA-Lite as root user!
	Installing as root may cause some permission problems while running
	as a normal user in the future.
-------------------------------------------------------------------
```

However, this is not true.

## Branch or installer method

`develop` branch, OSX 10.9.5, Python 2.7.11

## How to reproduce

Check out a fresh copy of the `develop`, create a new virtualenv environment, and follow the [instructions](https://github.com/learningequality/ka-lite/wiki/Getting-started) until step 8.

This seems to stem from `setup.py`, and was already noted in the code:

```python
        # Tried not to be os-specific, but ... hey. :-/
        # benjaoming: This doesn't work, why is 502 hard coded!? Root is normally
        # '0' And let's not care about stuff like this, people can be free to
        # run this as root if they want :)
        if not is_windows() and hasattr(os, "getuid") and os.getuid() == 502:
            print(
                "-------------------------------------------------------------------")
            print("WARNING: You are installing KA-Lite as root user!")
            print(
                "\tInstalling as root may cause some permission problems while running")
            print("\tas a normal user in the future.")
            print(
                "-------------------------------------------------------------------")
            if options["interactive"]:
                if not raw_input_yn("Do you wish to continue and install it as root?"):
                    raise CommandError("Aborting script.\n")
```

radinamatic:
Could you please edit your issue  - fill in the rest of the issue template and delete non applicable parts - so we can better understand the actual problem? Thanks! :smiley:

indirectlylit:
slipped and submitted before complete. updated now.

benjaoming:
Nicely spotted, @indirectlylit.. I always wondered why someone had put 502 there.

It doesn't make sense.

On a Mac OSX, the first user is 501 and the second user is 502. Someone must have wrongly assumed that this would be the root user on *any* system!?

We should use a check like `os.geteuid() == 0`.

benjaoming:
Fixed in #5222

PR #5222, opened 2016-07-27
test getuid() == 0 instead of arbitrary 502

benjaoming:
## Summary

Seems like an old bug and some kind of untested code that's been introduced a long time ago.

## TODO

## Reviewer guidance

If you're a Mac expert the please feel free to jump in.

Otherwise please feel convinced by this:

```
➜  ka-lite git:(fix-uid-check) sudo python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> import os
>>> os.getuid()
0
```

## Issues addressed

#5032

### DIFF
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -257,11 +257,8 @@ class Command(BaseCommand):
                 "--------------------------------------------------------------------------------")
             raw_input("Press [enter] to continue...")
 
-        # Tried not to be os-specific, but ... hey. :-/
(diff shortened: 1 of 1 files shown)


benjaoming:
screenshot from 2016-07-28 18 53 41

benjaoming:
Since this works fine, I'm merging it.. it might interfere with deployments that deliberately use root, but using `--noinput` silences the prompt.

@j-schwarz just a heads up on this...

PR #5225, opened 2016-07-28
Pragma support

aronasorman:
Adds SQLite pragma support for content databases.

- Upgrades our peewee version to add pragma support for `SQLiteDatabase` calls.
- Adds a new setting in `base.py`, called `CONTENT_DB_SQLITE_PRAGMAS`, which should be a list of two-tuples. By default it's an empty list, but you can override it like so:

```python
CONTENT_DB_SQLITE_PRAGMAS = [("journal_mode", "MEMORY"),
                             ("temp_store", "2")]
```

depending on your situation. See [this link](https://www.sqlite.org/pragma.html) for a list of pragmas you can put here.

### DIFF
--- a/kalite/settings/base.py
+++ b/kalite/settings/base.py
@@ -530,6 +530,9 @@ LOGOUT_URL = "/securesync/api/user/logout/"
 # 18 threads seems a sweet spot
 CHERRYPY_THREAD_COUNT = getattr(local_settings, "CHERRYPY_THREAD_COUNT", 18)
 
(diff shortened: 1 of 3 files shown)


benjaoming:
@aronasorman tests are breaking, perhaps rebase to current 0.16.x will fix it...

aronasorman:
Rebased! Let's see what happens with the test now.

benjaoming:
Looks good, let's assume updating Peewee is a good idea and have Pragma support.

Do you have any advice for the release notes, like a link that documents its usage?

Issue #5211, opened 2016-07-26
Unclean shutdown in OS X installer


arceduardvincent:
## Summary

I can install it successfully and start the server but when I load it in browser the server will quit. 

## Branch or installer method

 - Installer: OSX
 - Version: 0.16.7b3
 - Branch: 0.16.x

## Screenshots
![f052b7f2-3e69-11e6-9087-7492e24338db](https://cloud.githubusercontent.com/assets/4099119/17150750/34435fae-53a3-11e6-96f7-1752b11597f7.png)

benjaoming:
We have to leave this as a "known issue" for now since there are so many bug fixes in 0.16.7 at this point, it makes no sense to delay it any further.

arceduardvincent:
@cryil and I discuss about this. We agree on that this issue will remain as "known issue" for now. We apologise for the delay of this fix. We will follow up this fix as soon as possible.

benjaoming:
@arceduardvincent could this be related to the way that Pyrun is built for the installer? It seems a similar issue has been solved previously regarding downloads... https://github.com/learningequality/installers/issues/100

benjaoming:
But the mystery of this issue is also a great motivational factor for: https://github.com/learningequality/installers/issues/98

arceduardvincent:
@benjaoming Yeah, maybe similar on the video downloads issue. 
However the mystery on this `Unclean shutdown`,  if we downgrade `requests>=1.0,<1.2.3`. 
The `Unclean shutdown` will not replicate.

arceduardvincent:
I think we will venture some alternatives for `pyrun` like [PEX](https://pypi.python.org/pypi/pex) or `Virtualenv` perhaps. 
What are your thoughts @cpauya?

benjaoming:
@arceduardvincent - I've just checked out [the documentation](http://docs.python-requests.org/en/v1.1.0/user/advanced/#streaming-requests) for `requests` 1.1.0 and it seems it has the `stream=True` option for `get()` which was my main concern since I just refactored all the video download stuff to use requests.

We could try to change to use 1.1.0?

The problem must somehow be in the Python libraries that requests expects and that Pyrun supplies.... brokenly...

benjaoming:
I would be okay with that, as long as we just test the basic things that invoke requests: Content pack and video downloads and the legacy assessment item download.

arceduardvincent:
@benjaoming Ok working on it now. So we can have a OS X installer release too.

[...]

PR #5226, opened 2016-07-29
Fix missing return statement in user_agent helper function

benjaoming:
## Summary

There was a missing return statement such that the `user-agent` header on various calls to our central hub was being set to `None`.

@arceduardvincent - it would be nice if you could test if this could be the cause of your mysterious pyrun crashes.

## TODO

## Reviewer guidance

N/A

## Issues addressed

Possibly #5211


arceduardvincent:
@benjaoming yes I will take a look on this and test it.
Many thanks for the help.

benjaoming:
Thanks @arceduardvincent !

Merging this now since the fix is obvious for other reasons (reporting version stats to our server!).

PR #5256, opened 2016-08-09
Enable simplified login

### LINKED ISSUE #5255: Activating simplified login results in blank login modal
## Summary

Simplified login is a feature that is supposed to allow users to login with username only. However, when this feature is activated, it produces a blank login modal and the error `Uncaught ReferenceError: setGetParamDict is not defined` in the console. This issue was previously closed in #3553 but seems to have  resurfaced. It is not clear in which release of kalite it first re-emerged.

## System information
 - Operating system: Ubuntu 16.04
 - Version: 0.16.8
 - Browser: Chrome

## How to reproduce

1.  Add SIMPLIFIED_LOGIN = True to settings.py
2. Restart kalite server
3. Attempt to log in

## Screenshots
untitled

## Real-life consequences 
When using kalite with very young learners, they often forget their passwords. It is easier to have them type in their username only in order to login. This will save alot of time and frustration for the young ones.
Thanks!

arceduardvincent:
## Summary

Hi @aronasorman, @benjaoming and @radinamatic 
When we minify the javascript in the `bundle_common.js` the `setGetParamDict` is not accessible in `set_autocomplete` function. 
I declare `setGetParamDict` inside the  `set_autocomplete` function so it can be accessible.

## Issues addressed

-Fixes #5255

screen shot 2016-08-10 at 12 00 50 am

screen shot 2016-08-10 at 12 41 15 am

### DIFF
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -345,7 +345,7 @@ var LoginView = BaseView.extend({
         return this.student_usernames.indexOf(this.$("#id_username").val()) > -1;
     },
 
-    set_login_button_state: function() {
+    set_login_button_state: function(event) {
         if (this.admin || this.check_user_in_list()) {
             this.$(".login-btn").removeAttr("disabled");
         } else {
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -278,6 +278,7 @@ var LoginView = BaseView.extend({
             facility: this.facility,
             is_teacher: false
         };
+        var setGetParamDict = require("utils/get_params").setGetParamDict;
         var url = setGetParamDict(window.sessionModel.get("USER_URL"), data);
         api.doRequest(url, null, {
             cache: true,
--- a/kalite/distributed/static/js/distributed/user/views.js
+++ b/kalite/distributed/static/js/distributed/user/views.js
@@ -320,7 +320,7 @@ var LoginView = BaseView.extend({
         // has finished loading.
         this.set_login_button_state();
         this.$(".password-btn").removeAttr("disabled");
-        this.$("#id_username").on( "autocompletesearch", this.set_login_button_state);
+        this.$("#id_username").on(this.set_login_button_state);
     },
 
     key_user: function(event) {


arceduardvincent:
This `PR` is still work in progress found out some remaining issue here after I enable the `SIMPLIFIED_LOGIN`. The login button is disabled. 
screen shot 2016-08-10 at 9 33 59 pm

arceduardvincent:
Hi @aronasorman . Can you take a look on my fix?
The image attach below demonstrate that the user drop down is now functional.
screen shot 2016-08-16 at 6 53 34 am

kalite/distributed/static/js/distributed/user/views.js
```diff
@@ -319,7 +320,7 @@ var LoginView = BaseView.extend({
         // has finished loading.
         this.set_login_button_state();
         this.$(".password-btn").removeAttr("disabled");
-        this.$("#id_username").on( "autocompletesearch", this.set_login_button_state);
+        this.$("#id_username").on(this.set_login_button_state);
```
benjaoming: That looks weird? The `on` method is missing an event label, so it will use the `set_login_button_state` as event label, which will just cause no events to be caught. It's equivalent to removing the whole line?

http://api.jquery.com/on/

benjaoming:
It seems you've found and solved the issue, wonderful :)

I have one comment/question in the review.

Also, don't mind if you squash your commits and force-push your changes. It's to have better commit histories. Try this guide: http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit

benjaoming:
Ping @arceduardvincent :)

We can have this in 0.16.9 because once #5261 is merged, we just release another patch update...

kalite/distributed/static/js/distributed/user/views.js
arceduardvincent: Hi @benjaoming I will try to fix that one.
Yeah your right about that comment. If I remove that `event`. It causing an error.

I will try to figure out to fix and fix the issue.
Thanks for the tip.

benjaoming: @arceduardvincent did you figure out which event label to use? :)

benjaoming: You could use this event trigger instead?

https://api.jquery.com/keypress/

benjaoming:
Closing, will open a new PR with problems addressed so it can be part of 0.16.9.

PR #5267, opened 2016-09-05
Release notes for 0.16.9

benjaoming:
## Summary

0.16.9 will happen!


benjaoming:
Hahahaha... despite all the attempts to improve test quality, now there's a new random test failure.. `     IntegrityError: column id is not unique`, possibly a race condition in securesync.

Issue #5277, opened 2016-09-13
Server not starting in KA-Lite 0.16.9 


yogeshmr:
## Summary

After upgrading KA-Lite from 0.16.7 to 0.16.9, the server is not starting.
![image](https://cloud.githubusercontent.com/assets/21152041/18474791/72731c56-79e1-11e6-982b-92903608457a.png)

## System information
 - Operating system:  Windows 7
 - Version: 0.16.9
 - Browser: Chrome

## The following result is obtained in the command prompt

![image](https://cloud.githubusercontent.com/assets/21152041/18474837/a27b0bf2-79e1-11e6-8eba-2d06cb27d627.png)

Lots of orphan .pyc files get created and I am deleting them and trying but still in vain.

The requests folder in the below path is empty:
C:\Python27\share\kalite\python-packages\requests

benjaoming:
Thanks for reporting quickly and with a neat traceback! :)

The problem is apparently common: http://stackoverflow.com/a/32988728/405682

Could you try the two proposed solutions?

Using your default `pip` command should install python packages in the same environment that KA Lite uses, but it's hard to say if it'll work. In case none of the two commands work, try removing the entire folder `dist-packages/requests` (not the same as the previous `python-packages`) and the install the `requests` library straight from pip.

yogeshmr:
@benjaoming 

I deleted the contents of the requests folder in the path "C:\Python27\share\kalite\dist-packages\requests" and I think that fixed the issue.

Although I tried other things given in http://stackoverflow.com/a/32988728/405682 before deleting the contents of the requests folder.

Thank you very much for helping!!!!!!!

Below provides the snapshot of the commands that I tried before and after deleting the contents of the dist-packages\requests folder:

![image](https://cloud.githubusercontent.com/assets/21152041/18479216/d4023e7e-79f1-11e6-81e3-7e24dd145ee3.png)


![image](https://cloud.githubusercontent.com/assets/21152041/18479246/f4e8aa42-79f1-11e6-9cff-db1f56da0cf2.png)

![image](https://cloud.githubusercontent.com/assets/21152041/18479294/172f88aa-79f2-11e6-9990-a8439c39285d.png)


At last, the "kalite start" command worked:

![image](https://cloud.githubusercontent.com/assets/21152041/18479323/377ebeb4-79f2-11e6-8ecc-b289dfee09db.png)

yogeshmr:
The last image in the above comment should be 

[...]

Issue #5264, opened 2016-08-29
Downgrading requests causes issues with old versions of pip


mrpau-eduard:
## Summary

Edit: @benjaoming

The downgrading of `requests` causes issues with pip both system-wide and inside virtualenvs (in which, however, pip isn't needed unless user installs additional pkgs from PyPi)

This is known to affect older versions of pip, 1.5.6 especially

The fix will arrive in 0.17.

---

The original issue here is we downgrade requests to fix the [Unclean shutdown](https://github.com/learningequality/ka-lite/issues/5211#event-747204501) on `OS X`.
Spawn issue after the downgrade [Ka-lite installed with PIP break the latter](https://github.com/learningequality/ka-lite/issues/5257).

## System information

 - Operating system: `OS X`, `Debian Jessie`
 - Version: 0.16.

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

* [Ka-lite installed with PIP break the latter](https://github.com/learningequality/ka-lite/issues/5257)
* [Unclean shutdown in OS X installer ](https://github.com/learningequality/ka-lite/issues/5211#event-747204501)

mrpau-eduard:
We need to replace `Pyrun` by [PEX](https://github.com/pantsbuild/pex) base on the spawning issue. `PEX` will isolate the virtual environment for ka-lite to prevent some issue like this. Then we need upgrade the `requests` to `requests==2.10.0`

benjaoming:
@mrpau-eduard - it's unclear to me what the issue on Debian Jessie is?

But I agree, we need to get requests upgraded ASAP because of the pip breakage.

benjaoming:
Fixed in #5285

`requests` is now in version 2.11.1 in `develop`, will be released in 0.17 - prereleased in 0.17b1 on PyPi

PR #5278, opened 2016-09-14
Update requests version to 2.10.0

mrpau-eduard:
## Summary

Since our major user is windows we need to fix this #5277 issue. Updating the **requests** will solve this issue. 

Note: 
After upgrading the requests some issue will respawn on the KA Lite for OS X. Such as [Unclean shutdown in OS X installer](https://github.com/learningequality/ka-lite/issues/5211).
PEX will be implemented soon after **0.17.x** release then the OS X issue is solve.

/cc: @benjaoming 

## TODO

- [ ] Test the upgrade process in the previous release.

## Issues addressed

- [ ]  [Server not starting in KA-Lite 0.16.9](https://github.com/learningequality/ka-lite/issues/5277)
- [ ] [Downgrading requests causes issues with old versions of pip](https://github.com/learningequality/ka-lite/issues/5264)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ peewee==2.6.4
 python-dateutil==2.4.2
 django-dbbackup==2.3.2
 ifcfg==0.9.3
-requests==1.1.0
+requests==2.10.0


requirements.txt
```diff
@@ -11,4 +11,4 @@ peewee==2.6.4
 python-dateutil==2.4.2
 django-dbbackup==2.3.2
 ifcfg==0.9.3
-requests==1.1.0
+requests==2.10.0
```
cpauya: There's a version `2.11.1` on [PyPi](https://pypi.python.org/pypi/requests/) so maybe you can use that instead?

benjaoming:
This should not target 0.16.x, see #5280

Issue #5270, opened 2016-09-07
Caught another bad automated test (continuous integration issue)


benjaoming:
## Summary

This is not a major issue, but here is yet another one of these randomly failing BDD tests.

## System information

 - Operating system: Circle CI

## Traceback or relevant snippet from server.log

```
[INFO] [2016-09-06 20:36:34,728] django.request: HTTP Request /api/coachreports/logs/ - Response: 200
    Given I am on the tabular report                            # kalite/coachreports/features/steps/coachreports.py:83 2.120s
    When I click on the Hide Tabular Report button              # kalite/coachreports/features/steps/coachreports.py:191
    When I click on the Hide Tabular Report button              # kalite/coachreports/features/steps/coachreports.py:191 0.297s
    Then I should not see the tabular report anymore            # kalite/coachreports/features/steps/coachreports.py:183
    Then I should not see the tabular report anymore            # kalite/coachreports/features/steps/coachreports.py:183 0.055s
      Traceback (most recent call last):
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1173, in run
          match.run(runner.context)
        File "/home/ubuntu/virtualenvs/venv-2.7.9/lib/python2.7/site-packages/behave/model.py", line 1589, in run
          self.func(context, *args, **kwargs)
        File "/home/ubuntu/ka-lite/kalite/coachreports/features/steps/coachreports.py", line 185, in impl
          assert_no_element_by_css_selector(context, "#displaygrid")
        File "/home/ubuntu/ka-lite/kalite/testing/behave_helpers.py", line 114, in assert_no_element_by_css_selector
          _assert_no_element_by(context, By.CSS_SELECTOR, css_value, wait_time)
        File "/home/ubuntu/ka-lite/kalite/testing/behave_helpers.py", line 96, in _assert_no_element_by
          raise KALiteTimeout
      KALiteTimeout

```

benjaoming:
Fixed in #5284

PR #5284, opened 2016-09-20
Add an extra second because it evidently may take more than a second …

benjaoming:
## Summary

Adding an extra second to wait for a button to show up, might fix unexpected timeout when asserting that element disappears when clicking a hide button
## Issues addressed

#5270


benjaoming:
After adding an extra second, incidentally the BDD test still failed. My patience is out. Old reports indicate that even bumping it to 30 seconds had given problems in the past.

Issue #5321, opened 2016-10-06
 Progress bar is missing when downloading a Kiswahili videos.


mrpau-richard:
## Summary

Progress bar is missing when downloading a Kiswahili videos.
## System information
- Mac OS X 10.11.6
- 0.16.9
## How to reproduce
1. download a Kiswahili language pack at http://pantry.learningequality.org/downloads/ka-lite/0.16/content/contentpacks/
2. Used it in KA Lite.
3. Download any Kiswahili videos.
## Screenshots

Downloading Kiswahili videos.
![screen shot 2016-10-06 at 5 28 58 pm](https://cloud.githubusercontent.com/assets/8663934/19147397/8928f70c-8bea-11e6-94c8-97546536bd2d.png)

Downloading English videos.
![screen shot 2016-10-06 at 5 31 37 pm](https://cloud.githubusercontent.com/assets/8663934/19147454/c6c236b4-8bea-11e6-8a03-708007b6ea6b.png)

/cc: @julianharty

benjaoming:
@mrpau-richard - see any javascript errors or errors in your server.log?

mrpau-richard:
@benjaoming I have this console error. 
![screen shot 2016-10-07 at 6 04 34 am](https://cloud.githubusercontent.com/assets/8663934/19174957/57e0b042-8c65-11e6-99ae-485f98b8c785.png)

benjaoming:
@mrpau-richard - you should also expand the `SyntaxError` so it's possible to actually read the traceback :)

benjaoming:
Thanks for the fix @mrpau-eduard - changes will be implemented for 0.17

mrpau-eugene:
@benjaoming -- @mrpau-eduard explained to me that the progress bar doesn't show when downloading `sw` videos. So, I checked on `updates/base.js` first to see if I can do anything there. Tried some stuffs. But to make the long story short, the English text of the string is:

    Overall progress: %(percent_complete)5.1f%% complete (%(cur_stage)d of %(num_stages)d)

and the current Swahili translation is

    Maendeleo yaliyokidhiri:%(percent_complete)5.1%%kamilika%(cur_stage)d ya%(num_stages)d

I noticed that there was a missing `f` in `5.1%` so it should've been `5.1f%`.

Having that change will now show the progress bar when downloading Swahili videos.

mrpau-eugene:
[...]

PR #5323, opened 2016-10-07
Syntax Error sprintf unexpected placeholder

mrpau-eduard:
## Summary
Hi @benjaoming and @aronasorman 
This fixes the issue of the unexpected placeholder which causes the progress bar will not show.
Reference for fmt: 
- https://golang.org/pkg/fmt/ 
- http://www.diveintojavascript.com/projects/javascript-sprintf

## Issues addressed

- #5321 

screen shot 2016-10-07 at 5 53 46 pm


benjaoming:
Thanks @mrpau-eduard - but since in practice the formatting of `%(cur_stage)f of %(num_stages)f` may now contain unlimited decimal digits, the fix has to be fixed :)

Since stages are always natural numbers, you could keep `d` and instead call `parseInt` on the values of `progress_log.cur_stage_num` and `progress_log.total_stages`.

mrpau-eduard:
Hi @benjaoming thanks for the tips and reviewing my `PR`.
I update my `PR` and implement the proper formatting that you suggested.

PR #5340, opened 2016-11-21
Change language via url redirect

needlestack:
## Summary

In multi-lingual environments, it is not user friendly for each user to change the default language manually with the drop-down box on the bottom of the page. This patch allows setting the language through a link, which then redirects the user to the requested content. We use this patch in RACHEL for our multi-lingual installations. You can see it in action here, where we link to English, French, and Spanish versions on the same server:

  http://rachelfriends.org/previews/rachelplus-en/
  http://rachelfriends.org/previews/rachelplus-es/
  http://rachelfriends.org/previews/rachelplus-fr/

It would be nice if this functionality were included in the official release.

## TODO

- No tests existed for this module - I don't know how to write tests for python/django
- I have added documentation (there was none before)
- New dependencies in this file: from django.shortcuts import redirect


benjaoming:
Thanks @needlestack - would you be able to resubmit this PR against our `develop` branch?

I think it sounds like a valuable contribution, and am really happy to see things from Rachel make it upstream :+1: 

Regarding tests, I can assist you with that once I can see the PR being applied cleanly on develop.

kalite/i18n/api_views.py
```diff
@@ -3,34 +3,52 @@
...
-            logging.debug("setting session language to %s" % lang_code)
-            request.session["default_language"] = lang_code
+    if lang_code != request.session.get("default_language"):
+        logging.debug("setting session language to %s" % lang_code)
+        request.session["default_language"] = lang_code
 
-        set_request_language(request, lang_code)
+    set_request_language(request, lang_code)
 
+    if returnUrl == '':
```
benjaoming: IMO better to say `if not returnUrl`

kalite/i18n/api_views.py
```diff
@@ -3,34 +3,52 @@
...
+    # by allowing GET requests we can use this script to switch
+    # languages on the fly with a link like this:
+    #   http://my.host:8008/api/i18n/set_default_language/?lang=en&returnUrl=/learn/khan/math
+    # That would change the language for the current user. You
+    # can also switch the default language for all users with
+    # the addition of '&allUsers=1' argument
+    if request.method == 'GET':
+        data = request.GET
+        if 'returnUrl' in data:
+            returnUrl = data['returnUrl']
```
benjaoming: This is a small thing, you don't need to bother - but in Python, camelCase is almost entirely never used for variable names :) But nevermind in this case :)

needlestack:
Ah, that makes much more sense. I have done so (I believe) here:
https://github.com/learningequality/ka-lite/pull/5342

I also made the change to "if not returnUrl" as suggested.

Let me know if I've got it right, and what the next steps are. Thanks!

Jonathan Field
jfield@worldpossible.org

benjaoming:
Thanks for making a new PR, closing this one :)

PR #5342, opened 2016-11-24
ability to change language via url redirect, added docstring

needlestack:
## Summary

(resubmitting against develop branch per [request](https://github.com/learningequality/ka-lite/pull/5340#issuecomment-262745332))

In multi-lingual environments, it is not user friendly for each user to change the default language manually with the drop-down box on the bottom of the page. This patch allows setting the language through a link, which then redirects the user to the requested content. We use this patch in RACHEL for our multi-lingual installations. You can see it in action here, where we link to English, French, and Spanish versions on the same server:

  http://rachelfriends.org/previews/rachelplus-en/
  http://rachelfriends.org/previews/rachelplus-es/
  http://rachelfriends.org/previews/rachelplus-fr/

It would be nice if this functionality were included in the official release.

## TODO

- No tests existed for this module - I don't know how to write tests for python/django
- I have added documentation (there was none before)
- New dependencies in this file: from django.shortcuts import redirect

### DIFF
--- a/kalite/i18n/api_views.py
+++ b/kalite/i18n/api_views.py
@@ -3,34 +3,67 @@ import json
 from django.conf import settings; logging = settings.LOG
 from django.utils.translation import gettext as _
 from django.views.decorators.csrf import csrf_exempt
+from django.shortcuts import redirect
 
 from .base import get_default_language, set_default_language, set_request_language
 from fle_utils.internet.classes import JsonResponse, JsonResponseMessageError
 from fle_utils.internet.decorators import api_handle_error_with_json
 
-
 @csrf_exempt
 @api_handle_error_with_json
 def set_server_or_user_default_language(request):
-    if request.method == 'GET':
-        return JsonResponseMessageError(_("Can only handle default language changes through POST requests"), status=405)
+    """This function sets the default language for either the server or user.
+    It is accessed via HTTP POST or GET.
 
+    Required Args (POST or GET):
+        lang (str): any supported ISO 639-1 language code
+
+    Optional Args (GET):
+        returnUrl (str): the URL to redirect the client to after setting the language
+        allUsers (bool): when true, set the the default language for all users,
+                         when false or missing, set the language for current user
+
+    Returns:
+        JSON status, unless a returnUrl is provided, in which case it returns
+        a redirect when successful
+
+    Example:
+        To set the current user's language to Spanish and send them to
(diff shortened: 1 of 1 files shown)


benjaoming:
This looks good, except we now have a new test failure :) This is reported in Circle CI...

```
FAIL: test_responses (kalite.distributed.tests.url_tests.AllUrlsTest)
  File "/home/ubuntu/ka-lite/kalite/distributed/tests/url_tests.py", line 138, in test_responses
    check_urls(module.urlpatterns)
  File "/home/ubuntu/ka-lite/kalite/distributed/tests/url_tests.py", line 95, in check_urls
    check_urls(pattern.url_patterns, prefix=new_prefix)
  File "/home/ubuntu/ka-lite/kalite/distributed/tests/url_tests.py", line 95, in check_urls
    check_urls(pattern.url_patterns, prefix=new_prefix)
  File "/home/ubuntu/ka-lite/kalite/distributed/tests/url_tests.py", line 127, in check_urls
    url=url, status_code=response.status_code))
AssertionError: 500 not found in [200, 302, 400, 401, 404, 405] : /api/i18n/set_default_language/ gave status code 500
```

I believe the test may be caused by calling the page by `GET` without the `lang` param set. You should probably just fail gracefully or return a 302 redirect?

You can run the failing test like this: `bin/kalite manage test distributed.AllUrlsTest`

needlestack:
Ah, good idea. I have made this change and committed to this branch.

One problem, though -- I this change tested manually, but I was not able to run the test code as directed. When trying against my branch, I got "ImportError: No module named kalite" and when I tried under the master branch I got "ImportError: No module named dbbackup". Sorry, I'm not familiar enough with python and django!

needlestack:
Not sure what that codecov/patch error means - any suggestions appreciated...

benjaoming:
@needlestack thanks that's great!! I would think that now we have the tests working, then it's safe to merge. Judging from the previous fail then the tests do reach part of the function.

codecov/patch is the metric for the code that's affected - I think it means that tests has only reached 4% of the changed lines.

Overall, it stacks up, but it means that you have contributed code where I have to rely on your assertion that it's working and my own ability to read code :)

But it's all good! I will write something for the release notes and possibly add a quick remark to ensure that future devs are able to understand why this code is there.

needlestack:
Great - thanks for your help. Hope others find this useful!

PR #5346, opened 2016-11-29
Cautiously upgrade CherryPy to 3.3.0 - fixes #5317

### LINKED ISSUE #5317: JS files are still cached in browser history after upgrade
## Summary

Coach and Teacher have been using version 0.16.5 over 4 months.

The exercise and videos stops loading after a version update (0.16.7 or 0.16.9) on the servers, this happens on the tablets which the learners use too.

But strangely on deleting the Browsing history the problem seams to get solved both on the coach server and the learner tablets. The browser used is Chrome.  

This is a big bummer :( in schools since not only the coach has to delete the browsing history but also all the tablets that has been serving KA lite needs to be clear the browser and we have around 3000+ tablets across the cities.

Is there a workaround for this? I am not sure if this can be addressed using the KA lite, could be a browser issue with cache and cookies doing their part. 
## System information

_Please specify the KA Lite version you were using and your operating system!_
- Operating system: Ubuntu 14.04LTS
- Version: 0.16.X
- Browser: Chrome
## How to reproduce
1. Use any version of KA lite over few months on Chrome browser.
2. Update the server to any later versions.
3. Access the videos to notice that the exercise does not load.
## Screenshots

Will get while we come across this problem in next coach server that we update.

This issue is prominent on the tablets that the kids have been using because of which it creates disruption in the classroom, when coach uses the updated server.

benjaoming:
## Summary

CherryPy doesn't document all changes, but this upgrade immediately meant I could see cache invalidation go back to normal, fixing #5317

Since it's an upstream bug, we don't write tests, we expect dependencies to work :)

Tested:

1. Changes in files => HTTP 200
1. Unchanged files => HTTP 304

## TODO

## Issues addressed

#5317

### DIFF
--- a/docs/installguide/release_notes.rst
+++ b/docs/installguide/release_notes.rst
@@ -17,6 +17,7 @@ New features
    :url-issue:`5342` -
    (Thanks: Jonathan Field)
 
+
 Bug fixes
 ^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


benjaoming:
Tests passing, merging

Issue #5336, opened 2016-11-15
Contentpack errors on 0.17.0b2


radinamatic:
## Summary

Some contentpack errors on 0.17 beta release.

## System information

 - Operating system: Ubuntu 16.10 64bit
 - Version: 0.17.0b2
 - Browser: Firefox

## Screenshots

Alert to update language pack right after installation, so far so good...

![0 17-cps1](https://cloud.githubusercontent.com/assets/1457929/20316157/0b7e0068-ab61-11e6-9c8d-fd21ecb8f088.png)

But no way to update once Admin is on Languages page:

![0 17-cps2](https://cloud.githubusercontent.com/assets/1457929/20316181/2219b538-ab61-11e6-9b5a-178d7783a42c.png)

Also, `Delete` button is a bit misnomer here: why would somebody delete the **only** contentpack/language installed...? 

Installed other languages, they all present the alert to update at login, but result apparently "Up-to-date":

![0 17-cps6](https://cloud.githubusercontent.com/assets/1457929/20316553/a255e5b8-ab62-11e6-82bc-771c75c79208.png)

![0 17-cps5](https://cloud.githubusercontent.com/assets/1457929/20316319/a13719a0-ab61-11e6-9df4-6eb93f083381.png)

No `pt-PT` contentpack that I did see yesterday after installing 0.16.9 on Windows XP:

![0 17-cps3](https://cloud.githubusercontent.com/assets/1457929/20316193/28f21e2c-ab61-11e6-98ff-fadf00102b20.png)

Swahili, Zulu & Xhosa language packs have no available videos for download, and load an empty topic tree in the sidebar:

![0 17-cps4](https://cloud.githubusercontent.com/assets/1457929/20316221/4f8d8062-ab61-11e6-920a-37a79d842f9e.png)

#### No available subtitles anywhere!!!

Not sure if this should go in the separate issue?

benjaoming:
Regarding the delete button, please see #5232  and #4864

benjaoming:
All of the content packs are just copied raw from 0.16, I think (but haven't investigated) that this is somehow related, i.e. the old version tag in the JSON metadata prompts the UI javascript to assume an upgrade is available.

benjaoming:
...but we should start building new content packs for 0.17 testing, this could be done as soon as we ensure the current state of things.. I don't feel too on top of the latest content pack updates for various language bugs and whether they are done. But we could also decide to move all efforts to 0.17 language packs.

[...]

Issue #5318, opened 2016-10-04
Ship w/ empty content.db, handle blank default installations


benjaoming:
## Summary

We ship a broken default content pack (English content.db w/o assessment items or subtitles, means that you can click exercises but they're broken) - that alone is a solid enough reason to stop building this content pack and instead fix KA Lite.

The other reason is the 90 MB of almost meaningless data (unpacked, ~50 MB download size) that we bundle into the source distribution. This used to make sense, but since it's now entirely overwritten by the content pack, it seems like an irresponsible waste of resources IMO.

This means we should auto-redirect to the Manage tab when user hasn't installed any content.db.

Issue about removing en-minimal.zip from content pack maker:
https://github.com/fle-internal/content-pack-maker/issues/40

## TODO

 - [ ] Make tests etc. function with a fixture based on an empty content.db
 - [ ] The whole app should function without ANY content.db at all
 - [ ] Add sensitive messages explaining the lack of a content db
 - [ ] Consolidate the display of installed content pack, as it probably suffers from some adjustments made to accommodate the legacy "minimal" contentpack

benjaoming:
+ the danger of having wrong versions of the bundled English minimal content pack means that there is no "Upgrade" button visible unless we edit metadata on the server Manually. So we can't release without addressing this.

benjaoming:
It seems that the database and possibly also subtitle files are affected by upgrades, I'm suspecting that the pre-bundled minimal content pack is to blame, but should investigate further.

As of now, it seems certain that when I switch from 0.17b2 with the latest full `en` content pack with correct subtitle annotations and to `develop`, I get all the subtitles wrongly annotated, something that looks like a pre-bundled minimal content.db.

benjaoming:
The .whl source files have reduced from `52 MB` til `31 MB`, expecting similar results in debian packages and installers because they contain the source distribution in which the English content.db is overwritten after installation.

It should also have a performance-wise impact because content annotation no longer takes place twice (at first setup, then after content pack installation).

benjaoming:
Fixed in  #5350

PR #5350, opened 2016-12-01
[WIP] Empty contentpack for distribution and testing

benjaoming:
## Summary

We should finally kill off the English content pack, as the only purpose it serves is to have a function content tree in a blank installation -- for which users should download a full English content pack, anyways.

This will also greatly improve test speeds AFAIK.

## TODO

- [x] Will need to deal with whatever assumptions our tests have been based on by writing a bit of fixtures on top of the empty content db.
- [x] Make sure the Language tab communicates that it's an *empty* English language pack.
- [x] Make sure the Language tab tells users downloading other languages that they MUST install the English language pack.
- [x] ~After factoring out illegal `<a>` elements, make sure `cursor:pointer` is restored~ Restored because of accessibility concerns, @radinamatic look, you didn't even have to point it out :)
- [x] Ensure that the test content.db does not get rebuild before every BDD test, just build it once.

## Reviewer guidance

Please comment if you have any concerns about testing with no content data :) IMO we need to speed up the tests significantly, so going from 92MB test db to ~0 MB is pretty awesome.

## Issues addressed

#5336 #5318


benjaoming:
All BDD tests are actually working now, other tests are failing because they haven't got a content.db with fixtures.

benjaoming:
Update on English content pack vs. empty content.db:

If there is an empty content.db (i.e. no content packs installed), the user will now be displayed this warning regardless of the situation. For instance, if you have 1 other language, it will still show up. I think this is the intention of all content packs as they do not contain any static exercise data files.

screenshot from 2016-12-11 19-34-50

PR #5364, opened 2016-12-16
Fix logging issue on windows

mrpau-richard:
## Summary

Hi @benjaoming & @radinamatic  This fixes #5057

@radinamatic  Here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMU3ppRGl4RmR6QTg) for testing the KA Lite windows logging.

PR #5368, opened 2016-12-19
Windows logging

benjaoming:
## Summary

Adding a few things to #5364


benjaoming:
Apparently after moving the probing and printing of server address, `kalite start` now returns before it's actually running. That's why `kalite start && kalite status` would fail.

PR #5378, opened 2016-12-27
Fix format issue of sprintf unexpected placeholder.

mrpau-eduard:
## Summary
Hi @radinamatic and @benjaoming this fixes issue downloading videos in `sw` language packs.
screen shot 2016-12-27 at 1 50 01 pm

## Issues addressed
#5323

See image below for demonstration for this quick fix.
screen shot 2016-12-27 at 3 37 29 pm

### DIFF
--- a/kalite/updates/static/js/updates/updates/base.js
+++ b/kalite/updates/static/js/updates/updates/base.js
@@ -196,7 +196,7 @@ function updateDisplay(process_name, progress_log) {
         select_update_elements(process_name, ".progressbar-current").progressbar({value: 100*progress_log.stage_percent});
         select_update_elements(process_name, ".progressbar-overall").progressbar({value: 100*progress_log.process_percent});
 
-        select_update_elements(process_name, "#stage-summary").text(sprintf(gettext("Overall progress: %(percent_complete)5.1f%% complete (%(cur_stage)d of %(num_stages)d)"), {
+        select_update_elements(process_name, "#stage-summary").text(sprintf(gettext("Overall progress: %(percent_complete)5.1f%% complete (%(cur_stage)d of %(num_stages)d) "), {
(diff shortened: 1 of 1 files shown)


kalite/updates/static/js/updates/updates/base.js
```diff
@@ -196,7 +196,7 @@ function updateDisplay(process_name, progress_log) {
         select_update_elements(process_name, ".progressbar-current").progressbar({value: 100*progress_log.stage_percent});
         select_update_elements(process_name, ".progressbar-overall").progressbar({value: 100*progress_log.process_percent});
 
-        select_update_elements(process_name, "#stage-summary").text(sprintf(gettext("Overall progress: %(percent_complete)5.1f%% complete (%(cur_stage)d of %(num_stages)d)"), {
+        select_update_elements(process_name, "#stage-summary").text(sprintf(gettext("Overall progress: %(percent_complete)5.1f%% complete (%(cur_stage)d of %(num_stages)d) "), {
```
benjaoming: Wouldn't it work to escape the parenthesis instead, `\)` ? That way, you don't affect a string that's translated.

mrpau-eduard: Hi @benjaoming thanks for the tips. I just replace the `White space` with a scape(`Newline`).
Can you take a look again?

benjaoming: What I mean is: If you don't change the string as it's read by the translation msg collector, the change might not affect translators. Adding a blank space or newline will definitely change the translation msg. Adding an escape character might not.

Is it because it doesn't work with an escape character?

benjaoming:
@mrpau-eduard opened another PR with the suggested change, thanks for finding the issue!

PR #5389, opened 2017-01-19
Update docs regarding current roadmap

benjaoming:
## Summary

We should make our strategy visible to community.

I'm not sure if this is the best way, but at least it'll be better than nothing (I don't think we've really mentioned our thoughts for the future development of KA Lite anywhere in public?)

We should also encourage more community members to join the development since KA Lite really depends on community - for those who want to continue their deployments of KA Lite and cannot move to Kolibri.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,11 +64,13 @@ browsers, operating systems etc. We are also still optimizing regarding
 performance issues.
 
 If you are creating a new deployment at this very moment, feel assured that
-KA Lite is still alive and will be maintained for the rest of 2017.
+KA Lite is still alive and will be maintained for the rest of 2017, after which
+point we will be recommending that you migrate to Kolibri.
 
-In case you need KA Lite to continue to grow its features, you should join the
-community. In other words, we (Learning Equality) encourages you (community
-members), to feel empowered and take responsibility for the future of KA Lite.
+In the meantime, if you need new features in KA Lite, we welcome you to join
+the community and contribute. In other words, we (Learning Equality) encourages
+you (community members), to feel empowered and take responsibility for the
+future of KA Lite.
 
 
 Connect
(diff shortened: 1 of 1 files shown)


radinamatic:
I tota11y agree @benjaoming, nice write-up! Maybe we should be thinking of making a blog post on this subject, since we are nearing Kolibri release date? 
@jamalex

README.rst
```diff
@@ -49,6 +49,28 @@ Get involved!
 -  Read more about the project's motivation at `Introducing KA Lite, an offline version of Khan
    Academy <http://jamiealexandre.com/blog/2012/12/12/ka-lite-offline-khan-academy/>`__.
 
+Roadmap
+-------
+
+During 2017, Learning Equality is launching the successor of KA Lite. It's
+called `Kolibri <http://github.com/learningequality/kolibri>`__ and will have
```
jamalex: Wondering whether it might be better to link to https://learningequality.org/kolibri/ for now, to avoid ambitious people trying to install from Github and then being disappointed (no channels, etc).

README.rst
```diff
@@ -49,6 +49,28 @@ Get involved!
...
+educational sources.
+
+Because of the reported popularity of KA Lite, we are continuing our
+efforts to support deployments by providing fixes problems that may occur and
+thus directly effect current usage. For instance issues related to new
+browsers, operating systems etc. We are also still optimizing regarding
+performance issues.
+
+If you are creating a new deployment at this very moment, feel assured that
+KA Lite is still alive and will be maintained for the rest of 2017.
```
jamalex: perhaps add ", after which point we will be recommending that you migrate to Kolibri"

README.rst
```diff
@@ -49,6 +49,28 @@ Get involved!
...
+
+In case you need KA Lite to continue to grow its features, you should join the
```
jamalex: Perhaps "In the meantime, if you need new features in KA Lite, we welcome you to join the community and contribute."

Issue #5319, opened 2016-10-04
django-tastypie for 0.17 cannot be a dependency link


benjaoming:
## Summary

Seems that some old version of pip will mis-behave, and the dependency links are scheduled to be removed in Pip 9.
## System information
- Version: 0.17
## Traceback or relevant snippet from server.log

```

pip install ka-lite --pre

Collecting ka-lite

  Downloading ka_lite-0.17.0b1-py2-none-any.whl (54.7MB)

    100% |████████████████████████████████| 54.7MB 19kB/s 

Collecting cherrypy==3.2.2 (from ka-lite)

  Downloading CherryPy-3.2.2.zip (464kB)

    100% |████████████████████████████████| 471kB 969kB/s 

Collecting django-tastypie==0.12.2dev0 (from ka-lite)

  Could not find a version that satisfies the requirement django-tastypie==0.12.2dev0 (from ka-lite) (from versions: 0.9.6, 0.9.7, 0.9.9, 0.9.10, 0.9.11, 0.9.12, 0.9.13, 0.9.14, 0.9.15, 0.9.16, 0.10.0, 0.11.0, 0.11.1, 0.12.0, 0.12.1, 0.12.2, 0.13.0, 0.13.1, 0.13.2, 0.13.3)

No matching distribution found for django-tastypie==0.12.2dev0 (from ka-lite)
```
## How to reproduce

Happens on OSX

radinamatic:
@benjaoming 
Hmmm, strange because I already had the 0.17 installed previously on that Ubuntu machine, and I can't remember it gave me the same issue... :thinking:

benjaoming:
You'll only see the error on pip 9 + `ka-lite` (not `ka-lite-static`).

radinamatic:
pip version 8.1.2 seems to fail too...
```
ubmooc@ubmooc-VirtualBox:~$ pip --version
pip 8.1.2 from /usr/lib/python2.7/dist-packages (python 2.7)
ubmooc@ubmooc-VirtualBox:~$ pip install ka-lite --pre
Collecting ka-lite
  Downloading ka_lite-0.17.0b3-py2-none-any.whl (32.0MB)
    100% |████████████████████████████████| 32.0MB 54kB/s 
Collecting requests==2.11.1 (from ka-lite)
  Downloading requests-2.11.1-py2.py3-none-any.whl (514kB)
    100% |████████████████████████████████| 522kB 2.0MB/s 
[...]

PR #5391, opened 2017-01-19
Remove artifacts from dependency links

benjaoming:
Dependency links are not support in Pip 9, which is already released. But we haven't needed them since #5319 was solved.

 - Removing this should ensure that `ka-lite-static` builds on Pip 9.
 - Also added a make target `sdist` for installers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,9 @@ release: dist man
 	echo "Uploading dist/* to PyPi, using twine"
 	twine upload -s --sign-with gpg2 dist/*
 
+sdist: clean docs assets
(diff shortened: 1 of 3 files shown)


benjaoming:
This should be safe to merge - if it causes issues, they will show up during building of installers, however since Pip 9 is already out, it's also quite critical to get rid of this deprecated call.

radinamatic:
Tested the `pip install` this morning, all's well! 👍 

virtualbox_ubuntu 16 10_23_01_2017_14_39_00

virtualbox_ubuntu 16 10_23_01_2017_14_41_52

Issue #5348, opened 2016-11-30
Device registration: User message in case of connection failure


radinamatic:
## Suggested change

Old:

``` 
 <p>{% trans "The device needs to have internet access when first being registered. Please ensure it is connected to the internet, and then" %} <a href="." class="refresh-link">{% trans "refresh this page" %}</a>.</p>
``` 

New:

``` 
<p>{% trans "The device needs internet access in order to be registered. Please ensure you are connected to the internet, and then" %} <a href="." class="refresh-link">{% trans "refresh this page" %}</a>. If you are still getting this message, it might indicate a temporary problem with the central server. Please wait a few minutes and try again later.</p>
``` 

...in `python-packages/securesync/templates/securesync/register_public_key_client.html`

## Summary

~KA Lite seems unable to connect to register the device in order to download the videos & languages. Refreshing the page has no effect, and the machine is correctly connected to Internet (see the loaded elpais.com in the other window).~

~Is `staging.learningequality.org` where the device register happens? It seems to be running OK, so the culprit might be other.~

## Screenshots

![virtualbox_windows 7_30_11_2016_13_46_49](https://cloud.githubusercontent.com/assets/1457929/20752815/9486453e-b703-11e6-90f5-01631c996870.png)

benjaoming:
Just experienced exactly the same on FF, Ubuntu 0.16.9 installation.

@radinamatic are you able to reproduce this for 0.16.9 in all cases? It might be a server-side issue..

benjaoming:
Also, try running the server with `kalite start --foreground` in order to get more logging and see what it says at the exact moments of registration.

radinamatic:
Apparently, one-click registration on WXP is working today:

![virtualbox_wxp_14_12_2016_14_47_52](https://cloud.githubusercontent.com/assets/1457929/21184398/9b790480-c20c-11e6-8e44-d70819a63efe.png)


![virtualbox_wxp_14_12_2016_14_49_17](https://cloud.githubusercontent.com/assets/1457929/21184404/a136de74-c20c-11e6-816c-739c1f256849.png)

radinamatic:
Works on Ubuntu too, fresh 0.16.9 install:

![virtualbox_gubuntu_14_12_2016_15_04_36](https://cloud.githubusercontent.com/assets/1457929/21184915/d5877f56-c20e-11e6-9c99-993c9484e8dd.png)

[...]

PR #5403, opened 2017-02-03
Add additional logging during AJAX request failures + CSP middleware for debugging

benjaoming:
## Summary

Since the actual issue was due to upstream HTTP headers and `OPTIONS` returning 405, there was nothing to fix.

Instead, I leave us with some:

1) Removes some homebrewed `JsonAsHTML` debug middleware that was really breaking stuff in debug mode.
1) Additional JS console output in case of failure
1) Fixing broken debug-toolbar due to fetching online resources
1) A CSP middleware that's only enabled with DEBUG=True and blocks online sources so we won't make mistakes in the future
1) Disables `tota11y` by default, but enables it with `USE_TOTA11Y=True` in settings @radinamatic 

## TODO

## Issues addressed

#5348

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,7 @@ assets:
 	node build.js
 	KALITE_HOME=.kalite_dist_tmp bin/kalite manage syncdb --noinput
 	KALITE_HOME=.kalite_dist_tmp bin/kalite manage migrate
+	rm -rf kalite/database/templates/
 	mkdir -p kalite/database/templates/
 	cp .kalite_dist_tmp/database/data.sqlite kalite/database/templates/
 	bin/kalite manage retrievecontentpack empty en --foreground --template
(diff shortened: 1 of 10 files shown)


benjaoming:
```
$.ajax({url: "http://pantry.learningequality.org/downloads/ka-lite/0.17/content/contentpacks/all_metadata.json"}).success(function(resp) {console.log("succ")}).fail(function(resp) {console.log("fail: " + resp.state())})
```

> fail: rejected

Interesting.. it fails with "rejected". I have cleared all my browser cache in case it had to do with cached OPTIONS 405s.

However, this works:

```
$.ajax({url: "http://mysafeinfo.com/api/data?list=englishmonarchs&format=json"}).success(function(resp) {console.log("success")}).fail(function(resp) {console.log("fail: " + resp.state())})
```

> success

So I'm assuming this has to do with a problem with FF or our setup. I'm still not convinced about our Nginx settings :)

benjaoming:
Well well well, trying a clean FF profile makes it work.

I can't make it work by clearing my cache with CTRL+SHIFT+DEL, just a heads up to anyone encountering the issue.

Issue #5401, opened 2017-02-02
Device registration not compulsory to download videos and languages - redirect to device registration page not working


radinamatic:
## Summary

Contrary to what used to happen in 0.16 releases, 0.17 does not require the user to register the device in order to download videos and languages.

## System information

*Please specify the KA Lite version you were using and your operating system!*

 - Operating system: all
 - Version: 0.17b5
 - Browser: all

## How to reproduce

1. Fresh install 
1. Login as Admin
1. Go to Video or Language tab - you are not redirected to Device Registration

## Screenshots

(I'll try to grab a screencast)

## Real-life consequences 

Users not registering devices and our stats are not up-to-date.

benjaoming:
Confirmed - will try to see if I can plug the hole by adding a good old redirect on the language and video update pages. There's no harm in this, since these pages are almost useless on a device that doesn't have a connection.

In case a device has videos and content packs, but no internet connection and isn't registered, then the pages are useful for deleting stuff. But it should hardly matter... if someone deploys by copying `.kalite` folders without registering devices, I can't see why they would make use of the delete function from UI afterwards.

radinamatic:
Unable to register with today's 0.17b5 installer on Windows 7.

There's no alert message at all, even though the device is reported as not registered, refreshing the register page does nothing, no errors in the console...

![w7b](https://cloud.githubusercontent.com/assets/1457929/22784670/407497ba-eed1-11e6-8e9b-b50b9866641a.gif)

benjaoming:
Fixed in #5404  :)

benjaoming:
@radinamatic the actual explanation is that _if_ a device is offline, we don't require registration. Since the online detection part was broken due to a redirect on `kalite.learningequality.org`, and the function that was supposed to handle and assess the redirect wasn't configured to do so, the function `i_am_online` assumed the device was offline even though it was online :)

radinamatic:
Working correctly on the latest `.deb`:

#### Online, trying to download a language:

![virtualbox_ubuntu 16 10_10_02_2017_11_03_38](https://cloud.githubusercontent.com/assets/1457929/22822833/4c32b162-ef81-11e6-92e0-ad8d8ddaa5d7.png)

#### Offline, trying to download a language:

[...]

PR #5404, opened 2017-02-09
Simplify am_i_online which broke because of a 301 on kalite.learninequality.org

benjaoming:
## Summary

It's fine that kalite.learningequality.org redirects. However, with the mess that `am_i_online` was, it wasn't actually doing the test right.

So I simplified it to a function that doesn't accept arguments.

Also, it doesn't check the root URL since this one redirects, but looks for a well-known API endpoint... which hopefully, we won't remove :)

 * Outputs logging from `fle_utils`
 * Increases reliability of `am_i_online`
 * Add release note

## TODO

## Reviewer guidance

Not expecting anyone to jump in on this, but feel free, @jamalex and @aronasorman - since this is a last-minute release blocker that will ship shortly.

## Issues addressed

#5401


benjaoming:
As a curiosity this didn't show up with `DEBUG=True`, because it uses `staging.learningequality.org` @radinamatic

Issue #5123, opened 2016-05-25
Make error messages friendlier


benjaoming:
## Summary

In #4875 @j-schwartz noted this important issue in KA Lite:

> From the perspective of a first time computer user, or at least first time KA-lite user, you have 90% fear of breaking something and 10% excitement.
> 
> There are lots of kids I have seen, who if they saw that sorry, something has gone wrong message, after first watching a video, will immediately close the browser and walk away for fear they have done something wrong.
> 
> The investment in computers / ka-lite / time on hardware is massive relative to school budgets, and the fear of breaking something is one of the primary obstacles I see us trying to overcome to increase adoption.
## Solving this issue

There are error messages in many places, both Javascript and Python. But we should especially target those that can face students.

This happens when error handling Javascript exceptions, missing content, broken exercises.

But we still cannot drop exception information. So make sure this goes in the Javascript console if it's a JS error or in the application log if it's a Python error.
## Example

![image](https://cloud.githubusercontent.com/assets/9270388/15523994/6c038a10-21d3-11e6-87d0-bf17f7088ba3.png)

Message should be displayed in a more graphical and with a friendlier text, here's a suggestion:

```
Hey there! Unfortunately a technical error has occured which is not your fault! 
This exercise had problems displaying itself. But don't worry, there are over
25,000 exercises, and a dedicated team is always looking to make each and
every one of them function. Let's hope in a future update, that this exercise will
be fixed.
```
## Branch or installer method
- Branch: develop
## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

Can have very severe consequences even if it's an anticipated error.

j-schwartz:
This is great. Thanks guys!  I might add somewhere in there, "this is not
your fault"

benjaoming:
Added that in the example, once someone starts working on this, I trust that they will also work on the contents of the error messages themselves!

radinamatic:
Very good idea, I'm proposing a slightly more simple phrasing:

```
Hey, the error you are seeing **is not your fault**! 

This exercise may have a technical problem, and we will work on to have it corrected soon.

Please don't worry, and select another from 25,000 exercises in here to continue learning!
```

Since we are aiming this message for the young learners and our content is crowd-translated, let's make the English original as simple as possible 😉

** Ideally, the **is not your fault** would be styled differently.

benjaoming:
[...]

PR #5405, opened 2017-02-10
Additional fixes for device registration

benjaoming:
## Summary

Due to various issues reported, I'm trying to affirm that device registration works through a bit of fumbling.

 * Add an additional test case to clear that *online* presence is correctly detected.

## Issues addressed

#5401 
#5123


benjaoming:
New error page (before clicking textarea)

screenshot from 2017-02-10 20-29-44

After clicking text area:

screenshot from 2017-02-10 20-30-23

radinamatic:
:heart_eyes:

benjaoming:
Merging this, because then at least we can know about the other issue on Mac/Windows that we haven't been able to troubleshoot through logs etc.

PR #5406, opened 2017-02-13
Remove debug exception expression, + test case

benjaoming:
Sorry about that @mrpau-eduard @mrpau-richard 

I was rushing along in some kind of fantasy world where we had test cases for pages like the registration page.

facepalm

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -20,7 +20,7 @@ test:
     - make assets
     - make docs
     - kalite start --traceback -v2
-    - sleep 4s  # Necessary for server to be ready
+    - sleep 6s  # Necessary for server to be ready
     - kalite status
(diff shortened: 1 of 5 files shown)


benjaoming:
@radinamatic did you report suspicion that `i_am_online` is unstable?

At least, the latest test failure seems to support the claim:

```
FAIL: test_online (kalite.updates.tests.regression_tests.RegistrationRedirectTestCase)
  File "/home/ubuntu/ka-lite/kalite/updates/tests/regression_tests.py", line 28, in test_online
    self.assertEqual(am_i_online(), google_is_online)
AssertionError: False != True
```

The above test has worked every time since the change was made, now fails...

Issue #5407, opened 2017-02-13
Sidebar list doesn't scroll on touch


nedrichards:
## Summary

*When using KA Lite on a touchscreen device touch scroll events drop through to the underlying page rather than scrolling long sidebar lists (mouse wheel scrolls work)*

## System information

*Please specify the KA Lite version you were using and your operating system!*

 - Operating system: Linux
 - Version: 0.16.5
 - Browser: Chromium 55, Google Chrome 56

## How to reproduce

1. Visit a URL such as http://localhost:8008/learn/khan/humanities/medieval-world/latin-western-europe/romanesque1/romanesque-art/
2. Attempt to scroll the arts and humanities list.
3. Notice that the page behind the overlay scrolls.

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

*It makes using KA Lite on an interactive whiteboard in a classroom environment somewhat difficult.*

benjaoming:
Thanks for reporting this!

@66eli77 you being the sidebar expert, would you be able to fix or recommend what to do in this case? I don't have a touch device in my development kit, so I find it a bit tedious to reproduce and solve.

benjaoming:
@nedrichards Did you btw notice the little overlay scrollbar? It's possible to drag it using the mouse, so I was thinking if it's possible to also drag it on a touch device?

![screenshot from 2017-02-13 19-31-49](https://cloud.githubusercontent.com/assets/374612/22897238/5074d634-f223-11e6-85f0-ff5e8f3a8394.png)

benjaoming:
Also of interest: If I use the mouse's scroll wheel on the contents of a sidebar panel, it scrolls correctly. However, scrolling with the mouse wheel having the cursor placed on top of the overlay scrollbar of a sidebar panel scrolls the full window.

nedrichards:
I was unable to tap or drag the overlay scrollbar with my finger (or with a handy nearby stylus)

benjaoming:
Roger - I think this sounds relevant for a patch release.

66eli77:
I cannot reproduce the issue on android 4.4.2 tablet(chrome browser 50)  running kalite 0.16.9, will try 0.16.5

benjaoming:
@66eli77 - I think the sidebar is something that hasn't been changed much since 0.15?

@nedrichards could you give some more details about which OS your tables etc are running?

66eli77:
[...]

PR #5410, opened 2017-02-14
prevent underneath body element from scrolling when scroll the sideba…

66eli77:
…r or sidebar-fade area

## Summary

Fix the issue #5407 
Only tested on Mac chrome and android chrome browsers, would need more testing on different devices, especially on Linux where the issue is originally discovered.

## TODO

- [ ] Tested on Linux
- [x] Tested on Firefox
- [x] Tested on Safari
- [x] Tested on Edge
- [x] Tested on IE10

### DIFF
--- a/kalite/distributed/static/css/distributed/sidebar.less
+++ b/kalite/distributed/static/css/distributed/sidebar.less
@@ -37,7 +37,7 @@ Nav Panel
 }
 
 div.sidebar-fade {
-    position: absolute;  /* makes the div go into a position that’s absolute to the browser viewing area */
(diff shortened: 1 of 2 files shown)


66eli77:
A separate issue, seems clicking on sidebar item not working on IE9. @benjaoming does kalite 0.17 support IE9?

benjaoming:
@66eli77 yes, we do :)

http://ka-lite.readthedocs.io/en/latest/installguide/system_requirements.html

benjaoming:
It sounds like from @nedrichards that it's been tested for Edge, so I've marked it as Done.

Merging now, this will be part of the 0.17.1 release.

Issue #4104, opened 2015-07-15
Factor out PROJECT_PATH


benjaoming:
The notion of `PROJECT_PATH` was tied to the old one-dimensional single-dir-based distribution where the `kalite` package was in limbo between being a django app and a django project.

It didn't really make itself understandable, neither, because the project was either the top-most level `/` or the `/kalite/` dir.

Everywhere that `PROJECT_PATH` is used, it reveals some level of implicit structure. We need to go through all the occurrences and replace them with the new settings strategy #4054

On a secondary note, PROJECT_PATH was actually never a "setting", it was never supposed to be changed.

```
✗ ack-grep --python PROJECT_PATH kalite

kalite/playlist/management/commands/checkplaylistentries.py
22:        all_playlists = json.load(open(os.path.join(settings.PROJECT_PATH, 'playlist/playlists.json')))

kalite/basetests/tests.py
6:PROJECT_PATH = os.path.realpath(os.path.join(THIS_PATH, "..", ".."))
11:        os.path.join(PROJECT_PATH),
12:        os.path.join(PROJECT_PATH, "kalite"),
13:        os.path.join(PROJECT_PATH, "python-packages"),
315:        content_path = os.path.realpath(os.path.join(PROJECT_PATH, "content"))
320:        khan_path = os.path.realpath(os.path.join(PROJECT_PATH, "data", "khan"))
325:        khan_path = os.path.realpath(os.path.join(PROJECT_PATH, "data", "khan"))
351:        scripts_path = os.path.realpath(os.path.join(PROJECT_PATH, "scripts"))

kalite/distributed/management/commands/apacheconfig.py
34:            "project_path": settings.PROJECT_PATH,

kalite/distributed/management/commands/nginxconfig.py
76:        self.stdout.write(config_template % {"root_path": os.path.realpath(settings.PROJECT_PATH + "/../"),

kalite/distributed/management/commands/setup.py
359:        # call_command("clean_pyc", interactive=False, verbosity=options.get("verbosity"), path=os.path.join(settings.PROJECT_PATH, ".."))
468:                        os.path.join(settings.PROJECT_PATH, "..", "bin", kalite_executable))
471:                        os.path.join(settings.PROJECT_PATH, "..", "bin", "windows", "kalite.bat"))

kalite/distributed/management/commands/screenshots.py
184:            self.output_path = os.path.join( os.path.realpath(os.path.join(settings.PROJECT_PATH, '..')),

[...]

PR #5429, opened 2017-04-04
Remove legacy code, PROJECT_PATH and related test case code_tests

benjaoming:
## Summary

Removes broken code in 'setup' management command

Targeting for 0.17.1 because there are no current plans for 0.18, and this is just a legacy cleanup.

## TODO

## Reviewer guidance

As noted in the release notes, `PROJECT_PATH` was not even configurable. It was used wrongly, too. Its default value was the `KALITE_HOME` environment, but some places of usage assumed it would be a "project root", which is a remnant of when KA Lite was distributed as a static directory.

## Issues addressed

#4104


kalite/distributed/management/commands/setup.py
```diff
@@ -477,15 +477,10 @@ def handle(self, *args, **options):
...
 
             kalite_executable = 'kalite'
-            if not spawn.find_executable('kalite'):
-                if os.name == 'posix':
-                    start_script_path = os.path.realpath(
-                        os.path.join(settings.PROJECT_PATH, "..", "bin", kalite_executable))
-                else:
-                    start_script_path = os.path.realpath(
-                        os.path.join(settings.PROJECT_PATH, "..", "bin", "windows", "kalite.bat"))
-            else:
```
benjaoming: This part of the cleanup makes sense because `PROJECT_PATH` wasn't set to a directory containing the `bin/` directory, but rather then `~/.kalite` directory as we discontinued having the source checkout as user data directory.

benjaoming: In other words: It's removing code that would be broken anyways.

PR #5438, opened 2017-04-10
Add logging to setup and initialize_kalite

benjaoming:
## Summary

This also (once again) prints the server address at command line ( a regression from #5364 )

## TODO

If not all TODOs are marked, this PR is considered WIP (work in progress)

- [ ] Has documentation been written/updated?
- [ ] Have you written release notes for the upcoming release?

## Reviewer guidance

Let me know if there are problems in this approach. I don't see much of a difference, except I forced ALL of the loggers in `kalite.distributed.management.commands` to user a logger WITHOUT formatting, expecting their outputs to be generally targeted for command line displaying.

However, they are also subject to be printed in the server log.

## Issues addressed

#5408

Issue #5408, opened 2017-02-14
Put all output from `kalite manage setup` in log


benjaoming:
## Summary

Some of the output during setup is just added via `print` statements and `sys.stderr`, we should log everything the proper way so it goes into `server.log`.

## System information

All systems

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

It's hard to properly debug problems pertaining initial setup. Currently, we have an issue with certificates, and it's impossible to tell if they've been generated or not.

benjaoming:
Fixed in #5441

PR #5441, opened 2017-04-12
 Add logging to setup and initialize_kalite (2)

benjaoming:
Taking over from #5438

## Summary

This also (once again) prints the server address at command line ( a regression from #5364 )

## TODO

## Reviewer guidance

Let me know if there are problems in this approach. I don't see much of a difference, except I forced ALL of the loggers in `kalite.distributed.management.commands` to user a logger WITHOUT formatting, expecting their outputs to be generally targeted for command line displaying.

However, they are also subject to be printed in the server log.

## Issues addressed

#5408

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,7 +17,8 @@ test:
     - sudo mv geckodriver /home/ubuntu/bin
     - export PATH="$PATH:/home/ubuntu/bin"
   override:
-    - make assets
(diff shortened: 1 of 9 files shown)


benjaoming:
Doesn't seem like it's possible to trick Codecov into a new coverage report, leaving it here.

Issue #5473, opened 2017-06-16
Build artifacts for KA Lite using Buildkite.


mrpau-eduard:
## Summary
Build a `python packages ` and `installers` so it will easily to grab. 
This will allow user and developer to test the installer once they have pull request and commits on there `PR`.

## Checklist:
- [ ] Python Packages
- [ ] Pex file
- [ ] Windows Installer
- [ ] Debian Installer 
- [ ] Upload artifacts on the Google storage.
- [ ] Post a comment on the `PR` with the artifact media link to download.

PR #5474, opened 2017-06-16
Build python packages and pex file using buildkite

mrpau-eduard:
## Summary
This is implemented also on kolibri and will implement also on ka-lite.

## TODO

- [x] Python packages 
- [x] Pex file
- [x] Build windows installer

## Issues addressed
#5473


Makefile
```diff
@@ -70,6 +70,9 @@ coverage:
 	coverage run --source kalite bin/kalite test
 	coverage report -m
 
+writeversion:
```
benjaoming: We don't use this for KA Lite, it's a Kolibri thing :)

Makefile
```diff
@@ -106,7 +109,7 @@ sdist: clean docs assets
 	python setup.py sdist --formats=$(format) --static
 	python setup.py sdist --formats=$(format)
 
-dist: clean docs assets
+dist: writeversion clean docs assets
```
benjaoming: Can be removed, too

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:xenial
+
+# install latest python and nodejs
+RUN apt-get -y update --fix-missing
+RUN apt-get install -y software-properties-common curl
+RUN add-apt-repository ppa:voronov84/andreyv
+RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
```
benjaoming: This is interesting - AFAIK, we've been using the old legacy Node (0.12). Does this work for you?

Under any circumstances, I don't think we should have it here for building Pex files. This node version isn't used for our tests, so it can't really be trusted.

Makefile
```diff
@@ -118,3 +121,16 @@ dist: clean docs assets
 
 install: clean
 	python setup.py install
+
+pex:
+	ls dist/ka_lite-*.whl | while read whlfile; do pex $$whlfile -o dist/kalite-`unzip -p $$whlfile kalite/VERSION`.pex -m kalite --python-shebang=/usr/bin/python; done
```
benjaoming: Instead of writing something in `kalite/VERSION` and then reading back the file through piped output from an unzip command, I would suggest that you just replace `unzip -p $$whlfile kalite/VERSION` with `git describe --tags` ;)

mrpau-eduard:
Hi @benjaoming done updating the `PR` can you take a look again.
you can check the build artifact here https://buildkite.com/learningequality/ka-lite/builds/69#9ebc63e1-3142-4fd4-8ce7-411a38d9ffe0

radinamatic:
@mrpau-eduard I grabbed the `exe` artifact to test but I'm guessing that it is not fully functional yet, since it skipped the whole `ka-lite-static` part of installation and server could not start afterwards:

win7 start running - oracle vm virtualbox_428

radinamatic:
@mrpau-eduard I see that you requested a review, but is there a new build apart from the one I tested yesterday?

mrpau-eduard:
Hi @radinamatic I am currently replicating the issue on the windows installer. 
I found some issue during installing the `whl` file, but If I install the`whl` file on the terminal manually it is working.

radinamatic:
Ok, ping me when there is a build for me to poke and prod! :wink:

mrpau-eduard:
I was able to replicate the issue: 
screen shot 2017-08-04 at 7 55 43 pm

The issue here is `ka_lite_static-0.17.2.dev1-py2-none-any.whl` is a wrong filename. Since the ka-lite-installer for windows assume that the filename is correct. It uses the `kalite.version` as the build identifier for each installer.

I try to bump the VERSION `0.17.2dev1` into `0.17.2` and it work. 
You can grab the installer here 
https://buildkite.com/learningequality/ka-lite/builds/73#17bc8ae0-fba9-4dd5-b755-80070fdc90f4

radinamatic:
@mrpau-eduard Looking good! With this latest build everything seems to install correctly, and no static files are missing! :+1: 

Is it my impression or the installation takes longer now? I tested on Windows 7... :thinking:

Makefile
benjaoming: Thinking about this again, I think you should change it before we start (ab)using cloud systems - after all, unnecessary computation is a waste of resources, whether we pay for them or not :)

Makefile
```diff
@@ -118,3 +119,20 @@ dist: clean docs assets
...
+
+
+dockerwriteversion:
+	python -c "import kalite; print(kalite.__version__)" > kalite/VERSION
```
benjaoming: Please remove this step, for the reason stated above (reading it back out via `unzip` is heavy) + introducing some new version maintaining mechanism isn't really ideal.

You can use `git describe --tags` or  `python -c "import kalite; print(kalite.__version__)"`

kalite/version.py
```diff
@@ -3,7 +3,7 @@
 # Must also be of the form N.N.N for internal use, where N is a non-negative integer
 MAJOR_VERSION = "0"
 MINOR_VERSION = "17"
-PATCH_VERSION = "2dev1"
+PATCH_VERSION = "2"
```
benjaoming: This should remain untouched.

Makefile
mrpau-eduard: Yes, I will remove that steps above since we didn't use pex for windows installer also.

benjaoming:
Great work @mrpau-eduard - Now that it's merged to 0.17.x, will you be able to try it out with Buildkite and switch it on for PRs?

It would be nice if the builds are made available through PR status - and *not* with a comment for every commit :) @rtibbles has made this work for Kolibri.

Issue #5477, opened 2017-06-23
Missing msgids in version 0.17 content packs


polilinux-fsrechia:
## Summary

Translation files for version 0.17 at crowdin are incomplete (at least for pt-BR). Even though crowdin shows 100% translated content for pt-BR 0.17, some very basic msg ids are missing.

## How to reproduce

Download .po files for both version 0.16 and 0.17. Notice that the string "Teach" is not present in the 0.17 translation file:

```
fsrechia@linux-t2je:/mnt/sda9> grep Teach 0.17-django*.po 
fsrechia@linux-t2je:/mnt/sda9> grep Teach 0.16-django*.po 
0.16-djangojs.po:msgid "Teach"
fsrechia@linux-t2je:/mnt/sda9> 
```

I've noticed this with files downloaded today (June 23rd 2017)

## Workaround:

Merge translation files from versions 0.16 and 1.17:

```
$ msgcat 0.17-django.po 0.17-djangojs.po 0.16-django.po 0.16-djangojs.po > unified.po
$ msgfmt unified.po -o unified.mo
```

and use `unified.mo` as both `backend.mo` and `frontend.mo`

## Real-life consequences

Standard pt-BR version of KA Lite 0.17.1 has some very basic translation strings missing. I am not sure if other languages are affected

benjaoming:
Thanks @polilinux-fsrechia for the feedback -- I'm going to investigate this now by comparing the msgids on CrowdIn with what I can generate locally from the `0.17.x` branch.

@mrpau-eugene would you be able to participate in solving this issue?

benjaoming:
Trying to run `kalite manage makemessages -l en` - works fine, however when generating the JavaScript part `--domain=djangojs`, problems occur on `0.17.x`, which makes me believe that the commands were not run on this branch:

```
kalite manage makemessages -l=en --domain="djangojs"
processing language =en
Traceback (most recent call last):
  File "ka-lite/kalite/packages/bundled/django/core/management/base.py", line 224, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "ka-lite/kalite/packages/bundled/django/core/management/base.py", line 263, in execute
    output = self.handle(*args, **options)
  File "ka-lite/kalite/packages/bundled/django/core/management/base.py", line 393, in handle
    return self.handle_noargs(**options)
  File "ka-lite/kalite/i18n/management/commands/makemessages.py", line 42, in handle_noargs
[...]

PR #5485, opened 2017-07-22
Translation updates

benjaoming:
## Summary

 * Documentation
 * Make target `msgids` for generating translation sources
 * Do not hide "IGNORE_PATTERNS" in a Python file, moved them to the Makefile
 * Adds the sources to the repo so we can track them moving forwards

## TODO

## Reviewer guidance

Let me know if I missed something in the way we generate msgids!

## Issues addressed

#5477


radinamatic:
OK, we are not done with this, yet...

This is what the Coach tools looks like after updating to the latest content pack in Spanish

virtualbox_ubuntu17 04_28_07_2017_19_06_18

Strings still in English:

`All selected` inside the topic selector and the `Submit` button.
`Search` string inside the search field.
`N/A` value for the summary of total hours on content compared to total hours connected.
`Key` string for the tabular report colors. 

Disregard the strings with orange overlay, as that is a different issue ([already reported](https://github.com/learningequality/ka-lite/issues/5494)), which we have to correct nevertheless... :wink: 

@mrpau-eduard @benjaoming I believe we should correct this too, before releasing 0.17.2.

benjaoming:
@radinamatic 

Some of these examples are unfortunately from JS libraries that aren't translated:

 * `All selected` is from `node_modules/bootstrap-multiselect/dist/js/bootstrap-multiselect.js`
 * `Search` is from `node_modules/bootstrap-multiselect/dist/js/bootstrap-multiselect.js`

I've opened #5496 for the two others.

PR #5496, opened 2017-07-29
Mark missing msgids for translation

benjaoming:
## Summary

@radinamatic found a couple of missing things.


kalite/coachreports/static/js/coachreports/tabular_reports/hbtemplates/tabular-view.handlebars
```diff
@@ -1,7 +1,7 @@
 <div>
     <div class="row">
         <div class="col-md-6 col-xs-12">
-        <label id="legendMargin" for="legend">Key</label>
+        <label id="legendMargin" for="legend">{{_ "Color legend"}}</label>
```
radinamatic: @benjaoming I'm not sure if this is the best solution, I would keep the string `Key`, it's simpler and already present on the page:

color-legend

benjaoming: The word "Key" will confuse translators, that's why I replaced it - sadly it's very long, but it shows up okay.

A secondary issue is that the legend is shown twice which IMO is plain silly :)

image

PR #5503, opened 2017-08-18
Re-arrange CI test build steps

benjaoming:
Move a bunch of dependency setup that affects all containers to the dependency build so the build fails when any of these steps fail

Before, our builds would continue even though JS asset building had failed. Which created a lot of confusing errors in Selenium tests.


benjaoming:
Since this passes tests, I'm merging it!

PR #5506, opened 2017-08-19
Fix database move message

benjaoming:
## Summary

The current setup command wrongly claims that a database file exists and will be backed up on a clean release:

```
$ rm -rf ~/.kalite
$ kalite start --foreground
Running 'kalite start' in foreground...
Generating a new secret key file ~/.kalite/secretkey.txt...

Setting up KA Lite; this may take a few minutes; please wait!

                                     
   _   __  ___    _     _ _          
  | | / / / _ \  | |   (_) |         
  | |/ / / /_\ \ | |    _| |_ ___    
  |    \ |  _  | | |   | | __/ _ \   
  | |\  \| | | | | |___| | ||  __/   
  \_| \_/\_| |_/ \_____/_|\__\___|   
                                     
https://learningequality.org/ka-lite/
                                     
         version 0.17.2
                                     
(Re)moving database file to temp location, starting clean install. Recovery location: /tmp/tmpkPYEPM
```

### DIFF
--- a/kalite/distributed/management/commands/setup.py
+++ b/kalite/distributed/management/commands/setup.py
@@ -332,32 +332,29 @@ class Command(BaseCommand):
 
         install_clean = not database_exists
 
-        if database_file:
-            if not database_exists:
-                install_clean = True
+        if database_exists:
+            # We found an existing database file.  By default,
+            #   we will upgrade it; users really need to work hard
+            #   to delete the file (but it's possible, which is nice).
+            logger.info(
+                "-------------------------------------------------------------------\n"
+                "WARNING: Database file already exists!\n"
+                "-------------------------------------------------------------------"
+            )
+            if not options["interactive"] \
+               or raw_input_yn("Keep database file and upgrade to KA Lite version %s? " % VERSION) \
+               or not raw_input_yn("Remove database file '%s' now? " % database_file) \
+               or not raw_input_yn("WARNING: all data will be lost!  Are you sure? "):
+                install_clean = False
+                logger.info("Upgrading database to KA Lite version %s" % VERSION)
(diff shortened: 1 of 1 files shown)


kalite/distributed/management/commands/setup.py
```diff
@@ -398,7 +395,7 @@ def handle(self, *args, **options):
         ########################
 
         # Move database file (if exists)
-        if install_clean and database_file and os.path.exists(database_file):
+        if install_clean and database_file and database_exists:
```
benjaoming: This is the essential part, but I believe that the expression `install_clean and database_file and database_exists` is pretty messed up when looking at the definitions of these vars above.

benjaoming:
@mrpau-eugene could you check if I got it right? It's a bit complicated, would be nice with another pair of eyes to verify :)

kalite/distributed/management/commands/setup.py
```diff
@@ -332,32 +332,29 @@ def handle(self, *args, **options):
...
+            #   to delete the file (but it's possible, which is nice).
+            logger.info(
+                "-------------------------------------------------------------------\n"
+                "WARNING: Database file already exists!\n"
+                "-------------------------------------------------------------------"
+            )
+            if not options["interactive"] \
+               or raw_input_yn("Keep database file and upgrade to KA Lite version %s? " % VERSION) \
+               or not raw_input_yn("Remove database file '%s' now? " % database_file) \
+               or not raw_input_yn("WARNING: all data will be lost!  Are you sure? "):
```
benjaoming: It occurs to me that I need several minutes to understand what the above does and how `install_clean` is set.

Issue #5516, opened 2017-09-06
Some *.js files are not being included for translation


mrpau-eugene:
## Summary

Strings inside */bundle_modules/ folder aren't being included for translation even when wrapped by the `gettext` function.

Sample issue here: #5514

benjaoming:
Fixed in #5517

Issue #5514, opened 2017-09-01
Missing translations on Manage Languages


radinamatic:
## Summary

Still missing some translations on Manage Languages page.

## System information

 - Operating system: Windows 7 & Ubuntu 16.10
 - Version: 0.17.3
 - Browser: all

## Screenshots

![ubuntu3 start running - oracle vm virtualbox_588](https://user-images.githubusercontent.com/1457929/29988961-6aa5de8c-8f70-11e7-9073-3f370a2ee6ed.png)

![w7 start running - oracle vm virtualbox_585](https://user-images.githubusercontent.com/1457929/29988968-73a3bab8-8f70-11e7-9c4e-c33308773d68.png)

![w7 start running - oracle vm virtualbox_586](https://user-images.githubusercontent.com/1457929/29988970-82cd6f5c-8f70-11e7-95ab-abf2b4079463.png)

benjaoming:
Fixed in #5517 - and new msgids

This has added 40 new strings and removed 3. So it will be substantial IMO. @radinamatic could you give an overview of the most important places that were untranslated?

benjaoming:
@radinamatic @mrpau-eugene I would like to close this issue with a final verdict about how serious the issue is, for instance an idea of where the missing translations mainly feature, like are they important pages that are hard to understand for the intended user?

Also, have they been translated now for the major languages? I can see that Spanish is completed..

radinamatic:
@benjaoming All the strings from #5517 have been translated on Crowdin, but some of them did not make it into 0.17.3. Not a major tragedy, and this can be closed!

![w7 0 17 3 running - oracle vm virtualbox_629](https://user-images.githubusercontent.com/1457929/30695445-86270326-9ed8-11e7-81e1-ff3312809e4e.png)

benjaoming:
@radinamatic - per #5518 which will be released in 0.17.4, we'll move the translations into the release itself (see the issue), so I was just wondering how much this glitch pushes our priorities.. but I think for now, we'll just see how quickly we can get it resolved and release 0.17.4 - seems there's no panic.

radinamatic:
Agreed, @benjaoming, since this is on Admin side, no panic involved! :+1:

PR #5517, opened 2017-09-06
Missing translations for *.js files

mrpau-eugene:
## Summary

@rtibbles mentioned that the `IGNORE_PATTERNS` inside the `Makefile` ignores the source files for translation. We don't want this to happen.

Also, thanks @benjaoming for introducing the `make msgids` to me 😄 
## Issues addressed

#5514 
#5516


benjaoming:
Oh #facepalm, well there's a pile of additional translations to be dealt with! Nicely spotted, I'll merge and upload the files to CrowdIn.

If we see additional translation activity, we can consider to inject the newly built translations files directly into existing content packs to avoid rebuilding them.

benjaoming:
Btw. as the commit message said "Missing translations for *.js files" - this isn't true, it was a few .js messages in webpack bundles.

mrpau-eugene:
>Btw. as the commit message said "Missing translations for *.js files" - this isn't true, it was a few .js messages in webpack bundles.

Oh, yeah! You're right @benjaoming my bad.

Issue #5518, opened 2017-09-06
Move interface translations out of content packs


benjaoming:
## Summary

We call them "content packs" for a reason.

Interface translations tie directly to the software release and should as such be released with the software and not sitting separately in these rather huge .zip files for content packs, which we'd rather *avoid* having to update.

I order to fix this, we need to:

 * [ ] Stop unpacking `.po` files from content packs
 * [ ] Remove `~/.kalite/locale` from `LOCALE_PATHS`, it shouldn't be necessary anymore
 * [ ] Setup crowdin-client to point to our correct CrowdIn resources and add it to `make assets` target in order to have translations for all languages downloaded and built into releases.

I'm targeting this for 0.17, because I see that we can fix this:

1. without rebuilding content packs, just leaving the content packs as they are - 0.17.4+ won't use .po resources from content packs, and everything <0.17.3 will.
1. being backwards compatble
1. avoid building content packs again once the 40 strings in #5517 are translated

benjaoming:
There was an unfortunate and confusing error in the description, sorry about this @mrpau-eugene 

>  ~Ensure that new path `~/.kalite/project/locale` works~

Becomes:

>  Ensure that new path `kalite/project/locale` works

PR #5524, opened 2017-10-02
[WIP] Stop unpacking .po files from content packs

mrpau-eugene:
## TODO

 * [x] Remove old functionality: Stop unpacking `.po` files from content packs
 * [x] Remove old functionality: Remove `~/.kalite/locale` from `LOCALE_PATHS`, it shouldn't be necessary anymore
 * [ ] Ensure that new path `kalite/project/locale` works

## Issues addressed
#5518


kalite/settings/base.py
```diff
@@ -188,16 +188,15 @@
...
 
 LOCALE_PATHS = (USER_WRITABLE_LOCALE_DIR, KALITE_APP_LOCALE_DIR)
-if not os.path.exists(USER_WRITABLE_LOCALE_DIR):
-    os.mkdir(USER_WRITABLE_LOCALE_DIR)
-
 DEFAULT_DATABASE_DIR = os.path.join(USER_DATA_ROOT, "database",)
-if not os.path.exists(DEFAULT_DATABASE_DIR):
-    os.mkdir(DEFAULT_DATABASE_DIR)
+
+ensure_dir(USER_WRITABLE_LOCALE_DIR)
```
benjaoming: This functionality shouldn't actually be here. It's just a settings file.

Imagine that the default setting is broken (e.g. unwritable location), then even changing the setting wouldn't work.

`extract_content_pack_metadata()` seems like a better place for `USER_WRITABLE_LOCALE_DIR` to be created.

kalite/settings/base.py
```diff
@@ -188,16 +188,15 @@
...
+USER_WRITABLE_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'project', 'locale')
+KALITE_APP_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'project', 'locale')
 
 LOCALE_PATHS = (USER_WRITABLE_LOCALE_DIR, KALITE_APP_LOCALE_DIR)
-if not os.path.exists(USER_WRITABLE_LOCALE_DIR):
-    os.mkdir(USER_WRITABLE_LOCALE_DIR)
-
 DEFAULT_DATABASE_DIR = os.path.join(USER_DATA_ROOT, "database",)
-if not os.path.exists(DEFAULT_DATABASE_DIR):
-    os.mkdir(DEFAULT_DATABASE_DIR)
```
benjaoming: I prefer this: `USER_DATA_ROOT` has to exist at this point. If it doesn't, the above breaks.

kalite/settings/base.py
```diff
@@ -188,16 +188,15 @@
 # the user running kalite and should be in a user-data
 # storage place.
 
-USER_WRITABLE_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'locale')
-KALITE_APP_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'locale')
+from fle_utils.general import ensure_dir
```
benjaoming: `fle_utils` is something we would like to get rid of... but don't have time to factor out.. it's just a good idea not to use it anymore.. but below you'll see that it's good to not use `ensure_dir` in this context anyways....

benjaoming:
Great that coverage will improve because affected (removed!) code has no coverage :D

kalite/settings/base.py
mrpau-eugene: Hmm. I didn't know that we will be removing `fle_utils`.. anyway, I used the `ensure_dir` since I get an error if I do this:

```python
def extract_content_pack_metadata(zf, lang):
    lang = lcode_to_django_lang(lang)
    modir = get_po_filepath(lang)

    if not os.path.exists(modir):
        os.mkdir(modir)
```
```
OSError: [Errno 2] No such file or directory: '/Users/me/.kalite/project/locale/en/LC_MESSAGES'
```

kalite/distributed/management/commands/retrievecontentpack.py
```diff
@@ -176,29 +174,18 @@ def process_content_pack(self, zf, lang):
 
 
 def extract_content_pack_metadata(zf, lang):
+    lang = lcode_to_django_lang(lang)
+    modir = get_po_filepath(lang)
+    if not os.path.exists(modir):
+        os.makedirs(modir)
```
benjaoming: The `modir` is no longer relevant for content packs, so I'm not sure what this is doing here? The variable is not even used :)

kalite/settings/base.py
```diff
@@ -188,14 +188,12 @@
 # the user running kalite and should be in a user-data
 # storage place.
 
-USER_WRITABLE_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'locale')
-KALITE_APP_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'locale')
+USER_WRITABLE_LOCALE_DIR = os.path.join(USER_DATA_ROOT, 'project', 'locale')
```
benjaoming: We are no longer serving `django.mo` etc files from content packs. Content packs will only contain subtitles.

Meanwhile, another important aspect is that we cannot and should not change location folders of existing unpacked subtitles, because that will mean that existing content packs have to be re-installed - that cannot happen in an upgrade.

This PR should only remove everything that has to do with translation files (`.mo`) from content packs and instead use the location of `kalite/project/locale`, which is a non-existing location until you put something there :)

I think you have confused it with `USER_DATA_ROOT`, which is to be understood as `~/.kalite`

benjaoming: I think `USER_WRITABLE_LOCALE_DIR` and `KALITE_APP_LOCALE_DIR` are entirely useless paths and you should factor them out completely.

The definition of `get_subtitle_path` doesn't seem to rely on these at all, `.kalite/locale` was only for translation files IIRC.

mrpau-eugene:
@benjaoming I just went back again on this PR, while reading the original issue again, I just wanted to know if the `makemessages` command should point to `kalite/project/locale` and not `kalite/locale` anymore?

mrpau-eugene:
Originally, the instructions are:

* Remove old functionality: Stop unpacking .po files from content packs
* Remove old functionality: Remove `~/.kalite/locale` from `LOCALE_PATHS`, it shouldn’t be necessary anymore
* Ensure that new path `kalite/project/locale` works


But there were other factors that affects the old process and I cannot just simply remove the `~/.kalite/locale` because there are some files being stored in there such as the `{lang}_metadata.json`, so I was reading the codebase and initially came up with these new procedures:

* Remove old functionality: Stop unpacking .po files from content packs
* Remove old functionality: Remove `~/.kalite/locale` from `LOCALE_PATHS`, it shouldn’t be necessary anymore
* The above contains `{lang}_metadata.json` so we should move the `{lang}_metadata.json` to a new folder, maybe `~/.kalite/language-metadata/`
* Change `update_jsi18n_file` output path to somewhere inside `kalite/project`
* Change the `makemessages` command output path to `kalite/project/locale`
* For existing upgrades, copy the `{lang}_metadata.json` to the new metadata folder and delete the old one.
* Ensure that new path `kalite/project/locale` works

I’m not quite sure if it is correct and it still might be missing something. Mind checking these procedures @benjaoming ?

benjaoming:
Hi @mrpau-eugene 

> But there were other factors that affects the old process and I cannot just simply remove the `~/.kalite/locale` because there are some files being stored in there such as the `{lang}_metadata.json`

Ah no, the description doesn't say to remove the directory, just to remove it from the settings so we are sure translations still living there (from old content pack installations) are ignored:

> Remove old functionality: Remove `~/.kalite/locale` from `LOCALE_PATHS`, it shouldn’t be necessary anymore

benjaoming:
Hi @mrpau-eugene - I'll try to continue work over in #5537, will perhaps pass it back to you later -- the communication regarding how to understand the issue that I posted was a bit heavy for me to keep up with, but the task is also very much entangled in our old understanding of how KA Lite works -- so hopefully this will be the last headache about translations stuff, before we can move on to a more modern CrowdIn-based approach.

mrpau-eugene:
@benjaoming sure! just ping me anytime :)

Issue #5523, opened 2017-09-28
Simplified login does not work if there are more than 1,000 users


mrpau-eugene:
## Summary

It was originally a report from Edulution's feedback:

> We noticed that this is due to the API request (securesync/api/user/facility<facility_id>&is_teacher=false) only returning 1000 objects from the students table at the login screen. To overcome this, we simply deleted inactive student accounts at each facility until they were less than 1000. However, this is only a temporary solution. A more permanent fix to enable >1000 learners to login via simplified login even would be appreciated

## System information

*Please specify the KA Lite version you were using and your operating system!*

 - Operating system: ?
 - Version: 0.16.?
 - Browser: ?

## Real-life consequences (anything community should be aware, for instance how it affects your deployment)

We don't want users to only login when it's less than a thousand users registered.

benjaoming:
FYI @Aypak :)

PR #5527, opened 2017-10-20
Set max_limit to 0 in order to fetch all users.

mrpau-eugene:
## Summary

Reference: https://stackoverflow.com/a/18477897

FYI @benjaoming @Aypak 

It seems that tastypie automatically limits the API request to `1000` and I have no idea why do they do that.

I was not able to replicate the issue when there are 1,000 users registered in a facility. But I think this fixes the issue since it is already returning the number of expected users

`{"meta": {"limit": 0, "offset": 0, "total_count": 1011}`

## TODO

## Issues addressed
#5523

### DIFF
--- a/docs/installguide/release_notes.rst
+++ b/docs/installguide/release_notes.rst
@@ -12,6 +12,11 @@ to read the release notes.
 0.17.4 (unreleased)
 -------------------
 
+Bug fixes
+^^^^^^^^^
+
+* Simplified login is now working when there are 1,000 or more users registered in a facility. :url-issue:`5523`
+
 Developers
 ^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


benjaoming:
Ping @Aypak 

Great work, it does seem very likely to fix the issue.

But okay, so it's very much a hotfix, because it means that JS is loading 1000+ usernames in from an API request in order to process a simplified login, right? So the behavior us very bad in that sense...

benjaoming:
Could you add a release note about it? We can include it in the upcoming 0.17.4 patch release.

rtibbles:
Oh tastypie...

Aypak:
Thanks a million!

PR #5532, opened 2017-11-02
Check HTTP Response for BDD tests

mrpau-eugene:
## Summary

Just check the HTTP response if it is 200
## TODO

- [x] Use response status code

## Reviewer guidance

## Issues addressed


benjaoming:
Looks good! I'm not sure how many times, but this has definitely been a cause of head aches in the past :)

Issue #5544, opened 2017-11-20
PEX test file build script + buildkite integration


benjaoming:
## Summary

To speed up testing a bit, an automated PEX build via Buildkite will be a good starting point:

 * [ ] A make target for a `.buildkite/pex.sh` script creating `dist/ka-lite-static<version-info>.pex`
 * [ ] Activate script on Buildkite
 * [ ] Get PR status feedback through bot

Versions:

I think you can stick output of `git describe` in the file name. We don't have the same clever automatic dev versioning in KA Lite as Kolibri, but at least having a unique file name would be nice.

## System information

*Please specify the KA Lite version you were using and your operating system!*

 - Operating system: ?
 - Version: current 0.17.x branch
 - Browser: ?

mrpau-eugene:
@benjaoming I was wondering, would it be okay to use `git describe`?

I'm on the `0.17.x` branch and executed `git describe` and then it gave me this `v0.17.3-30-g86e1ea6e2`.

Doesn't it make confusion since the KA-Lite Version is `0.17.4`, but we will get something like `ka-lite-static-v0.17.3-30-g86e1ea6e2.pex`?

benjaoming:
@mrpau-eugene garh, you are right, it's because we aren't tagging pre-releases

let's change the work flow, so pre-releases get merged into `0.17.x` and only `master` serves as the "current stable". That way, we can tag and track our pre-releases in a linear way.

benjaoming:
I can start doing tags for pre-releases once #5545 is finalized and a new pre-prelease is out..

benjaoming:
@mrpau-eugene 0.17.x is now tagged, so if you apply `git describe`, you should get meaningful output now :)

mrpau-eugene:
@benjaoming thanks!

benjaoming:
Yay! Fixed in #5551 :) ( which was merged in #5552 )

PR #5551, opened 2017-12-01
[WIP] Build .pex files to speed up testing

mrpau-eugene:
## Summary

@benjaoming quite off from the instructions but I made it similar to how kolibri generates the pex and installers. I can make it as instructed (.buildkite/pex.sh) if this approach is not appropriate

## TODO

 * [x] Generate pex file `dist/ka-lite-static<version-info>.pex`
 * [ ] Activate script on Buildkite
 * [ ] Get PR status feedback through bot

Should we also upload the files to google cloud storage?

## Reviewer guidance

## Issues addressed

#5544

### DIFF
--- a/.buildkite/build_whl.sh
+++ b/.buildkite/build_whl.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+make dockerenvdist
+buildkite-agent artifact upload 'dist/*.whl'
+buildkite-agent artifact upload 'dist/*.tar.gz'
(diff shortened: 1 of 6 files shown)


benjaoming:
@mrpau-eugene I found that the current Buildkite settings skip third-party PRs for security reasons.. which is actually a good idea.. however, just for now, let's switch it on so we can start using it and testing it some more.

Would you mind opening a new PR? Buildkite skips it despite that I close and re-open because it has seen the commit before methinks...

PR #5554, opened 2017-12-05
Build pex file and separate windows installer pipeline

mrpau-eugene:
test PR to see if it autobuilds on buildkite


benjaoming:
It does!! :tada: :smile:

benjaoming:
@mrpau-eugene I think this looks very useful already, maybe if we can get a thumbs up from @radinamatic that the built artifacts (can be found on Buildkite ATM) are working, then we could merge.

@mrpau-eugene One very nice to have would  be if we can have the artifacts available same way as for Kolibri (in the build status), so people don't have to login to see them on Buildkite. Do you know how to do this?

mrpau-eugene:
> One very nice to have would be if we can have the artifacts available same way as for Kolibri (in the build status), so people don't have to login to see them on Buildkite. Do you know how to do this?

@benjaoming I'm not quite sure how but I got some reference from kolibri repo. @rtibbles pointed me to this code on Kolibri https://github.com/learningequality/kolibri/blob/develop/.buildkite/upload_artifacts.py#L123 but I need an access token that has the right permissions (not quite sure which permissions are needed) in order to have something like that.

benjaoming:
Oh right, you mean this line? `ACCESS_TOKEN = os.getenv("GITHUB_ACCESS_TOKEN")`

benjaoming:
@mrpau-eugene I have set both `GOOGLE_APPLICATION_CREDENTIALS` and `GITHUB_ACCESS_TOKEN`, in case they don't work, let me know by PM.

mrpau-eugene:
@benjaoming yup! that's pretty much it. I also want to know how to build the debian installers manually. Can you give me some help?

mrpau-eugene:
@benjaoming github status and being uploaded to google cloud is working now

benjaoming:
GREAT WORK @mrpau-eugene :tada:

mrpau-eugene:
@benjaoming Thanks 😄 

Btw, we should change the access token as soon as LE Bot is ready..

benjaoming:
@mrpau-eugene yes, did someone say they would create it btw? otherwise, I can do that today!

mrpau-eugene:
@benjaoming I think it was @rtibbles who mentioned it that they were talking about creating an LE bot

PR #5559, opened 2017-12-22
master into 0.17.x

benjaoming:
Merging master back into 0.17.x to explore the issue with `git describe` and if we can manage to somehow have a release flow where we can make benefit of the command or not.


benjaoming:
This does not make the `v0.17.4` tag appear on the 0.17.x branch using `git describe`, so I'm going to continue with the plan of just putting the current KA Lite version and a commit hash in Buildkite artifacts.

PR #5565, opened 2018-01-30
Fix wrong remote_size when annotating content

mrpau-eugene:
## Summary

Take this for example:

After a successful scan, some of the Topics' `remote_size` becomes `0`

screen shot 2018-01-30 at 1 24 42 pm

Therefore, when ticking the `Math` topic, it might look something like this:
image

@benjaoming this was the one I found. Could you test this one?

## TODO

## Reviewer guidance

## Issues addressed


benjaoming:
This change didn't make sense, the logic expression was a "not ((a and b) or b)" which would be the same as just saying "not b".

PR #5592, opened 2019-01-12
Always raise an error when importing from django.contrib.gis

benjaoming:
## Summary

This is an attempt to address the issue in https://github.com/iiab/iiab/issues/1382

There are version strings of libgeos system libraries that seem to make the loading crash.

## TODO

## Reviewer guidance

Let's assume that our tests would break, since this is about load-time issues.. not request/response related.

## Issues addressed

https://github.com/iiab/iiab/issues/1382


benjaoming:
Ay, just ran into this on a normal 18.04

PR #5595, opened 2019-07-15
central-develop into 0.17.x

benjaoming:
## Summary

These changes are necessary to align the current central server with the current KA Lite release. Some changes to the `control_panel` app are necessary to keep data export working for many of the larger deployments. More specifically, they have to generate CSV exports in an async environment, which will be done following a change to the central server itself to just use this upcoming 0.17.6.dev release.

This PR flags a few problematic changes that we need to fix for things to work in both the distributed and the central server environments

## TODO

## Reviewer guidance

I would definitely not mind a quick look from @jamalex and @rtibbles if you happen to have time.

## Issues addressed


kalite/coachreports/static/css/coachreports/exercise_mastery_view.css
```diff
@@ -0,0 +1,158 @@
...
+.exercise-name {
+    height: 100px;
+    text-align: center;
+    vertical-align: none;
+    padding: 0px;
+}
+
+#exercise-mastery {
+    font-size: 10px
+}
```
benjaoming: This might be a feature lost from 0.16.x - or it might be unwanted in 0.17?

kalite/coachreports/static/js/coachreports/exercise_mastery_view.js
```diff
@@ -0,0 +1,74 @@
...
+            }
+        },
+        onSelect: function(select, dtnode) {
+            var selKeys = $.map(dtnode.tree.getSelectedNodes(), function(dtnode){
+              return dtnode.data.key;
+            });
+            window.location.href = setGetParam(window.location.href, "playlist", selKeys);
+        }
+    });
+});
```
benjaoming: This might be a feature lost from 0.16.x - or it might be unwanted in 0.17?

kalite/control_panel/api_resources.py
```diff
@@ -282,6 +282,23 @@ def alter_list_data_to_serialize(self, request, to_be_serialized):
...
+                for i in xrange(101,104):
+                    if StoreTransactionLog.objects.filter(user=user, context_id=i, context_type="unit_points_reset", purchased_at__gte=bundle.data["timestamp_first"]).count() > 0:
+                        bundle.data["unit"] = i
+                        break
+
+            # For entries we are not sure about the unit, we keep them as 0, mostly chances are that the unit would be the current_unit.
+            # As we can't predict the current unit on the central server, its better to have the value as 0.
+            # We can't predict the current unit because in some database we have unit_point_reset gift card for unit 101 whereas the current unit is also 101.
+            # So we can't find current_unit by saying that the first unit that doesn't have the unit_point_reset gift card is current_unit.
+
```
benjaoming: I don't understand what all this is. Never heard about the "gift card" part nor the `purchased_at__gte` in the query further up? We don't have a model `StoreTransactionLog` anywhere, so I think all of this should go.

kalite/distributed/static/js/distributed/bundle_modules/base.js
```diff
@@ -14,6 +14,9 @@ require("browsernizr/test/canvas");
 require("browsernizr/test/touchevents");
 var Modernizr = require("browsernizr");
 
+// Expose this as a global object for use in central server inline JS.
+global.getCookie = require("utils/get_cookie");
+
```
benjaoming, line 19: This needs to be configurable and only switched on in the central server.

kalite/packages/bundled/fle_utils/django_utils/command.py
```diff
@@ -118,6 +118,73 @@ def call_command_async(cmd, *args, **kwargs):
...
+            cleaned_val = unicode(val).replace(" ", "")
+            cmd += (u"%s%s=%s" % (prefix, key, cleaned_val),)
+
+    # we also need to change the environment to point to the the local
+    # kalite settings. This is especially important for when the
+    # central server calls this function, as if we don't change this,
+    # kalitectl.py wil look for centralserver.settings instead of
+    # kalite.settings.
+    new_env = os.environ.copy()
+    new_env["DJANGO_SETTINGS_MODULE"] = kwargs.get("settings") or "kalite.settings"
```
benjaoming: This should ideally be configurable and only switched on in the central server. The whole function is only used in a test, though, `centralserver/central/tests/utils/distributed_server_factory.py`.

kalite/packages/bundled/securesync/devices/api_views.py
```diff
@@ -49,15 +49,17 @@ def register_device(request):
     try:
         if not client_device.verify():
             # We've been getting this verification error a lot, even when we shouldn't. Send more details to us by email so we can diagnose.
-            msg = "\n\n".join([request.body, client_device._hashable_representation(), str(client_device.validate()), client_device.signed_by_id, client_device.id, str(request)])
+            import inspect
+            d = client_device
+            msg = "\n\n".join([str(msg) for msg in [request.body, d._hashable_representation(), d.__dict__, d.__class__, inspect.getfile(d.__class__), local_version, d.description, d.signed_by_id, d.id, request]])
```
benjaoming: This looks like debugging noise?

benjaoming (commented):
Adding some notes on whether things should really go in.

kalite/coachreports/static/js/coachreports/exercise_mastery_view.js
benjaoming: The two `exercise_master_view` changes are from: https://github.com/learningequality/ka-lite/pull/3033

@aronasorman noted:

> We can revisit the PRs merged into nalanda by looking at the nalanda-bugfixes milestone, and then porting whatever PRs are relevant to the develop branch.

benjaoming: The CSS lives here:

https://github.com/learningequality/ka-lite/blob/master/kalite/coachreports/static/css/coachreports/tabular_view.less

I've searched the codebase for occurrences of "exercise_mastery_view", and as there are none, I will assume that the JS file is unused.

kalite/control_panel/api_resources.py
benjaoming: `StoreTransactionLog` is a model from `kalite.store` which is an application we removed.

kalite/packages/bundled/fle_utils/django_utils/command.py
benjaoming: Since all this code is used from the central server's test suite, I'm moving it there.

kalite/packages/bundled/securesync/devices/api_views.py
benjaoming: Reverting back to the former debug email.

kalite/packages/bundled/securesync/management/commands/register.py
```diff
@@ -61,7 +61,7 @@ def handle(self, *args, **options):
         csrftoken = s.cookies['csrftoken_central']
         login_data = dict(username=options["username"], password=options["password"], csrfmiddlewaretoken=csrftoken, next='/')
         r = s.post(login_url, data=login_data, headers={"Referer": login_url})
-        assert r.status_code == 200, "Error logging into central server: " + r.content
+        assert r.status_code == 200 and "Incorrect user" not in r.content, "Error logging into central server: " + r.content
```
benjaoming, line 64: The change looks like a desired precision from the commit msg: 34a27be96b2feb8bc09582ef401ced3cb2959ddb

kalite/distributed/static/js/distributed/bundle_modules/base.js
benjaoming: Okay, this might actually be harmless to the rest of the KA Lite application, wouldn't you think so @rtibbles ?

kalite/packages/bundled/securesync/api_client.py
```diff
@@ -13,9 +13,17 @@
...
+        
+        if not host:
+            if settings.CENTRAL_SERVER_URL:
+                self.url = settings.CENTRAL_SERVER_URL
+            else:
+                self.url = "%s://%s/" % (settings.SECURESYNC_PROTOCOL, settings.CENTRAL_SERVER_HOST)
+        else:
+            parsed_url = urllib2.urlparse.urlparse(host)
+            self.url = "%s://%s" % (self.parsed_url.scheme, self.parsed_url.netloc)
+
```
benjaoming, line 26: I'm not sure why the latter used to call `urllib2.urlparse.urlparse`, it seems that somehow it checks if the `host` is a valid URL with a scheme even though the kwarg was called `host`.. I don't quite grasp this. The `central-develop` branch had this, which would cause a regression on 0.17.x:

```
    def __init__(self, host=None, require_trusted=True, verbose=True):
        self.url = host or settings.CENTRAL_SERVER_URL
        self.parsed_url = urllib2.urlparse.urlparse(self.url)
```

benjaoming:
Having tested this to the point of registration and by introspecting settings, it seems that the securesync client connects correctly. I had another issues fixed in #5597 - but this should be fine.

PR #5627, opened 2020-05-23
Circle CI 2.1 conf

benjaoming:
## Summary

Buildkite and Circle CI is restored, however Selenium-based tests now fail. The failures seem to be related to new breakages, but not the Selenium/geckodriver updates which seem to work.

* Nodejs 6.x was used as 6.17 before, but the index of the Debian repository is somehow broken and reports 6.14 as the latest version
* It's not possible anymore to use parallelism in Circle CI the way we did since we cannot share state between the containers as we did before
* Had to do a lot of experimentation to get a working combination of Firefox, Geckodriver and Selenium up and running again
* Many elements of Selenium tests have been disabled as they caused false negatives
* The coverage dip in the CodeCov report is not related to disabled tests, as it is just a few tests that have been disabled, and those were about validating for correctness, not executing code. Not sure why the coverage has gone down.
* Circle CI 2.x containers is not amazing.. it's just a maze.

## Background

Circle CI has killed the 1.0 API that we were using. So we needed to either go back to Travis CI or start using their 2.0. The changes introduced here have no new value other than just extending the lifespan of the automated tests and builds. It's very sad the way that Circle CI has removed their old API and forced us into this.

The reason why this problem went under the radar was that the Circle CI integration was simply shut down, so we kept getting :heavy_check_mark: on our builds because Buildkite was still building artifacts.. but alas, no testing was done.

It also turned out that the combination of geckodriver+Selenium+Firefox wasn't reproducible in the new Circle CI docker containers.. I couldn't make it work, sticking to the same Selenium version and going through a series of different geckodriver+Firefox combinations, including the ones we had used before.

The automated tests are the only means to introduce changes to a somewhat large codebase that will be difficult to support in case of breakage.

https://circleci.com/sunset1-0/

>  The CircleCI 1.0 cloud infrastructure is no longer available as of March 15, 2019.

## TODO

## Reviewer guidance

*If your PR has a significant size, give the reviewer some helpful remarks*

## Issues addressed


benjaoming (commented):
No code was changed, just tests and build scripts. And a bit of developer documentation. Hence, I will self-merge this in order to have functional builds restored for other release PRs.

Issue #4408, opened 2015-09-10
Content Rating appears before user has engaged with the content at all


rtibbles:
Branch: 0.15.x

Expected Behaviour: We should only elicit feedback about content once a user has actually engaged with it in some way.

Current Behaviour: We ask them for feedback before they have pressed play or attempted to answer any questions.

MCGallaspy:
How would you prioritize this? Could it potentially wait for 0.16? It doesn't seem outside the realm of possibility for 0.15, but it's a slightly more substantial change.

rtibbles:
I think we could triage it until 0.16.

MCGallaspy:
I think this was fixed? Will try to find PR.

MCGallaspy:
Not fixed. Question @richard: What does it mean for content to be "interacted with"? Can I say that a rating should only be available if the user has either a VideoLog or an ExerciseLog for the content item being viewed? Do we create ExerciseLogs alike for Khan exercises and Perseus assessment items?

rtibbles:
Yes and yes.

MCGallaspy:
@rtibbles you suggested moving rating logic from the `ContentAreaView` to the `ContentWrapperView`, in order to take advantage of that latter's log model and avoid making redundant requests. The simplest solution is to make a second request. An alternate solution is to use the fact that `TopicContainerOuterView` contains both views, and so a reference to `ContentWrapperView` can be made from the `ContentAreaView` through their mutual parent. Alternate alternate solution is to remove the `ContentAreaView` entirely. Moving the rating logic to another view could mean changing its layout (or at least where its container element is positioned).

The solutions are listed by increasing complexity. If I make a second request, is it an unmergeable offense in your opinion?

MCGallaspy:
I had started to address this here: https://github.com/MCGallaspy/ka-lite/commit/52de9a5e0faf8a4a5362ad222e2ce53e668d36f9

Commit msg:
```text
Modify should_show_rating, now returns a jQuery promise
Starts to address #4408. First, should_show_rating becomes a promise
since it *might* need to make a request. Secondly, it attempts to
determine whether a user has *interacted* with a content item before,
by fetching the associated XLogCollection. (X is Video or Exercise.)

Next step is to call show_rating if a *new* XLog is created when
initially one wasn't found -- though we may wish to take a different
approach entirely which requires only *one* request to the
XLogCollection api endpoint by folding ContentArea into another view.
```

[...]

PR #5635, opened 2025-11-08
Fix: Defer feedback prompt until after user engagement #4408

0xgaurav:
### Summary
Currently, the content feedback form is displayed before the user interacts with any learning content. This causes premature prompts for feedback.

### Expected Behaviour
Feedback should only be requested **after** the user has engaged with the content (e.g., played a video, clicked, or attempted an exercise).

### Implementation Details
- Added event listeners (`click`, `keypress`, `play`) in `rating/views.js`.
- Feedback view now renders only after the first engagement.
- Listeners are removed once triggered to prevent duplicate rendering.

### Testing
1. Load any content page.
2. Verify that no feedback prompt appears initially.
3. Click or start playback → feedback appears.
4. Confirm normal rating and text feedback flow still works.

### Related Issue
Addresses behaviour mentioned in branch `0.15.x` regarding premature feedback.


rtibbles:
Hi @0xgaurav - thanks for your interest in our open source projects. Unfortunately, KA Lite is no longer maintained. I will not merge this pull request, as it would not go anywhere, and besides, it is targeted at the wrong branch.

If you or someone with whom you are working is currently using KA Lite still - I would recommend that they use Kolibri, our successor platform. You can find more information about it at learningequality.org/kolibri - if you are interested in contributing to Kolibri, please see github.com/learningequality/kolibri .

If KA Lite usage is still required, I would recommend deploying your fix from your own fork. Thank you!

